PR #4, opened 2015-04-22
Implement selectors

kenzierocks:
See https://github.com/SpongePowered/Sponge/pull/243.

### DIFF
--- a/src/main/java/org/spongepowered/common/text/selector/SpongeSelectorFactory.java
+++ b/src/main/java/org/spongepowered/common/text/selector/SpongeSelectorFactory.java
@@ -1,0 +1,279 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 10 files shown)


zml2008:
This needs to be updated to set the appropriate fields in the GameRegistry, because currently none of this stuff is hooked up.

kenzierocks:
Looks like I lost that in the copy. Porting ASAP.

kenzierocks:
Alright, I think that should be good. If you merge the API and update the ref on master, I'll rebase this.

src/main/java/org/spongepowered/common/text/selector/SpongeSelectorFactory.java
```diff
@@ -0,0 +1,268 @@
...
+        if (!selector.startsWith("@")) {
+            throw new IllegalArgumentException("Invalid selector " + selector);
+        }
+        // If multi-character types are possible, this handles it
+        int argListIndex = selector.indexOf('[');
+        if (argListIndex < 0) {
+            argListIndex = selector.length();
+        }
+        String typeStr = selector.substring(1, argListIndex);
+        Optional<SelectorType> type = Sponge.getGame().getRegistry().getType(SelectorType.class, typeStr);
```
zml2008: This part won't work with the keys you registered selectors under though?

src/main/resources/common_at.cfg
```diff
@@ -109,3 +109,5 @@ public net.minecraft.entity.EntityLivingBase field_70755_b # entityLivingToAttac
 public-f net.minecraft.entity.EntityLivingBase field_94063_bt # _combatTracker
 
 public net.minecraft.util.FoodStats field_75126_c # foodExhaustionLevel
+
```
zml2008: Also you have to update the AT. Please actually test this...

src/main/java/org/spongepowered/common/text/selector/SpongeSelectorFactory.java
```diff
@@ -0,0 +1,268 @@
...
+    @SuppressWarnings("unchecked")
+    @Override
+    public <T, V> Set<Argument<T>> createArguments(ArgumentHolder<? extends ArgumentType<T>> type, V value) {
+        Set<Argument<T>> set = Sets.newLinkedHashSet();
+        if (type instanceof SpongeArgumentHolder.SpongeVector3) {
+            Set<Function<V, T>> extractors = ((SpongeArgumentHolder.SpongeVector3<V, T>) (Object) type).extractFunctions();
+            Set<? extends ArgumentType<T>> types = type.getTypes();
+            Iterator<Function<V, T>> extIter = extractors.iterator();
+            Iterator<? extends ArgumentType<T>> typeIter = types.iterator();
+            for (; extIter.hasNext() && typeIter.hasNext();) {
```
Mumfrey: This is normally called a `while` loop. Why the bizarre use of `for` ?

kenzierocks: I think I used to initialize the iterators in it. I'll patch it.

Aaron1011:
@kenzierocks: I'd recommend commiting on a branch other then `master`. I'll save you a lot of headache when pulling in other changes.

kenzierocks:
I normally do, this was done in a rush.

zml2008:
Could you try to get this fixed up @kenzierocks?

kenzierocks:
Nuking this in favor of new branch,

Issue #3, opened 2015-04-21
Update the README.md


gabizou:
Currently, the README file is a copy pasta of the Sponge implementation. We should document what SpongeCommon really is etc.

Zidane:
@Grinch typically cleans up READMEs and polishes them. Gonna have him PR one to cleanup each project for us.

PR #13, opened 2015-04-27
README.md Update

Grinch:
Addresses issue #3.


gabizou:
Possibly elaborate how this is a common repository for the current (two) implementations of SpongeAPI?

Deamon5550:
I would think that we would be probabl wanting to remove the information on running from this repo's readme, as running requires one of the implementations not just this repo.

Grinch:
@Deamon5550 Where in this README does it state how to run it?

@gabizou Never really been good with explaining things so if you have a better description to put there, let me know and I'll make the change.

Deamon5550:
@Grinch Sorry my mistake

gabizou:
@Grinch I think @Zidane had a very good description of it.

PR #18, opened 2015-05-02
Common implementation of networking API

simon816:
Taken from https://github.com/SpongePowered/Sponge/pull/226 but sharing some common implementations.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/network/MixinPacketBuffer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/network/MixinPacketBuffer.java
@@ -1,0 +1,372 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/common/network/AbstractNetworkManager.java
```diff
@@ -0,0 +1,52 @@
...
+import net.minecraft.network.PacketBuffer;
+import net.minecraft.network.play.server.S3FPacketCustomPayload;
+import org.spongepowered.api.network.ChannelListener;
+
+import java.util.List;
+
+public abstract class AbstractNetworkManager {
+
+    // Set by the implementation on startup
+    public static AbstractNetworkManager instance = null;
```
Zidane: Surely this could be handled better...you could initialize it in SpongeGame?

Also, the naming is off. Either we should do stuff like AbstractGame or SpongeGame or CommonGame...but pick one and don't deviate.

src/main/java/org/spongepowered/common/mixin/core/network/MixinPacketBuffer.java
```diff
@@ -0,0 +1,372 @@
...
+
+    public ChannelBuf cbuf$slice() {
+        return (ChannelBuf) new PacketBuffer(this.buf.slice());
+    }
+
+    public ChannelBuf cbuf$slice(int index, int length) {
+        return (ChannelBuf) new PacketBuffer(this.buf.slice(index, length));
+    }
+
+    public byte[] cbuf$array() {
```
gabizou: You might have to mark a few of these methods as `@Intrinsic` in the event we're mixing in a method that exists with the same method signature.

simon816:
This PR is finished. If anyone has any comments please say as this will be merged soon

src/main/java/org/spongepowered/common/mixin/core/network/MixinPacketBuffer.java
simon816: All methods use the underlying buf field so there shouldn't be any self-referencing methods anywhere

PR #19, opened 2015-05-03
[WIP] Implement human entity

simon816:
Migrated most functionality from https://github.com/SpongePowered/Sponge/pull/207

Would be great to get some feedback on how I've implemented it to see whether anything can be improved.
More testing needs to be done, especially when there's many players and humans.

### DIFF
--- a/src/main/java/org/spongepowered/common/entity/living/HumanEntity.java
+++ b/src/main/java/org/spongepowered/common/entity/living/HumanEntity.java
@@ -1,0 +1,276 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.spongepowered.common.entity.living;
+
+import com.google.common.base.Optional;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import com.mojang.authlib.GameProfile;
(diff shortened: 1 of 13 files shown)


src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -209,6 +221,28 @@ public String getSubjectCollectionIdentifier() {
     }
 
     @Override
+    public String getCustomName() {
+        return this.getGameProfile().getName();
+    }
+
+    @Override
+    public void setCustomName(String name) {
+        throw new UnsupportedOperationException("Cannot set the custom name a player");
```
me4502: Missing 'of'

src/main/java/org/spongepowered/common/entity/living/HumanEntity.java
```diff
@@ -0,0 +1,205 @@
...
+import java.util.Map;
+import java.util.Random;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+
+/*
+ * Notes
+ *
+ * The label above the player's head is always visible unless the human is in
+ * a team with invisible labels set to true. Could we leverage this at all?
```
Aaron1011: I think this should be left up to plugins, considering that the Scoreboard API allows setting nametag visibility. Automatially modifying it could cause problems, in addition to giving the false impression that a player's nametag can be hidden like any other mob's.

src/main/java/org/spongepowered/common/registry/SpongeGameRegistry.java
```diff
@@ -397,7 +394,7 @@
     private final Map<String, CookedFish> cookedFishMappings = Maps.newHashMap();
     private final Map<String, DyeColor> dyeColorMappings = Maps.newHashMap();
     private final Map<String, Art> artMappings = Maps.newHashMap();
-    private final Map<String, EntityType> entityTypeMappings = Maps.newHashMap();
+    protected final Map<String, EntityType> entityTypeMappings = Maps.newHashMap();
```
ZephireNZ: Any reason this was changed to protected?

simon816: It's not very good I know but the Sponge mod has to use a different entity class, see [SpongeModGameRegistry:120](https://github.com/SpongePowered/Sponge/pull/207/files#diff-fc191454fb273d4eafe538517be30af5R120)

ST-DDT: Maybe a protected method?
That way you can apply some verification to the params.

simon816: This is internal sponge stuff, no need to verify IMO. Some fields are even public, it's just for the implementation to make use of.

SonarBeserk: There is always a need to verify imo. Seems better to have a protected method and keep the map private.

src/main/java/org/spongepowered/common/entity/living/HumanEntity.java
simon816: A player's nametag cannot be hidden in the same way, but what about a human? A human could have properties just like any other living entity.

src/main/java/org/spongepowered/common/registry/SpongeGameRegistry.java
simon816: I believe @Mumfrey is going to clean up `SpongeGameRegistry` at some point so this will hopefully be improved when that happens.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -307,6 +318,31 @@ public boolean spawnEntity(Entity entity) {
...
+     * @author Simon816
+     *
+     * Same as original method except an added check for instanceof HumanEntity
+     * because human entities should not count as a player.
+     */
+    @Overwrite
+    public boolean spawnEntityInWorld(net.minecraft.entity.Entity entityIn) {
+        int cx = MathHelper.floor_double(entityIn.posX / 16.0D);
+        int cz = MathHelper.floor_double(entityIn.posZ / 16.0D);
+        boolean force = entityIn.forceSpawn || (entityIn instanceof EntityPlayer && !(entityIn instanceof HumanEntity));
```
Zidane: Same thing here as with Sponge. Overwrite shouldn't be needed.

PR #25, opened 2015-05-07
Implement Rcon API. Fixes SpongePowered/Sponge#254.

zml2008:
Implementation of SpongePowered/SpongeAPI#650.

I am unsure of whether the locations the events are being called from are the best -- especially the login event. Should that be called before authentication? How would we go about sending any text sent to the command source back to the client (flush the buffer?)

Also this fixes the issue of no line breaks in rcon output, which is kinda nice

### DIFF
--- a/src/main/java/org/spongepowered/common/Sponge.java
+++ b/src/main/java/org/spongepowered/common/Sponge.java
@@ -29,12 +29,29 @@ import static com.google.common.base.Preconditions.checkState;
 import static org.spongepowered.common.configuration.SpongeConfig.Type.GLOBAL;
 
 import com.google.inject.Injector;
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.server.dedicated.DedicatedServer;
 import org.apache.logging.log4j.Logger;
 import org.spongepowered.api.Game;
+import org.spongepowered.api.Platform;
 import org.spongepowered.api.plugin.PluginContainer;
+import org.spongepowered.api.service.ProviderExistsException;
+import org.spongepowered.api.service.command.CommandService;
+import org.spongepowered.api.service.command.SimpleCommandService;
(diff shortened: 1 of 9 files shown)


src/main/java/org/spongepowered/common/mixin/core/network/rcon/MixinRConConsoleSource.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param component
+     */
+    @Inject(method = "addChatMessage", at = @At("RETURN"))
+    public void addNewlines(IChatComponent component, CallbackInfo ci) {
+        this.buffer.append('\n');
+    }
+
+    @Override
+    public String getName() {
+        return getCommandSenderName();
```
simon816: should qualify `this` at all times, I recommend  making your IDE automatically do it.

zml2008: not for methods though?

Minecrell: Ye, it's not required for methods so this is correct

src/main/java/org/spongepowered/common/Sponge.java
```diff
@@ -74,6 +91,43 @@ public Sponge(Injector injector, Game game, Logger logger,
 
         this.plugin = checkNotNull(plugin, "plugin");
         this.minecraftPlugin = checkNotNull(minecraftPlugin, "minecraftPlugin");
+
```
Minecrell: It's not a good idea to put this directly in the constructor. In SpongeVanilla for example the Sponge class will be constructed even before the `MinecraftServer` gets called so this will just cause issues if something references the Minecraft server here in any way. Better place this in something like a `load()` method so the implementation can control when this is called.

PR #28, opened 2015-05-09
Implemented PlayerMoveEvent

thomas15v:


gabizou:
Did you test this?

thomas15v:
I tested it off course :smile:.
I tested the following:
  - Canceling the event. result: Makes it impossible to move.
  - Changing the location. result: teleports you to the new location.
  - Changing the rotation. result: whoops. fixing that now.
note: I am unsure about the injection point. Its like 1 line lower than where I wanted it to be. Not that it changes much to the behaviour.

src/main/java/org/spongepowered/common/mixin/core/server/MixinNetHandlerPlayServer.java
```diff
@@ -187,4 +193,32 @@ public void processCommandBlock(C17PacketCustomPayload packetIn, CallbackInfo ci
 
     }
 
+    @Inject(method = "processPlayer", at = @At(value = "FIELD", target = "net/minecraft/network/NetHandlerPlayServer.hasMoved", ordinal = 2), cancellable = true, remap = false)
```
simon816: I don't think `remap` should equal `false` here. (Correct me if I'm wrong)

thomas15v: I honestly don't understand the remap thing of mixins. Only thing I understood of it, is that it will make updating sponge easier.

simon816: No, read the javadoc on the `@Inject` annotation. remap is whether it should remap the method/class/field names in production (an obfuscated environment). It needs to convert the friendly names to SRG names.

thomas15v: Oh. But why does the mixins library fails to find that [field](https://travis-ci.org/SpongePowered/SpongeCommon/jobs/61914584#L468)? 
I checked it and HasMoved isn't a [forge](https://github.com/MinecraftForge/MinecraftForge/blob/f0d96b7c50cb1aceb101c5e9c0d78806b87e1240/patches/minecraft/net/minecraft/network/NetHandlerPlayServer.java.patch) thingy :worried:.

Deamon5550: I believe your reference to the field should be dot separated not slash separated and it needs to include the type information. eg.

`net.minecraft.network.NetHandlerPlayServer.hasMoved:Z`

Your method reference also needs to include more information see [this](https://github.com/SpongePowered/SpongeCommon/blob/206575ac29e767c6645353101303e320fe360ae6/src/main/java/org/spongepowered/common/mixin/core/entity/vehicle/MixinEntityBoat.java#L76) for an example.

modwizcode:
Shouldn't changing the rotation teleport the player with rotation?

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -239,4 +243,31 @@ public String dontNormalizeStringBecauseSomeMojangDevSucks(String input) {
...
+    @Inject(method = "processPlayer", at = @At(value = "FIELD", target = "net.minecraft.network.NetHandlerPlayServer.hasMoved:Z", ordinal = 2), cancellable = true)
+    public void proccesPlayerMoved(C03PacketPlayer packetIn, CallbackInfo ci){
+        if (packetIn.isMoving() || packetIn.getRotating()) {
+            Player player = (Player) playerEntity;
+            Vector3d fromrot = player.getRotation();
+            Location from = player.getLocation();
+            Vector3d torot = new Vector3d(packetIn.getYaw(), packetIn.getPitch(), 0);
+            Location to = new Location(player.getWorld(), packetIn.getPositionX(), packetIn.getPositionY(), packetIn.getPositionZ());
+
+            double delta = Math.pow(from.getX() - to.getX(), 2) + Math.pow(from.getX() - to.getY(), 2) + Math.pow(from.getX() - to.getZ(), 2);
```
simon816: `double delta = from.sub(to).lengthSquared()` would be neater
Am I right in thinking this should be square rooted?
If so, `double delta = from.sub(to).length()`

kenzierocks: The delta is only internal so not square rooting would be ok.

ST-DDT: > `Math.pow(from.getX() - to.getY(), 2)`

Why `X1 - Y2` ?

`double delta = from.distanceSquared(to)` is even shorter and does not create a new `Vector3d` object.

DDoS: The delta is fine squared as long as the variables it is compared to also are squared. This saves a square root operation. I would strongly recommend calling it `deltaSquared` and having `1f / 256` written as `(1f / 16) * (1f / 16)` to make the constant clearer.

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -239,4 +243,31 @@ public String dontNormalizeStringBecauseSomeMojangDevSucks(String input) {
...
+            float deltaAngle = Math.abs((float) (fromrot.getX() - torot.getX()) + Math.abs((float) (fromrot.getY() - torot.getY())));
+
+            if ((delta > 1f / 256 || deltaAngle > 10f) && !playerEntity.isDead) {
```
modwizcode: What are these magic numbers?

PR #35, opened 2015-05-15
Fixing sync scheduler not firing after the appropriate delay

### LINKED ISSUE #33: SyncScheduler has incorrect delay compensation
When using the SynchronousScheduler#runTaskAfter method (implemented by SyncScheduler), the delay given by the command in ticks is sometimes not used. For example, when using this code in a command:
```java
final long st1 = System.currentTimeMillis();
// Start a timeout schedule
plugin.getGame().getSyncScheduler().runTaskAfter(plugin, new Runnable() {
    public void run() {
        source.sendMessage(Texts.of(TextColors.AQUA, (System.currentTimeMillis() - st1)));
    }
}, 200);
```
Usually a server averages about 20tps, so this delay would be 10 seconds, however this value returns 51 milliseconds.

Doing a similar test with 2000 ticks (which is usually about 100 seconds) gives about 90000 milliseconds, which is closer to the estimated value.

**EDIT: After some testing, it's been determined that this issue only persists inside of a command. Running this during ServerStartingEvent or otherwise yields expected values.**

Sponge dev version: 1.8-1371-2.1DEV-443
Forge version: 1.8-11.14.1.1371

RobertHerhold:
When a task is scheduled inside of a command executor, tasks do not get run after the appropriate delay. This PR fixes that by setting the timestamp at the appropriate time when the task is created. This fixes #33 (please read for complete details)

If you want to reproduce, checkout [my fork of SpongeCommon](https://github.com/RobertHerhold/SpongeCommon) and build it. I tested this with [BLWarps](https://github.com/BlockLaunch/BLWarps/releases/tag/v1.0.0), my warping plugin. To reproduce:
* Use both the BLWarps plugin and my fork of SpongeCommon
* In game, make a warp: "/warp add mywarp"
* Warp to the newly created warp: "/warp mywarp". After about 5 seconds, you will be warped.
* If you want proof, just run the same thing without my fork of SpongeCommon and compare results. When running "/warp mywarp", you will be warped immediately rather than after 5 seconds.

This is the same as SpongePowered/Sponge#225. However, the reviewers tested this in a different environment than my plugin (not scheduling a task from a command), and I do take responsibility for not having detailed reproduction procedures

### DIFF
--- a/src/main/java/org/spongepowered/common/service/scheduler/SyncScheduler.java
+++ b/src/main/java/org/spongepowered/common/service/scheduler/SyncScheduler.java
@@ -252,6 +252,7 @@ public class SyncScheduler implements SynchronousScheduler {
         if (nonRepeatingTask == null) {
             Sponge.getLogger().warn(SchedulerLogMessages.CANNOT_MAKE_TASK_WARNING);
         } else {
+            nonRepeatingTask.setTimestamp(this.counter);
             resultTask = this.schedulerHelper.utilityForAddingTask(this.taskMap, nonRepeatingTask);
         }
 
@@ -306,6 +307,7 @@ public class SyncScheduler implements SynchronousScheduler {
         if (nonRepeatingTask == null) {
             Sponge.getLogger().warn(SchedulerLogMessages.CANNOT_MAKE_TASK_WARNING);
         } else {
+            nonRepeatingTask.setTimestamp(this.counter);
             resultTask = this.schedulerHelper.utilityForAddingTask(this.taskMap, nonRepeatingTask);
         }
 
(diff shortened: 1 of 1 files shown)


Grinch:
I can confirm that this resolves the issue. Receiving consistent and accurate timing with these changes.

simon816:
I have incorporated this fix into my scheduler cleanup PR.
When adding a task, the timestamp it set, see [`SchedulerBase#addTask`](https://github.com/SpongePowered/SpongeCommon/pull/47/files#diff-168ace8114caeb27d02b38cad72bc18eR73)

Zidane:
We'll be rolling this up into #47 . Thanks for working with us to point to the exact issue.

PR #37, opened 2015-05-17
Fix all cases where entities would crash when creating or spawning

simon816:
This fixes crashes that occur when trying to create an entity, or spawn an entity directly after creation.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
@@ -35,12 +35,27 @@ import com.google.common.base.Optional;
 import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableList;
 import net.minecraft.block.state.IBlockState;
+import net.minecraft.entity.EntityHanging;
+import net.minecraft.entity.boss.EntityDragonPart;
+import net.minecraft.entity.effect.EntityLightningBolt;
+import net.minecraft.entity.item.EntityArmorStand;
+import net.minecraft.entity.item.EntityEnderPearl;
+import net.minecraft.entity.item.EntityFallingBlock;
+import net.minecraft.entity.item.EntityItem;
+import net.minecraft.entity.item.EntityPainting;
+import net.minecraft.entity.item.EntityPainting.EnumArt;
+import net.minecraft.entity.player.EntityPlayerMP;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -277,28 +288,41 @@ public boolean isBlockPowered(int x, int y, int z) {
...
+            return Optional.absent();
+        }
+
+        net.minecraft.world.World world = (net.minecraft.world.World) (Object) this;
+
+        // Not all entities have a single World parameter as their constructor
+        if (entityClass.isAssignableFrom(EntityLightningBolt.class)) {
+            entity = (Entity) new EntityLightningBolt(world, position.getX(), position.getY(), position.getZ());
+        } else if (entityClass.isAssignableFrom(EntityEnderPearl.class)) {
+            EntityArmorStand tempEntity = new EntityArmorStand(world, position.getX(), position.getY(), position.getZ());
```
Zidane: While hilarious, unsure if we want an armorstand to be the projectile source of an EnderPearl when spawned this way :P.

simon816: I meant to unset this after the constructor and have an `UnknownProjectileSource` as the default. I'll fix this now.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -277,28 +294,71 @@ public boolean isBlockPowered(int x, int y, int z) {
...
+            return Optional.absent();
+        }
+
+        net.minecraft.world.World world = (net.minecraft.world.World) (Object) this;
+
+        // Not all entities have a single World parameter as their constructor
+        if (entityClass.isAssignableFrom(EntityLightningBolt.class)) {
+            entity = (Entity) new EntityLightningBolt(world, x, y, z);
+        } else if (entityClass.isAssignableFrom(EntityEnderPearl.class)) {
+            EntityArmorStand tempEntity = new EntityArmorStand(world, x, y, z);
```
Zidane: Still meant for this to be EntityArmorStand?

simon816: Yes, the constructor must take an `EntityLivingBase` and I think that an armor stand is the cheapest object to construct.

PR #42, opened 2015-05-19
[WIP] A few extra events being called

me4502:
This PR is adding a few extra event calls.

* BlockRandomTickEvent
* EntityLeashEvent
* EntityUnleashEvent
* LeafDecayEvent
* EntityBreedEvent


src/main/java/org/spongepowered/common/mixin/core/block/MixinBlock.java
```diff
@@ -119,6 +128,15 @@ public void setTickRandomly(boolean tickRandomly) {
         this.needsRandomTick = tickRandomly;
     }
 
+    @Overwrite
+    public void randomTick(World worldIn, BlockPos pos, IBlockState state, Random random) {
+        BlockRandomTickEvent event = SpongeEventFactory.createBlockRandomTick(Sponge.getGame(), null, new Location((Extent)worldIn, VecHelper.toVector(pos)));
```
Zidane: Need to provide a reasoning for this Overwrite or, better yet, do an inject.

src/main/java/org/spongepowered/common/mixin/core/entity/living/MixinEntityLiving.java
```diff
@@ -79,6 +92,43 @@ public void setCanPickupItems(boolean canPickupItems) {
         this.canPickUpLoot = canPickupItems;
     }
 
+    @Overwrite
+    public final boolean interactFirst(EntityPlayer playerIn) {
```
Zidane: Need to provide a reason for this overwrite or, better yet, do an inject.

src/main/java/org/spongepowered/common/mixin/core/block/MixinBlock.java
me4502: The entire method here is rewritten, is an @Inject capable of adding a return to the injected method? If so, I could do a head inject

simon816: Make it `cancellable` in the `@Inject`, a `CallbackInfo` object in the parameter can then be used to return by calling `cancel()` on it.

me4502:
I'm getting a LinkageError with this, not sure why exactly - I'm assuming guava isn't correctly being loaded into the LaunchWrapper classloader.

Here is the error,

http://pastebin.com/UebkAxCm

Paging @Mumfrey and @Zidane for assistance.

simon816:
@Me4502 I believe that is the same as this issue: https://github.com/SpongePowered/Sponge/issues/266#issuecomment-1034035
I think @Mumfrey is right in thinking it is [this class loader](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/util/event/factory/ClassGeneratorProvider.java#L100-L109) that is causing the problem.

src/main/java/org/spongepowered/common/mixin/core/block/MixinBlock.java
```diff
@@ -119,6 +127,15 @@ public void setTickRandomly(boolean tickRandomly) {
         this.needsRandomTick = tickRandomly;
     }
 
+    @Inject(method = "randomTick", at = @At(value = "HEAD"), locals = LocalCapture.CAPTURE_FAILEXCEPTION, cancellable = true)
+    public void callRandomTickEvent(World world, BlockPos pos, IBlockState state, Random rand, CallbackInfo ci) {
+        BlockRandomTickEvent event = SpongeEventFactory.createBlockRandomTick(Sponge.getGame(), null, new Location((Extent)world, VecHelper.toVector(pos)));
```
Zidane: Mark this as final and add a TODO to fix cause

src/main/java/org/spongepowered/common/mixin/core/block/MixinBlockLeaves.java
```diff
@@ -0,0 +1,63 @@
...
+@NonnullByDefault
+@Mixin(BlockLeaves.class)
+public abstract class MixinBlockLeaves extends MixinBlock {
+
+    @Inject(method = "updateTick", at = @At(value = "INVOKE", target = "destroy(Lnet/minecraft/world/World;Lnet/minecraft/util/BlockPos;)V"), locals = LocalCapture.CAPTURE_FAILEXCEPTION, cancellable = true)
+    public void callLeafDecay(World worldIn, BlockPos pos, IBlockState state, Random rand, CallbackInfo ci, byte b0, int i, int j, int k, int l, byte b1, int i1, int j1, int k1) {
+        Location block = new Location((Extent) worldIn, VecHelper.toVector(pos));
+        BlockSnapshot postChange = block.getSnapshot();
+        postChange.setBlockState(BlockTypes.AIR.getDefaultState());
+        LeafDecayEvent event = SpongeEventFactory.createLeafDecay(Sponge.getGame(), null, block, postChange);
```
Zidane: See comment above

me4502:
Should be all good now.

PR #43, opened 2015-05-20
Implement ClientConnectEvent

simon816:
Implementation of https://github.com/SpongePowered/SpongeAPI/pull/663

Tested and works as expected

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/server/network/MixinNetHandlerLoginServerAnonThread.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/server/network/MixinNetHandlerLoginServerAnonThread.java
@@ -1,0 +1,52 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/common/mixin/core/server/network/MixinNetHandlerLoginServerAnonThread.java
```diff
@@ -0,0 +1,50 @@
...
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import org.spongepowered.common.interfaces.IMixinNetHandlerLoginServer;
+
+@Mixin(targets = "net/minecraft/server/network/NetHandlerLoginServer$2")
+public class MixinNetHandlerLoginServerAnonThread extends Thread {
+
+    @Shadow private NetHandlerLoginServer this$0;
```
simon816: Where can I find the SRG name for this synthetic field? @Mumfrey?

JBYoshi: I'm pretty sure you need to find the obfuscated class name, then look it up by the field name `a`. `a` is always mapped to the first field, and `this$0` is always the first field in the class, at least according to `javac`.

simon816: @JBYoshi Didn't work unfortunately.
The notch name for `NetHandlerLoginServer$2` is `rs`, but the field `rs.a` doesn't exist

JBYoshi: It should be `field_151286_b`. Look up the field in the notch/srg mappings.

JBYoshi: Okay, not so sure about that. Further searching makes it look like it's supposed to be `field_180221_a`.

JBYoshi: I confused `NetHandlerLoginServer` and `NetHandlerPlayServer`. It is `field_180221_a`.

simon816: Yep that works. It's interesting that that field does not appear in MCPBot searches.

PR #46, opened 2015-05-24
Make Human an EntityCreature

simon816:
As suggested by @Zidane, the human entity could be an `EntityCreature` (server side). The advantage is that it is possible to control it with AI (anything extending `EntityLiving` has support for AI).

The downside is that the human won't have things that you would assume a player has.
While going through `EntityPlayer` to find things that should be included, I formulated this list of things to consider:
* Inventory (main and armor)
* Sleeping
* XP
* Food
* Capabilities
 - invulnerable
 - flying
 - etc.
* Sneaking
* Sprinting
* keepInventory
* Drop items
* Dig blocks

These things are programmed in `EntityPlayer` but  the logic will need to be copied to `HumanEntity` if they are to be possible.
I have overridden some methods (e.g. `getSwimSound`, `getPortalCooldown`) to replicate `EntityPlayer` behaviour. The above list is of things that are more difficult (and a lot more copy/pasting) so they need to be discussed before I go ahead and add them.


Zidane:
@simon816 

I went ahead and pushed this up to a branch 

https://github.com/SpongePowered/SpongeCommon/tree/feature/entity/human

We'll continue work there.

simon816:
Should this be left open for discussion on the above topics? Or maybe that could be discussed in the API issue https://github.com/SpongePowered/SpongeAPI/issues/502

PR #47, opened 2015-05-24
Cleanup the two schedulers

### LINKED ISSUE #33: SyncScheduler has incorrect delay compensation

### LINKED ISSUE #36: Tasks will not stop if canceled during run()
Both Async and Sync Schedulers use the same similar code to process every task, every run.

```java
boolean bTaskStarted = startTask(task);
if (bTaskStarted) {
    task.setState(ScheduledTask.ScheduledTaskState.RUNNING);
```
startTask() ends up calling run() on the task as expected but the task is always set to the RUNNING state once completed, which means if task.cancel() is called during run() of the task, it will be ignored and set back to RUNNING, and never stop.

### LINKED ISSUE #67: Scheduled tasks don't throw exception output!
I don't know if it happens everywhere. 
But I runned a task with 
```getAsyncScheduler().runRepeatingTask()``` And the runnable had an exception. And I spended 2 hours figuring out why the code didn't worked.

### PR DESCRIPTION
**Pull request by simon816**
Remove duplication. Tidy things up. Removed javadocs that are in the API.
Made it easier to understand (for me at least).
Fixes #33
Closes #35
Fixes #36 
Fixes #67
(At least I hope it fixes them :) )
Fixed searching tasks by name, it was comparing the given name to the owner plugin name, not the task name.
Also fixed a bug I discovered when scheduling multiple async tasks at the same time. Basically the time the thread slept for was calculated incorrectly. Now works fine.

Tested and can confirm it schedules tasks correctly, test script available here: http://pastebin.com/XP9X9zhc

With the new design where there's only one scheduler, I have made that scheduler hold instances of the `SyncScheduler` and `AsyncScheduler`. Either one is selected using a `delegate` method depending on the parameter passed in. It then continues to add the task in the scheduler as before.

### CODE REVIEW

File: src/main/java/org/spongepowered/common/service/scheduler/AsyncScheduler.java [hunk 1]
```diff
@@ -194,520 +113,73 @@ private void recalibrateMinimumTimeout() {
...
-            resultTask = Optional.of((Task) task);
-        } finally {
-            this.lock.unlock();
+            this.condition.await(this.minimumTimeout, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+            // The taskMap has been modified; there is work to do.
+            // Continue on without handling the Exception.
+        } catch (IllegalMonitorStateException e) {
+            Sponge.getLogger().error("The scheduler internal state machine suffered a catastrophic error.  Check #spongedev");
+            Sponge.getLogger().error(e.toString());
```
Minecrell: Make sure to give the exception to the logger above, then you can leave out this line. Also a better error message of what exactly happened would be good

src/main/java/org/spongepowered/common/service/scheduler/SchedulerBase.java
```diff
@@ -0,0 +1,260 @@
...
+     * Begin the execution of a task. Exceptions are caught and logged.
+     *
+     * @param task The task to start
+     */
+    protected void startTask(ScheduledTask task) {
+        task.setState(ScheduledTask.ScheduledTaskState.RUNNING);
+        try {
+            this.executeTaskRunnable(task.getRunnable());
+        } catch (Exception ex) {
+            Sponge.getLogger().error("The Scheduler tried to run the Task, but the Runnable could not be started.");
```
Minecrell: Log just one message here and add the exception directly. Also manage the plugin and task name etc when reporting the error, similar to this for example: https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/event/SpongeEventManager.java#L250

src/main/java/org/spongepowered/common/service/scheduler/AsyncScheduler.java
```diff
@@ -194,520 +113,73 @@ private void recalibrateMinimumTimeout() {
...
-
-        if (repeatingTask == null) {
-            Sponge.getLogger().warn(SchedulerLogMessages.CANNOT_MAKE_TASK_WARNING);
-        } else {
-            resultTask = utilityForAddingAsyncTask(repeatingTask);
-        }
-
-        return resultTask;
+    protected void executeTaskRunnable(Runnable runnable) {
+        this.executor.submit(runnable);
```
Minecrell: This doesn't handle errors correctly because the task isn't executed in this thread and this will (possibly) return before it was executed

simon816: I have compensated for this by wrapping the runnable in another runnable, https://github.com/SpongePowered/SpongeCommon/pull/47/files#diff-168ace8114caeb27d02b38cad72bc18eR245

simon816:
I believe I have tested this enough to say that it works as expected. The code used to test is available here: http://pastebin.com/JzAPAgEv

src/main/java/org/spongepowered/common/service/scheduler/AsyncScheduler.java
```diff
@@ -40,64 +39,10 @@
...
- * getAsyncScheduler(); Plugin authors never cause the creation of a scheduler,
- * the scheduler (Asynchronous and {@link SyncScheduler} are each Singletons.
- * </p>
- *
- *
- */
-public class AsyncScheduler implements AsynchronousScheduler {
+public class AsyncScheduler extends SchedulerBase implements AsynchronousScheduler {
+
+    public static final AsyncScheduler INSTANCE = new AsyncScheduler();
```
gratimax: Generally speaking, practice dictates to not deal with the instance field in a singleton directly but expose it through a `getInstance()` method, which is what was done previously.

src/main/java/org/spongepowered/common/service/scheduler/AsyncScheduler.java
```diff
@@ -106,19 +51,20 @@
...
                 stateMachineBody();
             }
-        }).start();
+        });
+        thread.setName("Sponge Async Scheduler Thread");
+        thread.setDaemon(true);
+        thread.start();
     }
 
     private void stateMachineBody() {
```
gratimax: Since your changes, this name has lost a lot of its meaning and is probably best being called a loop of some sort

src/main/java/org/spongepowered/common/service/scheduler/SchedulerBase.java
```diff
@@ -0,0 +1,265 @@
...
+            if (!matcher.matches()) {
+                it.remove();
+            }
+        }
+
+        return matchingTasks;
+    }
+
+    @Override
+    public Set<Task> getScheduledTasks() {
```
gratimax: You can make the return type here more specific so that you don't to cast to SchedulerTask elsewhere

simon816: I would have done, except this is implementing a method in `SchedulerQuery `

src/main/java/org/spongepowered/common/service/scheduler/AsyncScheduler.java
simon816: meh. Because it's `final` I don't see any reason to need a getter, I often prefer field access over methods. But I'll change it back to `getInstance` anyway.

ST-DDT: IIRC I heart something like singletons are bad (for important things like testing). 
Source: Google Dev Talk.
I can search you the YT link if you want.

Zidane:
@simon816 

Have you tested this and its good to go?

src/main/java/org/spongepowered/common/service/scheduler/AsyncScheduler.java
```diff
@@ -106,608 +51,139 @@
...
-            this.condition.signalAll();
-            resultTask = Optional.of((Task) task);
-        } finally {
-            this.lock.unlock();
+            this.condition.await(this.minimumTimeout, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException ignored) {
+            // The taskMap has been modified; there is work to do.
+            // Continue on without handling the Exception.
+        } catch (IllegalMonitorStateException e) {
+            Sponge.getLogger().error("The scheduler internal state machine suffered a catastrophic error. Check #spongedev", e);
```
Zidane: No need to put #spongedev ...

src/main/java/org/spongepowered/common/service/scheduler/SchedulerBase.java
```diff
@@ -0,0 +1,265 @@
...
+
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+abstract class SchedulerBase implements SchedulerQuery {
```
Zidane: In the API, SchedulerQuery should be renamed SchedulerService.

Zidane:
Another thing I talked with @Minecrell about was the fact that the API has the two Scheduler interfaces yet I don't see why a plugin should know to make the distinction.

Instead, this seems like the perfect candidate for a TaskBuilder of sorts that can build your task and submit it to the Scheduler.

```
final Task task = this.game.getTaskBuilder()
    .async(true)
    .execute(...)
    .delay(TimeUnit.MILISECONDS, 10)
    .repeatEvery(TimeUnit.MILISECONDS, 10)
    .build()

this.game.getScheduler().submit(this, task);
```

@SpongePowered/developers Thoughts?

Grinch:
@Zidane Your example looks pretty sane and clean for use.

kashike:
@Zidane Your example looks good, but I would reverse the position of `TimeUnit`.
```java
final Task task = this.game.getTaskBuilder()
    .async(true)
    .execute(new Runnable() {})
    .delay(10, TimeUnit.MILISECONDS)
    .repeatEvery(10, TimeUnit.MILISECONDS)
    .side(Platform.Side.SERVER) // Once Sponge is on the client, this can determine if to run on client or not
    .build()

this.game.getScheduler().submit(this, task);
```

src/main/java/org/spongepowered/common/service/scheduler/AsyncScheduler.java
```diff
@@ -106,608 +51,139 @@
     private final Condition condition = this.lock.newCondition();
     // The dynamic thread pooling executor of asynchronous tasks.
     private ExecutorService executor;
-    // Query actor for task information
-    private SchedulerHelper schedulerHelper;
 
     private AsyncScheduler() {
```
ST-DDT: Does this need to be private?
Since it is async it does not really matter how much instances you create and use at the same time, does it? Sure Sponge itself only needs one but there is a difference between having one and enforcing a singleton.

simon816: It will make no difference to plugins whether it's a singleton or not and I don't see a problem with it being one. There's no use-case for more than one instance from what I can see.

Zidane:
```
final Task task = this.game.getTaskBuilder()
    .async()
    .execute(new Runnable() {})
    .delay(10, TimeUnit.MILISECONDS)
    .repeatEvery(10, TimeUnit.MILISECONDS)
    .side(Platform.Side.SERVER) // Once Sponge is on the client, this can determine if to run on client or not
    .build()

this.game.getScheduler().submit(this, task);
```

I removed the setting of the boolean value and instead made it a toggle to on. I also included Kashike's suggestion.

Maybe roll this up into an API Scheduler cleanup @simon816 ?

simon816: Going to update https://github.com/SpongePowered/SpongeAPI/pull/697, therefore this PR will be on hold until the scheduler API is finalized.

simon816:
I have updated this to the new design. A new test script is available here: http://pastebin.com/XP9X9zhc

src/main/java/org/spongepowered/common/service/scheduler/SpongeTaskBuilder.java
```diff
@@ -0,0 +1,110 @@
...
+
+    public SpongeTaskBuilder() {
+        this.syncType = ScheduledTask.TaskSynchroncity.SYNCHRONOUS;
+    }
+
+    @Override
+    public TaskBuilder async() {
+        this.syncType = ScheduledTask.TaskSynchroncity.ASYNCHRONOUS;
+        if (this.tickDelay != -1) {
+            this.delay = this.tickDelay * 50;
```
Zidane: 50 means...what? Magic numbers are always bad.

src/main/java/org/spongepowered/common/service/scheduler/SchedulerBase.java
```diff
@@ -0,0 +1,195 @@
...
+
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+
+abstract class SchedulerBase {
+
+    // The simple queue of all pending (and running) ScheduledTasks
+    private final Map<UUID, ScheduledTask> taskMap = new ConcurrentHashMap<UUID, ScheduledTask>();
```
Zidane: Could do Maps.newConcurrentHashMap() here. Just nitpicking :p.

src/main/java/org/spongepowered/common/service/scheduler/SpongeScheduler.java
```diff
@@ -0,0 +1,168 @@
...
+    private SchedulerBase getDelegate(Task task) {
+        if (task.isAsynchronous()) {
+            return this.asyncScheduler;
+        } else {
+            return this.syncScheduler;
+        }
+    }
+
+    private SchedulerBase getDelegate(ScheduledTask.TaskSynchroncity syncType) {
+        switch (syncType) {
```
Zidane: This is a case of form form.

If syncType == ASYNCHRONOUS return that scheduler else return sync one. No reason to even let this be another value.

src/main/java/org/spongepowered/common/service/scheduler/SchedulerBase.java
simon816: Technically `new ConcurrentHashMap()` defaults to a `concurrencyLevel` of 16 whereas `Maps.newConcurrentMap()` defaults to 4 but I doubt that will make a difference.

Deamon5550: Well if you really fell you need that level of control then use [MapMaker](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/MapMaker.html) (which is what `Maps.newConcurrentHashMap()` actually calls anyway).

Zidane: @Deamon5550 

Be on TS in a hr, got some questions for ya :p

me4502:
This is looking good. From my initial testing, it appears to make the schedulers more stable than they already are.

randombyte-developer:
I would be very happy if someone could merge this or is there something to wait for? https://github.com/SpongePowered/SpongeAPI/pull/697 is ok I think.

PR #51, opened 2015-05-28
Implement pagination service.

zml2008:
Fixes SpongePowered/SpongeAPI#547.

This implements the pagination service, providing a simple method for plugins to provide paginated output. This implementation provides some useful features like text width calculation based on MC's font's character widths, and clickable text objects for next/previous page.

Screenshot:
pagination

Commands added:

- `/page [id] next` - Print next page 
- `/page [id] prev` - Print previous page (Not supported by Iterable-type paginations)
- `/page [id] cur` - Reprint current page
- `/page [id] page <number>` - Print the page with a specific number (not supported by Iterable-type paginations)


src/main/java/org/spongepowered/common/service/pagination/SpongePaginationService.java
```diff
@@ -0,0 +1,218 @@
...
+        public UUID getLastUuid() {
+            return this.lastUuid;
+        }
+    }
+    final ConcurrentMap<Class<? extends CommandSource>, PaginationCalculator<?>> calculators = new ConcurrentHashMap<Class<? extends
+            CommandSource>, PaginationCalculator<?>>();
+    final ConcurrentMap<CommandSource, SourcePaginations> activePaginations = new MapMaker().weakKeys().makeMap();
+
+    public SpongePaginationService() {
+        Sponge.getGame().getCommandDispatcher().register(Sponge.getPlugin(), CommandSpec.builder()
```
kashike: Is it possible to __not__ register the command if the pagination service is not used? The API doesn't allow removing services (or changing the provider) for some stupid reason, and removing commands isn't exposed without casting.

PR #57, opened 2015-05-30
Implemented the luminescence functions in the context of a Sponge World

DarkArc:


src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -610,6 +617,16 @@ public Context getContext() {
     }
 
     @Override
+    public byte getLuminanceFromSky(int x, int y, int z) {
+        return (byte) Math.max(0, getLightFor(EnumSkyBlock.SKY, new BlockPos(x, y, z)) - getSkylightSubtracted());
```
DarkArc: NMS lacks a sever side accessible function to fetch the light level, with the time of day/weather factored in, thus this must be subtracted, and a floor is imposed to prevent negative light levels.

src/main/java/org/spongepowered/common/mixin/core/world/extent/MixinExtent.java
```diff
@@ -163,6 +163,26 @@ public Location getFullBlock(int x, int y, int z) {
     }
 
     @Override
+    public byte getLuminance(Vector3i position) {
+        return getLuminance(position.getX(), position.getY(), position.getZ());
+    }
+
+    @Override
+    public byte getLuminance(int x, int y, int z) {
+        return (byte) Math.max(getLuminanceFromGround(x, y, z), getLuminanceFromSky(x, y, z));
```
DarkArc: Similarly, NMS lacks a sufficient server side function to provide the light level present at a block, thus, the higher of the two values is used, as per Minecraft's current behavior.

DarkArc:
While this is implemented as a byte currently per the current API, it seems like it would be best if the API were changed to use a more flexible int in the case of Mojang adding a greater range of light levels.

gabizou:
@DarkArc I don't mind changing light to be an int vs a byte. @SpongePowered/developers agree?

me4502:
I don't see why not to.

DarkArc:
RIP, whiped it out with a bad push GG PR.

DarkArc:
Intellij's local history saved the day...

DarkArc:
Now depends on [SpongeAPI/723](https://github.com/SpongePowered/SpongeAPI/pull/723)

PR #58, opened 2015-05-31
Implement XML text format

zml2008:
Implementation of [XML text format](https://github.com/SpongePowered/SpongeAPI/pull/708), mostly provided by TomyLobo

### Remaining implementation issues
- [ ] Preserving whitespace
- [x] Error handling -- squash everything?

### DIFF
--- a/src/main/java/org/spongepowered/common/text/xml/A.java
+++ b/src/main/java/org/spongepowered/common/text/xml/A.java
@@ -22,45 +22,29 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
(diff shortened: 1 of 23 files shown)


src/main/java/org/spongepowered/common/text/xml/A.java
```diff
@@ -0,0 +1,50 @@
...
+import org.spongepowered.api.text.action.TextActions;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlRootElement
+public class A extends Element {
```
simon816: I think that these miniature classes (`A`, `B`, `I`, ...) could be contained within a class `Elements` as `public static class`es because they are too small IMO for a separate class each.

PR #64, opened 2015-06-04
Fix leaf

jamierocks:
/me <3 leaf.

But seriously why did no one do this?!

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/block/MixinBlockLeaves.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/block/MixinBlockLeaves.java
@@ -67,7 +67,7 @@ import java.util.Random;
 @Mixin(BlockLeaves.class)
 public abstract class MixinBlockLeaves extends MixinBlock implements IMixinBlockTree {
 
-    @Inject(method = "updateTick", at = @At(value = "INVOKE", target = "net/minecraft/block/BlockLeaves.destroy (Lnet/minecraft/world/World;Lnet/minecraft/util/BlockPos;)V"), locals = LocalCapture.CAPTURE_FAILEXCEPTION, cancellable = true)
+    @Inject(method = "updateTick", at = @At(value = "INVOKE", target = "net/minecraft/block/BlockLeaves;destroy (Lnet/minecraft/world/World;Lnet/minecraft/util/BlockPos;)V"), locals = LocalCapture.CAPTURE_FAILEXCEPTION, cancellable = true)
     public void callLeafDecay(World worldIn, BlockPos pos, IBlockState state, Random rand, CallbackInfo ci, byte b0, int i, int j, int k, int l, byte b1, int i1, int j1, int k1) {
         Location block = new Location((Extent) worldIn, VecHelper.toVector(pos));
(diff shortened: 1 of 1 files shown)


modwizcode:
Sorry was going to do it right now actually :P

Zidane:
Cause no one has messed with this besides the person who originally pushed this.

Mumfrey:
@jamierocks the new format is wrong as well :/

jamierocks:
@modwizcode Feel free to close this, and commit yourself if you like :P

simon816:
There's two more cases of this a wrong injection target (introduced by 97029e3cbd9a6eaa003e24e0b9ad35951a9d3ef1)
[`MixinEntityLiving.callLeashEvent`](https://github.com/SpongePowered/SpongeCommon/blob/97029e3cbd9a6eaa003e24e0b9ad35951a9d3ef1/src/main/java/org/spongepowered/common/mixin/core/entity/living/MixinEntityLiving.java#L95) and [`MixinEntityLiving.callUnleashEvent`](https://github.com/SpongePowered/SpongeCommon/blob/97029e3cbd9a6eaa003e24e0b9ad35951a9d3ef1/src/main/java/org/spongepowered/common/mixin/core/entity/living/MixinEntityLiving.java#L104)

me4502:
Hmm... When I was originally testing them, using ; caused them to not be called at all. After switching it to ., it started working.

modwizcode:
Working on this now. It's not as simple as I hoped.

Mumfrey:
@me4502 `MemberInfo` supports all these formats:

```php
    // Bare class name in bytecode descriptor style
    Lpackage/name/here/ClassName;

    // Bare class name with method/field name
    Lpackage/name/here/ClassName;memberName

    // Bare class name with method name and descriptor
    Lpackage/name/here/ClassName;memberName(IIJJZZ)V

    // Bare class name with field name and descriptor
    Lpackage/name/here/ClassName;memberName:Z

    // Bare class name in unqoted style
    package/name/here/ClassName

    // Bare class name with method/field name
    package/name/here/ClassName.memberName

    // Bare class name with method name and descriptor
    package/name/here/ClassName.memberName(IIJJZZ)V

    // Bare class name with field name and descriptor
    package/name/here/ClassName.memberName:Z

    // Friendly class name in human-readable style
    package.name.here.ClassName

    // Friendly class name with method/field name
    package.name.here.ClassName.memberName

    // Friendly class name with method name and descriptor
    package.name.here.ClassName.memberName(IIJJZZ)V

    // Friendly class name with field name and descriptor
    package.name.here.ClassName.memberName:Z
    
    // Unqualified member name
    memberName

    // Unqualifier method name with descriptor
    memberName(III)V

    // Unqualified field name with descriptor
    memberName:Z
```

So you see, the semicolon isn't a descriptor, it's the end mark for the `L` bytecode-style descriptor

me4502:
Ah.. So which one should I be using for Sponge?

Mumfrey:
Whatever you feel like. Idiomatically we tend to use the 3rd variant for fully-qualified references, mainly because it's just easiest when you're copying the descriptor from the bytecode outline plugin view.

PR #68, opened 2015-06-13
Introduce multi-world support to the entire Sponge ecosystem.

Zidane:
This graduates up bloodmc's amazing work in Sponge to SpongeCommon so
that SpongeVanilla will have this capability as well. I've also cleaned up
a few spots such as removing some overwrites.

Fix END and NETHER GeneratorTypes not generating a Vanilla The_End or
Nether respectively.

Fix Vanilla DIM-1 and DIM1 not creating the correct config folder on first
generation.

Fix sending incorrect dimension to client (fixes teleports), fix teleport
to world with same client dimension with Vanilla client. Fix generator
modifiers not being applied in WorldBuilder

Signed-off-by: Chris Sanders <zidane@outlook.com>

### DIFF
--- a/src/main/java/org/spongepowered/common/interfaces/IMixinEntity.java
+++ b/src/main/java/org/spongepowered/common/interfaces/IMixinEntity.java
@@ -26,7 +26,6 @@ package org.spongepowered.common.interfaces;
 
 import net.minecraft.entity.Entity;
 import net.minecraft.nbt.NBTTagCompound;
-import org.spongepowered.api.world.Location;
 
 public interface IMixinEntity {
 
@@ -48,9 +47,9 @@ public interface IMixinEntity {
 
     void inactiveTick();
 
-    NBTTagCompound getSpongeData();
+    NBTTagCompound getEntityData();
 
-    boolean teleportEntity(Entity entity, Location location, int currentDim, int targetDim, boolean forced);
+    NBTTagCompound getSpongeData();
 
     /**
      * Read extra data (SpongeData) from the entity's NBT tag.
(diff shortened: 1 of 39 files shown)


src/main/java/org/spongepowered/common/interfaces/IMixinEntity.java
```diff
@@ -48,9 +47,9 @@
 
     void inactiveTick();
 
-    NBTTagCompound getSpongeData();
+    NBTTagCompound getEntityData();
```
simon816: Unused method

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -326,6 +326,11 @@ public void reset() {
     }
 
     @Override
+    public boolean usesCustomClient() {
+        return false;
```
simon816: What if a forge client connects to the vanilla server? Should this remain false?

src/main/java/org/spongepowered/common/mixin/core/server/MixinServerConfigurationManager.java
```diff
@@ -24,48 +24,367 @@
...
+        {
+            chatcomponenttranslation = new ChatComponentTranslation("multiplayer.player.joined", new Object[] {playerIn.getDisplayName()});
+        }
+
+        chatcomponenttranslation.getChatStyle().setColor(EnumChatFormatting.YELLOW);
+
+        // Fire PlayerJoinEvent
+        final PlayerJoinEvent event = SpongeImplEventFactory.createPlayerJoin(Sponge.getGame(), (Player) playerIn, ((Player) playerIn).getLocation(),
+                SpongeTexts.toText(chatcomponenttranslation), ((Player) playerIn).getMessageSink());
+        Sponge.getGame().getEventManager().post(event);
```
simon816: Pretty sure this won't post to Forge's event bus. Will need to look into that in `SpongeModEventManager`

Zidane: @simon816 

@Aaron1011 is going to look into triggering them on the Forge event bus when this is called.

Otherwise, this is an easy fix to simply fire the event in the creation method and return it. We can still check cancellation status if needed.

PR #69, opened 2015-06-17
Implementation of FurnaceConsumFuelEvent

Piratdu52:
Actually i can't test this because furnace.getData throw an abstract method error,  so in the state is more like a WIP don't hesitate to provide advice and require modifications.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/block/tiles/MixinTileEntityFurnace.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/block/tiles/MixinTileEntityFurnace.java
@@ -22,23 +22,38 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
+
 package org.spongepowered.common.mixin.core.block.tiles;
 
 import static org.spongepowered.api.data.DataQuery.of;
 
+import net.minecraft.tileentity.TileEntityFurnace;
 import org.spongepowered.api.block.tileentity.TileEntityType;
 import org.spongepowered.api.block.tileentity.TileEntityTypes;
 import org.spongepowered.api.block.tileentity.carrier.Furnace;
 import org.spongepowered.api.data.DataContainer;
+import org.spongepowered.api.data.manipulator.tileentity.FurnaceData;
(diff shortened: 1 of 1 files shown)


Grinch:
You may want to format your code a bit. As it stands, it is extremely difficult to read.

Piratdu52:
I have checked your guidelines here, https://docs.spongepowered.org/en/contributing/guidelines.html and all seem okay. Maybe you want more vertical space inside my method ? If not please hilight an issue.

twizmwazin:
The spacing is inconsistent. You have an extra space at line 96, and are omitting spaces around operators and curly brackets everywhere. I'm not sure if I'm missing something here, but why are you casting to Object before casting to the correct class in lines 84 and 85?

Piratdu52:
Ok, I see. Will edit that later.

The (Object) is here for compilation purpose because I cast a net.micraft ItemStack to Sponge ItemStack or TileEntityFurnace to Furnace. Without I can't compile.

ST-DDT:
Try using the auto formatter from your Ide.
In the extra folder (from the API) there are formatter files you can import and use.

src/main/java/org/spongepowered/common/mixin/core/block/tiles/MixinTileEntityFurnace.java
```diff
@@ -56,4 +71,51 @@ public DataContainer toContainer() {
...
+                    remainingFuel= event.getRemainingFuel().get();
+                    remainingFuel.setQuantity(remainingFuel.getQuantity()+1); //add one because The mojang class will decrement this
+                }
+                else {
+                    ItemStackBuilder itemStackBuilder= Sponge.getGame().getRegistry().getItemBuilder();
+                    remainingFuel= itemStackBuilder.itemType(ItemTypes.SPONGE).quantity(1).build(); //set  a arbitrary itemStack with quantity=1,
+                                                                                                    // the Mojang class will delete it
+                }
+                this.furnaceItemStacks[1] = (net.minecraft.item.ItemStack) remainingFuel;
+                return;
```
JBYoshi: Did you mean for this to return from update()? If so, set `@Inject(cancellable=true)` and replace this with `ci.cancel()`.

Zidane:
This event isn't correct anymore due to new events system.

PR #72, opened 2015-06-22
Fix mixin problems.

JBYoshi:
- There are some explosive mixins that were written but not registered. I have registered them and removed a method that was implemented in the explosive mixin.
- There were two duplicate EnumArt mixins. I have removed the one in mixin.core.entity.hanging.art (which also wasn't registered).


src/main/java/org/spongepowered/common/mixin/core/entity/vehicle/minecart/MixinEntityMinecartTNT.java
```diff
@@ -33,9 +32,4 @@
 @NonnullByDefault
 @Mixin(EntityMinecartTNT.class)
 public abstract class MixinEntityMinecartTNT extends MixinEntityMinecart implements MinecartTNT {
-
-    @Override
-    public FuseData getFuseData() {
```
Zidane: What was the reasoning for removing this?

Zidane: ignore comment above.

Zidane:
Thanks for this @JBYoshi :).

JBYoshi:
You're welcome.

PR #74, opened 2015-06-26
Implement ignoring players when skipping the night

caseif:
This PR implements [SpongeAPI #735](https://github.com/SpongePowered/SpongeAPI/pull/735), which adds API methods for ignoring certain `Humans` when determining whether the night should be skipped due to all players being asleep.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
@@ -24,16 +24,20 @@
  */
 package org.spongepowered.common.mixin.core.world;
 
+import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.scoreboard.ScoreboardSaveData;
 import net.minecraft.util.BlockPos;
 import net.minecraft.village.VillageCollection;
 import net.minecraft.world.World;
 import net.minecraft.world.WorldServer;
 import net.minecraft.world.WorldSettings;
+import org.spongepowered.api.entity.living.Human;
 import org.spongepowered.api.util.annotation.NonnullByDefault;
 import org.spongepowered.api.world.GeneratorType;
 import org.spongepowered.api.world.GeneratorTypes;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Overwrite;
+import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.Surrogate;
@@ -43,9 +47,14 @@ import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
 import org.spongepowered.common.interfaces.IMixinScoreboardSaveData;
 import org.spongepowered.common.interfaces.IMixinWorld;
 
+import java.util.Iterator;
+import java.util.List;
+
 @NonnullByDefault
 @Mixin(WorldServer.class)
 public abstract class MixinWorldServer extends MixinWorld {
+
+    @Shadow private boolean allPlayersSleeping;
 
     @Inject(method = "createSpawnPosition(Lnet/minecraft/world/WorldSettings;)V", at = @At("HEAD"), cancellable = true)
     public void onCreateSpawnPosition(WorldSettings settings, CallbackInfo ci) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -75,4 +84,38 @@ public void onPostInit(CallbackInfoReturnable<World> ci) {
         IMixinWorld world = (IMixinWorld) ci.getReturnValue();
         world.updateWorldGenerator();
     }
+
+    @Overwrite
+    public void updateAllPlayersSleepingFlag() {
+        this.allPlayersSleeping = false;
+        if (!this.playerEntities.isEmpty())
+        {
```
me4502: This should be on the previous line.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -75,4 +84,37 @@ public void onPostInit(CallbackInfoReturnable<World> ci) {
...
+
+    @Overwrite
+    public void updateAllPlayersSleepingFlag() {
+        this.allPlayersSleeping = false;
+        if (!this.playerEntities.isEmpty()) {
+            int ignoredPlayers = 0;
+            int sleepingPlayers = 0;
+            for (EntityPlayer entityPlayer : this.playerEntities) {
+                if (entityPlayer.isSpectator()
+                        || (entityPlayer instanceof Human && ((Human)entityPlayer).isSleepingIgnored())) {
```
simon816: Humans are no longer types of `EntityPlayer`. They are their own entity type, [`EntityHuman`](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/entity/living/human/EntityHuman.java#L74) which extends `EntityCreature`. But as I said in your API PR the sleep flag is only for players.

caseif:
@simon816 My reasoning here was that `EntityPlayerMP` both extends `EntityPlayer` and implements `Human`, and would therefore be picked up by this comparison. However, given the comment on the API PR, I'll change it accordingly.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
caseif: @simon816 My reasoning here was that `EntityPlayerMP` both extends `EntityPlayer` and implements `Human`, and would therefore be picked up by this comparison. However, given the comment on the API PR, I'll change it accordingly.

modwizcode:
Can you bring this up to date again?

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -75,4 +84,37 @@ public void onPostInit(CallbackInfoReturnable<World> ci) {
         IMixinWorld world = (IMixinWorld) ci.getReturnValue();
         world.updateWorldGenerator();
     }
+
+    @Overwrite
+    public void updateAllPlayersSleepingFlag() {
```
modwizcode: Can be solved with an @Inject

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -75,4 +84,37 @@ public void onPostInit(CallbackInfoReturnable<World> ci) {
...
+                    && (sleepingPlayers >= this.playerEntities.size() - ignoredPlayers));
+        }
+    }
+
+    @Overwrite
+    public boolean areAllPlayersAsleep() {
+        if ((this.allPlayersSleeping) && (!this.isRemote)) {
+            for (EntityPlayer entityPlayer : this.playerEntities) {
+                boolean ignore = entityPlayer instanceof Player && ((Player)entityPlayer).isSleepingIgnored();
+                if (!ignore && (entityPlayer.isSpectator() || !entityPlayer.isPlayerFullyAsleep())) {
```
modwizcode: Just use `@Redirect` on the `isPlayerFullyAsleep()` call to check if the player's sleeping status is ignored. That might not be a valid override for every call to that method, but in this case it will be able to do the same task.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -75,4 +84,37 @@ public void onPostInit(CallbackInfoReturnable<World> ci) {
...
+
+    @Overwrite
+    public void updateAllPlayersSleepingFlag() {
+        this.allPlayersSleeping = false;
+        if (!this.playerEntities.isEmpty()) {
+            int ignoredPlayers = 0;
+            int sleepingPlayers = 0;
+            for (EntityPlayer entityPlayer : this.playerEntities) {
+                if (entityPlayer.isSpectator()
+                        || (entityPlayer instanceof Player && ((Player)entityPlayer).isSleepingIgnored())) {
```
modwizcode: @SpongePowered/developers is this instanceof a valid check or the correct way to check this? Also I have a feeling that overriding the specific `isSpectator` call might be even better than an `@Inject` but it also seems a little bit like a hack.

simon816: The instanceof looks fine

ryantheleach:
I could have sworn it was recorded on github somewhere, but there was a discussion with amaranth in IRC that an event may have been more appropriate then giving players sleeping ignored status.

It might still be useful to throw an event.

modwizcode:
A few conversations were had about that. And again the dialogue never made
it to GitHub. It's should just be the event.

Zidane:
@bloodmc 

This ought to be the event instead and this likely reverted.

PR #80, opened 2015-07-02
Expand the Command API. Implements SpongePowered/SpongeAPI#744

JBYoshi:
This pull request:
- Implements SpongePowered/SpongeAPI#744
- Adds `WrapperCommandSource` to provide a `CommandSource` for mod-provided `ICommandSender`s
- Adds a /help command as Sponge uses its own command registering system which NMS doesn't know about
- Adds `IMixinCommandSender` and `IMixinCommandSource` to allow separation of the `ICommandSender` and `CommandSource` objects (for example, `EntityMinecartCommandBlock` is the `CommandSource`, but an inner class is the `ICommandSender`)
- Extracts names of inner classes used in commands to constants, to reduce work when updating MC versions
- Adds `sponge.debug.hover-stacktrace` support to exceptions in `ICommand`s
- Update permissions, descriptions, and help for NMS commands
- Move code from multiple command source mixins into shared mixins

~~Note: When I was implementing `PaginationBuilder.page()`, I found that it was easier to simply remove `IterablePagination` and just move `ListPagination` into `ActivePagination`. Although it takes a bit more memory, it is easier both to code and to use.~~ Removed after feedback from zml.


src/main/java/org/spongepowered/common/SpongeBootstrap.java
```diff
@@ -73,6 +75,7 @@ public static void initializeServices() {
         } catch (ProviderExistsException e) {
             Sponge.getLogger().warn("Non-Sponge CommandService already registered: " + e.getLocalizedMessage());
         }
+        Sponge.getGame().getCommandDispatcher().register(Sponge.getPlugin(), new SpongeHelpCommand(), "help");
```
Zidane: This could be registered below where the Sponge command is.

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "SpongeAPI"]
 	path = SpongeAPI
-	url = https://github.com/SpongePowered/SpongeAPI.git
+	url = https://github.com/JBYoshi/SpongeAPI.git
```
Zidane: Looks like you accidentally pushed a module change here.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -245,6 +228,10 @@ public PlayerConnection getConnection() {
         return (PlayerConnection) this.playerNetServerHandler;
     }
 
+    public void setBedLocation(@Nullable Location location) {
+        this.spawnChunk = location != null ? VecHelper.toBlockPos(location.getPosition()) : null;
```
Zidane: This seems to be out of the scope of this PR?

Zidane:
Besides the comments I've made, you've impressed me which few actually do. Nice job, definitely could use a few more guys like you who clearly understand our implementation.

With that out of the way, more questions:
* Got a picture of the Help command in action?
* Any reason why the Help command was written as a CommandCallable? Could it use CommandSpec?

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
JBYoshi: It was. Not sure how it got there. Should I remove it?

.gitmodules
JBYoshi: So I don't change those when I'm doing a pull request? Sorry. This is my first pull request with submodules, so I wasn't sure how to do that.

Zidane: Just leave them out. Its not a big deal and easily fixable by whoever merges this but its easily missed and then we have to hurry and make a commit before everyone and their brother ends up pulling your fork XD.

JBYoshi: :) Fixed.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
JBYoshi: Due to the fact that it's not in the AT anymore (but it would need to be) I'll remove it.

JBYoshi:
@Zidane I actually forgot about `CommandSpec` when writing the help command. I've added it on my local repo, but I still need to test it. Something seems to be ignoring the access transformer's `EntityList.classToStringMapping`, leaving it private. Once it's fixed and tested, I'll push the commit and send a picture.

> Besides the comments I've made, you've impressed me which few actually do. Nice job, definitely could use a few more guys like you who clearly understand our implementation.

Thanks.

src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
```diff
@@ -133,41 +135,50 @@ public PluginContainer getOwner() {
...
     public Text getUsage(CommandSource source) {
-        final ICommandSender mcSender =
-                source instanceof ICommandSender ? (ICommandSender) source : new WrapperICommandSender(source);
+        final ICommandSender mcSender = WrapperICommandSender.of(source);
         String usage = this.command.getCommandUsage(mcSender);
-        if (usage.startsWith("/") && usage.contains(" ")) {
-            usage = usage.substring(usage.indexOf(" "));
+        Translation translation = Sponge.getGame().getRegistry().getTranslationById(usage).get();
+        if (source instanceof Player) {
+            usage = translation.get(((Player) source).getLocale());
```
zml2008: can't do this we don't control all CommandSources.

src/main/java/org/spongepowered/common/command/SpongeHelpCommand.java
```diff
@@ -0,0 +1,151 @@
...
+    private static final Comparator<CommandMapping> COMMAND_COMPARATOR = new Comparator<CommandMapping>() {
+
+        @Override
+        public int compare(CommandMapping o1, CommandMapping o2) {
+            return o1.getPrimaryAlias().compareTo(o2.getPrimaryAlias());
+        }
+    };
+
+    @Override
+    public CommandResult process(final CommandSource source, String arguments) throws CommandException {
```
zml2008: we have a command API that does arg parsing for you. use it

src/main/java/org/spongepowered/common/command/SpongeHelpCommand.java
```diff
@@ -0,0 +1,151 @@
...
+
+        if (Sponge.getGame().getCommandDispatcher().containsAlias(arguments)) {
+            source.sendMessage(getDescription(source, Sponge.getGame().getCommandDispatcher().get(arguments).get()));
+            return CommandResult.success();
+        }
+
+        int page = 1;
+        if (!arguments.isEmpty()) {
+            try {
+                page = Integer.parseInt(arguments);
```
zml2008: the pagination system is designed to eliminate int page arguments

src/main/java/org/spongepowered/common/command/SpongeHelpCommand.java
```diff
@@ -0,0 +1,151 @@
...
+            } catch (NumberFormatException e) {
+                throw new CommandException(Texts.of("No such command: " + arguments));
+            }
+        }
+
+        PaginationBuilder builder = Sponge.getGame().getServiceManager().provide(PaginationService.class).get().builder();
+        builder.header(Texts.builder("Available commands:").color(TextColors.DARK_GREEN).build());
```
zml2008: .title()

src/main/java/org/spongepowered/common/command/SpongeHelpCommand.java
JBYoshi: Zidane mentioned that. I'm working on it, but I haven't pushed it since I haven't tested it yet.

src/main/java/org/spongepowered/common/command/SpongeHelpCommand.java
JBYoshi: @zml2008 This is meant for compatibility with vanilla. Vanilla has page arguments in /help.

src/main/java/org/spongepowered/common/command/SpongeHelpCommand.java
JBYoshi: @zml2008 Got it. Will be coming with the CommandSpec commit.

src/main/java/org/spongepowered/common/command/SpongeHelpCommand.java
zml2008: well that usage is silly (and very much an implementation leak in the API), the `/page next` command exists for a reason.

src/main/java/org/spongepowered/common/mixin/core/command/MixinCommandExecuteAtSender.java
```diff
@@ -35,106 +39,132 @@
...
+@Mixin(targets = IMixinCommandSender.EXECUTE_COMMAND_SENDER)
 @NonnullByDefault
-public abstract class MixinCommandExecuteAtSender implements CommandSource, ICommandSender {
-    @Shadow(aliases = "val$sender")
-    private ICommandSender field_174802_b;
+public abstract class MixinCommandExecuteAtSender implements ProxySource, IMixinCommandSource, IMixinCommandSender {
+
+    @Shadow(aliases = "val$entity") private net.minecraft.entity.Entity field_174804_a;
+    @Shadow(aliases = "val$sender") private ICommandSender field_174802_b;
+    private MessageSink sink = Sponge.getGame().getServer().getBroadcastSink();
```
zml2008: seems like it makes more sense to get this from the source -- why have a separate one here? (or not at least default to the source's current sink)?

src/main/java/org/spongepowered/common/service/pagination/ActivePagination.java
```diff
@@ -39,42 +40,36 @@
...
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.UUID;
 
 /**
  * Holds logic for an active pagination that is occurring.
  */
-abstract class ActivePagination {
+class ActivePagination {
```
zml2008: Nope.

The point of having separate pagination types is so that plugins wanting to send really long lists to players don't have to store everything in memory. not getting rid of that

additionally, the changes to next/previous page texts get rid of nice unicode arrows are not needed

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationService.java
```diff
@@ -102,36 +102,14 @@ void registerCommandOnce() {
         if (this.commandRegistered.compareAndSet(false, true)) {
             Sponge.getGame().getCommandDispatcher().register(Sponge.getPlugin(), CommandSpec.builder()
                     .description(t("Helper command for paginations occurring"))
-                    .arguments(new ActivePaginationCommandElement(t("pagination-id")))
```
zml2008: This reduces usability for console and other environments where the clickable text is not available.

src/main/java/org/spongepowered/common/command/SpongeHelpCommand.java
kashike: @zml2008 I find it very helpful to give a page number if I want to.

zml2008: `/page [num]` command exists for that

src/main/java/org/spongepowered/common/command/WrapperCommandSource.java
```diff
@@ -0,0 +1,171 @@
...
+        if (sender instanceof IMixinCommandSender) {
+            return ((IMixinCommandSender) sender).asCommandSource();
+        }
+        if (sender instanceof WrapperICommandSender) {
+            return ((WrapperICommandSender) sender).source;
+        }
+        if (sender.getCommandSenderEntity() != null) {
+            return new WithEntity(sender);
+        }
+        if (!new Vec3(0, 0, 0).equals(sender.getPositionVector())) {
```
zml2008: extract this into a constant?

src/main/java/org/spongepowered/common/mixin/core/command/MixinCommandHandler.java
```diff
@@ -0,0 +1,53 @@
...
+import org.spongepowered.common.command.WrapperCommandSource;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
+@Mixin(CommandHandler.class)
+public abstract class MixinCommandHandler {
+
+    @Inject(method = "tryExecute", at = @At(value = "INVOKE", target = "Lnet/minecraft/command/ICommandSender;addChatComponent(Lnet/minecraft/util/IChatComponent)V;"))
+    public void onCommandError(ICommandSender sender, String[] args, ICommand command, String input, Throwable error, ChatComponentTranslation comp) {
```
zml2008: hm, would it be possible to throw an InvocationCommandException or so in this method? that would let us reduce code duplication.

src/main/java/org/spongepowered/common/mixin/core/command/MixinICommandSender.java
```diff
@@ -0,0 +1,56 @@
...
+import net.minecraft.server.MinecraftServer;
+import org.spongepowered.api.util.command.CommandCallable;
+import org.spongepowered.api.util.command.CommandMapping;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.common.Sponge;
+import org.spongepowered.common.command.MinecraftCommandWrapper;
+import org.spongepowered.common.interfaces.IMixinCommandSender;
+
+@Mixin(value = {EntityPlayerMP.class, MinecraftServer.class, RConConsoleSource.class, CommandBlockLogic.class}, targets = IMixinCommandSender.SIGN_CLICK_SENDER)
+public abstract class MixinICommandSender implements ICommandSender, IMixinCommandSender {
```
zml2008: this seems to clean up a lot of the rough edges between MC and Sponge CommandSources very nicely. However, it adds a bunch of various interfaces. Would you be able to add a package-info.java that explains which interfaces&mixins are responsible for which components? would make it easier for others to get up to speed on this bit of the impl which does some complicated wrapping.

src/main/java/org/spongepowered/common/mixin/core/command/MixinCommandExecuteAtSender.java
JBYoshi: This was a bit confusing. Delegating to the source sounds good.

src/main/java/org/spongepowered/common/service/pagination/ActivePagination.java
JBYoshi: @zml2008 Unfortunately, the Unicode arrows didn't work when I tried it. In addition, using an Iterable makes the per-page stuff harder.

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationService.java
JBYoshi: @zml2008 Could you please explain?

src/main/java/org/spongepowered/common/mixin/core/command/MixinCommandHandler.java
JBYoshi: This would require some fairly complex code to interact between MinecraftCommandWrapper and MixinCommandHandler. Not sure how to do that.

src/main/java/org/spongepowered/common/mixin/core/command/MixinICommandSender.java
JBYoshi: Sounds good. I'll take care of that tomorrow.

src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
JBYoshi: This is a bit complex. The translation maps to something like "/give <player> <item> [amount]", but the method should return "<player> <item> [amount]". Hacky, but this is the best I can do.

src/main/java/org/spongepowered/common/command/SpongeHelpCommand.java
JBYoshi: Just a question: what about people who don't know about the /page command?

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationService.java
JBYoshi: @zml2008 What do you mean?

src/main/java/org/spongepowered/common/command/WrapperCommandSource.java
JBYoshi: Done, will be pushing tomorrow.

JBYoshi:
Thanks for the feedback, guys. I will be working more on this tomorrow.

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationService.java
zml2008: Requires providing a specific page number rather than allowing `/page next` or `/page prev` to go to previous or next page as was possible before.

src/main/java/org/spongepowered/common/command/SpongeHelpCommand.java
ST-DDT: @zml2008
Shall the user type `/help` and then `/page 5` instead of just `/help 5`?
Im pretty sure a lot of plugins will offer a 'fix' for this unexpected behaviour. Especially for console like interfaces without click support.

src/main/java/org/spongepowered/common/mixin/core/command/MixinCommandHandler.java
JBYoshi: Actually, I just got an idea on how to do it. Working on it.

JBYoshi: Wait, no. The Minecraft commands can throw multiple exceptions as tryExecute can be called multiple times (see MinecraftCommandWrapper), but the Sponge commands are only called once and therefore cannot throw multiple exceptions. Never mind.

src/main/java/org/spongepowered/common/mixin/core/command/MixinICommandSender.java
JBYoshi: Pushed.

src/main/java/org/spongepowered/common/command/WrapperCommandSource.java
```diff
@@ -0,0 +1,171 @@
...
+import org.spongepowered.common.util.VecHelper;
+
+public class WrapperCommandSource extends SpongeSubject implements CommandSource {
+
+    final ICommandSender sender;
+    private final MemorySubjectData data;
+
+    private WrapperCommandSource(ICommandSender sender) {
+        this.sender = sender;
+        this.data = new MemorySubjectData(Sponge.getGame().getServiceManager().provide(PermissionService.class).get());
```
zml2008: I think it would be preferable to delegate to permissions plugins for this if possible (but have to figure out how to do identifiers, and an appropriate subject type (maybe `minecraft`?)).

src/main/java/org/spongepowered/common/service/pagination/ActivePagination.java
```diff
@@ -94,16 +95,19 @@ public UUID getId() {
...
+
     protected abstract boolean hasPrevious(int page);
+
     protected abstract boolean hasNext(int page);
+
     protected abstract int getTotalPages();
 
     public void nextPage() throws CommandException {
-        specificPage(++this.currentPage);
+        specificPage(this.currentPage + 1);
```
zml2008: ? this doesn't seem to advance currentPage at all -- so keeps on looping the same page.

src/main/java/org/spongepowered/common/mixin/core/command/MixinCommandHandler.java
zml2008: well for Throwables, I think it makes sense to abort execution entirely. A non-CommandException indicates a programming error, not a handled failure -- so things that should never happen in a production server (well, this is mojang, so ehhhh) and that warrant some extra 'lets put on the breaks now ok', which would stop execution. 
I'm planning on adding to the error handling to add error report generation, would be nice to get that with MC commands automatically too.

src/main/java/org/spongepowered/common/service/pagination/ActivePagination.java
JBYoshi: @zml2008 It gets set back onto currentPage inside specificPage(). See line 130.

src/main/java/org/spongepowered/common/mixin/core/command/MixinCommandHandler.java
JBYoshi: Done.

src/main/java/org/spongepowered/common/command/WrapperCommandSource.java
JBYoshi: @zml2008 Yeah, identifiers for these are pretty tough.

src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
zml2008: Yeah -- I think the best solution to this might just be to add a getLocale method to CommandSource -- it's something I've wanted to do for a while and the idea keeps sounding better.

Issue #23, opened 2015-05-04
Ongoing Minor Issue List


Deamon5550:
__Minor Issues List__

- [ ] Discover Minor Issues
- [ ] Build link in README should be linking to the SpongeCommon build, not the Sponge build.

__Checkstyle warnings__

File | Problem
---- | ----
Unqualified instance access | [MixinChunk:154](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/mixin/core/world/MixinChunk.java#L154)
Unqualified instance access | [MixinChunk:206](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/mixin/core/world/MixinChunk.java#L206)
Unqualified instance access | [MixinWorldType:117](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldType.java#L117)
Unqualified instance access | [SpongeGame:113](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/SpongeGame.java#L113)
Statement unnecessarily nested within else clause | [AbstractTileBuilder:143](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/data/builders/block/tile/AbstractTileBuilder.java#L143)
Statement unnecessarily nested within else clause | [ChatComponentIterable:46](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/text/ChatComponentIterable.java#L46)
Statement unnecessarily nested within else clause | [CustomWorldChunkManager:132](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/world/gen/CustomWorldChunkManager.java#L132)
Statement unnecessarily nested within else clause | [DataUtil:36](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/data/utils/DataUtil.java#L36)
Statement unnecessarily nested within else clause | [MinecraftCommandWrapper:137](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java#L137)
[...]

PR #81, opened 2015-07-03
[minor] Fixed two generics related warnings (seen in Eclipse)

vorburger:


Deamon5550:
Could you write this as a comment on #23 rather than a PR, thanks.

twizmwazin:
@Deamon5550 I don't see the reason to close a pull request immediately and tell them to report it as an issue for someone else to fix. That seems silly and redundant.

ST-DDT:
@twizmwazin 
Every merge requires the team to checkout the fix and test whether it actually does what it should. 
For small none bug fixes this is a really heavy time overhead.
To save the team that time the OCD list exists.

twizmwazin:
@ST-DDT That kinda defeats a large idea behind an open source, community-driven project. Rather than encouraging community members to contribute you point them to an complaint box.  
This being a trivial thing, the only thing I feel is really necessary is check if the warning can be fixed rather than suppressed.

Deamon5550:
@twizmwazin I really don't understand your comment about this not being open-source. If you want to submit a PR for minor issues then by all means submit a PR for minor issues, you can get the entries of the OCD list, the couple hundred checkstyle warnings thrown by gradle, and I'm sure if you just start looking you'll find even more places where the code style could use some cleanup or a comment has bad grammar.

However I'll save you right now and suggest that you do not do this as you will have a nightmare of rebasing and merge conflicts to handle every time someone pushes something to master before your PR is accepted and a nightmare for everyone else having to rebase their branches across your changes once they are in.

This is why we have the minor issue list so that we can track these small fixes and fix them during a freeze of the repository where we don't have to worry about having to rebase anything (usually just before version releases).

twizmwazin:
@Deamon5550 I'm not saying that the project isn't open source, as it clearly and very much is. Although a basic concept behind open source is the idea of community contributions to a project, you seem to promote making an issue for a collaborator to fix rather than allow a community member to actually provide a fix. In the case of something as small as this, I would think double checking and merging this would take just about as long as actually writing the fix yourself. But since he has already made the commit and opened the pull request, why not just double check and merge the commit now rather than put it off for someone else to do again down the road?

Zidane:
@twizmwazin 

We only do mass fixes for "OCD Problems" during codebase freezes. That way we don't have to rebase someone's PR they did for fixes during times of potentially high commit volume.

Its far easier to dedicate a team member to do one massive fell swoop of this vs. getting a community cleanup pr mergable in the midst of other PRs/contributions.

PR #84, opened 2015-07-03
Implement PermissionDescription API

ST-DDT:
Adds https://github.com/SpongePowered/SpongeAPI/issues/601 (low priority)
Implements https://github.com/SpongePowered/SpongeAPI/pull/747 (description + discussion)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "SpongeAPI"]
 	path = SpongeAPI
-	url = https://github.com/SpongePowered/SpongeAPI.git
+	url = https://github.com/ST-DDT/SpongeAPI.git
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "SpongeAPI"]
 	path = SpongeAPI
-	url = https://github.com/SpongePowered/SpongeAPI.git
+	url = https://github.com/ST-DDT/SpongeAPI.git
```
ryantheleach: This probably shouldn't be included in the PR.

ST-DDT: Otherwise the automated PR build/test support does not work.

kenzierocks: To make it easier you can do it like I did with my Selectors PR, I kept the code in one commit and the API ref/module changes in another commit. This way the team can manipulate it easier.

ST-DDT: Good advice. I'll change this later today.

PR #85, opened 2015-07-06
[WIP] Added implementation of additional world management API's

modwizcode:
Implementation of https://github.com/SpongePowered/SpongeAPI/pull/751

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java
@@ -42,6 +42,7 @@ import net.minecraft.world.storage.ISaveHandler;
 import net.minecraft.world.storage.WorldInfo;
 import net.minecraftforge.fml.relauncher.Side;
 import net.minecraftforge.fml.relauncher.SideOnly;
+import org.apache.commons.io.FileUtils;
 import org.apache.logging.log4j.Logger;
 import org.spongepowered.api.Platform;
 import org.spongepowered.api.Server;
@@ -64,10 +65,7 @@ import org.spongepowered.asm.mixin.Overwrite;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.common.Sponge;
 import org.spongepowered.common.event.SpongeImplEventFactory;
-import org.spongepowered.common.interfaces.IMixinMinecraftServer;
-import org.spongepowered.common.interfaces.IMixinSubject;
-import org.spongepowered.common.interfaces.IMixinWorldInfo;
-import org.spongepowered.common.interfaces.IMixinWorldProvider;
+import org.spongepowered.common.interfaces.*;
 import org.spongepowered.common.text.SpongeTexts;
 import org.spongepowered.common.text.sink.SpongeMessageSinkFactory;
 import org.spongepowered.common.world.DimensionManager;
@@ -75,6 +73,7 @@ import org.spongepowered.common.world.SpongeDimensionType;
 import org.spongepowered.common.world.storage.SpongeChunkLayout;
 
 import java.io.File;
+import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java
```diff
@@ -64,10 +65,7 @@
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.common.Sponge;
 import org.spongepowered.common.event.SpongeImplEventFactory;
-import org.spongepowered.common.interfaces.IMixinMinecraftServer;
-import org.spongepowered.common.interfaces.IMixinSubject;
-import org.spongepowered.common.interfaces.IMixinWorldInfo;
-import org.spongepowered.common.interfaces.IMixinWorldProvider;
+import org.spongepowered.common.interfaces.*;
```
JBYoshi: I don't think the checkstyle allows wildcard imports.

modwizcode: No it doesn't that's an automatic change from intellij I'll have to clean
that up.

src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java
```diff
@@ -531,6 +530,33 @@ protected void prepareSpawnArea(WorldServer world) {
     }
 
     @Override
+    public Optional<World> copyWorld(World originalWorld, String copyName) {
+        try {
+            FileUtils.copyDirectory(getWorldDir(originalWorld.getName()), getWorldDir(copyName));
```
JBYoshi: If someone changes the world immediately before cloning it, the changes won't be saved to disk. You may want to save the world before doing this.

modwizcode: It will have to be unloaded first is the best solution to that. I can't
guarantee no modifications between flush to disk and copy otherwise.

JBYoshi: My suggestion:
1. Save the world to disk.
2. Disable the world saving.
3. Copy the world.
4. Re-enable world saving.

modwizcode: No because players could modify the world in-between, it wouldn't be saved,
I might add that as a separate copy but it shouldn't be the default kind,
because the most likely case I see is duplication of a minigame world to
clean up after use. And in that case you're going to want players out of
the world anyway.

ryantheleach: Wouldn't the changes since the last auto-save be a larger set of changes on average then the changes between an autosave that you triggered immediately before copying?

modwizcode: Yes but having an inconsistent api seems a bit strange.

ST-DDT: Maybe we should just mention that in the API docs.
`This method does not save any pending changes. If you want to copy the uptodate world consider saving it first.`

JBYoshi: Also, this wouldn't work very well for the "root" world (dimension 0) as it copies all the data, not just the current world.

modwizcode: Might wanna hold of commenting on this till I push the more proper
implementation. This was a sample that worked well enough but wasn't really
thought about in a huge amount of detail.

ryantheleach: Actually I think we can do better. Looking at how blocks are marked as dirty for clients to receive the new block changes, world.worldAcesses is a list of observers of the world. If you are doing asynchronous copying creating and registering an IWorldAccess would allow you to receive any updates that happened since the copy was started.

That said I'm not sure which way would be better give the plugin requesting a copy an exact copy at the time it ended, a partial possibly incomplete and out of sync copy.

Neither of those really seem ideal when they more then likely want a copy at the time that the API was called. But having a complete sync'd copy at the time that the future returns seems better then having an out of sync copy that could cause item/entity duplication.

modwizcode:
Because this issue could take awhile to work out, I'm going to split the offline import functionality into a separate PR that can be merged quicker.

gabizou:
@bloodmc I know this isn't quite the highest priority, but, thoughts on this PR as it stands?

modwizcode:
I may have time to clean this up if given feedback.

gabizou:
@modwizcode If you want, could you rebase the PR off master as there are conflicts needing to be resolved?

modwizcode:
Sure thing. Be nice to get back into some Sponge work after being so busy.

bloodmc:
@modwizcode Copying worlds need to be done async. There is no reason why this should run off the main thread.

modwizcode:
Agreed. I'll start making some slow progress on updating this. Gotta
reestablish myself in the codebase.

simon816:
Oh didn't realize there was a PR. Well I went ahead and implemented these methods in https://github.com/SpongePowered/SpongeCommon/commit/8db71e6bd71262f83628981be3726ab06601d950 so this can be closed now.

PR #86, opened 2015-07-06
Implement simulated methods using a fake player, controlled by a PlayerSimulator

simon816:
**SpongeCommon** | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/473) 

A few methods in the API for `Extent` describe method that simulate the act of a player doing something i.e. interacting and digging.
Due to how vanilla calculates the result of these actions, the only feasible solution I have come up with is to use an actor; a fake player.
This is heavily influenced from Forge's FakePlayer, but is designed to only be used internally.

There are some potential memory leak considerations when using the actor, as well as the performance.
The player object has a reference to the world, so this needs to be removed after usage to stop the world from staying in memory if unloaded.
One way to overcome this is to create a new instance of `SpongeActor` every time it's needed. But the construction is expensive so I'd like to avoid doing this.

SpongeActor could be a singleton, which would work at the moment, but if Mojang ever create threaded worlds this would be problematic. So at the moment I'm storing the Actors in a `Map<UUID, SpongeActor>`, and creating one actor per world. But then what if the world is unloaded? Now there's a dangling actor that won't be cleaned up.
What about a `Map<UUID, WeakReference<SpongeActor>>` or subscribe to world unload and remove the entry?
Basically I'd like feedback on what's a good solution as I don't know.

Now, there's a second issue. Due to SpongeActor being a Player, it will end up posting various PlayerEvents which shouldn't happen (but in some cases, like Forge's PlayerEvent.BreakSpeed it should be fired). So basically, what to do about this?

Closes SpongePowered/Sponge#319


DDoS:
If you want a map value to be dropped when all references of the key are gone, you use a WeakHashMap. It has weak keys, which won't prevent the object from being GC'd. When the key gets collected, the mapping is removed. If the value has a reference to the key though you won't benefit from it. The world won't get collected because the player is holding on to it, which means the key is always present and the mapping is never removed. You need to arrange this so the player gets removed from world with the others when it is unloaded.

Also I suggest using something else than the word "Actor" as that already refers to a programming concept and could cause confusion. You should use something like SpongeSimulatedPlayer.

progwml6:
@simon816 forge has a fake player class which should be used on that implementation so mods don't need special casing for this fake player

simon816:
Hmm yes, good point. Obviously the class hierarchy couldn't make that possible as-is in Sponge/SpongeCommon, I'll have to create an interface or something.

me4502:
This is looking good, apart from the already mentioned changes.

These were my intentions when I created these APIs, I'm pretty sure they are all already implemented in this PR, but it's good to have them here if anything is changed.

* Simulate it as if it's done as a player
* Call the events so that other plugins and mods can listen for it
* Provide the most accurate way to destroy a block with item drops, unlike Bukkit's break method which didn't show special drops etc.

ryantheleach:
@me4502 So if you wanted to make a plugin that caused the diamond pick to mine a 3x3 area, Would the break happen on behalf of the player, or on behalf of sponge?

The reason I ask is say if you want to start mining an area that is protected, would the player need permission or would SpongeFakePlayer end up needing permission for the area?

So lets assume this is fixed by allowing plugins to specify the player that is breaking the block, this would then cause the player to be successfully able to mine in their own protected area.

Suddenly an anti-cheat plugin gets excited because the player is breaking blocks faster then you can in the base game, it doesn't know how to handle it.

So is there a way that we can attribute the block changes to a player, whilst exposing the fact that it's on behalf of a fakeplayer, or another plugin allowing the player to break those blocks to fakeplayer / proxyplayer aware plugins?

simon816:
I have updated this in response to feedback.
In order to make it extend `FakePlayer` in Sponge, the following will be added to Sponge's initialization
```java
        PlayerSimulatorFactory.instance = new PlayerSimulatorFactory() {

            @Override
            protected EntityPlayerMP createPlayer(WorldServer world) {
                return new SpongeFakePlayer(world);
            }
        };
```
where `SpongeFakePlayer extends FakePlayer`

There's still some open issues here:
* The permission of the player
* Making it fire events like `BlockBreakEvent` rather than `PlayerBreakBlockEvent`
* For block interaction (`interactBlock`, `interactBlockWith`) should the side of the block be a parameter in the API? Currently I'm using `NORTH` simply because there has to be something.

simon816:
I have come up with a way to fire the correct events for the simulated player, see https://github.com/SpongePowered/Sponge/compare/master...feature/fake-player
This will 'downgrade' `PlayerBreakBlockEvent` to `BlockBreakEvent` if the player is a `FakePlayer`. The same effect can be done in SpongeVanilla too.
@me4502, @zml2008, anyone, what can we do about the permission of the player?
It would be a good idea to be able to control what the player can/cannot do. How could a plugin specify what permissions it must have.
Lets say a plugin creates a 'turtle' which can dig blocks (something like computercraft)
If the player is a subject, it could have a set of permissions, something like:
```
SimulatedPlayer: {
    "com.example.turtle"
    "net.example.robot"
}
```
etc
So in terms of the API could we specify the permissions needed `digBlock(Set<String>)` or something.

I'm going to add a Direction parameter to `interactBlock` in the API because there needs to be a way to choose what side of the block the simulator is interacting with.

me4502:
We could have a system of 'permission sets', that basically act as a group of permissions that can be set on something. This way we could have PermissionSets.ALL, etc

Zidane:
@me4502 

A Group is really a glorified "PermissionSet"

@simon816 

You should just define permissions for actions simulated players can do and clearly document them *somewhere*. Then server admins can know to enable them. By default, all simulated player perms should be set to FALSE.

@SpongePowered/developers ^

Zidane:
@simon816 

The biggest thing is how to have simulated players fit into the permissions framework. Subjects would normally be stored via UUID...for simulated players to fit they would need a UUID that no legitimate player could ever have. Needless to say, it gets real messy fast.

@zml2008 any ideas?

zml2008:
Would have to have the fake player be like a sysem-typd subject. -- so an identifier of Like fake-player

ryantheleach:
This is still assuming you want a single global fake-player for Sponge. Where possible the subject methods on a fakePlayer should delegate to the player most appropriate.

Zidane:
@ryantheleach 

simulated player is per world.

ryantheleach:
You still get my point though right?

JBYoshi:
About the weak reference stuff... why not use a `LoadingCache` with `weakValues()` set?

Zidane:
@simon816 

I want to switch gears with this PR. We need to re-think this and instead expose creating your own FakePlayer to SpongeAPI so you can specify name of it and other plugins can check for it. Just the same as the Forge system.

@bloodmc

JBYoshi:
Might be worth thinking about allowing these methods to be called with any `Player` (not just a `FakePlayer`) as a parameter. However, it would still be useful to have the option of using a `FakePlayer`.

gabizou:
@simon816 How well tested is this both in single player (try to cover almost all corner cases) and in server side?

simon816:
I've only tested this on a dedicated server in debug but it worked as expected.
I'm still waiting on @DDoS to finish up the API changes, https://github.com/SpongePowered/SpongeAPI/commit/d0bc27252b81e2288a2a19681585bac112437a74#commitcomment-13578083

src/main/java/org/spongepowered/common/world/SimulatedPlayer.java
```diff
@@ -0,0 +1,123 @@
...
+import net.minecraft.server.management.ItemInWorldManager;
+import net.minecraft.stats.StatBase;
+import net.minecraft.util.DamageSource;
+import net.minecraft.util.IChatComponent;
+import net.minecraft.world.WorldServer;
+
+import java.util.Set;
+
+// Based on Forge's FakePlayer
+public class SimulatedPlayer extends EntityPlayerMP {
```
gabizou: By extending `EntityPlayerMP`, the simulated player is going to be a `Player` instance, is it not?

simon816: Correct. That's how it has to work and how it's done in Forge.

simon816:
Updated to most recent changes

simon816:
OK this implementation should be good to go. I got the cause to transfer over to the events that are fired. (Interact event's tweaked in SpongeForge).

gabizou:
Couldn't you just handle it by default in Spongecommon so SpongeVanilla doesn't have to have any changes made to work? It makes sense in most cases where Common has the base implementation and Forge just overrides the implementation necessary.

simon816:
Normally that would work fine, and I have done that where possible but because the interact event is handled differently on Forge/Vanilla. It wouldn't be too hard just to modify `MixinItemInWorldManager` on both impls.

gabizou:
Can you link the PR's for SpongeForge/Vanilla then?

simon816:
SpongePowered/SpongeForge#473 SpongePowered/SpongeVanilla#198

src/main/java/org/spongepowered/common/world/FakePlayer.java
```diff
@@ -0,0 +1,194 @@
...
+            player.theItemInWorldManager.setWorld((WorldServer) world);
+            player.posX = x;
+            player.posY = y;
+            player.posZ = z;
+            player.onGround = true;
+            player.inventory.setItemStack((net.minecraft.item.ItemStack) itemInHand);
+            player.inventory.mainInventory[player.inventory.currentItem] = (net.minecraft.item.ItemStack) itemInHand;
+
+            // Set up cause capture
+            causeLock.lock();
```
simon816: I wasn't sure if putting a lock here was fine or not. I just don't want the `currentCause` to get overridden by anything while the simulation is processing.

JBYoshi: This might be a problem if an event listener wants to do some simulated methods of its own. Maybe a `Stack` (or a similar class) would suffice?

simon816: Yeah I did think of a stack. I'll change it to a stack actually

Aaron1011:
@simon816: What's the status of this?

simon816:
Not touched this in a while but it's functionally complete. Only thing that I need to do is fix merge conflicts unless other comments arise.

gabizou:
Has this been tested with cause tracking to see what sort of things happen?

simon816:
Yes, it was designed to work properly with the cause tracking system. I should test it again as it was before the new year the last I looked at this

gabizou:
@simon816 As #502 is nearing completion (most everything is working currently, just a few phases that aren't processing their captures), I'd like to see if the simulator will be using the new system if possible. If you can familiarize yourself with it to fully understand what is going ion a majority of the time.

simon816:
I have just re-tested this and fixed issues since I last rebased.

Here are the results of events fired when performing some actions:
(Only `ChangeBlockEvent` and `InteractBlockEvent` were being listened for)

__Digging a grass block__
```java
org.spongepowered.api.event.block.ChangeBlockEvent$Break {
    cause=Cause[{Name=PlayerSimulated, Object={
        com.mojang.authlib.GameProfile@c71ae16[id=ac1750f0-e653-4f06-aef6-1437df64f6dd, name=testplugin, properties={}, legacy=false]
    }}], cancelled=false, transactions=[Transaction {
        original=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(281, 3, 24), blockState=minecraft:grass[snowy=false], extendedState=minecraft:grass[snowy=false]},
        default=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(281, 3, 24), blockState=minecraft:air, extendedState=minecraft:air},
        custom=null, valid=true
    }], targetWorld=net.minecraft.world.WorldServer@70e36a53
}
```

__Mining a stone block with iron pickaxe__
```java
org.spongepowered.api.event.block.ChangeBlockEvent$Break {
    cause=Cause[{Name=PlayerSimulated, Object={
        com.mojang.authlib.GameProfile@c71ae16[id=ac1750f0-e653-4f06-aef6-1437df64f6dd, name=testplugin, properties={}, legacy=false]
    }}], cancelled=false, transactions=[Transaction {
        original=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(279, 5, 25), blockState=minecraft:stone[variant=stone], extendedState=minecraft:stone[variant=stone]},
        default=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(279, 5, 25), blockState=minecraft:air, extendedState=minecraft:air},
        custom=null, valid=true
    }], targetWorld=net.minecraft.world.WorldServer@70e36a53
}
```

__Interacting with lever__
```java
org.spongepowered.api.event.block.InteractBlockEvent$Secondary {
    originalUseBlockResult=UNDEFINED, originalUseItemResult=UNDEFINED, interactionPoint=Optional[(0.0, 0.0, 0.0)], targetSide=UP,
    cause=Cause[{Name=PlayerSimulated,Object={
            com.mojang.authlib.GameProfile@c71ae16[id=ac1750f0-e653-4f06-aef6-1437df64f6dd, name=testplugin, properties={}, legacy=false]
        }}], cancelled=false, useItemResult=UNDEFINED, useBlockResult=UNDEFINED,
    targetBlock=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(284, 4, 22), blockState=minecraft:lever[facing=up_x, powered=true], extendedState=minecraft:lever[facing=up_x, powered=true]}
}
```
```java
org.spongepowered.api.event.block.ChangeBlockEvent$Modify {
    cause=Cause[{Name=PlayerSimulated, Object={
        com.mojang.authlib.GameProfile@c71ae16[id=ac1750f0-e653-4f06-aef6-1437df64f6dd, name=testplugin, properties={}, legacy=false]
    }}], cancelled=false, transactions=[Transaction {
        original=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(284, 4, 22), blockState=minecraft:lever[facing=up_x, powered=true], extendedState=minecraft:lever[facing=up_x, powered=true]},
        default=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(284, 4, 22), blockState=minecraft:lever[facing=up_x, powered=false], extendedState=minecraft:lever[facing=up_x, powered=false]},
        custom=null, valid=true
    }], targetWorld=net.minecraft.world.WorldServer@70e36a53
}
```

__Interacting with end portal holding ender eye__
```java
org.spongepowered.api.event.block.InteractBlockEvent$Secondary {
    originalUseBlockResult=UNDEFINED, originalUseItemResult=UNDEFINED, interactionPoint=Optional[(0.0, 0.0, 0.0)], targetSide=UP,
    cause=Cause[{Name=PlayerSimulated, Object={
            com.mojang.authlib.GameProfile@c71ae16[id=ac1750f0-e653-4f06-aef6-1437df64f6dd, name=testplugin, properties={}, legacy=false]
        }}], cancelled=false, useItemResult=UNDEFINED, useBlockResult=UNDEFINED,
     targetBlock=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(282, 4, 20), blockState=minecraft:end_portal_frame[eye=false, facing=south], extendedState=minecraft:end_portal_frame[eye=false, facing=south]}
}
```
```java
org.spongepowered.api.event.block.ChangeBlockEvent$Modify {
    cause=Cause[{Name=PlayerSimulated, Object={
        com.mojang.authlib.GameProfile@c71ae16[id=ac1750f0-e653-4f06-aef6-1437df64f6dd, name=testplugin, properties={}, legacy=false]
    }}], cancelled=false, transactions=[Transaction {
        original=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(282, 4, 20), blockState=minecraft:end_portal_frame[eye=false, facing=south], extendedState=minecraft:end_portal_frame[eye=false, facing=south]},
        default=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(282, 4, 20), blockState=minecraft:end_portal_frame[eye=true, facing=south], extendedState=minecraft:end_portal_frame[eye=true, facing=south]},
        custom=null, valid=true
    }], targetWorld=net.minecraft.world.WorldServer@70e36a53
}
```
__Placing a stone block__
```java
org.spongepowered.api.event.block.ChangeBlockEvent$Place {
    cause=Cause[{Name=PlayerSimulated, Object={
        com.mojang.authlib.GameProfile@c71ae16[id=ac1750f0-e653-4f06-aef6-1437df64f6dd, name=testplugin, properties={}, legacy=false]
    }}], cancelled=false, transactions=[Transaction {
        original=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(279, 4, 18), blockState=minecraft:air, extendedState=minecraft:air},
        default=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(279, 4, 18), blockState=minecraft:stone[variant=stone], extendedState=minecraft:stone[variant=stone]},
        custom=null, valid=true
    }], targetWorld=net.minecraft.world.WorldServer@70e36a53
}
```

__Hitting grass__
```java
org.spongepowered.api.event.block.InteractBlockEvent$Primary {
    interactionPoint=Optional.empty, targetSide=UP, cause=Cause[{
        Name=PlayerSimulated, Object={
            com.mojang.authlib.GameProfile@c71ae16[id=ac1750f0-e653-4f06-aef6-1437df64f6dd, name=testplugin, properties={}, legacy=false]
    }}], cancelled=false, targetBlock=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(284, 3, 22), blockState=minecraft:grass[snowy=false], extendedState=minecraft:grass[snowy=false]}
}
```

__Hitting redstone wire__
```java
org.spongepowered.api.event.block.InteractBlockEvent$Primary {
    interactionPoint=Optional.empty, targetSide=UP, cause=Cause[{
        Name=PlayerSimulated, Object={
            com.mojang.authlib.GameProfile@c71ae16[id=ac1750f0-e653-4f06-aef6-1437df64f6dd, name=testplugin, properties={}, legacy=false]
    }}], cancelled=false,
    targetBlock=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(277, 4, 21), blockState=minecraft:redstone_wire[east=none, north=none, power=0, south=none, west=none], extendedState=minecraft:redstone_wire[east=none, north=none, power=0, south=none, west=none]}
}
```
```java
org.spongepowered.api.event.block.ChangeBlockEvent$Break {
    cause=Cause[{Name=PlayerSimulated, Object={
            com.mojang.authlib.GameProfile@c71ae16[id=ac1750f0-e653-4f06-aef6-1437df64f6dd, name=testplugin, properties={}, legacy=false]
     }}], cancelled=false, transactions=[Transaction {
        original=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(277, 4, 21), blockState=minecraft:redstone_wire[east=none, north=none, power=0, south=none, west=none], extendedState=minecraft:redstone_wire[east=none, north=none, power=0, south=none, west=none]},
        default=SpongeBlockSnapshot {worldUniqueId=54f24825-484f-4284-a65f-a30fd5526b16, position=(277, 4, 21), blockState=minecraft:air, extendedState=minecraft:air},
        custom=null, valid=true
    }], targetWorld=net.minecraft.world.WorldServer@70e36a53
}
```

gabizou:
@simon816 how about any resulting spawn entity events from harvesting those blocks?

simon816:
Oh yes, all events handled by the capture system are handed the same `Cause` object so any events should be alright

This was when digging a grass block
```java
org.spongepowered.api.event.item.inventory.DropItemEvent$Destruct {
    entities=[EntityItem['item.tile.dirt.default' / 566563, l='world', x=282.36, y=3.27, z=25.33]],
    cause=Cause[
        {Name=Source, Object={SpawnCause{SpawnType=SpongeSpawnType{ id=DroppedItem, name=DroppedItem }}}},
        {Name=PlayerSimulated, Object={com.mojang.authlib.GameProfile@6b187660[id=b440f5b7-30b6-4114-a899-8249bb5194bf, name=testplugin, properties={}, legacy=false]}}
    ],
    cancelled=false, targetWorld=net.minecraft.world.WorldServer@14f087e8,
    entitySnapshots=[SpongeEntitySnapshot {
        uniqueId=85fd6b92-8739-4bb8-b94c-918829333e5a, entityType=SpongeEntityType {
            id=minecraft: item, name=item, translation=SpongeTranslation {id=entity.Item.name}, modid=minecraft, class=net.minecraft.entity.item.EntityItem
        }, position=(282.356988966465, 3.273331105709076, 25.32516786456108), rotation=(0.0, 226.51947021484375, 0.0), scale=(1.0, 1.0, 1.0)
    }]
}
```

Aaron1011:
@simon816: What's the status of this?

theboomer:
Please provide an update on this issue status

simon816:
Updated.

simon816:
I've updated this to the new cause tracking system

gabizou:
@Aaron1011, while I'll still review the PR, if you can give this a look over as well as you're the other person most familiar with `CauseTracker`. Likewise @bloodmc if you can look this over as well.

src/main/java/org/spongepowered/common/event/tracking/PhaseContext.java
```diff
@@ -98,6 +99,13 @@ public PhaseContext add(@Nullable NamedCause namedCause) {
         return this;
     }
 
+    public PhaseContext addCause(Cause cause) {
```
gabizou: Yeah no... this isn't at all the way it should be used. If anything, just pass it in as an object.

src/main/java/org/spongepowered/common/event/tracking/PhaseContext.java
```diff
@@ -295,6 +303,14 @@ public boolean isComplete() {
         return Optional.empty();
     }
 
+    public Cause toCause() {
```
gabizou: This should never be used, for any reason, `PhaseContext` is for contextual system information during a phase, if you want to keep the cause, just keep the cause that the plugin provided.

src/main/java/org/spongepowered/common/event/tracking/phase/plugin/PluginPhase.java
```diff
@@ -45,6 +45,7 @@
         public static final IPhaseState BLOCK_WORKER = new BlockWorkerPhaseState();
         public static final IPhaseState CUSTOM_SPAWN = new PluginPhaseState();
         public static final IPhaseState CUSTOM_EXPLOSION = new CustomExplosionState();
+        public static final IPhaseState FAKEPLAYER = new FakeplayerState();
```
gabizou: Name the field appropriately (`FAKE_PLAYER`)

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -348,6 +353,56 @@ public boolean setBlock(int x, int y, int z, BlockState block, Cause cause) {
...
+        checkNotNull(cause, "cause");
+        Optional<Object> simulatedCause = cause.get(NamedCause.PLAYER_SIMULATED, Object.class);
+        checkArgument(simulatedCause.isPresent(), "Cause does not contain a NamedCause.PLAYER_SIMULATED object");
+        Object obj = simulatedCause.get();
+        checkArgument(obj instanceof GameProfile || obj instanceof User, "Simulated cause object is not a GameProfile or User");
+        return obj instanceof GameProfile ? (GameProfile) obj : (GameProfile) ((User) obj).getProfile();
+    }
+
+    @Override
+    public boolean hitBlock(int x, int y, int z, Direction side, Cause cause) {
```
gabizou: All of this should be in `MixinWorldServer`, not `MixinWorld`.

src/main/java/org/spongepowered/common/world/SimulatedPlayer.java
```diff
@@ -0,0 +1,127 @@
...
+
+    @Override
+    public Entity changeDimension(int dim) {
+        return this;
+    }
+
+    @Override
+    public void handleClientSettings(CPacketClientSettings packetIn) {
+    }
+}
```
gabizou, line 127: I'm pretty sure there are other methods that should be overridden to avoid certain corner cases where the server or game can crash (on death, player specific methods, etc.)

simon816: I took it from forge's [FakePlayer](https://github.com/MinecraftForge/MinecraftForge/blob/1.11.x/src/main/java/net/minecraftforge/common/util/FakePlayer.java#L45-L55), I think it covers the main cases. For the restricted amount of actions available, and no external access to the `SimulatedPlayer` there shouldn't be any unexpected method calls. (Note: in spongeforge I use forge's `FakePlayer` so mods can handle fake players properly)

src/main/java/org/spongepowered/common/event/tracking/PhaseContext.java
simon816: What's the best way to make all events in the fake player phase have the plugin-provided cause? I figured the cause should just get added to the context.

src/main/java/org/spongepowered/common/event/tracking/PhaseContext.java
simon816: I probably missed it but I somehow need to grab whatever the current cause is at the time an event should be fired.
See here: https://github.com/SpongePowered/SpongeCommon/pull/86/files#diff-feadb8e465f1fe25208169de8ca12a02R136

src/main/java/org/spongepowered/common/event/tracking/PhaseContext.java
gabizou: Yes, added to the context, but don't deconstruct the cause to the context and then re-construct it.... it's wasteful.

src/main/java/org/spongepowered/common/event/tracking/phase/plugin/PluginPhase.java
Cybermaxke: The same goes for the class name `FakePlayerState`

theboomer:
Does this system stand a chance to be implimented one day, or is it turning out to be two systems at odds with each other that has no way to resolve the battle/struggle to work in a consistant way?

bloodmc:
@simon816 Please update this PR for bleeding

src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
```diff
@@ -113,8 +115,21 @@ public EnumActionResult processRightClickBlock(EntityPlayer player, net.minecraf
...
 
+        Cause cause;
+        // Handle fake players
+        if (player instanceof FakePlayer) {
+            Optional<Object> simulated = ((IMixinWorldServer) player.world).getCauseTracker().getCurrentContext()
+                    .firstNamed(NamedCause.PLAYER_SIMULATED, Object.class);
+            if (simulated.isPresent()) {
+                cause = Cause.of(NamedCause.of(NamedCause.PLAYER_SIMULATED, simulated.get()));
+            } else {
+                cause = Cause.of(NamedCause.of(NamedCause.PLAYER_SIMULATED, ((EntityPlayer) player).getGameProfile()));
```
Zidane: How could this branch get hit?

src/main/java/org/spongepowered/common/world/FakePlayer.java
```diff
@@ -0,0 +1,203 @@
...
+            }
+            player.inventory.setItemStack(stack);
+            player.inventory.mainInventory.set(player.inventory.currentItem, stack);
+
+            if (!cause.containsNamed(NamedCause.SOURCE)) {
+                // Cause tracker REQUIRES source?
+                cause = cause.with(NamedCause.of(NamedCause.SOURCE, cause.get(NamedCause.PLAYER_SIMULATED, Object.class).get()));
+            }
+            // Set up cause capture
+            ((IMixinWorldServer) world).getCauseTracker().switchToPhase(PluginPhase.State.FAKEPLAYER, PhaseContext.start()
```
Zidane: Why do you need to switch to this phase when setting up the player?

@gabizou maybe I need a lesson on phase switching here.

src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
simon816: It's probably unlikely, but safer than doing Optional.get() on something.
If it's a fake player from a forge mod, the cause tracker probably wont have the necessary objects in the context

src/main/java/org/spongepowered/common/world/FakePlayer.java
simon816: I originally wrote this before we had a cause tracker, once we did, I adapted it to use a [cause stack](https://github.com/SpongePowered/SpongeCommon/commit/361a6017#diff-feadb8e465f1fe25208169de8ca12a02R102). After the cause tracking refactor I changed it to use a phase (I should mention that I don't really know how the current system works).

Essentially, the following needs to happen:
```
> API call (e.g. digBlock)
>> Setup fake player
>> Push given cause onto stack
>>> Make internal call
    * All events fired during this call will have the cause on the top of the stack *
    * another call to digBlock should be possible here because we use a stack *
<< Pop cause stack
<< Tear down fake player
< Return result
```

Maybe someone more knowledgeable on the cause tracking system will be able to see if I have done this correctly

Zidane: My confusion is not in the pushing of the cause (as that is correct) but where you are doing this.

Aren't FakePlayers singletons per world? If so, you would switchToPhase and toss the fakeplayer on the cause stack when the API call occurs.

Zidane: Hold on, I misread this.

setupPlayer is called each API action. Ignore all I said then, you are fine :p.

simon816:
I'll be updating this soon now the cause tracker changes have been merged.

simon816:
Updated to the new cause tracker (again)

gabizou:
Update and merge. Looks good to me otherwise.

src/main/java/org/spongepowered/common/world/FakePlayer.java
```diff
@@ -0,0 +1,195 @@
...
+            // Tear down cause capture
+            PhaseTracker.getInstance().completePhase(PhaseTracker.getInstance().getCurrentState());
+
+            player.inventory.clear();
+            player.world = null;
+            player.interactionManager.setWorld(null);
+        }
+
+        public boolean hit(World world, int x, int y, int z, Direction side, GameProfile profile) {
+            EntityPlayerMP player = setUpPlayer(profile, world, x, y, z, null);
```
gabizou: One thing I can imagine looking at this PR is that we can make this entirely contained within good `try catch` blocks like so:
```java
try (SpongeFakePlayer player = setUpPlayer(profile, world, x, y, z null)) {
  EnumFacing.....
  // yadda yadda yadda
}

```

In the sense that the SpongeFakePlayer would be `AutoCloseable` to call `tearDownPlayer` properly. This can also play into pushing onto the `CauseStackManager` for the game profile object etc.

ryantheleach:
This is now updated (badly, I need to spend time fixing stack frame corruption) and I plan on adopting this PR until it's reviewed / merged. 

<3 @me4502 (Thanks for the WE refactors!)

I've added a test plugin, merged old history, and attempted to fix merge conflicts.

The test plugin has exposed stack frame corruption, So I will likely squash, rebase, (on a different branch) and attempt to rewrite this for the latest cause tracker stuff. I'll need a hand.

Other then the clearly critical events (I havn't even looked at the forge side yet) the test plugin works as expected.

```
[21:11:17] [Server thread/WARN] [Sponge]:   Add -Dsponge.debugcauseframes to your startup flags to enable further debugging output.
[21:11:17] [Server thread/ERROR] [Sponge]: /********************************************************************************************************/
[21:11:17] [Server thread/ERROR] [Sponge]: /*                                    Cause Stack Frame Corruption!                                     */
[21:11:17] [Server thread/ERROR] [Sponge]: /********************************************************************************************************/
[21:11:17] [Server thread/ERROR] [Sponge]: /* Found 
 frames left on the stack. Clearing them all.                                                */
[21:11:17] [Server thread/ERROR] [Sponge]: /*                                                                                                      */
[21:11:17] [Server thread/ERROR] [Sponge]: /* Please add -Dsponge.debugcauseframes=true to your startup flags to enable further debugging output.  */
[21:11:17] [Server thread/ERROR] [Sponge]: /********************************************************************************************************/
[21:11:17] [Server thread/ERROR] [Sponge]: Could not pass InteractBlockEvent$Secondary$MainHand$Impl to Plugin{id=playersimulatedtest, name=PlayerSimulatedTest, version=0.0.0, description=Tests Fake Player methods., source=D:\s\SpongeForge\classes\production\testplugins_main}
java.lang.IllegalStateException: Cause Stack Frame Corruption! Attempted to pop a frame that was not on the stack.
	at org.spongepowered.common.event.SpongeCauseStackManager.popCauseFrame(SpongeCauseStackManager.java:249) ~[SpongeCauseStackManager.class:?]
	at org.spongepowered.common.event.SpongeCauseStackManager$CauseStackFrameImpl.close(SpongeCauseStackManager.java:443) ~[SpongeCauseStackManager$CauseStackFrameImpl.class:?]
	at org.spongepowered.mod.event.SpongeModEventManager.post(SpongeModEventManager.java:384) [SpongeModEventManager.class:?]
	at org.spongepowered.mod.event.SpongeModEventManager.extendedPost(SpongeModEventManager.java:443) [SpongeModEventManager.class:?]
	at net.minecraft.server.management.PlayerInteractionManager.processRightClickBlock(PlayerInteractionManager.java:1224) [PlayerInteractionManager.class:?]
	at net.minecraft.network.NetHandlerPlayServer.redirect$onProcessRightClickBlock$zin000(NetHandlerPlayServer.java:2500) [NetHandlerPlayServer.class:?]
	at net.minecraft.network.NetHandlerPlayServer.processTryUseItemOnBlock(NetHandlerPlayServer.java:767) [NetHandlerPlayServer.class:?]
	at net.minecraft.network.play.client.CPacketPlayerTryUseItemOnBlock.processPacket(CPacketPlayerTryUseItemOnBlock.java:68) [CPacketPlayerTryUseItemOnBlock.class:?]
	at net.minecraft.network.play.client.CPacketPlayerTryUseItemOnBlock.processPacket(CPacketPlayerTryUseItemOnBlock.java:13) [CPacketPlayerTryUseItemOnBlock.class:?]
	at org.spongepowered.common.network.PacketUtil.onProcessPacket(PacketUtil.java:92) [PacketUtil.class:?]
	at net.minecraft.network.PacketThreadUtil$1.redirect$onProcessPacket$zle000(PacketThreadUtil.java:539) [PacketThreadUtil$1.class:?]
	at net.minecraft.network.PacketThreadUtil$1.run(PacketThreadUtil.java:21) [PacketThreadUtil$1.class:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_144]
	at java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:266) [?:1.8.0_144]
	at java.util.concurrent.FutureTask.run(FutureTask.java) [?:1.8.0_144]
	at net.minecraft.util.Util.runTask(Util.java:53) [Util.class:?]
	at org.spongepowered.common.SpongeImplHooks.onUtilRunTask(SpongeImplHooks.java:292) [SpongeImplHooks.class:?]
	at net.minecraft.server.MinecraftServer.redirect$onRun$zjj000(MinecraftServer.java:4181) [MinecraftServer.class:?]
	at net.minecraft.server.MinecraftServer.updateTimeLightAndEntities(MinecraftServer.java:798) [MinecraftServer.class:?]
	at net.minecraft.server.MinecraftServer.tick(MinecraftServer.java:743) [MinecraftServer.class:?]
	at net.minecraft.server.integrated.IntegratedServer.tick(IntegratedServer.java:192) [IntegratedServer.class:?]
	at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:592) [MinecraftServer.class:?]
	at java.lang.Thread.run(Thread.java:748) [?:1.8.0_144]
[21:11:18] [Server thread/WARN] [Sponge]:   Add -Dsponge.debugcauseframes to your startup flags to enable further debugging output.
[21:11:18] [Server thread/ERROR] [Sponge]: /********************************************************************************************************/
[21:11:18] [Server thread/ERROR] [Sponge]: /*                                    Cause Stack Frame Corruption!                                     */
[21:11:18] [Server thread/ERROR] [Sponge]: /********************************************************************************************************/
[21:11:18] [Server thread/ERROR] [Sponge]: /* Found 
 frames left on the stack. Clearing them all.                                                */
[21:11:18] [Server thread/ERROR] [Sponge]: /*                                                                                                      */
[21:11:18] [Server thread/ERROR] [Sponge]: /* Please add -Dsponge.debugcauseframes=true to your startup flags to enable further debugging output.  */
[21:11:18] [Server thread/ERROR] [Sponge]: /********************************************************************************************************/
[21:11:18] [Server thread/ERROR] [Sponge]: Could not pass InteractBlockEvent$Secondary$MainHand$Impl to Plugin{id=playersimulatedtest, name=PlayerSimulatedTest, version=0.0.0, description=Tests Fake Player methods., source=D:\s\SpongeForge\classes\production\testplugins_main}
java.lang.IllegalStateException: Cause Stack Frame Corruption! Attempted to pop a frame that was not on the stack.
	at org.spongepowered.common.event.SpongeCauseStackManager.popCauseFrame(SpongeCauseStackManager.java:249) ~[SpongeCauseStackManager.class:?]
	at org.spongepowered.common.event.SpongeCauseStackManager$CauseStackFrameImpl.close(SpongeCauseStackManager.java:443) ~[SpongeCauseStackManager$CauseStackFrameImpl.class:?]
	at org.spongepowered.mod.event.SpongeModEventManager.post(SpongeModEventManager.java:384) [SpongeModEventManager.class:?]
	at org.spongepowered.mod.event.SpongeModEventManager.extendedPost(SpongeModEventManager.java:443) [SpongeModEventManager.class:?]
	at org.spongepowered.mod.event.SpongeModEventManager.post(SpongeModEventManager.java:404) [SpongeModEventManager.class:?]
	at org.spongepowered.common.event.SpongeEventManager.post(SpongeEventManager.java:447) [SpongeEventManager.class:?]
	at org.spongepowered.common.SpongeImpl.postEvent(SpongeImpl.java:225) [SpongeImpl.class:?]
	at org.spongepowered.mod.event.ForgeToSpongeEventFactory.onPostEnd(ForgeToSpongeEventFactory.java:513) [ForgeToSpongeEventFactory.class:?]
	at org.spongepowered.mod.event.SpongeModEventManager.post(SpongeModEventManager.java:316) [SpongeModEventManager.class:?]
	at net.minecraftforge.fml.common.eventhandler.EventBus.post(EventBus.java:665) [EventBus.class:?]
	at net.minecraftforge.fml.common.eventhandler.EventBus.post(EventBus.java:639) [EventBus.class:?]
	at net.minecraftforge.common.ForgeHooks.onItemRightClick(ForgeHooks.java:1111) [ForgeHooks.class:?]
	at net.minecraft.server.management.PlayerInteractionManager.processRightClick(PlayerInteractionManager.java:379) [PlayerInteractionManager.class:?]
	at net.minecraft.network.NetHandlerPlayServer.processTryUseItem(NetHandlerPlayServer.java:794) [NetHandlerPlayServer.class:?]
	at net.minecraft.network.play.client.CPacketPlayerTryUseItem.processPacket(CPacketPlayerTryUseItem.java:43) [CPacketPlayerTryUseItem.class:?]
	at net.minecraft.network.play.client.CPacketPlayerTryUseItem.processPacket(CPacketPlayerTryUseItem.java:9) [CPacketPlayerTryUseItem.class:?]
	at org.spongepowered.common.network.PacketUtil.onProcessPacket(PacketUtil.java:92) [PacketUtil.class:?]
	at net.minecraft.network.PacketThreadUtil$1.redirect$onProcessPacket$zle000(PacketThreadUtil.java:539) [PacketThreadUtil$1.class:?]
	at net.minecraft.network.PacketThreadUtil$1.run(PacketThreadUtil.java:21) [PacketThreadUtil$1.class:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_144]
	at java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:266) [?:1.8.0_144]
	at java.util.concurrent.FutureTask.run(FutureTask.java) [?:1.8.0_144]
	at net.minecraft.util.Util.runTask(Util.java:53) [Util.class:?]
	at org.spongepowered.common.SpongeImplHooks.onUtilRunTask(SpongeImplHooks.java:292) [SpongeImplHooks.class:?]
	at net.minecraft.server.MinecraftServer.redirect$onRun$zjj000(MinecraftServer.java:4181) [MinecraftServer.class:?]
	at net.minecraft.server.MinecraftServer.updateTimeLightAndEntities(MinecraftServer.java:798) [MinecraftServer.class:?]
	at net.minecraft.server.MinecraftServer.tick(MinecraftServer.java:743) [MinecraftServer.class:?]
	at net.minecraft.server.integrated.IntegratedServer.tick(IntegratedServer.java:192) [IntegratedServer.class:?]
	at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:592) [MinecraftServer.class:?]
	at java.lang.Thread.run(Thread.java:748) [?:1.8.0_144]
/********************************************************************************************************/
/*                                    Cause Stack Frame Corruption!                                     */
/********************************************************************************************************/
/* Found 
 frames left on the stack. Clearing them all.                                                */
/*                                                                                                      */
/* Please add -Dsponge.debugcauseframes=true to your startup flags to enable further debugging output.  */
/********************************************************************************************************/
/********************************************************************************************************/
[21:11:18] [Server thread/WARN] [Sponge]:   Add -Dsponge.debugcauseframes to your startup flags to enable further debugging output.
/*                                    Cause Stack Frame Corruption!                                     */
[21:11:18] [Server thread/ERROR] [Sponge]: /********************************************************************************************************/
/********************************************************************************************************/
/* Found 
[21:11:18] [Server thread/ERROR] [Sponge]: /*                                    Cause Stack Frame Corruption!                                     */
 frames left on the stack. Clearing them all.                                                */
/*                                                                                                      */
/* Please add -Dsponge.debugcauseframes=true to your startup flags to enable further debugging output.  */
/********************************************************************************************************/
[21:11:18] [Server thread/ERROR] [Sponge]: /********************************************************************************************************/
[21:11:18] [Server thread/ERROR] [Sponge]: /* Found 
 frames left on the stack. Clearing them all.                                                */
[21:11:18] [Server thread/ERROR] [Sponge]: /*                                                                                                      */
[21:11:18] [Server thread/ERROR] [Sponge]: /* Please add -Dsponge.debugcauseframes=true to your startup flags to enable further debugging output.  */
[21:11:18] [Server thread/ERROR] [Sponge]: /********************************************************************************************************/
[21:11:18] [Server thread/ERROR] [Sponge]: Could not pass InteractBlockEvent$Secondary$MainHand$Impl to Plugin{id=playersimulatedtest, name=PlayerSimulatedTest, version=0.0.0, description=Tests Fake Player methods., source=D:\s\SpongeForge\classes\production\testplugins_main}
java.lang.IllegalStateException: Cause Stack Frame Corruption! Attempted to pop a frame that was not on the stack.
	at org.spongepowered.common.event.SpongeCauseStackManager.popCauseFrame(SpongeCauseStackManager.java:249) ~[SpongeCauseStackManager.class:?]
	at org.spongepowered.common.event.SpongeCauseStackManager$CauseStackFrameImpl.close(SpongeCauseStackManager.java:443) ~[SpongeCauseStackManager$CauseStackFrameImpl.class:?]
	at org.spongepowered.mod.event.SpongeModEventManager.post(SpongeModEventManager.java:384) [SpongeModEventManager.class:?]
	at org.spongepowered.mod.event.SpongeModEventManager.extendedPost(SpongeModEventManager.java:443) [SpongeModEventManager.class:?]
	at org.spongepowered.mod.event.SpongeModEventManager.post(SpongeModEventManager.java:404) [SpongeModEventManager.class:?]
	at org.spongepowered.common.event.SpongeEventManager.post(SpongeEventManager.java:447) [SpongeEventManager.class:?]
	at org.spongepowered.common.SpongeImpl.postEvent(SpongeImpl.java:225) [SpongeImpl.class:?]
	at org.spongepowered.mod.event.ForgeToSpongeEventFactory.onPostEnd(ForgeToSpongeEventFactory.java:513) [ForgeToSpongeEventFactory.class:?]
	at org.spongepowered.mod.event.SpongeModEventManager.post(SpongeModEventManager.java:316) [SpongeModEventManager.class:?]
	at net.minecraftforge.fml.common.eventhandler.EventBus.post(EventBus.java:665) [EventBus.class:?]
	at net.minecraftforge.fml.common.eventhandler.EventBus.post(EventBus.java:639) [EventBus.class:?]
	at net.minecraftforge.common.ForgeHooks.onItemRightClick(ForgeHooks.java:1111) [ForgeHooks.class:?]
	at net.minecraft.server.management.PlayerInteractionManager.processRightClick(PlayerInteractionManager.java:379) [PlayerInteractionManager.class:?]
	at net.minecraft.network.NetHandlerPlayServer.processTryUseItem(NetHandlerPlayServer.java:794) [NetHandlerPlayServer.class:?]
	at net.minecraft.network.play.client.CPacketPlayerTryUseItem.processPacket(CPacketPlayerTryUseItem.java:43) [CPacketPlayerTryUseItem.class:?]
	at net.minecraft.network.play.client.CPacketPlayerTryUseItem.processPacket(CPacketPlayerTryUseItem.java:9) [CPacketPlayerTryUseItem.class:?]
	at org.spongepowered.common.network.PacketUtil.onProcessPacket(PacketUtil.java:92) [PacketUtil.class:?]
	at net.minecraft.network.PacketThreadUtil$1.redirect$onProcessPacket$zle000(PacketThreadUtil.java:539) [PacketThreadUtil$1.class:?]
	at net.minecraft.network.PacketThreadUtil$1.run(PacketThreadUtil.java:21) [PacketThreadUtil$1.class:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_144]
	at java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:266) [?:1.8.0_144]
	at java.util.concurrent.FutureTask.run(FutureTask.java) [?:1.8.0_144]
	at net.minecraft.util.Util.runTask(Util.java:53) [Util.class:?]
	at org.spongepowered.common.SpongeImplHooks.onUtilRunTask(SpongeImplHooks.java:292) [SpongeImplHooks.class:?]
	at net.minecraft.server.MinecraftServer.redirect$onRun$zjj000(MinecraftServer.java:4181) [MinecraftServer.class:?]
	at net.minecraft.server.MinecraftServer.updateTimeLightAndEntities(MinecraftServer.java:798) [MinecraftServer.class:?]
	at net.minecraft.server.MinecraftServer.tick(MinecraftServer.java:743) [MinecraftServer.class:?]
	at net.minecraft.server.integrated.IntegratedServer.tick(IntegratedServer.java:192) [IntegratedServer.class:?]
	at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:592) [MinecraftServer.class:?]
	at java.lang.Thread.run(Thread.java:748) [?:1.8.0_144]
```

simon816:
Thanks @ryantheleach !

Pulverizer:
I am on porting the Movecraft plugin to Sponge. This needed! Glad somebody else saw this being needed.
InteractableVolume#interactBlock(int x, int y, int z, Direction side, GameProfile profile) shows promise but would need to specify what InteractBlockEvent to trigger. Primary or Secondary? Main-hand or Off-hand?

simon816:
@Pulverizer `hitBlock` is a `Primary` interaction, `interactBlock` is `Secondary`. Main/off-hand didn't exist when the API was proposed. The API will need to be updated for different hands. You can assume it's the main hand for now.

src/main/java/org/spongepowered/common/event/tracking/phase/plugin/PluginPhase.java
```diff
@@ -35,6 +35,7 @@
         public static final IPhaseState<BasicPluginContext> CUSTOM_SPAWN = new BasicPluginState();
         public static final IPhaseState<BasicPluginContext> SCHEDULED_TASK = new ScheduledTaskPhaseState();
         public static final IPhaseState<BasicPluginContext> TELEPORT = new BasicPluginState();
+        public static final IPhaseState<BasicPluginContext> FAKE_PLAYER = new BasicPluginState();
```
gabizou, line 38: It might be benefiting to have a custom state here that will push the fake player as a source and still ensure the plugin container as the context so to speak. You can kinda view this by overriding the popularCauseFrame method

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -225,7 +227,8 @@
 import java.util.TreeSet;
 import java.util.UUID;
 
-import javax.annotation.Nullable;
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
```
gabizou, line 232: Moving imports.

ImMorpheus:

PR #87, opened 2015-07-07
Implement User, UserStorage and GameProfileResolver

simon816:
This implements the 3 interfaces as stated in the title.

Parts of the implementation of `User` was influenced from the discussion on SpongePowered/Sponge#122
Because `Player` extends `User` but `User` is not an `Entity`, there needs to be a separate new class which represents `User`, this is the `SpongeUser`.
Methods in `Player` that are defined and only defined in `User` are proxied to the underlying user object, this avoids duplication of code.

You may look at `SpongeUser` an go, "hang on, it doesn't implement `User`". Yes, this is intentional, it's to avoid making SpongeUser a massive collection of copied methods from various places (I experimented with that idea too, it was a mess).
Instead, it utilizes the power of mixins to mix into `SpongeUser` from existing mixins, thus avoids duplication. The actual implementation of `User` is done in `MixinSpongeUser`.

The second part of this PR is implementing a way to obtain `User` objects through `UserStorage`.
The vanilla layout of storing user data is such that there's a variety of places the data could be.
There's the `world/playerdata/{uuid}.dat`, `whitelist.json` and `banned-players.json`.
`UserDiscoverer` is a "binding" of these locations so `User` objects can be accessed uniformly.

Finally, I have implemented `GameProfileResolver` as a bonus treat :)

Closes SpongePowered/Sponge#122

### DIFF
--- a/src/main/java/org/spongepowered/common/service/profile/SpongeProfileResolver.java
+++ b/src/main/java/org/spongepowered/common/service/profile/SpongeProfileResolver.java
@@ -1,0 +1,260 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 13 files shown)


src/main/java/org/spongepowered/common/service/profile/SpongeProfileResolver.java
```diff
@@ -0,0 +1,260 @@
...
+        }
+
+        @Override
+        public GameProfile call() throws Exception {
+            if (this.id != null) {
+                return this.fromId(this.id);
+            } else if (this.name != null) {
+                return this.fromNames(Lists.newArrayList(this.name)).get(0);
+            }
+            throw new IllegalStateException("Impossibru");
```
me4502: Shouldn't we have a more professional error message?

simon816: Well it literally is impossible to get to this line because either `id` or `name` will not be null but I made it in an `else if` to make it more clear (an explicit null check),

simon816:
I will merge this in 2 days if there are no issues brought before then.

PR #93, opened 2015-07-14
Restore Pre-1.8 Server Spawn Protection Behavior closes #15

### LINKED ISSUE #15: Restore Pre-1.8 Server Spawn Protection Behavior
As some may know, in 1.8 the server disables spawn point protection if the server has no known OPs. In my opinion, this is an incredibly stupid design decision and, as evidenced by looking at various forum posts, many server admins were under a false impression that spawn protection broke.

Put simply, you shouldn't HAVE to have an OP specified for spawn protection work. Its an easy change to DedicatedServer to remove the if check.

@SpongePowered/developers thoughts?

ryantheleach:
@Zidane commented on 29 Apr https://github.com/SpongePowered/SpongeCommon/issues/15
>As some may know, in 1.8 the server disables spawn point protection if the server has no known OPs. In my opinion, this is an incredibly stupid design decision and, as evidenced by looking at various forum posts, many server admins were under a false impression that spawn protection broke.
>Put simply, you shouldn't HAVE to have an OP specified for spawn protection work.

Simple fix in case this is the solution you want to go with, Has been tested in a dev environment only.

I didn't add a configuration option because you can just disable the spawn protection anyway by setting spawn-protection in server.properties below 0.

At this stage it just comes down to whatever avoids the most confusion, having a default spawn protection, or not having it. Having it based on the number of OP's is just plain weird.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/server/MixinDedicatedServer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/server/MixinDedicatedServer.java
@@ -26,9 +26,12 @@ package org.spongepowered.common.mixin.core.server;
 
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.dedicated.DedicatedServer;
+import net.minecraft.server.management.UserListOps;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Overwrite;
 import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
 
 import java.io.File;
 import java.net.Proxy;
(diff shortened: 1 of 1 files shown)


Zidane:
@ryantheleach 

Thanks bud for this PR. Glad someone finally got to it!

@SpongePowered/developers Objections?

me4502:
This is a much needed change

src/main/java/org/spongepowered/common/mixin/core/server/MixinDedicatedServer.java
```diff
@@ -55,4 +58,18 @@ public void setGuiEnabled() {
         this.guiIsEnabled = false;
     }
 
+    /**
+     * @author ryantheleach
+     *
+     * Purpose: This disables the new check in minecraft 1.8 that checks the OP list to check if there is no operators
+     * Reasoning: There has been complaints about 1.8 breaking spawn protection see: https://github.com/SpongePowered/SpongeCommon/issues/15
+     * UserList.hasEntries() is named incorrectly in MCP, it should be isEmpty
```
kenzierocks: You might want to report this to [MCPBot-Issues](https://github.com/ModCoderPack/MCPBot-Issues/issues).

src/main/java/org/spongepowered/common/mixin/core/server/MixinDedicatedServer.java
```diff
@@ -55,4 +58,18 @@ public void setGuiEnabled() {
...
+     * The logic at time of writing is if OPList is empty, disable spawn protection. This has been replaced with if(false) disable spawn protection
+     */
+
```
kenzierocks: Not sure if this newline is part of the style?

src/main/java/org/spongepowered/common/mixin/core/server/MixinDedicatedServer.java
ryantheleach: https://github.com/ModCoderPack/MCPBot-Issues/issues/27

ryantheleach:
Build failing only on openjdk8 because travis couldn't find MCP

src/main/java/org/spongepowered/common/mixin/core/server/MixinDedicatedServer.java
Mumfrey: Or we could just change the mapping to `isEmpty`, is there any need to open an issue for such a minor thing?

zml2008:
imo this ops list check should be replaced with a permissions check, which would match op levels as well, but also allow configuration with a permissions plugin.

src/main/java/org/spongepowered/common/mixin/core/server/MixinDedicatedServer.java
kenzierocks: @mumfrey MCP changes that replace other ones need higher permissions, so we can't personally change it.

zml2008:
#128 replaces this with a permission-controlled system that will block even if no ops are set.

PR #94, opened 2015-07-15
Implement Data API 2.0

gabizou:
[Values API PR](https://github.com/SpongePowered/SpongeAPI/pull/685) | Data 2.0 Implementation PR

This implements the Values API with Data API adjustments. With this PR, a few things are resolved:

- Processors for both `DataManipulator`s and `Value`s such that processing delegation can take place accordingly
- Registration of `Key`s
- Better handling of mutable and immutable data
- Actual support for custom `DataManipulator`s.

The current ToDo list:

- [ ] Re-implement all DataManipulators 
 - [x] Implement Block based DataManipulators
 - [ ] Implement Entity based DataManipulators
 - [ ] Implement TileEntity based DataManipulators
 - [ ] Implement ItemStack based DataManipulators
 - [x] Implement Common DataManipulators
- [ ] Implement Data Processors
 - [x] Devise a system for handling automatic Key<?> -> Field reference setting system within the AbstractData class so as to avoid having ot delegate setting of values on DataManipulators to a ValueProcessor.
- [ ] Implement some BlockDataProcessors
 - [ ] Rewrite all IMixinBlockData interfaces for handling of the correct value types
 - [x] Rewrite actual mixins for above changes
- [x] Implement ValueProcessors
- [ ] Re-implement Data related Mixins


src/main/java/org/spongepowered/common/data/manipulator/mutable/common/AbstractData.java
```diff
@@ -0,0 +1,248 @@
...
+            return false;
+        }
+        final AbstractData other = (AbstractData) obj;
+        return Objects.equal(this.manipulatorClass, other.manipulatorClass);
+    }
+
+    @Override
+    public DataContainer toContainer() {
+        // Yep, empty container
+        return new MemoryDataContainer();
```
Mumfrey: Is it worth having a `MemoryDataContainer.EMPTY` or `DataContainer.EMPTY` singleton and always returning it in these circumstances? Effectively a *bottom type* for `DataContainer` ?

gabizou: Not likely since the `MemoryDataContainer` is mutable, making any changes to the `EMPTY` instance would likely result in a messy `EMPTY` that isn't really `EMPTY`.

src/main/java/org/spongepowered/common/data/value/mutable/SpongeWeightedEntityCollectionValue.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public SpongeWeightedEntityCollectionValue(Key<? extends BaseValue<WeightedCollection<WeightedEntity>>> key,
+                                               WeightedCollection<WeightedEntity> actualValue) {
+        super(key, actualValue);
+    }
+
+    @Override
+    public WeightedEntityCollectionValue filter(Predicate<? super WeightedEntity> predicate) {
+        final WeightedCollection<WeightedEntity> collection = new WeightedCollection<WeightedEntity>();
+        for (WeightedEntity entity : this.actualValue) {
```
Mumfrey: Could you use:

    collection.addAll(Collections2.<WeightedEntity>filter(this.actualValue, checkNotNull(predicate)));

here instead?

src/main/java/org/spongepowered/common/data/value/mutable/common/SpongeEntityValue.java
```diff
@@ -0,0 +1,113 @@
...
+    public SpongeEntityValue(Key<? extends BaseValue<Optional<Entity>>> key) {
+        super(key);
+    }
+
+    public SpongeEntityValue(Key<? extends BaseValue<Optional<Entity>>> key, Optional<Entity> actualValue) {
+        super(key, Optional.<Entity>absent());
+        if (actualValue.isPresent()) {
+            this.weakReference = new WeakReference<Entity>(actualValue.get());
+        } else {
+            this.weakReference.clear();
```
Mumfrey: Since this is a ctor, is this clause actually required? `weakReference` should contain a null `referent` at this point anyway.

gabizou: Good point.

src/main/java/org/spongepowered/common/data/manipulator/mutable/common/AbstractData.java
Mumfrey: Ah capisce, I thought they were immutable.

*In other news, why the hell are my github email notifications borked*. ugh.

gabizou: > Ah capisce, I thought they were immutable.

I was considering at one point to make an immutable `DataView` but I figured it would be highly likely that it would be nigh impossible with the nested structures involved. Possibly an `UnmodifiableDataView` of sorts?

gabizou:
@Zidane Implementation has been tested with a plugin specifically written to interact with the new Data API. The plugin code should be in the API PR description.

PR #99, opened 2015-07-21
Implementation of SneakingData

joshgarde:
SneakingData implemented


ryantheleach:
As a tip if you push to the same branch as an already open PR it will update the PR without you having to create another, it also preserves the history of what you have tried so whilst reviewing people can see how the PR has changed.

joshgarde:
Ah so this is why ``DataManipulator``s like ``SneakingData`` haven't been implemented yet.
https://github.com/SpongePowered/SpongeCommon/pull/94 - DataAPI 2.0

me4502:
Please update this to the current data API.

gabizou:
You need to `git rebase sponge/master`.

joshgarde:
SneakingData implementation - done & updated.

simon816:
You've rebased wrong and have a bunch of commits that shouldn't be there. This is (one of) the reasons why PRing on your `master` branch is a bad idea.
Needs to be fixed

joshgarde:
Updated pull request with new DataAPI and fixed commits - https://github.com/SpongePowered/SpongeCommon/pull/135

PR #100, opened 2015-07-21
Implement resource packs.

JBYoshi:
This PR implements the resource pack API. It requires my changes in SpongePowered/SpongeAPI#772.

Note: Although I reverted my split of `URLResourcePack` and `WorldResourcePack` in the API, I felt it would be easier to keep the split in the implementation because of some special casing for `level://` resource packs (Mojang simply puts `"level://"` in front of the relative path).

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
@@ -40,7 +40,9 @@ import net.minecraft.network.PacketBuffer;
 import net.minecraft.network.play.client.C03PacketPlayer;
 import net.minecraft.network.play.client.C12PacketUpdateSign;
 import net.minecraft.network.play.client.C17PacketCustomPayload;
+import net.minecraft.network.play.client.C19PacketResourcePackStatus;
 import net.minecraft.network.play.server.S3FPacketCustomPayload;
+import net.minecraft.network.play.server.S48PacketResourcePackSend;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.management.ServerConfigurationManager;
 import net.minecraft.tileentity.TileEntity;
@@ -60,9 +62,12 @@ import org.spongepowered.api.event.SpongeEventFactory;
 import org.spongepowered.api.event.block.tileentity.ChangeSignEvent;
 import org.spongepowered.api.event.cause.Cause;
 import org.spongepowered.api.event.entity.DisplaceEntityEvent;
+import org.spongepowered.api.event.entity.living.player.ResourcePackStatusEvent;
+import org.spongepowered.api.event.entity.living.player.ResourcePackStatusEvent.ResourcePackStatus;
 import org.spongepowered.api.event.network.ClientConnectionEvent;
 import org.spongepowered.api.network.ChannelBuf;
 import org.spongepowered.api.network.PlayerConnection;
+import org.spongepowered.api.resourcepack.ResourcePack;
 import org.spongepowered.api.text.Text;
 import org.spongepowered.api.text.Texts;
 import org.spongepowered.api.text.format.TextColors;
@@ -81,10 +86,13 @@ import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
 import org.spongepowered.common.Sponge;
 import org.spongepowered.common.event.SpongeImplEventFactory;
 import org.spongepowered.common.interfaces.IMixinNetworkManager;
+import org.spongepowered.common.interfaces.IMixinPacketResourcePackSend;
 import org.spongepowered.common.text.SpongeTexts;
 
 import java.net.InetSocketAddress;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 
 @Mixin(NetHandlerPlayServer.class)
@@ -99,6 +107,8 @@ public abstract class MixinNetHandlerPlayServer implements PlayerConnection {
 
     private boolean justTeleported = false;
     private Location<World> lastMoveLocation = null;
+
+    private final Map<String, ResourcePack> sentResourcePacks = new HashMap<String, ResourcePack>();
 
     @Override
     public Player getPlayer() {
@@ -362,6 +372,42 @@ public abstract class MixinNetHandlerPlayServer implements PlayerConnection {
         this.tmpQuitMessage = null;
         Sponge.getGame().getEventManager().post(event);
         event.getSink().sendMessage(event.getMessage());
+    }
+        
+    @Inject(method = "handleResourcePackStatus", at = @At("HEAD"))
+    public void onResourcePackStatus(C19PacketResourcePackStatus packet, CallbackInfo ci) {
+        String hash = packet.hash;
+        ResourcePackStatus status;
+        ResourcePack pack = this.sentResourcePacks.get(hash);
+        switch (packet.status) {
+            case ACCEPTED:
+                status = ResourcePackStatus.ACCEPTED;
+                break;
+            case DECLINED:
+                status = ResourcePackStatus.DECLINED;
+                break;
+            case SUCCESSFULLY_LOADED:
+                status = ResourcePackStatus.SUCCESSFULLY_LOADED;
+                break;
+            case FAILED_DOWNLOAD:
+                status = ResourcePackStatus.FAILED;
+                break;
+            default:
+                throw new AssertionError();
+        }
+        if (status.wasSuccessful().isPresent()) {
+            this.sentResourcePacks.remove(hash);
+        }
+        Sponge.getGame().getEventManager()
+                .post(SpongeEventFactory.createResourcePackStatusEvent(Sponge.getGame(), pack, (Player)this.playerEntity, status));
+    }
+
+    @Inject(method = "sendPacket", at = @At("HEAD"))
+    public void onResourcePackSend(Packet packet, CallbackInfo ci) {
+        if (packet instanceof IMixinPacketResourcePackSend) {
+            IMixinPacketResourcePackSend p = (IMixinPacketResourcePackSend) packet;
+            this.sentResourcePacks.put(p.setFakeHash(), p.getResourcePack());
+        }
     }
 
 }
--- a/src/main/java/org/spongepowered/common/mixin/core/network/packet/MixinS48PacketResourcePackSend.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/network/packet/MixinS48PacketResourcePackSend.java
@@ -1,0 +1,78 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.mixin.core.network.packet;
+
+import net.minecraft.network.play.server.S48PacketResourcePackSend;
+import org.spongepowered.api.resourcepack.ResourcePack;
+import org.spongepowered.asm.mixin.Mixin;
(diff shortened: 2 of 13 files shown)


src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -347,5 +357,43 @@ public void onDisconnectPlayer(IChatComponent reason, CallbackInfo ci) {
...
+                break;
+            case DECLINED:
+                status = ResourcePackStatus.DECLINED;
+                break;
+            case SUCCESSFULLY_LOADED:
+                status = ResourcePackStatus.SUCCESSFULLY_LOADED;
+                break;
+            case FAILED_DOWNLOAD:
+                // Best we can do :(
+                status = ResourcePackStatus.PACK_FILE_FORMAT_NOT_RECOGNIZED;
```
ST-DDT: We should rename this to FAILED or at least add it.

src/main/java/org/spongepowered/common/resourcepack/SpongeWorldResourcePack.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public String getUrlString() {
+        return "level://" + this.path;
+    }
+
+    @Override
+    public URI getUri() {
+        try {
+            return new URI("level://" + URLEncoder.encode(path, "UTF-8"));
```
ST-DDT: Why didn't you cache the value?

JBYoshi: I just wasn't thinking about it. Will do.

src/main/java/org/spongepowered/common/resourcepack/SpongeResourcePackFactory.java
```diff
@@ -0,0 +1,79 @@
...
+            return new SpongeURIResourcePack(uri, Optional.of(hasher.hash().toString()));
+        } catch (IOException e) {
+            FileNotFoundException ex = new FileNotFoundException(e.toString());
+            ex.initCause(e);
+            throw ex;
+        }
+    }
+
+    private static InputStream openStream(URI uri) throws IOException {
+        if (uri.toString().startsWith("level://")) {
```
Mumfrey: Given that the constant string `level://` appears so often in this PR, it ought to be a string constant somewhere.

src/main/java/org/spongepowered/common/resourcepack/SpongeWorldResourcePack.java
Mumfrey: Instead of using the string ctor, maybe use `URI.create` factory method instead, since you're just catching this exception further up the call hierarchy and wrapping it in an `IllegalArgumentException` anyway, which is what `URI.create` does.

src/main/java/org/spongepowered/common/resourcepack/SpongeWorldResourcePack.java
```diff
@@ -0,0 +1,64 @@
...
+public class SpongeWorldResourcePack extends SpongeResourcePack {
+
+    private final String path;
+    private final URI uri;
+
+    public SpongeWorldResourcePack(String levelUrl, String hash) {
+        super(hash);
+        this.path = levelUrl.substring("level://".length());
+        try {
+            this.uri = new URI("level://" + URLEncoder.encode(path, "UTF-8"));
```
Mumfrey: Again, you may as well just use `URI.create` here since you're just blindly catching and forwarding the checked exception.

src/main/java/org/spongepowered/common/resourcepack/SpongeResourcePack.java
```diff
@@ -0,0 +1,67 @@
...
+        return this.id;
+    }
+
+    @Override
+    public Optional<String> getHash() {
+        return this.hash;
+    }
+
+    public static SpongeResourcePack create(String url, String hash) throws URISyntaxException {
+        if (url.startsWith("level://")) {
```
Mumfrey: Maybe a `checkNotNull` on `url`?

src/main/java/org/spongepowered/common/mixin/core/network/packet/MixinS48PacketResourcePackSend.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public void setResourcePack(ResourcePack pack) {
+        this.pack = pack;
+        this.url = ((SpongeResourcePack) pack).getUrlString();
+        this.hash = pack.getHash().or("");
+    }
+
+    @Override
+    public void setFakeHash() {
+        if (hash.length() == 0) {
```
Mumfrey: Unqualified access to instance variable `hash`

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -347,5 +357,42 @@ public void onDisconnectPlayer(IChatComponent reason, CallbackInfo ci) {
...
+                status = ResourcePackStatus.SUCCESSFULLY_LOADED;
+                break;
+            case FAILED_DOWNLOAD:
+                status = ResourcePackStatus.FAILED;
+                break;
+            default:
+                throw new AssertionError();
+        }
+        if (status.wasSuccessful().isPresent()) {
+            sentResourcePacks.remove(hash);
```
Mumfrey: A few unqualified field accesses on `sentResourcePacks` here.

src/main/java/org/spongepowered/common/resourcepack/SpongeResourcePackFactory.java
```diff
@@ -0,0 +1,79 @@
...
+            return new FileInputStream(new File(DimensionManager.getCurrentSaveRootDirectory(), uri.toString().substring("level://".length())));
+        }
+        return uri.toURL().openStream();
+    }
+
+    @Override
+    public ResourcePack fromUriUnchecked(URI uri) {
+        return new SpongeURIResourcePack(Preconditions.checkNotNull(uri, "uri"), Optional.<String>absent());
```
Mumfrey: I know it's weird, but conventions for sponge are that the `Preconditions` calls should be statically imported.

src/main/java/org/spongepowered/common/resourcepack/SpongeResourcePack.java
JBYoshi: I don't think it's necessary. I've done `checkNotNull()` on all of the calls in `SpongeResourcePackFactory`.

JBYoshi:
There is one more feature I'd like to add - caching the last resource pack sent to the client and not sending it multiple times in a row. Thoughts?

kashike:
@JBYoshi I'm not sure that's a good idea, unless it is a parameter (`boolean ignoreCache`, for example).
If a client denies a resource pack the first time, but enables and the server tries to send it again, nothing will happen

JBYoshi:
@kashike 
> If a client denies a resource pack the first time, but enables and the server tries to send it again, nothing will happen

I have an idea that will take care of that:
- There are two fields per player, `pendingPack` and `activePack`. The start out as `null`.
- When a resource pack `pack` is sent, if `pendingPack` is the new pack or `pendingPack` is null while `activePack` is the new pack, nothing happens.
- When the client denies or fails a resource pack download, `pendingPack` is set to `null`.
- When the client accepts a resource pack, `activePack` is set to `pendingPack`, then `pendingPack` is set to `null`.

The only question is if the download fails but the client doesn't say so, as the client might still say it succeeded.

Mumfrey:
Given that the client's behaviour cannot be trusted, I don't see any benefit in what you're suggesting. Sending the resource pack is not an intensive operation and preventing duplicate messages doesn't seem like something we should really be doing, it should be up to plugins to determine the behaviour they desire.

Zidane:
@JBYoshi 

Your proposed changes in your above comment are making ResourcePacks stateful (as you discussed with the others in the other issue). We gain nothing in caching something that could ultimately be 100% wrong.

JBYoshi:
Sounds good.

JBYoshi:
The build works on my machine (the only reason it fails here is because I didn't change the submodule). Ready for merge. Any objections?

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -95,6 +103,8 @@
 
     private Location lastMoveLocation = null;
 
+    private final Map<String, ResourcePack> sentResourcePacks = new HashMap<String, ResourcePack>();
```
simon816: My preference is to have the field close to where it's used, especially for classes like NetHandlerPlayServer where there are lots of different handlers scattered about. That's why I put `tmpQuitMessage` where it is. Don't know if the rest of the Sponge team approve but that's my preference.

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -347,5 +357,42 @@ public void onDisconnectPlayer(IChatComponent reason, CallbackInfo ci) {
         Sponge.getGame().getEventManager().post(event);
         event.getSink().sendMessage(event.getNewMessage());
     }
+        
+    @Inject(method = "handleResourcePackStatus", at = @At("HEAD"))
+    public void onResourcePackStatus(C19PacketResourcePackStatus packet, CallbackInfo ci) {
+        String hash = packet.hash;
+        PlayerResourcePackStatusEvent.ResourcePackStatus status;
+        ResourcePack pack = sentResourcePacks.get(hash);
```
simon816: Need to qualify `this.` i.e. `this.sentResourcePacks`

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
Mumfrey: Fields should be at the top of the class.

src/main/java/org/spongepowered/common/resourcepack/SpongeResourcePack.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.UUID;
+
+public abstract class SpongeResourcePack implements ResourcePack {
+
+    private final Optional<String> hash;
+    private final String id = UUID.randomUUID().toString();
+
+    public static final int HASH_SIZE = 40;
+
+    public SpongeResourcePack(Optional<String> hash) {
```
simon816: `Optional` isn't supposed to be used for method parameters.

JBYoshi: I think this is OK as it's not part of the API (?).

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
simon816: @Mumfrey Most of the time, yes, but not necessarily all of the time. [Consulting the Google Style Guide](https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s3.4.2-class-member-ordering) suggests that there is not a defined way to order members, but order them so that it is logical. To me, it's more logical to keep the fields close to the methods because in this particular mixin fields will not be shared outside the cluster of methods they are for.

src/main/java/org/spongepowered/common/resourcepack/SpongeResourcePack.java
simon816: It is outside of the API but it doesn't feel right. Still, it's good to keep a consistency throughout the implementation as well as the API.

src/main/java/org/spongepowered/common/resourcepack/SpongeURIResourcePack.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public SpongeURIResourcePack(String uri, String hash) throws URISyntaxException {
+        super(hash);
+        this.uri = new URI(uri);
+        this.name = getName0();
+    }
+
+    private String getName0() {
+        String name = uri.getPath();
```
simon816: Qualify `this.uri`

src/main/java/org/spongepowered/common/resourcepack/SpongeWorldResourcePack.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    public String getUrlString() {
+        return LEVEL_PACK_PROTOCOL + this.path;
+    }
+
+    @Override
+    public URI getUri() {
+        return uri;
```
simon816: Qualify `this.uri`

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -347,5 +357,42 @@ public void onDisconnectPlayer(IChatComponent reason, CallbackInfo ci) {
...
+                status = ResourcePackStatus.FAILED;
+                break;
+            default:
+                throw new AssertionError();
+        }
+        if (status.wasSuccessful().isPresent()) {
+            this.sentResourcePacks.remove(hash);
+        }
+        Sponge.getGame().getEventManager()
+                .post(SpongeEventFactory.createPlayerResourcePackStatus(Sponge.getGame(), (Player)playerEntity, pack, status));
```
simon816: Qualify `this.playerEntity`

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
gabizou: @simon816 That would be a change from the field declarations across all mixins, which to be honest, doesn't need to happen. Regardless over personal preference, making mixins become conforming is much better than suggesting yet another formatting and style change, especially when @Mumfrey is the code style/format manager across the implementations and API.

Zidane: I've made and agree with Mumfrey's law here. Put them at the top of the class.

bloodmc:
@JBYoshi please rebase this PR and test to make sure it works. Let me know when you feel it is ready.
Thanks!

JBYoshi:
@bloodmc I've finished the rebase. I don't have time to test tonight, but I'll do some testing tomorrow.

JBYoshi:
@bloodmc I've had some problems testing, see #142.

JBYoshi:
@bloodmc Ready for merge.

bloodmc:
@JBYoshi please squash your commits.

JBYoshi:
@bloodmc Done.

PR #101, opened 2015-07-22
Register the Sponge Dyeable Data Processor as a builder

DarkArc:
It's not much, the data isn't properly applied still because of [this issue](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/item/ItemsHelper.java#L100). This at least prevents an exception from being throw when you try and get the builder though.


gabizou:
I hate to be a snob, but you know how much I hate merge conflicts, especially with #94 essentially rewriting everything in `org.spongepowered.common.data` package, effectively refining the registrations as well?

DarkArc:
Ah, I wasn't aware of the large scale changes coming with #94... well, I guess this is pretty redundant then.

gabizou:
> Ah, I wasn't aware of the large scale changes coming with #94... well, I guess this is pretty redundant then.

Well, now you're aware ;) I should have the implementation functioning by the weekend or so.

DarkArc:
Awesome :) I'll have to take a closer look at it when I get the chance.

PR #102, opened 2015-07-24
Implement the view API

DDoS:
This doesn't implement the thread-safe buffers, there will come later with the multithreaded world editing API (itself following the world editing API).

Tests can be found here: https://github.com/SpongePowered/Cookbook/tree/master/Plugin/TestSuite/src/main/java/org/spongepowered/cookbook/plugin

API implemented: https://github.com/SpongePowered/SpongeAPI/pull/762


DDoS:
This is complete.

PR #109, opened 2015-07-26
Fix event-related casts failing on the client side.

JBYoshi:
Fixes SpongePowered/Sponge#311 by not firing the event on the client thread
Fixes SpongePowered/Sponge#312 by only casting to `Entity`, not `Player`

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/living/MixinEntityLiving.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/living/MixinEntityLiving.java
@@ -34,7 +34,6 @@ import org.objectweb.asm.Opcodes;
 import org.spongepowered.api.data.DataContainer;
 import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.entity.living.Agent;
-import org.spongepowered.api.entity.player.Player;
 import org.spongepowered.api.event.SpongeEventFactory;
 import org.spongepowered.api.event.entity.EntityLeashEvent;
 import org.spongepowered.api.event.entity.EntityUnleashEvent;
@@ -94,7 +93,7 @@ public abstract class MixinEntityLiving extends MixinEntityLivingBase implements
 
     @Inject(method = "interactFirst", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/EntityLiving;setLeashedToEntity(Lnet/minecraft/entity/Entity;Z)V"), locals = LocalCapture.CAPTURE_FAILEXCEPTION, cancellable = true)
     public void callLeashEvent(EntityPlayer playerIn, CallbackInfoReturnable<Boolean> ci, ItemStack itemstack) {
-        final EntityLeashEvent event = SpongeEventFactory.createEntityLeash(Sponge.getGame(), this, (Player)playerIn);
+        final EntityLeashEvent event = SpongeEventFactory.createEntityLeash(Sponge.getGame(), this, (Entity)playerIn);
         Sponge.getGame().getEventManager().post(event);
         if(event.isCancelled()) {
             ci.cancel();
--- a/src/main/java/org/spongepowered/common/mixin/core/item/MixinItemFishingRod.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/item/MixinItemFishingRod.java
@@ -54,10 +54,12 @@ public abstract class MixinItemFishingRod extends Item {
             player.fishEntity.handleHookRetraction();
         } else {
             EntityFishHook fishHook = new EntityFishHook(world, player);
-            if (!Sponge.getGame().getEventManager()
+            if (world.isRemote || !Sponge.getGame().getEventManager()
                     .post(SpongeEventFactory.createPlayerCastFishingLineEvent(Sponge.getGame(), (Player) player, (FishHook) fishHook))) {
                 world.playSoundAtEntity(player, "random.bow", 0.5F, 0.4F / (itemRand.nextFloat() * 0.4F + 0.8F));
-                world.spawnEntityInWorld(fishHook);
+                if (!world.isRemote) {
+                    world.spawnEntityInWorld(fishHook);
+                }
 
                 player.swingItem();
                 player.triggerAchievement(StatList.objectUseStats[Item.getIdFromItem(this)]);


src/main/java/org/spongepowered/common/mixin/core/item/MixinItemFishingRod.java
```diff
@@ -54,10 +54,12 @@ public ItemStack onItemRightClick(ItemStack itemStack, World world, EntityPlayer
             player.fishEntity.handleHookRetraction();
         } else {
             EntityFishHook fishHook = new EntityFishHook(world, player);
-            if (!Sponge.getGame().getEventManager()
+            if (world.isRemote || !Sponge.getGame().getEventManager()
```
Zidane: This is tempting to pull but we'd like this event to work client-side.

The bigger issue is that Player is missed into EntityPlayerMP.

@Aaron1011

JBYoshi: I assume you mean `EntityPlayerSP`?

JBYoshi: Honestly, it seems a bit confusing to fire events on the client thread at all. For example, if you run this code on the client:
```java
import org.spongepowered.api.event.Subscribe;
import org.spongepowered.api.event.world.WorldLoadEvent;

@Plugin(id = "test", name = "Test", version = "1.0")
public class Test {
	@Subscribe
	public void onWorldLoad(WorldLoadEvent e) {
		System.out.println(e.getWorld());
	}
}
```
you get `WorldClient`s printed as well as `WorldServer`s.

Zidane: @JBYoshi 

That is why the Platform class exists; to know when you are on the server or the client.

JBYoshi: I know, but I don't think plugin developers would think about it.

ryantheleach: Could you make a ClientAware plugin marker interface for plugins to receive "client side client" events? Otherwise they would only receive the "server side" client events? Or would that be too messy?

Zidane: @JBYoshi 

They'll have to learn.

@ryantheleach 

Way too messy. Almost all events would have that marker. If anything, I'd love @Subscribe to be changed to allow you to target a specific platform; as in your handler wouldn't be in the server or client's handlers list at all. Kinda how @SideOnly behaves in FML.

ryantheleach: I was thinking the marker would be on the plugin not the event. But if Subscribe defaulted to server with the option of being set to both or the client that could work as well.

JBYoshi: I agree with @ryantheleach. But we still have to figure out how to take care of `EntityPlayerSP`.

JBYoshi: I just thought of another way to use per-side event handlers:
- `enum EventSide { CLIENT, SERVER, BOTH, DEFAULT }`
- If an `EventSide` is present (and not `DEFAULT`) on the `@Subscribe` annotation, that value applies.
- Otherwise, if it is present on an annotation on the class (should this be added?), that value applies.
- Otherwise, if it is present on the `@Plugin` annotation, that value applies.
- Otherwise, it defaults to `BOTH`.

@Zidane @ryantheleach How does that sound?

Zidane: Introduction of another enum is un-necessary. The one available in Platform would suffice.

Scratch bullet 3, don't want to go down the route of "Side Only classes" at this time.

JBYoshi: @Zidane 
> Introduction of another enum is un-necessary. The one available in Platform would suffice.

How do you handle `BOTH` or `DEFAULT`? Values in annotations can't be set to `null`.

> Scratch bullet 3, don't want to go down the route of "Side Only classes" at this time.

Sounds good. If someone wants to do something like that, either something on the `@Plugin` annotation would suffice, or they can register the listener only on the client.
```java
if (game.getPlatform().getType() == Platform.Type.CLIENT) { 
    game.getEventManager().register(this, new ClientSideListener());
}
```

Zidane: The lack of specifying the property in the @Subscribe would equate to it being whatever would normally happen otherwise.

There are some events that will never fire on one side so introducing BOTH is equally un-necessary.

progwml6: I would prefer to force plugin developers to not register things on the client instead of going down the per side method/class stripping while we still have no client api's

It is my opinion that plugins should be informing users that they are useless on a particular side instead of just causing the errors and crashes that would end up occurring if you added side attributes to @plugin or put side only annotations on your main plugin class.

JBYoshi: @Zidane 
> The lack of specifying the property in the `@Subscribe` would equate to it being whatever would normally happen otherwise.

However, a value would have to be specified in either the `Subscribe.java` or the method using `@Subscribe`, and as far as I know, there's no way to differentiate between them.

> There are some events that will never fire on one side so introducing BOTH is equally un-necessary.

But aren't there some events that fire on both sides?

JBYoshi: @progwml6 This is not about receiving events only on the client or only on the dedicated server, this is about receiving events only on the client *thread* or server *thread*.

src/main/java/org/spongepowered/common/mixin/core/entity/living/MixinEntityLiving.java
```diff
@@ -94,7 +93,7 @@ public void setCanPickupItems(boolean canPickupItems) {
 
     @Inject(method = "interactFirst", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/EntityLiving;setLeashedToEntity(Lnet/minecraft/entity/Entity;Z)V"), locals = LocalCapture.CAPTURE_FAILEXCEPTION, cancellable = true)
     public void callLeashEvent(EntityPlayer playerIn, CallbackInfoReturnable<Boolean> ci, ItemStack itemstack) {
-        final EntityLeashEvent event = SpongeEventFactory.createEntityLeash(Sponge.getGame(), this, (Player)playerIn);
+        final EntityLeashEvent event = SpongeEventFactory.createEntityLeash(Sponge.getGame(), this, (Entity)playerIn);
```
JBYoshi: @Zidane Do you think this part could be merged? It shouldn't affect anything, except for fixing the crash.

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemFishingRod.java
Zidane: mmmm good point @JBYoshi . While the property in the annotation can be optional, the value must be set in the annotation's class itself. DEFAULT could be added to Type in platform (but only really makes sense in events).

Type.DEFAULT - External code doesn't care about the side, can do it on either/or.
Type.SERVER - External code only wants to execute on server thread
Type.CLIENT - External code only wants to execute on client thread

This is what it would mean for the @Subscribe annotation.

progwml6: I would think both or common would be a better name for the third entry

ST-DDT: We already have an `UNKNOWN` value, why not just use that?

JBYoshi: @Zidane @ST-DDT Only one question: `DEFAULT` vs `BOTH`.

ST-DDT: If i had to choose between those two I would use `BOTH` as describes more precisely what it does.
`DEFAULT` is just the default, but it does not say anything about the real content, you have to clarify this with javadocs. In addition to that from the definition its easier to understand when using `BOTH`

`side() default BOTH`
"Oh its an subjscription to BOTH sides by default."
vs
`side() default DEFAULT`
"I can see that its the default but I don't get what its meaning is."

JBYoshi: Why not add both? (As in `DEFAULT` and `BOTH`, not just `BOTH`.) Maybe `DEFAULT` should be called `INHERIT` or something similar.

Zidane: Except, again, its un-needed to have two more for events.

Really, by default it is DEFAULT which means "I don't give a flip about what side I get ran against, just run this code". Otherwise, I can choose to only run on the CLIENT or the SERVER.

JBYoshi: What if the plugin says its events are for the server thread, but an event handler in that plugin wants to be called on the client thread and the server thread?

JBYoshi:
Closing as it has been added by 428f9c508f71e25a9b9dd0da6557629288209eac and bd4c60ad43ac3cba1b1906d5c85efa9771a95086.

PR #111, opened 2015-07-27
Remove un-needed interface in MixinSubject.

JBYoshi:
`MixinSubject` implements `IMixinCommandSource`. However, it doesn't implement or use any of the methods from it.


simon816:
This is more suited for the minor issues list, #23

Issue #118, opened 2015-08-04
Unexpected world-teleportation behavior when "allow-nether" is set to false


zRaptor:
Teleporting to newly-generated worlds does not seem to work properly when `allow-nether` is set to `false` in `server.properties`.

This problem can be replicated by running the WorldsTest plugin that is available in the Cookbook. It is expected that, after the server has started, typing `/tpworld skylands` will take you to the Skylands world. Instead, you are teleported to a void of absolutely nothing. The world terrain *does* seem to be generated properly, in case that needs to be ruled out - opening the world in MCEdit shows the terrain.

Sponge: 1.8-1499-2.1DEV-536
Forge: v11.14.3.1450
Java: 1.8.0_51

simon816:
See https://github.com/SpongePowered/Sponge/issues/287
"allow-nether" really means "allow multi-dimensions"

zRaptor:
Ah, thanks - didn't see that issue. That's a rather strange design decision imo, but I'll go with it.

Closing this in favor of the issue you linked.

JBYoshi:
IMO teleporting the player to an empty world seems like a bad decision. I think it would be better to just report an error (kick the player? throw an exception?).

ST-DDT:
There shouldnt be a world to teleport him to.
Because you need a world to teleport a player if you don't have that world you cannot be teleported to it.

JBYoshi:
I think this was accidentally closed? The above commit was supposed to close SpongePowered/Sponge#118.

PR #121, opened 2015-08-05
Fix world loading when allow-nether is false

JBYoshi:
This pull request prevents worlds, other than the overworld, from loading if `allow-nether` in server.properties is set to false. See #118.


me4502:
I'm not sure if I agree with this.

That option disables the vanilla nether, which is still possible with multi world support.

JBYoshi:
@me4502 According to blood in SpongePowered/SpongeCommon#287:

> We will be handling "allow-nether" as allowMultipleDimensions to be consistent with Forge. 

And:
> I pushed a new branch 'world-update' which has the functionality you need via our world configs. To disable a world or dimension, locate the config under ./config/Sponge and set 'world-enabled' to false.

Zidane:
@JBYoshi 

Not sure if I want to even honor that setting via Vanilla. @bloodmc do we want to keep this to our configs or allow this as a "vanilla-override"

JBYoshi:
@Zidane @bloodmc Status?

bloodmc:
@JBYoshi Please update this branch with current master.

JBYoshi:
@bloodmc Done.

PR #124, opened 2015-08-09
Add Weather Events

AIOOB:
I implemented ```LightningStrikeEvent``` here instead of mixing it into ```EntityStruckByLightningEvent``` in Sponge as they are different events (the forge one deals with the entity being struck, the sponge one the lightning striking entities).

### DIFF
--- a/src/main/java/org/spongepowered/common/block/SpongeBlockSnapshot.java
+++ b/src/main/java/org/spongepowered/common/block/SpongeBlockSnapshot.java
@@ -379,6 +379,26 @@ public class SpongeBlockSnapshot implements BlockSnapshot {
     }
 
     @Override
+    public int hashCode() {
+        return Objects.hashCode(this.worldUniqueId, this.pos, this.blockState, this.extraData);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+        final SpongeBlockSnapshot other = (SpongeBlockSnapshot) obj;
+        return Objects.equal(this.worldUniqueId, other.worldUniqueId)
+                && Objects.equal(this.pos, other.pos)
+                && Objects.equal(this.blockState, other.blockState)
+                && Objects.equal(this.extraData, other.extraData);
+    }
+
+    @Override
     public String toString() {
         return Objects.toStringHelper(this)
                 .add("worldUniqueId", this.worldUniqueId)
--- a/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
@@ -128,8 +128,12 @@ import org.spongepowered.api.entity.living.player.Player;
 import org.spongepowered.api.entity.living.player.User;
 import org.spongepowered.api.entity.projectile.EnderPearl;
 import org.spongepowered.api.entity.projectile.source.ProjectileSource;
+import org.spongepowered.api.entity.weather.Lightning;
 import org.spongepowered.api.event.Cancellable;
+import org.spongepowered.api.event.Event;
 import org.spongepowered.api.event.SpongeEventFactory;
+import org.spongepowered.api.event.SpongeEventFactoryUtils;
+import org.spongepowered.api.event.action.LightningEvent;
 import org.spongepowered.api.event.block.ChangeBlockEvent;
 import org.spongepowered.api.event.block.NotifyNeighborBlockEvent;
 import org.spongepowered.api.event.cause.Cause;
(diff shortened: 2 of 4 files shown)


src/main/java/org/spongepowered/common/event/weather/SpongeWeatherChangeEvent.java
```diff
@@ -0,0 +1,85 @@
...
+import org.spongepowered.api.Game;
+import org.spongepowered.api.event.AbstractEvent;
+import org.spongepowered.api.event.weather.WeatherChangeEvent;
+import org.spongepowered.api.util.annotation.NonnullByDefault;
+import org.spongepowered.api.world.weather.Weather;
+import org.spongepowered.api.world.weather.WeatherUniverse;
+import org.spongepowered.common.Sponge;
+
+@NonnullByDefault
+public class SpongeWeatherChangeEvent extends AbstractEvent implements WeatherChangeEvent {
```
Zidane: This shouldn't be implemented but rather placed into the SpongeEventFactory.

src/main/java/org/spongepowered/common/mixin/core/entity/weather/MixinEntityLightningBolt.java
```diff
@@ -57,12 +71,31 @@ public ExpirableData getExpiringData() {
...
-    public void onOnUpdate(CallbackInfo ci) {
+    public void onDoDamage(CallbackInfo ci) {
         if (this.effect) {
             ci.cancel();
         }
     }
 
+    @Inject(method = "onUpdate()V", at = @At(value = "INVOKE",
+            target = "Lnet/minecraft/world/World;setBlockState(Lnet/minecraft/util/BlockPos;Lnet/minecraft/block/state/IBlockState;)Z"))
+    public void onStrikeBlock(World world, BlockPos pos, IBlockState blockState) {
```
simon816: Need to fully qualify `World` (`net.minecraft.world.World`) because the imported world is the sponge API world.

src/main/java/org/spongepowered/common/mixin/core/entity/weather/MixinEntityLightningBolt.java
```diff
@@ -57,12 +71,31 @@ public ExpirableData getExpiringData() {
     @Inject(method = "onUpdate()V", at = {@At(value = "NEW", args = "class=net.minecraft.util.BlockPos"),
             @At(value = "NEW", args = "class=net.minecraft.util.AxisAlignedBB")}, cancellable = true
         )
```
me4502: This should be on the previous line

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -521,6 +523,20 @@ public void forecast(Weather weather, long duration) {
         }
     }
 
+    private Weather prevWeather = getWeather();
```
me4502: @SpongePowered/developers Should this be at the top of the class?

Zidane: Yes

bloodmc:
@AIOOB These injections do not work. Please test before creating a pull request.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -539,6 +559,23 @@ public void forecast(Weather weather, long duration) {
         }
     }
 
+    @Inject(method = "updateWeather", at = @At(value = "RETURN"))
+    public void onUpdateWeatherReturn(CallbackInfo ci) {
+        Weather weather = getWeather();
+        if (this.prevWeather != weather) {
```
Aaron1011: `prevWeather` can be null here- it might be necessary to update the API to make `initalWeather an `Optional`

src/main/java/org/spongepowered/common/block/SpongeBlockSnapshot.java
```diff
@@ -365,6 +365,26 @@ public int getUpdateFlag() {
     }
 
     @Override
+    public int hashCode() {
+        return Objects.hashCode(this.worldUniqueId, this.pos, this.blockState, this.extraData);
```
gabizou: This has nothing to do with weather events.

simon816:
@AIOOB Will this be getting an update soon?

AIOOB:
Yes, I have been really busy recently but after the 11th I should be able to finish everything. Currently the weather events are all working fine but lightning is presenting more of an issue since the entity captures have been added.

src/main/java/org/spongepowered/common/block/SpongeBlockSnapshot.java
AIOOB: It does due to the contains check in EntityLightningBolt for the block transaction as the objects are not necessarily the same I needed to implement this.

AIOOB:
I have checked this as much as I can at the moment and it seems to be done (I cannot check server side as my computer is not powerful enough), any suggestions are welcome. There are a few oddities: this only registers the change in weather one tick after the weather changes due to minecraft not assigning a time to the duration of a weather event until it has started and it also delays the effect of lightning until it dies due to the requirements of the sponge event, which is noticeable if you are looking for it.

AIOOB:
Fixes #311

Issue #62, opened 2015-06-03
Implementation of the InteractEntityEvent


TBlueF:
According to [this](https://github.com/SpongePowered/Sponge/issues/274) Event Progress Tracker and my Plugin, the **PlayerInteractEntityEvent** is not jet implemented.
_Deamon5550 said [here](https://github.com/SpongePowered/Sponge/issues/272): "Opening issues like these does help us prioritize which events we're implementing." So i thougt i'll open this issue, since this is the last missing event i need to finish my plugin and start my server :)_

For the sake of completeness:
_Sponge Build: 1.8-1371-2.1DEV-467
Forge Build: forge-1.8-11.14.1.1397
Client Type/Version (Forge/Vanilla and version) : Vanilla 1.8.6
Installed Plugins and/or Mods: Sponge, My Plugin, (and DynmapForge)
Java Version: 1.8.0_40
Operating System: Windows 7 64 bit_

(Reopens [this Issue](https://github.com/SpongePowered/Sponge/issues/286))

RafVDL:
I quickly wrote this (it needs to be inserted into MixinNetHandlerPlayServer):
```java
@Inject(method = "processUseEntity", at = @At(value = "INVOKE", target =
        "net/minecraft/network/play/client/C02PacketUseEntity.getAction()" +
                "Lnet/minecraft/network/play/client/C02PacketUseEntity$Action;", ordinal = 0), cancellable = true, locals = LocalCapture.CAPTURE_FAILSOFT)
public void processUseEntity(C02PacketUseEntity packetIn, CallbackInfo info, WorldServer worldserver, Entity
        entity, boolean flag, double d0) {
    EntityInteractionType type = null;
    Vector3d location = null;
    switch (packetIn.getAction()) {
        case ATTACK:
            type = EntityInteractionTypes.ATTACK;
            break;
        case INTERACT_AT:
            location = new Vector3d(packetIn.getHitVec().xCoord, packetIn.getHitVec().yCoord, packetIn.getHitVec().zCoord);
        case INTERACT:
            type = EntityInteractionTypes.USE;
            break;
    }
    
    if (type == null) return;

    if (Sponge.getGame().getEventManager().post(SpongeEventFactory.createPlayerInteractEntity(Sponge.getGame(),
            this.getPlayer(), (org.spongepowered.api.entity.Entity) entity, type, location))) {
        info.cancel();
    }
}
```
[...]

PR #126, opened 2015-08-09
Add PlayerInteractEntityEvent

pagoru:
This code implements the player entity event.
See https://github.com/SpongePowered/SpongeCommon/issues/62


src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -347,5 +350,32 @@ public void onDisconnectPlayer(IChatComponent reason, CallbackInfo ci) {
         Sponge.getGame().getEventManager().post(event);
         event.getSink().sendMessage(event.getNewMessage());
     }
+    
+    @Inject(method = "processUseEntity", at = @At(value = "INVOKE", target =
+             "net/minecraft/network/play/client/C02PacketUseEntity.getAction()Lnet/minecraft/network/play/client/C02PacketUseEntity$Action;",
+             ordinal = 0), cancellable = true, locals = LocalCapture.CAPTURE_FAILSOFT)
+    public void processUseEntity(C02PacketUseEntity packetIn, CallbackInfo info, WorldServer worldserver, Entity entity, boolean flag, double d0) {
```
simon816: I believe the `boolean flag` parameter needs the `@Coerce` annotation. `double` is fine I think

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -347,5 +350,32 @@ public void onDisconnectPlayer(IChatComponent reason, CallbackInfo ci) {
...
+        EntityInteractionType type = null;
+        Vector3d location = null;
+        switch (packetIn.getAction()) {
+            case ATTACK:
+                type = EntityInteractionTypes.ATTACK;
+                break;
+            case INTERACT_AT:
+                location = new Vector3d(packetIn.getHitVec().xCoord, packetIn.getHitVec().yCoord, packetIn.getHitVec().zCoord);
```
simon816: `location = VecHelper.toVector(packetIn.getHitVec());` is neater

simon816: Also, you're not setting the `type` variable in this block, should be `EntityInteractionTypes.USE`

Zidane:
@simon816 

Biggest issue I see with this PR (besides what you've pointed out) is the fact that Forge already has an event for this (pretty sure it does) and therefore this really should go into SpongeVanilla (and mixed into the Forge event in Sponge)

pagoru:
I fucked up the commit... D=
I need remember the "branches"...

PR #128, opened 2015-08-13
Implement API for controlling spawn protection

zml2008:
Additionally, this change assigns the permission
minecraft.spawn-protection.override to override spawn protection for a
specific user.

@bloodmc How am I supposed to do config handling for this change?

### DIFF
--- a/src/main/java/org/spongepowered/common/config/SpongeConfig.java
+++ b/src/main/java/org/spongepowered/common/config/SpongeConfig.java
@@ -841,6 +841,10 @@ public class SpongeConfig<T extends SpongeConfig.ConfigBase> {
         @Setting(value = WORLD_FLOWING_LAVA_DECAY, comment = "Lava behaves like vanilla water when source block is removed")
         private boolean flowingLavaDecay = false;
 
+        @Setting(value = "spawn-protection-radius", comment = "Radius in blocks around spawn to protect from being built in. -1 means user server "
+                + "default")
+        private int spawnProtectionRadius = -1;
+
         @Setting(value = WORLD_ENABLED, comment = "Enable if this world should be registered.")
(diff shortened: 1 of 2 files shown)


ryantheleach:
Whilst I dislike the vanilla behaviour of disabling spawn protection if the op list is empty, forcing it to be enabled for all worlds by default isn't a sane solution either. This will cause issues in the nether with unlightable portals within the spawn protection potentially trapping playesr.

I'd suggest either make the default 0 spawn protection, or only the primary world being protected.

JBYoshi:
IMO the default world's spawn protection should default to the same as vanilla and the other worlds should default it to off.

gabizou:
@zml2008 Anything on updating this or whatnot?

Minecrell:
@zml2008 Do we still need this with https://github.com/SpongePowered/SpongeCommon/commit/6b2e66d5fcebae2cad7ab0e0d31c0c0b40b8ab3a in master?

zml2008:
On March 19, 2016 5:55:21 PM GMT+07:00, Minecrell <notifications@github.com> wrote:
>@zml2008 Do we still need this with
>https://github.com/SpongePowered/SpongeCommon/commit/6b2e66d5fcebae2cad7ab0e0d31c0c0b40b8ab3a
>in master?
>
>---
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/SpongePowered/SpongeCommon/pull/128#issuecomment-198682339

Possibly, it adds configurability for spawn protection so things are more flexible for plugins

parlough:
Going to close this for now as it has been sitting here since early 2016. I will work on something for this soon as there have been various issues related to this.

PR #133, opened 2015-08-17
[WIP] TargetedLocationDataProcessor implementation

kamcio96:
You should check it ;)

### DIFF
--- a/src/main/java/org/spongepowered/common/data/processor/data/TargetedLocationDataProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/data/TargetedLocationDataProcessor.java
@@ -1,0 +1,136 @@
+package org.spongepowered.common.data.processor.data;
+
+import com.google.common.base.Optional;
+import net.minecraft.entity.Entity;
+import net.minecraft.entity.EntityLiving;
+import net.minecraft.pathfinding.PathPoint;
+import org.spongepowered.api.data.*;
+import org.spongepowered.api.data.key.Key;
+import org.spongepowered.api.data.key.Keys;
+import org.spongepowered.api.data.manipulator.immutable.ImmutableTargetedLocationData;
+import org.spongepowered.api.data.manipulator.mutable.TargetedLocationData;
+import org.spongepowered.api.data.merge.MergeFunction;
+import org.spongepowered.api.data.value.BaseValue;
+import org.spongepowered.api.service.persistence.InvalidDataException;
+import org.spongepowered.api.world.Location;
+import org.spongepowered.api.world.World;
+import org.spongepowered.common.Sponge;
+import org.spongepowered.common.data.DataProcessor;
+import org.spongepowered.common.data.manipulator.immutable.ImmutableSpongeTargetedLocationData;
+import org.spongepowered.common.data.manipulator.mutable.SpongeTargetedLocationData;
+import org.spongepowered.common.data.util.DataUtil;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class TargetedLocationDataProcessor implements DataProcessor<TargetedLocationData, ImmutableTargetedLocationData> {
+
+    @Override
+    public boolean supports(DataHolder dataHolder) {
+        return dataHolder instanceof Entity;
+    }
+
+    @Override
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/data/processor/data/TargetedLocationDataProcessor.java
```diff
@@ -0,0 +1,136 @@
...
+public class TargetedLocationDataProcessor implements DataProcessor<TargetedLocationData, ImmutableTargetedLocationData> {
+
+    @Override
+    public boolean supports(DataHolder dataHolder) {
+        return dataHolder instanceof Entity;
+    }
+
+    @Override
+    public Optional<TargetedLocationData> from(DataHolder dataHolder) {
+        if(dataHolder instanceof EntityLiving) {
```
gabizou: TagetedLocation should also consider `EyeofEnder`s, `Compasses`, etc.

JBYoshi: `Compass`es all point to the same location (as far as each client is concerned; each client can be given a different compass target). It's not per-`ItemStack`, it's per-`Player`.

gabizou: Right, for Players, you might want to set the fake "spawn location" then for the compass to point.

JBYoshi: @gabizou But what if you have multiple compasses in your hotbar? You can see both, but they would have to point in the same direction, regardless of the position each one is set to.

gabizou: @JBYoshi If you have multiple compasses, they use the "respawn" location for the player, however, we can use the `TargetedLocationData` as I've been suggesting for a while now.

src/main/java/org/spongepowered/common/data/processor/data/TargetedLocationDataProcessor.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @Override
+    public ImmutableTargetedLocationData createImmutable() {
+        return null; // new ImmutableSpongeTargetedLocationData() ?
+    }
+
+    @Override
+    public Optional<TargetedLocationData> createFrom(DataHolder dataHolder) {
+        return from(dataHolder); //something more?
```
Aaron1011: You always want to return a DataMapiulator here if the data holder is applicable. The difference between this method and `from` is that while `from` can return absent when passed a supported data holder (if no pathfinding path is set, for example), this method shouldn't.

src/main/java/org/spongepowered/common/data/processor/data/TargetedLocationDataProcessor.java
JBYoshi: @gabizou I think I'm just misunderstanding what you said. It seemed like you said that `TargetedLocationData` would apply to `ItemStack`s containing compasses.

gabizou: @JBYoshi `TargetedLocationData` does have some javadocs as to the "expected" use cases:
```
An {@link DataManipulator} handling the supposed targetted
{@link Location}. Usually for the case of {@link EnderPearl}s, the
targeted {@link Location} is where the {@link EnderPearl} will
move towards until it's expiration time. In the case of
{@link ItemStack}s of type {@link ItemTypes#COMPASS}, the targeted
{@link Location} is where the compass will point towards.
```

gabizou:
You also need to register in the `SpongeGameRegistry` and in the `KeyRegistry`.

gabizou:
Please update implementation.

kamcio96:
I want create new PR for this data. Where minecraft store player's compass target?

JBYoshi:
Minecraft doesn't store the compass target on the server. It just sends a packet to the client when the spawn position changes.

kamcio96:
Thanks, I will use mixin ;)

PR #134, opened 2015-08-17
[WIP] TameableData Implementation.

ryantheleach:
I'm unsure about the API bump commit, will probably remove it when I finish the rest of the implementation tomorrow. Just wanted to throw in what I had so far.


SpongeAPI
```diff
@@ -1 +1 @@
-Subproject commit 14c4512bb44ae77f76c5c148d182ac81087bae31
+Subproject commit cff02b98a0bae3d6cbc0c43a6021b21b563528ab
```
gabizou: Why are you changing the submodule?

kenzierocks: Perhaps it's https://github.com/SpongePowered/SpongeAPI/commit/a8023468dc1eca47e904d997e763929982e67c9a?

ryantheleach: If it was, wouldn't the commit hash map?

JBYoshi: After taking a look, it seems the commit is exactly the same as SpongePowered/SpongeAPI@0933b33a7f2723b8f1fc9183ce1d8374cd20a7af, which happens to be the commit that `feature/data` represents right now. Not sure what went wrong.

gabizou: @ryantheleach, @JBYoshi I wouldn't always directly reference other branches as rebases will always change the commit hash. In this case, since `TameableData` does have changes in the branch, I understand the submodule change.

SpongeAPI
```diff
@@ -1 +1 @@
-Subproject commit f164da66e3d0294f30379810ad9b82afa798a519
+Subproject commit 391a04ceaa7d0af2783b5986bf76e65dc692cdd2
```
gabizou: Shouldn't have to update the submodule.

ryantheleach: I attempted to remove the changes made to the sub module using the instructions given in IRC to hassan.

ryantheleach:
I'm unable to re-target the branch this PR is to, so I'll have to recreate it.

PR #151, opened 2015-09-01
[WIP] Add ScheduledExecutorService

Kiskae:
Companion PR to SpongePowered/SpongeAPI#829

Moves the implementation to SpongeCommon so some required information is available.

### DIFF
--- a/src/main/java/org/spongepowered/common/service/scheduler/SchedulerExecutionService.java
+++ b/src/main/java/org/spongepowered/common/service/scheduler/SchedulerExecutionService.java
@@ -240,6 +240,8 @@ public class SchedulerExecutionService extends AbstractExecutorService implement
         protected void done() {
             super.done();
 
+            if (isCancelled()) return;
+
             try {
                 get();
             } catch (InterruptedException ignored) {
--- a/src/main/java/org/spongepowered/common/service/scheduler/SchedulerExecutionService.java
+++ b/src/main/java/org/spongepowered/common/service/scheduler/SchedulerExecutionService.java
@@ -53,7 +53,14 @@ public class SchedulerExecutionService extends AbstractExecutorService implement
 
     @Override
     public void shutdown() {
-        //NOOP
+        // Since this class is delegating its work to SchedulerService
+        // and we have no way to stopping execution without keeping
+        // track of all the submitted tasks, it makes sense that
+        // this ExecutionService cannot be shut down.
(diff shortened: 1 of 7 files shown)


Kiskae:
Did have to update the API submodule for this PR, so the recent Java8 changes probably broke this build.

Kiskae:
Replaced currentTimeMillis with nanoTime as per https://github.com/SpongePowered/SpongeAPI/pull/829#issuecomment-136861606 , should improve long-term stability of the system.

src/main/java/org/spongepowered/common/service/scheduler/SchedulerExecutionService.java
```diff
@@ -0,0 +1,296 @@
...
+            // A repeating task that is done but hasn't been cancelled has
+            // failed exceptionally. Following the contract of
+            // ScheduledExecutorService, this means the task has to be stopped.
+            if (!isCancelled() && this.owningTask != null) {
+                this.owningTask.cancel();
+            }
+
+            // The ThrowingFutureTask will make sure the exception gets thrown,
+            // if there is any.
+            super.done();
```
PieGuy128: The one thing that I could see being a problem here is that you're making a call to get() when no value has been set.  Would this not make for some really long hangs?

Kiskae: done() is the method that is called last when handling the resolving of the Future, so the state will either have a value, an exception or be cancelled.

PieGuy128: I know, but seeing as you're extending ThrowingFutureTask whose done() method calls get() when no value has been set is what I'm concerned about.  Sure, it will throw any exceptions that have been thrown by the underlying task should they arise, but if no exception has been thrown, then then "listening" thread will block.  As per the [FutureTask JavaDocs][1], get() will wait for a result to be set.

Hopefully that clarifies my concerns!
Cheers!

[1]: http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/FutureTask.html#get--

Kiskae:
I believe the current state of the PR has completely and correctly converted the previous millisecond-based implementation to nanoseconds, but a secondary review would probably be a good idea :P

Kiskae:
Its been merged, so closing PR

PR #161, opened 2015-09-09
Stop :uploadArchive from uploading non-existent files

lukegb:
For some reason :uploadArchive keeps trying to upload sdkZip in SpongeVanilla, even where it doesn't exist. This patch forces gradle to check that sdkZip actually executed before adding it to the archives artifacts.

There might be a better way of doing this?

### DIFF
--- a/gradle/implementation.gradle
+++ b/gradle/implementation.gradle
@@ -151,10 +151,11 @@ task sdkZip(type: Zip) {
            ])
         }
     }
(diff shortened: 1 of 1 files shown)


lukegb:
This is actually just broken - it will never build sdkZip even if it's supported.

lukegb:
Updated my branch with a proper fix that should work for both Sponge and SpongeVanilla.

PR #164, opened 2015-09-12
implementation of LoreData and PagedData

lucdon:
It is 100% completed and I did test it. It should be fully functional.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/key/KeyRegistry.java
+++ b/src/main/java/org/spongepowered/common/data/key/KeyRegistry.java
@@ -68,6 +68,8 @@ public class KeyRegistry {
         keyMap.put("max_air", makeSingleKey(Integer.class, Value.class, of("MaxAir")));
         keyMap.put("remaining_air", makeSingleKey(Integer.class, Value.class, of("RemainingAir")));
         keyMap.put("game_mode", makeSingleKey(GameMode.class, Value.class, of("GameMode")));
+        keyMap.put("item_lore", makeListKey(Text.class, of("ItemLore")));
+        keyMap.put("book_pages", makeListKey(Text.class, of("BookPages")));
         RegistryHelper.mapFields(Keys.class, keyMap);
     }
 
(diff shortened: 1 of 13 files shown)


src/main/java/org/spongepowered/common/data/key/KeyRegistry.java
```diff
@@ -1,74 +1,76 @@
-/*
- * This file is part of Sponge, licensed under the MIT License (MIT).
```
gabizou: Yeah.. this shouldn't be happening. please fix your formatting changes. i.e. don't change our formatting.

src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongePagedData.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public ImmutableSpongePagedData(List<Text> pages) {
+        super(ImmutablePagedData.class);
+        this.pages = ImmutableList.copyOf(pages);
+        registerGetters();
+    }
+
+    @Override
+    public ImmutableListValue<Text> pages() {
+        return ImmutableDataCachingUtil.getValue(ImmutableSpongeListValue.class, Keys.BOOK_PAGES, this.pages, ImmutableList.of(Texts.of()));
```
gabizou: No. don't cache this.

src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongePagedData.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Override
+    public ImmutableSpongePagedData copy() {
+        return new ImmutableSpongePagedData(pages);
```
gabizou: `return this;`

src/main/java/org/spongepowered/common/data/util/NbtDataUtil.java
```diff
@@ -1,115 +1,191 @@
-/*
```
gabizou: Again here...

PR #173, opened 2015-09-23
Implement Elder data

ZephireNZ:
Implementation of `ElderData`.

Implementation of `ImmutableElderData`

Implementation of the `ElderDataBuilder`
- [x] In the `build()` method, checked if the container contained the keys required, if it doesn't, return `Optional.empty()`

Implementation of the `ElderDataProcessor`(s):

Registration:
- [x] Registered the `Key` correctly by `ELDER_GUARDIAN` in the `KeyRegistry`

Test Plugin code:

```java
@Plugin(id = "data-test", name = "DataTest")
public class DataTestPlugin {

    @Listener
    public void onSpawn(SpawnEntityEvent event) {
        if (entity.supports(ElderData.class)) {
            final ElderData elderData = entity.getOrCreate(ElderData.class).get();
            elderData.set(Keys.ELDER_GUARDIAN, !elderData.elder().get());
            entity.offer(elderData);
        }
        final Optional<Value<Boolean>> optional = entity.getValue(Keys.ELDER_GUARDIAN);
        if (optional.isPresent()) {
            if (optional.get().get()) {
                System.out.println("Holy cow! An Elder Guardian!");
            } else {
                System.out.println("Oh... just another Guardian.");
            }
        }
    }
    
    @Listener
    public void onDeath(DestructEntityEvent.Death event) {
        if(entity instanceof Guardian) {
            if(entity.get(Keys.ELDER_GUARDIAN).get()) {
                System.out.println("Elder guardian died!");
            }
        }
    }
}
```


src/main/java/org/spongepowered/common/data/builder/manipulator/mutable/entity/ElderDataBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+import org.spongepowered.api.service.persistence.InvalidDataException;
+import org.spongepowered.common.data.manipulator.mutable.entity.SpongeElderData;
+
+public class ElderDataBuilder implements DataManipulatorBuilder<ElderData, ImmutableElderData> {
+
+    @Override
+    public Optional<ElderData> build(DataView container) throws InvalidDataException {
+        if (container.contains(Keys.ELDER_GUARDIAN.getQuery())) {
+            Optional<Boolean> elder = container.getBoolean(Keys.ELDER_GUARDIAN.getQuery());
+            if (elder.isPresent()) {
```
gabizou: you just checked if the container contains the query....

Also, the contract is that if the container doesn't have the information to build the manipulator (or `DataSerializable`), throw an `InvalidDataException`. Not doing so kinda breaks the contract here.

src/main/java/org/spongepowered/common/data/processor/data/entity/ElderDataProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+import org.spongepowered.api.data.manipulator.mutable.entity.ElderData;
+import org.spongepowered.api.data.manipulator.mutable.entity.ScreamingData;
+import org.spongepowered.common.data.manipulator.mutable.entity.SpongeElderData;
+import org.spongepowered.common.data.manipulator.mutable.entity.SpongeScreamingData;
+import org.spongepowered.common.data.processor.common.AbstractEntityDataProcessor;
+import org.spongepowered.common.data.util.DataUtil;
+
+import java.util.Map;
+
+public class ElderDataProcessor extends AbstractEntityDataProcessor<EntityGuardian, ElderData, ImmutableElderData> {
```
gabizou: Better to extend `AbstractEntitySingleDataProcessor` since you focus on one single piece of data, the `boolean` value of whether a `Guardian` is an elder or not.

src/main/java/org/spongepowered/common/data/processor/value/entity/ElderValueProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    protected Value<Boolean> constructValue(Boolean defaultValue) {
+        return new SpongeValue<Boolean>(Keys.ELDER_GUARDIAN, false, defaultValue);
+    }
+
+    @Override
+    public Optional<Boolean> getValueFromContainer(ValueContainer<?> container) {
+        if (container.supports(Keys.ELDER_GUARDIAN)) {
```
gabizou: no..... don't call `container.supports` ever.... You're already in the value processor, you shouldn't call anything else that would require the `ValueContainer` to re-search for the very same `ValueProcessor` for a simple check.

Replacing with `if (supports(container)) {` is acceptable, though alternatively, you could just use `if (container instanceof EntityGuardian) {`.

src/main/java/org/spongepowered/common/data/processor/value/entity/ElderValueProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Override
+    public boolean supports(ValueContainer<?> container) {
+        return container instanceof EntityGuardian;
+    }
+
+    @Override
+    public DataTransactionResult offerToStore(ValueContainer<?> container, Boolean value) {
+        final ImmutableValue proposedValue = new ImmutableSpongeValue(Keys.ELDER_GUARDIAN, value);
+        final ImmutableValue oldValue = getApiValueFromContainer(container).get().asImmutable();
```
gabizou: You're getting the previous value before you're checking whether the `container` is an `EntityGuardian`.

src/main/java/org/spongepowered/common/data/builder/manipulator/mutable/entity/ElderDataBuilder.java
ZephireNZ: Okay, in that case should I change ScreamingData in this PR too then? That's also incorrect, which is where this came from haha

src/main/java/org/spongepowered/common/data/key/KeyRegistry.java
```diff
@@ -112,6 +112,7 @@ public static void registerKeys() {
         keyMap.put("item_enchantments", makeListKey(ItemEnchantment.class, of("ItemEnchantments")));
         keyMap.put("banner_patterns", makeListKey(PatternLayer.class, of("BannerPatterns")));
         keyMap.put("banner_base_color", makeListKey(DyeColor.class, of("BannerBaseColor")));
+        keyMap.put("elder", makeSingleKey(Boolean.class, Value.class, of("Elder")));
```
gabizou: Did you test this?

src/main/java/org/spongepowered/common/data/builder/manipulator/mutable/entity/ElderDataBuilder.java
gabizou: Yes. Any PR that isn't taking advantage of the various reductions in boilerplate code with `DataProcessor`s in general I'm requiring to be updated. I've considered nagging, if not just handling the issues myself, but I want the contributors to understand exactly what they need to be doing to implement the processors correctly.

ZephireNZ:
```
        if(container.contains(Keys.ELDER_GUARDIAN.getQuery())) {
            Boolean elder = DataUtil.getData(container, Keys.ELDER_GUARDIAN);
            return Optional.<ElderData>of(new SpongeElderData(elder));
        }
        return Optional.absent();
```

That's the better way, corect?

gabizou: The better way is simply:
```java
public Optional<ElderData> build(DataView container) throws InvalidDataException {
  final boolean isElder = DataUtil.getData(container, Keys.ELDER_GUARDIAN);
  return Optional.<ElderData>of(new SpongeElderData(isElder));
}
```

ZephireNZ:
Hopefully this should satisfy the test plugin requirements, it gets and offers data with manipulators and values:
```java
@Plugin(id = "data-test", name = "DataTest")
public class DataTestPlugin {

    @Listener
    public void onSpawn(SpawnEntityEvent event) {
        if(entity instanceof Guardian) {
            entity.offer(Keys.ELDER_GUARDIAN, true);
        }
        if(entity instanceof Enderman) {
            ScreamingData data = entity.get(ScreamingData.class).get();
            data.screaming().set(true);
            entity.offer(data);
        }
    }

    @Listener
    public void onDeath(DestructEntityEvent.Death event) {
        if(entity instanceof Guardian) {
            if(entity.get(Keys.ELDER_GUARDIAN).get()) {
                event.setMessage(Texts.of("Elder Guardian died!"));
            }
        }
        if(entity instanceof Enderman) {
            ScreamingData data = entity.get(ScreamingData.class).get();
            if(data.screaming().get()) {
                event.setMessage(Texts.of("AAAAAAAAAAAH!"));
            }
        }
    }
}
```

src/main/java/org/spongepowered/common/data/key/KeyRegistry.java
```diff
@@ -112,6 +112,7 @@ public static void registerKeys() {
         keyMap.put("item_enchantments", makeListKey(ItemEnchantment.class, of("ItemEnchantments")));
         keyMap.put("banner_patterns", makeListKey(PatternLayer.class, of("BannerPatterns")));
         keyMap.put("banner_base_color", makeListKey(DyeColor.class, of("BannerBaseColor")));
+        keyMap.put("elder", makeSingleKey(Boolean.TYPE, Value.class, of("Elder")));
```
gabizou: `!"elder".equalsIgnoreCase("ELDER_GUARDIAN")` which is why I was hoping that by testing with the proper methods (including manipulating the data through keys of the `ElderData` `DataManipulator`), you'd find that there's an issue with the key being registered here :wink: 

Basically, the keymap,  you're not putting the right field name for the key you're wanting to register: [`Keys.ELDER_GUARDIAN`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/data/key/Keys.java#L214).

ZephireNZ: Oh, my mistake - assuming it was the name of the DataQuery, whoops.

Only just got an instance going so couldn't test - builds were failing because I hadn't updated the DamageSource Access Transformers.

gabizou:
> Hopefully this should satisfy the test plugin requirements, it gets and offers data with manipulators and values:

As I mentioned in the latest comment about the `KeyRegistry`, you shouldn't just test the processors, but the manipulators, because manipulators don't just handle a data value, they deal with it through a `Key`. If the `Key` is `null`, there's going to be exceptions down the line.

What you could do with the onSpawn:
```java
  @Listener
  public void onSpawn(SpawnEntityEvent event) {
    if (entity.supports(ElderData.class)) {
      final ElderData elderData = entity.getOrCreate(ElderData.class).get();
      elderData.set(Keys.ELDER_GUARDIAN, elderData.elder().get() ? false : true);
      entity.offer(elderData);
    }
    final Optional<Value<Boolean>> optional = entity.getValue(Keys.ELDER_GUARDIAN);
    if (optional.isPresent()) {
      if (optional.get()) {
        System.out.println("Holy cow! An Elder Guardian!");
      } else {
        System.out.println("Oh... just another Guardian.");
      }
    }
  }


```

src/main/java/org/spongepowered/common/data/builder/manipulator/mutable/entity/ElderDataBuilder.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public ElderData create() {
+        return new SpongeElderData(false);
+    }
+
+    @Override
+    public Optional<ElderData> createFrom(DataHolder dataHolder) {
+        if (dataHolder.supports(Keys.ELDER_GUARDIAN)) {
```
gabizou: Don't delegate from the data builder to the value processor here. This should be exactly the same as the data processor's `createFrom(DataHolder)` method.

ZephireNZ:
Moving the Screaming data to a separate PR, seems better than lumping these two together.

gabizou:
Can you update the description as to what this PR is exactly trying to do?

ZephireNZ:
Hopefully the description is more clear now.

I think I just need to update the caching to use the new `.of` and it should be good.

PR #175, opened 2015-09-24
Properly fix MixinPlayerSelector

JBYoshi:
This is the proper fix for 8594d233290e42344cd06771b76fd660830c537d. When I wrote the commands PR, I set up `WrapperCommandSender` and `WrapperCommandSource` to convert between `ICommandSender`s and `CommandSource`s. This not only supports separation of the `CommandSender` and `ICommandSource`, but it also supports mod-added `ICommandSender`s and plugin-added `CommandSource`s.

@bloodmc


bloodmc:
Ok thanks. I'll pull it when I get time.

src/main/java/org/spongepowered/common/data/manipulator/mutable/entity/SpongeFoodData.java
```diff
@@ -126,7 +126,6 @@ public void setFoodExhaustion(double foodExhaustionLevel) {
         this.foodExhaustionLevel = (float) foodExhaustionLevel;
     }
 
-    @Override
```
gabizou: don't remove these.

PR #184, opened 2015-10-03
Implement AuthorData

lucdon:
the test i ran: 
````
@Plugin(id = "test", name = "test", version = "1.0-SNAPSHOT")
public class Test {

    private Logger logger;
    private Game game;

    @Inject
    public Test(Logger logger) {
        this.logger = logger;
    }

    @Listener
    public void preInit(GamePreInitializationEvent event) {
        this.game = event.getGame();
        this.logger.info("enabling author test");
    }

    @Listener
    public void enable(GameStartingServerEvent event) {
        event.getGame().getEventManager().registerListener(this, ClientConnectionEvent.Join.class, new EventListener<ClientConnectionEvent.Join>() {

            public void handle(Join event) throws Exception {
                Player player = event.getTargetEntity();
                // book 1
                ItemStack book = Test.this.game.getRegistry().createItemBuilder().itemType(ItemTypes.WRITTEN_BOOK).quantity(1).build();
                book.offer(Keys.BOOK_AUTHOR, Texts.of("a very awesome writer"));
                Test.this.logger.info("author created");
                Test.this.logger.info("author get: " + Texts.toPlain(book.getOrNull(Keys.BOOK_AUTHOR)));
                player.setItemInHand(book);
                // book 2
                AuthorData data = Test.this.game.getRegistry().getManipulatorRegistry().getBuilder(AuthorData.class).get().create();
                Test.this.logger.info("created authorData builder");
                data.set(Keys.BOOK_AUTHOR, Texts.of(TextColors.RED, "a nice guy from spongeDev team"));
                Test.this.logger.info("author set");
                ItemStack book2 = Test.this.game.getRegistry().createItemBuilder().itemType(ItemTypes.WRITTEN_BOOK).itemData(data).quantity(1).build();
                player.setItemInHand(book2);

            }

        });
    }
}
````

and of course the result: 

````
[01:20:48] [Server thread/INFO] [test]: author created
[01:20:48] [Server thread/INFO] [test]: author get: a very awesome writer
[01:20:48] [Server thread/INFO] [test]: created authorData builder
[01:20:48] [Server thread/INFO] [test]: author set
````

picture: 
test

### DIFF
--- a/src/main/java/org/spongepowered/common/data/builder/manipulator/mutable/item/ItemAuthorDataBuilder.java
+++ b/src/main/java/org/spongepowered/common/data/builder/manipulator/mutable/item/ItemAuthorDataBuilder.java
@@ -24,7 +24,6 @@
  */
 package org.spongepowered.common.data.builder.manipulator.mutable.item;
 
-import com.google.common.base.Optional;
 import net.minecraft.init.Items;
 import net.minecraft.item.ItemStack;
 import org.spongepowered.api.data.DataHolder;
@@ -39,6 +38,8 @@ import org.spongepowered.api.text.Texts;
 import org.spongepowered.common.data.manipulator.mutable.item.SpongeAuthorData;
 import org.spongepowered.common.data.util.DataUtil;
 import org.spongepowered.common.data.util.NbtDataUtil;
+
+import java.util.Optional;
 
 public class ItemAuthorDataBuilder implements DataManipulatorBuilder<AuthorData, ImmutableAuthorData> {
 
@@ -59,7 +60,7 @@ public class ItemAuthorDataBuilder implements DataManipulatorBuilder<AuthorData,
     public Optional<AuthorData> createFrom(DataHolder dataHolder) {
         if (dataHolder instanceof ItemStack) {
             if (((ItemStack) dataHolder).getItem() != Items.writable_book || ((ItemStack) dataHolder).getItem() != Items.written_book) {
-                return Optional.absent();
+                return Optional.empty();
             }
             if (((ItemStack) dataHolder).getTagCompound() == null || ((ItemStack) dataHolder).getTagCompound().hasKey(NbtDataUtil.ITEM_BOOK_AUTHOR)) {
                 return Optional.<AuthorData>of(new SpongeAuthorData());
@@ -68,7 +69,7 @@ public class ItemAuthorDataBuilder implements DataManipulatorBuilder<AuthorData,
             final Text author = Texts.json().fromUnchecked(json);
             return Optional.<AuthorData>of(new SpongeAuthorData(author));
         } else {
-            return Optional.absent();
+            return Optional.empty();
         }
     }
 
--- a/src/main/java/org/spongepowered/common/data/builder/manipulator/mutable/item/ItemAuthorDataBuilder.java
+++ b/src/main/java/org/spongepowered/common/data/builder/manipulator/mutable/item/ItemAuthorDataBuilder.java
@@ -32,7 +32,6 @@ import org.spongepowered.api.data.key.Keys;
 import org.spongepowered.api.data.manipulator.DataManipulatorBuilder;
 import org.spongepowered.api.data.manipulator.immutable.item.ImmutableAuthorData;
 import org.spongepowered.api.data.manipulator.mutable.item.AuthorData;
-import org.spongepowered.api.service.persistence.InvalidDataException;
 import org.spongepowered.api.text.Text;
 import org.spongepowered.api.text.Texts;
 import org.spongepowered.common.data.manipulator.mutable.item.SpongeAuthorData;
@@ -45,12 +44,11 @@ public class ItemAuthorDataBuilder implements DataManipulatorBuilder<AuthorData,
 
     @Override
     public Optional<AuthorData> build(DataView container) {
-        try {
-            DataUtil.checkDataExists(container, Keys.BOOK_AUTHOR.getQuery());
+        if (container.contains(Keys.BOOK_AUTHOR.getQuery())) {
             final String json = DataUtil.getData(container, Keys.BOOK_AUTHOR, String.class);
             final Text author = Texts.json().fromUnchecked(json);
             return Optional.<AuthorData>of(new SpongeAuthorData(author));
-        } catch (InvalidDataException e) {
+        } else {
             return Optional.empty();
         }
     }
--- a/src/main/java/org/spongepowered/common/data/builder/manipulator/mutable/item/ItemAuthorDataBuilder.java
+++ b/src/main/java/org/spongepowered/common/data/builder/manipulator/mutable/item/ItemAuthorDataBuilder.java
@@ -1,0 +1,75 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data.builder.manipulator.mutable.item;
+
+import com.google.common.base.Optional;
+import net.minecraft.init.Items;
+import net.minecraft.item.ItemStack;
+import org.spongepowered.api.data.DataHolder;
+import org.spongepowered.api.data.DataView;
+import org.spongepowered.api.data.key.Keys;
+import org.spongepowered.api.data.manipulator.DataManipulatorBuilder;
+import org.spongepowered.api.data.manipulator.immutable.item.ImmutableAuthorData;
+import org.spongepowered.api.data.manipulator.mutable.item.AuthorData;
+import org.spongepowered.api.service.persistence.InvalidDataException;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.Texts;
+import org.spongepowered.common.data.manipulator.mutable.item.SpongeAuthorData;
+import org.spongepowered.common.data.util.DataUtil;
+import org.spongepowered.common.data.util.NbtDataUtil;
+
+public class ItemAuthorDataBuilder implements DataManipulatorBuilder<AuthorData, ImmutableAuthorData> {
+
+    @Override
+    public Optional<AuthorData> build(DataView container) throws InvalidDataException {
+        DataUtil.checkDataExists(container, Keys.BOOK_AUTHOR.getQuery());
+        final String json = DataUtil.getData(container, Keys.BOOK_AUTHOR, String.class);
+        final Text author = Texts.json().fromUnchecked(json);
+        return Optional.<AuthorData>of(new SpongeAuthorData(author));
+    }
+
+    @Override
+    public AuthorData create() {
+        return new SpongeAuthorData();
+    }
+
+    @Override
+    public Optional<AuthorData> createFrom(DataHolder dataHolder) {
+        if (dataHolder instanceof ItemStack) {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/main/java/org/spongepowered/common/data/builder/manipulator/mutable/item/ItemAuthorDataBuilder.java [hunk 1]
```diff
@@ -0,0 +1,75 @@
...
+import org.spongepowered.common.data.util.DataUtil;
+import org.spongepowered.common.data.util.NbtDataUtil;
+
+public class ItemAuthorDataBuilder implements DataManipulatorBuilder<AuthorData, ImmutableAuthorData> {
+
+    @Override
+    public Optional<AuthorData> build(DataView container) throws InvalidDataException {
+        DataUtil.checkDataExists(container, Keys.BOOK_AUTHOR.getQuery());
+        final String json = DataUtil.getData(container, Keys.BOOK_AUTHOR, String.class);
+        final Text author = Texts.json().fromUnchecked(json);
```
gabizou: Should likely surround with try catch and throw an `InvalidDataException` if and when `Text` will throw it's own exceptions.

src/main/java/org/spongepowered/common/data/processor/data/item/ItemAuthorDataProcessor.java
```diff
@@ -0,0 +1,116 @@
...
+
+        }, Keys.BOOK_AUTHOR);
+    }
+
+    @Override
+    public DataTransactionResult remove(DataHolder dataHolder) {
+        if (supports(dataHolder)) {
+            final DataTransactionBuilder builder = DataTransactionBuilder.builder();
+            final Optional<AuthorData> data = from(dataHolder);
+            if (data.isPresent()) {
```
gabizou: wait what? Why are you changing it if and only if the data is present?

src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeAuthorData.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Override
+    public ImmutableAuthorData copy() {
+        return this;
+    }
+
+    @Override
+    public DataContainer toContainer() {
+        return new MemoryDataContainer()
+                .set(Keys.BOOK_AUTHOR.getQuery(), this.getValue());
```
gabizou: You likely are intending to send to the `DataContainer` the JSON string representation of the Text object.

src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeAuthorData.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override
+    public ImmutableValue<Text> author() {
+        return new ImmutableSpongeValue<Text>(Keys.BOOK_AUTHOR, this.getValue());
```
gabizou: Is the value also always the default value?

lucdon:
@gabizou didnt even notice that the second argument of ImmutableSpongeValue was the default value. will fix it

src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeAuthorData.java
```diff
@@ -0,0 +1,80 @@
...
+
+    @Override
+    public DataContainer toContainer() {
+        return new MemoryDataContainer()
+                .set(Keys.BOOK_AUTHOR.getQuery(), Texts.json().to(this.getValue()));
+    }
+
+    @Override
+    public ImmutableValue<Text> author() {
+        return new ImmutableSpongeValue<>(Keys.BOOK_AUTHOR, Texts.of(NbtDataUtil.INVALID_TITLE), this.getValue());
```
Aaron1011: You should store this value in the constructor, and get it using `ImmutableValue.cachedOf` instead of instantiating it yourself.

src/main/java/org/spongepowered/common/data/builder/manipulator/mutable/item/ItemAuthorDataBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+import org.spongepowered.common.data.util.NbtDataUtil;
+
+import java.util.Optional;
+
+public class ItemAuthorDataBuilder implements DataManipulatorBuilder<AuthorData, ImmutableAuthorData> {
+
+    @Override
+    public Optional<AuthorData> build(DataView container) throws InvalidDataException {
+        try {
+            DataUtil.checkDataExists(container, Keys.BOOK_AUTHOR.getQuery());
```
Aaron1011: You shoudn't be throwing an exception, even though you catch it - just return `empty` if the key isn't present.

src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeAuthorData.java
```diff
@@ -0,0 +1,80 @@
...
+
+public class ImmutableSpongeAuthorData extends AbstractImmutableSingleData<Text, ImmutableAuthorData, AuthorData> implements ImmutableAuthorData {
+
+    public ImmutableSpongeAuthorData(Text value) {
+        super(ImmutableAuthorData.class, value, Keys.BOOK_AUTHOR);
+    }
+
+    @Override
+    public ImmutableAuthorData copy() {
+        return this;
```
saladoc: I'd rather prefer an actual copy.

src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeAuthorData.java
lucdon: @Aaron1011 is it possible to cache a text value? 

here in ImmutableSpongeDisplayNameData: 

````

    @Override
    public ImmutableValue<Text> displayName() {
        return new ImmutableSpongeValue<>(Keys.DISPLAY_NAME, this.displayName);
    }
````
the value isnt cached. so i assumed you can't cache a text value.

@gabizou it would be great if you clarify what we can cache and what we can't cache.

Comment (saladoc) [inline src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeAuthorData.java]: You can, but you shouldn't. You can cache `ImmutableDataManipulator`s and `ImmutableValue`s and you should do so if they do not have many possible values. 
For a `Text`, the amount of possible values is effectively infinite, so caching may easily result in massive overhead, For a value storing a `boolean` or a variant caching is perfect, since there are only few possible values and it is more likely for instances to get reused.

Comment (gabizou) [inline src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeAuthorData.java]: You cache values that have finite possible values. You don't cache things that can vary based on whatever plugins decide (such as text based things).

Comment (lucdon) [inline src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeAuthorData.java]: so the value of book_author should not be cached?

Comment (gabizou) [inline src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeAuthorData.java]: Correct.

File: src/main/java/org/spongepowered/common/data/builder/manipulator/mutable/item/ItemAuthorDataBuilder.java [hunk 8]
```diff
@@ -0,0 +1,80 @@
...
+import org.spongepowered.common.data.util.NbtDataUtil;
+
+import java.util.Optional;
+
+public class ItemAuthorDataBuilder implements DataManipulatorBuilder<AuthorData, ImmutableAuthorData> {
+
+    @Override
+    public Optional<AuthorData> build(DataView container) {
+        try {
+            DataUtil.checkDataExists(container, Keys.BOOK_AUTHOR.getQuery());
```
Aaron1011: This should first check if the key query is present, and return `Optional.empty` otherwise.

src/main/java/org/spongepowered/common/data/builder/manipulator/mutable/item/ItemAuthorDataBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public AuthorData create() {
+        return new SpongeAuthorData();
+    }
+
+    @Override
+    public Optional<AuthorData> createFrom(DataHolder dataHolder) {
+        if (dataHolder instanceof ItemStack) {
+            if (((ItemStack) dataHolder).getItem() != Items.writable_book || ((ItemStack) dataHolder).getItem() != Items.written_book) {
```
Aaron1011: This should be an AND, not an OR.

src/main/java/org/spongepowered/common/data/processor/data/item/ItemAuthorDataProcessor.java
```diff
@@ -0,0 +1,106 @@
...
+    @Override
+    public DataTransactionResult remove(DataHolder dataHolder) {
+        if (supports(dataHolder)) {
+            final DataTransactionBuilder builder = DataTransactionBuilder.builder();
+            final Optional<AuthorData> data = from(dataHolder);
+            try {
+                if (((ItemStack) dataHolder).getTagCompound() == null) {
+                    ((ItemStack) dataHolder).setTagCompound(new NBTTagCompound());
+                }
+                ((ItemStack) dataHolder).getTagCompound().setString(NbtDataUtil.ITEM_BOOK_AUTHOR, NbtDataUtil.INVALID_TITLE);
```
Aaron1011: Is this correct? I'm not sure that Forge handles that name specially.

src/main/java/org/spongepowered/common/data/processor/data/item/ItemAuthorDataProcessor.java
```diff
@@ -0,0 +1,106 @@
...
+    protected AuthorData createManipulator() {
+        return new SpongeAuthorData();
+    }
+
+    @Override
+    protected boolean set(ItemStack itemStack, Text value) {
+        if (itemStack.getTagCompound() == null) {
+            itemStack.setTagCompound(new NBTTagCompound());
+        }
+        itemStack.getTagCompound().setString(NbtDataUtil.ITEM_BOOK_AUTHOR, ((IMixinText) value).toLegacy('\247', Locale.ENGLISH));
```
Aaron1011: Use `Texts.legacy()` here instead.

lucdon:
@Aaron1011 i did test the remove method if i put there null instead of "" then minecraft will throw an exception

lucdon:
@Aaron1011 why did you add that label? i did test it:

````
  book2.remove(Keys.BOOK_AUTHOR);
````

added that line to the test up above. and the result was a written book without an author

**Comment by gabizou** [comment]
@lucdon needs testing usually means the reviewer will test it and then merge it.

File: src/main/java/org/spongepowered/common/data/processor/data/item/ItemAuthorDataProcessor.java [hunk 12]
```diff
@@ -0,0 +1,103 @@
...
+    @Override
+    public DataTransactionResult remove(DataHolder dataHolder) {
+        if (supports(dataHolder)) {
+            final DataTransactionBuilder builder = DataTransactionBuilder.builder();
+            final Optional<AuthorData> data = from(dataHolder);
+            try {
+                if (!((ItemStack) dataHolder).hasTagCompound()) {
+                    ((ItemStack) dataHolder).setTagCompound(new NBTTagCompound());
+                }
+                ((ItemStack) dataHolder).getTagCompound().setString(NbtDataUtil.ITEM_BOOK_AUTHOR, "");
```
Aaron1011: I don't think removing should be supported, since there's always an author present (even if it's just the empty string). Considering that it's not valid to have the author NBT tag missing, it should be up to users to set the author to the empty string, if they want to.

Zidane: Agree with @Aaron1011

src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeAuthorData.java
```diff
@@ -0,0 +1,80 @@
...
+
+    @Override
+    public DataContainer toContainer() {
+        return new MemoryDataContainer()
+                .set(Keys.BOOK_AUTHOR.getQuery(), Texts.json().to(this.getValue()));
+    }
+
+    @Override
+    public ImmutableValue<Text> author() {
+        return new ImmutableSpongeValue<>(Keys.BOOK_AUTHOR, Texts.of(), this.getValue());
```
Aaron1011: This value instance should be saved in a field. However, it should still be instantiated normally, since a `Text` shouldn't be used with `ImmutableDataCachingUtil`

src/main/java/org/spongepowered/common/data/processor/data/item/ItemAuthorDataProcessor.java
lucdon: @Aaron1011 what is wrong with giving people multiple options? now they can do data.set("") or they can do data.remove().

minecraft only checks if the author tag is present when they convert an editable book to a written book

and it is up to users to use the data.remove() method too. so i dont see what is wrong with supporting the remove method

Aaron1011: @lucdon: The client calls `ItemEditableBook#validBookTagContents` when displaying the book gui. If it returns false, which happens when the `author` tag isn't present, then `* Invalid book tag * is displayed.

I still think that it shouldn't be removable. From a Data API perspective, removing data means that it's no longer present in any way on the `DataHolder`. An example of this would be `DisplayNameData`, which is not present on animals by default.

Setting data to an "empty" value shouldn't be done through the `remove` method, because it's still present. Since empty string is a perfectly valid author (though it causes the vanilla client not to display any author), this creates the situation where `AuthorData` is still present after you remove it.

Aaron1011:
Closed in https://github.com/SpongePowered/SpongeCommon/commit/b680d8122b721e78803284cb00f5e32d1380ac02.

Thanks for your contribution!

PR #185, opened 2015-10-04
Update ScreamingData

ZephireNZ:
This updates ScreamingData with the following:

- Extends `AbstractEntitySingleDataProcessor` for simpler processor
- Supplies itself as a vanilla manipulator for endermen
- Keep the `ImmutableValue` for `ImmutableScreamingData`

### DIFF
--- a/src/main/java/org/spongepowered/common/data/processor/data/entity/ScreamingDataProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/data/entity/ScreamingDataProcessor.java
@@ -24,27 +24,25 @@
  */
 package org.spongepowered.common.data.processor.data.entity;
 
-import com.google.common.collect.ImmutableMap;
 import net.minecraft.entity.monster.EntityEnderman;
-import org.spongepowered.api.data.DataContainer;
 import org.spongepowered.api.data.DataHolder;
 import org.spongepowered.api.data.DataTransactionBuilder;
 import org.spongepowered.api.data.DataTransactionResult;
-import org.spongepowered.api.data.key.Key;
 import org.spongepowered.api.data.key.Keys;
 import org.spongepowered.api.data.manipulator.immutable.entity.ImmutableScreamingData;
 import org.spongepowered.api.data.manipulator.mutable.entity.ScreamingData;
+import org.spongepowered.api.data.value.immutable.ImmutableValue;
+import org.spongepowered.api.data.value.mutable.Value;
 import org.spongepowered.common.data.manipulator.mutable.entity.SpongeScreamingData;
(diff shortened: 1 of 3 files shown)


ZephireNZ:
Once I get Sponge actually building in my IDE, I'll get a test case up for this... -_-

src/main/java/org/spongepowered/common/data/builder/manipulator/mutable/entity/ScreamingDataBuilder.java
```diff
@@ -55,8 +52,8 @@ public ScreamingData create() {
 
     @Override
     public Optional<ScreamingData> createFrom(DataHolder dataHolder) {
-        if (dataHolder.supports(Keys.IS_SCREAMING)) {
-            return Optional.<ScreamingData>of(new SpongeScreamingData(dataHolder.get(Keys.IS_SCREAMING).get()));
+        if (dataHolder instanceof EntityEnderman) {
```
Grinch: I don't think it should be restricted to just enderman. While they are the only ones that have this at the moment it'd be nice to be able to use this a mod-added entity that may have this tag.

gabizou: Yes, but at the same time, this isn't actually used anywhere since `DataProcessor`s are usually how you'll get the data.

gabizou:
@ZephireNZ Could you rebase the changes? I'm honestly unsure as to what is supposed to actually change since #209 was merged.

ZephireNZ:
@gabizou, merged/rebased :)

gabizou:
@ZephireNZ rebase to remove the merge commits, they're a pita to deal with with the amount of merges.

ZephireNZ:
Fair enough, there we are

src/main/java/org/spongepowered/common/data/processor/data/entity/ScreamingDataProcessor.java
```diff
@@ -53,29 +51,23 @@ protected ScreamingData createManipulator() {
...
+    protected Optional<Boolean> getVal(EntityEnderman entity) {
+        return Optional.of(entity.isScreaming());
     }
 
     @Override
-    public Optional<ScreamingData> fill(DataContainer container, ScreamingData screamingData) {
-        screamingData.set(Keys.IS_SCREAMING, DataUtil.getData(container, Keys.IS_SCREAMING));
-        return Optional.of(screamingData);
+    protected ImmutableValue<Boolean> constructImmutableValue(Boolean value) {
+        return new ImmutableSpongeValue<>(Keys.IS_SCREAMING, false, value);
```
gabizou: Just for future reference, this should be cached.

PR #190, opened 2015-10-08
The Complete OCD Pull Request: SpongeCommon 3.0 Edition

phroa:
[SpongeAPI PR](https://github.com/SpongePowered/spongeapi/pull/860) | SpongeForge PR (tbd) | SpongeVanilla PR (tbd)

Remaining errors: dumbass checkstyle that can't parse method references; javadoc issues a dev should probably look at; trivial stuff (ironic, I know)

An OCD issue tracker for common/forge/vanilla in addition to the API would have helped :/


src/main/java/org/spongepowered/common/mixin/core/data/MixinCustomDataHolder.java
```diff
@@ -107,9 +108,9 @@ public boolean hasManipulators() {
     @Override
     public List<DataManipulator<?, ?>> getCustomManipulators() {
         final List<DataManipulator<?, ?>> list = Lists.newArrayList();
-        for (DataManipulator<?, ?> manipulator : this.manipulators) {
-            list.add(manipulator.copy());
-        }
+        list.addAll(this.manipulators.stream()
+                .map(DataManipulator::copy)
```
gabizou: No. Remember what I said about anonymous classes?

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -208,9 +209,7 @@ public void spawnParticles(ParticleEffect particleEffect, Vector3d position, int
             double dz = this.posZ - position.getZ();
 
             if (dx * dx + dy * dy + dz * dz < radius * radius) {
-                for (Packet packet : packets) {
-                    this.playerNetServerHandler.sendPacket(packet);
-                }
+                packets.forEach(this.playerNetServerHandler::sendPacket);
```
gabizou: Again no.

src/main/java/org/spongepowered/common/mixin/core/entity/projectile/MixinEntityFishHook.java
```diff
@@ -177,7 +178,9 @@ public int handleHookRetraction() {
 
             if (event.getItemStackTransaction().isPresent()) {
                 ItemStackSnapshot itemSnapshot = event.getItemStackTransaction().get().getFinalSnapshot();
-                EntityItem entityitem1 = new EntityItem(this.worldObj, this.posX, this.posY, this.posZ, (net.minecraft.item.ItemStack) itemSnapshot.createStack());
+                EntityItem entityitem1 = new EntityItem(this.worldObj, this.posX, this.posY, this.posZ,
+
```
gabizou: what happened here?

src/main/java/org/spongepowered/common/mixin/core/item/inventory/MixinItemStack.java
```diff
@@ -344,9 +345,9 @@ public boolean hasManipulators() {
     @Override
     public List<DataManipulator<?, ?>> getCustomManipulators() {
         final List<DataManipulator<?, ?>> list = Lists.newArrayList();
-        for (DataManipulator<?, ?> manipulator : this.manipulators) {
-            list.add(manipulator.copy());
-        }
+        list.addAll(this.manipulators.stream()
+                .map(DataManipulator::copy)
```
gabizou: Again no.

src/main/java/org/spongepowered/common/mixin/core/block/MixinBlockState.java
```diff
@@ -103,9 +103,7 @@ public BlockSnapshot snapshotFor(Location<World> location) {
             .worldId(location.getExtent().getUniqueId());
         if (this.block.hasTileEntity() && location.getBlockType().equals(this.block)) {
             final TileEntity tileEntity = location.getTileEntity().get();
-            for (DataManipulator<?, ?> manipulator : tileEntity.getContainers()) {
-                builder.add(manipulator);
-            }
+            tileEntity.getContainers().forEach(builder::add);
```
gabizou: nope.

DDoS: I think we should have something to automatically detect inner classes (including anonymous classes, lambdas and method references) in mixins and refuse to compile. Maybe this could make part of Mixin itself? It could even be done with javaparser.

src/main/java/org/spongepowered/common/mixin/core/data/MixinCustomDataHolder.java
me4502: For my reference, what's the issue here?

DDoS: Lambdas are mostly syntactic sugar for anonymous inner classes. Mixins don't support these as it would be too complicated. The inner class gets discarded with it and you get a class not found exception at runtime.

gabizou: Mixins do NOT properly remap anonymous classes and will fail to apply at runtime due to the Mixin class being referenced but not available at runtime.

Zidane:
Closing this as this PR clearly wasn't tested as if it was it would have bombed the server out IMMEDIATELY due to the Java 8 stuff. It hasn't been secret knowledge that Java 8 anon-class generated code in Mixins won't work, @Mumfrey  hollered about it ages ago and the Java 8 branch that was just merged was stonewalled because people either forgot about it or didn't bother to remember it.

src/main/java/org/spongepowered/common/mixin/core/entity/projectile/MixinEntityFishHook.java
Phroa: fixed ✓

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
Phroa: all fixed ✓

Phroa:
You're right, it wasn't tested on a server :P

I'll start work on the servers probably this weekend. In the meantime I've fixed the mixin stuff so in theory they'll work now.

src/main/java/org/spongepowered/common/block/SpongeBlockSnapshot.java
```diff
@@ -210,6 +211,14 @@ public DataContainer toContainer() {
     }
 
     @Override
+    public <E> Optional<E> get(Key<? extends BaseValue<E>> key) {
```
gabizou: Why are you rearranging my methods?

src/main/java/org/spongepowered/common/block/SpongeBlockSnapshotBuilder.java
```diff
@@ -104,14 +105,28 @@ public SpongeBlockSnapshotBuilder from(Location<World> location) {
...
+            list.addAll(location.getContainers().stream()
+                    .map(DataManipulator::asImmutable)
+                    .collect(Collectors.toList()));
             this.manipulators = list;
         }
         return this;
     }
 
+    @Override
+    public SpongeBlockSnapshotBuilder from(BlockSnapshot holder) {
```
gabizou: And here.

src/main/java/org/spongepowered/common/data/ImmutableDataCachingUtil.java
```diff
@@ -70,15 +70,14 @@ private ImmutableDataCachingUtil() {}
         try {
             // Let's get the key
             return (T) (Object) ImmutableDataCachingUtil.manipulatorCache
-                .get(key,
```
gabizou: Don't touch the `data` package. I've got [changes](https://github.com/SpongePowered/SpongeCommon/compare/feature/data) that are still incubating.

src/main/java/org/spongepowered/common/block/SpongeBlockSnapshot.java
Phroa: Checkstyle prefers overrides to be next to each other. (And frankly I do too)

JBYoshi: You mean overwrites?

Phroa: probably, yeah

gabizou:
Closing this until the code freeze before beta

PR #191, opened 2015-10-08
Implement spawnWeatherEntity

clienthax:

PR #192, opened 2015-10-08
Add EntityWeatherEffect support for spawnEntity

clienthax:
Allows spawning of lightning bolts


hsyyid:
Nice

JBYoshi:
@clienthax What's the difference between this and #191?

clienthax:
@JBYoshi I didn't want to squash the commits, that one was originally done another way that needed api changed, but gab said to do this way instead (the old link is now showing current code however)

PR #195, opened 2015-10-11
Implementation of InvisibilityData

hsyyid:
Implementation of `InvisibilityData`.

Implementation of `ImmutableInvisibilityData`

Implementation of the `VALUE_PROCESSOR`(s):

Registration:
- [x] Registered the `Key` correctly by `FIELD_NAME` in the `KeyRegistry`

Test Plugin code:

```java
package io.github.hsyyid.essentialcmds;

import org.spongepowered.api.command.CommandException;
import org.spongepowered.api.command.CommandResult;
import org.spongepowered.api.command.CommandSource;
import org.spongepowered.api.command.args.CommandContext;
import org.spongepowered.api.command.args.GenericArguments;
import org.spongepowered.api.command.source.CommandBlockSource;
import org.spongepowered.api.command.source.ConsoleSource;
import org.spongepowered.api.command.spec.CommandExecutor;
import org.spongepowered.api.command.spec.CommandSpec;
import org.spongepowered.api.event.game.state.GameInitializationEvent;
import org.spongepowered.api.text.Texts;
import org.spongepowered.api.text.format.TextColors;

import java.util.Optional;

@Plugin(id = "VanishTest", name = "VanishTest", version = "0.1")
public class VanishTest {

    @Listener
    public void onGameInit(GameInitializationEvent event) {
        CommandSpec vanishCommandSpec = CommandSpec.builder().description(Texts.of("Vanish Command")).permission("essentialcmds.vanish.use")
                .arguments(GenericArguments.optional(GenericArguments.onlyOne(GenericArguments.player(Texts.of("player")))))
                .executor(new VanishExecutor()).build();

        Sponge.getGame().getCommandManager().register(this, vanishCommandSpec, "vanish");
    }

    public class VanishExecutor implements CommandExecutor {

        public CommandResult execute(CommandSource src, CommandContext ctx) throws CommandException {
            Optional<Player> target = ctx.<Player>getOne("player");

            if (!target.isPresent()) {
                if (src instanceof Player) {
                    Player player = (Player) src;

                    if (player.get(Keys.INVISIBLE).isPresent() && !player.get(Keys.INVISIBLE).get()) {
                        player.offer(Keys.INVISIBLE, true);
                        player.sendMessage(Texts.of(TextColors.GREEN, "Success! ", TextColors.YELLOW, "You are now invisible."));
                    } else if (player.get(Keys.INVISIBLE).isPresent() && player.get(Keys.INVISIBLE).get()) {
                        player.offer(Keys.INVISIBLE, false);
                        player.sendMessage(Texts.of(TextColors.GREEN, "Success! ", TextColors.YELLOW, "You are now visible."));
                    }
                } else if (src instanceof ConsoleSource) {
                    src.sendMessage(Texts.of(TextColors.DARK_RED, "Error! ", TextColors.RED, "Must be an in-game player to use /vanish!"));
                } else if (src instanceof CommandBlockSource) {
                    src.sendMessage(Texts.of(TextColors.DARK_RED, "Error! ", TextColors.RED, "Must be an in-game player to use /vanish!"));
                }
            } else if (target.isPresent() && src.hasPermission("vanish.others")) {
                Player player = target.get();

                if (player.get(Keys.INVISIBLE).isPresent() && !player.get(Keys.INVISIBLE).get()) {
                    player.offer(Keys.INVISIBLE, true);
                    src.sendMessage(Texts.of(TextColors.GREEN, "Success! ", TextColors.YELLOW, "Set player invisible."));
                    player.sendMessage(Texts.of(TextColors.GOLD, src.getName(), " has set you ", TextColors.GRAY, "invisible."));
                } else if (player.get(Keys.INVISIBLE).isPresent() && player.get(Keys.INVISIBLE).get()) {
                    player.offer(Keys.INVISIBLE, false);
                    src.sendMessage(Texts.of(TextColors.GREEN, "Success! ", TextColors.YELLOW, "Set player visible."));
                    player.sendMessage(Texts.of(TextColors.GOLD, src.getName(), " has set you ", TextColors.GRAY, "visible."));
                }
            }

        }
    }
}
```

### DIFF
--- a/src/main/java/org/spongepowered/common/data/manipulator/immutable/entity/ImmutableSpongeInvisibilityData.java
+++ b/src/main/java/org/spongepowered/common/data/manipulator/immutable/entity/ImmutableSpongeInvisibilityData.java
@@ -1,0 +1,45 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data.manipulator.immutable.entity;
+
+import org.spongepowered.api.data.key.Keys;
+import org.spongepowered.api.data.manipulator.immutable.entity.ImmutableInvisibilityData;
+import org.spongepowered.api.data.manipulator.mutable.entity.InvisibilityData;
+import org.spongepowered.api.data.value.immutable.ImmutableValue;
+import org.spongepowered.common.data.manipulator.immutable.common.AbstractImmutableBooleanData;
+import org.spongepowered.common.data.manipulator.mutable.entity.SpongeInvisibilityData;
+
+public class ImmutableSpongeInvisibilityData extends AbstractImmutableBooleanData<ImmutableInvisibilityData, InvisibilityData> implements ImmutableInvisibilityData {
+
+    public ImmutableSpongeInvisibilityData(boolean value) {
+        super(ImmutableInvisibilityData.class, value, Keys.INVISIBLE, SpongeInvisibilityData.class, false);
+    }
+
+    @Override
+    public ImmutableValue<Boolean> invisible() {
+        return getValueGetter();
+    }
+
+}
--- a/src/main/java/org/spongepowered/common/data/processor/dual/entity/InvisibilityDataProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/dual/entity/InvisibilityDataProcessor.java
@@ -1,0 +1,95 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data.processor.dual.entity;
+
+import net.minecraft.entity.Entity;
+import net.minecraft.entity.EntityTracker;
+import net.minecraft.world.WorldServer;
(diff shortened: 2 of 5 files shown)


hsyyid:
@gabizou Will rebase and fix conflicts right now.

@saladoc Yes it should. Need to make API changes to support that though.

ryantheleach:
So just to be 100% clear, this is *not* applying potion data "invisibility" but vanishing the player right? You send entity destruct packets to the client when they go invisible, but are movement / sound / other packets being suppressed?

hsyyid:
@ryantheleach Yes. I don't think invisibility data should be accounting for other packets. For Players it currently sets the player to invisible. However for the visible to players key, it may be sending that packet but it won't do anything yet.

ryantheleach:
Sound you could get away with, by listening for an event, and cancelling if the player the sound is caused by is invisible to the player receiving it is not such a big deal, and others might even want noises.

However, the client receiving *any* packets that reveal that there is an invisibleData player there, that don't otherwise have an event that allows you to cancel the packet updating the client *must* be prevented. It's the entire reason why VanishNoPacket was originally called vanishNoPacket. Because it prevented the packets from reaching hacked clients revealing vanished players.

@mbax

mbax:
InvisibilityData is *substantially* less useful if you don't block packets. Packet stopping can be almost entirely handled by just untracking the entity (be it another player or another entity) from the player. However, at least in 1.7 and earlier, there are also several other cases not checked against tracking.

Examples I remember:
* Some sounds, like step sounds
* Particle effects released by a player directly

Other cases, like walking across redstone ore, could be selectively blocked if you want to go mad, but the above examples certainly were more direct and not stoppable other ways.

kashike:
```
22:49:58        +kashike | HassanS6000: uh
22:50:01        +kashike | https://github.com/SpongePowered/SpongeCommon/pull/195
22:50:07        +kashike | that should support all entities, no?
22:50:36        +kashike | https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/InvisibilityData.java#L36
22:50:38     HassanS6000 | kashike, I suppose so
22:50:49        +kashike | looks like it only supports Player right now
22:51:12     HassanS6000 | kashike, Yea. I'll fix that soon.
```

hsyyid:
@kashike Yes I will fix it soon.

@mbax Thought you said untracking players specifically is NOT possible yet?

@ryantheleach I suppose so, but it's up to @gabizou

mbax:
@hsyyid because it's not implemented in SpongeCommon?

gabizou:
@hsyyid As @mbax is suggesting, we should have this encompass all that is needed so plugins like VanishNoPacket don't need to specifically write implementation dependent code to perform perfectly. If modifications to the vanish system need to actually be systems like Bans, Commands, etc. such that it isn't a part of data, but handled through a system to toggle various things, then so be it.

hsyyid:
@kashike Now supports all entities. :smile:

src/main/java/org/spongepowered/common/data/manipulator/immutable/entity/ImmutableSpongeInvisibilityData.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Override
+    public ImmutableSetValue<UUID> invisibleToPlayerIds() {
+        return new ImmutableSpongeSetValue<UUID>(Keys.INVISIBLE_TO_PLAYER_IDS, invisibleToPlayerIds);
+    }
+
+    @Override
+    public int compareTo(ImmutableInvisibilityData o) {
+        return ComparisonChain.start()
+//                .compare(o.invisibleToPlayerIds().get(), this.invisibleToPlayerIds)
```
kashike: I don't believe this is intended.

hsyyid:
@kashike Commented out until I get around to fixing it

src/main/java/org/spongepowered/common/data/processor/value/entity/InvisibleToPlayerIdsValueProcessor.java
```diff
@@ -0,0 +1,89 @@
...
+        return new SpongeSetValue<UUID>(Keys.INVISIBLE_TO_PLAYER_IDS, defaultValue);
+    }
+
+    @Override
+    protected boolean set(EntityPlayer container, Set<UUID> value) {
+        Player player = (Player) container;
+        InvisibilityData invisibleData = player.getOrCreate(InvisibilityData.class).get();
+        invisibleData.invisibleToPlayerIds().set(value);
+        player.offer(invisibleData);
+        for (UUID uuid : value) {
```
kashike: It would be nice to have this logic moved to `MixinEntityPlayer`.

JBYoshi:
@hsyyid Instead of sending “destroy” and “spawn” packets, why not simply add and remove the entity from the `EntityTracker`? This will prevent any packets regarding an invisible entity from being sent. Preventing the entity from being re-added and doing per-player visibility checks would be a bit difficult, but it's the easiest way to prevent other packets from being sent about the entity.

hsyyid:
@JBYoshi Good idea, I believe @mbax has suggested this as well.

Feeliiix:
@hsyyid When u finish it? Very need it.

hsyyid:
@gabizou Thought @mbax said it wasn't possible yet to remove players from the EntityTracker?

mbax:
@hsyyid 

No?

hsyyid:
@mbax Oh. Okay, fixing this up now

hsyyid:
@gabizou Fully tested and working, ready to merge, will fix Travis build when I get time.

src/main/java/org/spongepowered/common/data/processor/data/entity/InvisibilityDataProcessor.java
```diff
@@ -0,0 +1,88 @@
...
+import org.spongepowered.api.data.key.Keys;
+import org.spongepowered.api.data.manipulator.immutable.entity.ImmutableInvisibilityData;
+import org.spongepowered.api.data.manipulator.mutable.entity.InvisibilityData;
+import org.spongepowered.common.data.manipulator.mutable.entity.SpongeInvisibilityData;
+import org.spongepowered.common.data.processor.common.AbstractEntityDataProcessor;
+
+import java.util.Map;
+import java.util.Optional;
+
+public class InvisibilityDataProcessor extends AbstractEntityDataProcessor<Entity, InvisibilityData, ImmutableInvisibilityData> {
```
gabizou: You could likely turn this into a dual processor.

src/main/java/org/spongepowered/common/data/processor/dual/entity/InvisibilityDataProcessor.java
```diff
@@ -0,0 +1,91 @@
...
+        return new SpongeInvisibilityData();
+    }
+
+    @Override
+    protected boolean set(Entity entity, Boolean value) {
+        if (!entity.worldObj.isRemote) {
+            if (value) {
+                ((WorldServer) entity.worldObj).getEntityTracker().untrackEntity(entity);
+            } else {
+                ((WorldServer) entity.worldObj).getEntityTracker().trackEntity(entity);
```
clienthax: Need to check if the entity is already tracked here, otherwise it will explode.

JBYoshi:
@gabizou Is there any alternative to `INVISIBLE_TO_PLAYER_IDS`?

clienthax:
@JBYoshi its going to be done via context based data

PR #131, opened 2015-08-16
Implement Statistics

ST-DDT:
[[API]](https://github.com/SpongePowered/SpongeAPI/pull/958) [[Common]](https://github.com/SpongePowered/SpongeCommon/pull/131)

Implement the Achievement + Statistic API as specified in the [API](https://github.com/SpongePowered/SpongeAPI/tree/master/src/main/java/org/spongepowered/api/statistic).
Implement the DataAPI connectors.

**Tasks**
- [x] `Statistic`
  - [x] `BlockStatistic`
  - [x] `EntityStatistic`
  - [x] `ItemStatistic`
  - ~~`TeamStatistic`~~ - Does not exist at all (Only as Objective)
- [x] `Statistics`
- [x] `StatisticBuilder`
- [x] `StatisticFormat`
- [x] `StatisticFormats`
- [x] `StatisticGroup`
- [x] `StatisticGroups`
- [x] `StatisticData`
- [x] Call `StatisticChangeEvent`
- [x] `Achievement`
- [x] `Achievements`
- [x] `AchievementBuilder` ([needs API changes](https://github.com/SpongePowered/SpongeAPI/pull/958))
- [x] `AchievementData`
- [x] Call `AchievementEvent`
- [x] Create TestPlugin

**TestPlugin**
[Plugin.jar](https://dl.dropboxusercontent.com/u/16999313/Sponge/Statistics.jar)
[Plugin Sources](https://dl.dropboxusercontent.com/u/16999313/Sponge/Statistics.zip)

Commands:
- `/statisticdata`
- `/statistics`
- `/statisticgroups`
- `/statisticformats`
- `/achievementdata`
- `/achievements`

PR #199, opened 2015-10-11
[WIP] Add all CatalogTypes marked with TODO

Rain336:
# Info:
This PR add all CatalogTypes that are marked with `TODO` in the [SpongeGameRegistry](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/registry/SpongeGameRegistry.java) .

# TODO-List:
- [ ] Change Predicates Class?

# Test Plugin:
[Github Gist](https://gist.github.com/Rain336/54b95786831820ae09c5)
## Pictures:
Picture 1
Picture 2
Picture 3


src/main/java/org/spongepowered/common/data/type/SpongeEquipmentType.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public boolean equals(Object obj) {
+        if(obj == this)
+            return true;
+        if(obj instanceof EquipmentType) {
+            if(((EquipmentType) obj).getId().equals("ANY"))
+                return true;
+            if(((EquipmentType) obj).getId().equals(getId()))
+                return true;
```
Phroa: Make sure to put spaces after `if` tokens. Additionally, you're not checking for null. `if` statements should also always use braces.

src/main/java/org/spongepowered/common/data/type/SpongeEquipmentTypeWorn.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Override
+    public boolean equals(Object obj) {
+        if(super.equals(obj))
+            return true;
+        if(obj instanceof SpongeEquipmentTypeWorn) {
+            if(((EquipmentTypeWorn) obj).getId().equals("WORN"))
+                return true;
+        }
+        return false;
```
Phroa: Here too.

src/main/java/org/spongepowered/common/data/type/SpongeStatisticFormat.java
```diff
@@ -0,0 +1,93 @@
...
+        if (this.id.equals("DISTANCE")) {
+            double d0 = (double) value / 100.0D;
+            double d1 = d0 / 1000.0D;
+            return d1 > 0.5D ? this.decimalFormat.format(d1) + " km" : (d0 > 0.5D ? this.decimalFormat.format(d0) + " m" : value + " cm");
+        } else if (this.id.equals("TIME")) {
+            double d0 = (double) value / 20.0D;
+            double d1 = d0 / 60.0D;
+            double d2 = d1 / 60.0D;
+            double d3 = d2 / 24.0D;
+            double d4 = d3 / 365.0D;
```
Phroa: I'd probably collapse these. All the meaninglessly named variables are kinda silly imo.

src/main/java/org/spongepowered/common/data/type/SpongeStatisticFormat.java
```diff
@@ -0,0 +1,93 @@
...
+            return StatisticFormats.FRACTIONAL;
+        } else {
+            return StatisticFormats.COUNT;
+        }
+    }
+
+    @Override
+    public String format(long value) {
+        if (this.id.equals("DISTANCE")) {
+            double d0 = (double) value / 100.0D;
```
Phroa: I'm pretty sure the `(double)` cast is implicit since the RHS is a double.

src/main/java/org/spongepowered/common/data/type/SpongeStatisticFormat.java
```diff
@@ -0,0 +1,93 @@
...
+            double d1 = d0 / 1000.0D;
+            return d1 > 0.5D ? this.decimalFormat.format(d1) + " km" : (d0 > 0.5D ? this.decimalFormat.format(d0) + " m" : value + " cm");
+        } else if (this.id.equals("TIME")) {
+            double d0 = (double) value / 20.0D;
```
Phroa: here too

src/main/java/org/spongepowered/common/data/type/SpongeStatisticFormat.java
```diff
@@ -0,0 +1,93 @@
...
+            double d1 = d0 / 60.0D;
+            double d2 = d1 / 60.0D;
+            double d3 = d2 / 24.0D;
+            double d4 = d3 / 365.0D;
+            return d4 > 0.5D ? this.decimalFormat.format(d4) + " y" : (d3 > 0.5D ? this.decimalFormat.format(d3) + " d" :
+                    (d2 > 0.5D ? this.decimalFormat.format(d2) + " h" : (d1 > 0.5D ? this.decimalFormat.format(d1) + " m" : d0 + " s")));
+        } else if (this.id.equals("FRACTIONAL")) {
+            return this.decimalFormat.format((double) value * 0.1D);
```
Phroa: and here

src/main/java/org/spongepowered/common/data/type/SpongeStatisticFormat.java
```diff
@@ -0,0 +1,93 @@
...
+            double d0 = (double) value / 100.0D;
+            double d1 = d0 / 1000.0D;
+            return d1 > 0.5D ? this.decimalFormat.format(d1) + " km" : (d0 > 0.5D ? this.decimalFormat.format(d0) + " m" : value + " cm");
+        } else if (this.id.equals("TIME")) {
+            double d0 = (double) value / 20.0D;
+            double d1 = d0 / 60.0D;
+            double d2 = d1 / 60.0D;
+            double d3 = d2 / 24.0D;
+            double d4 = d3 / 365.0D;
+            return d4 > 0.5D ? this.decimalFormat.format(d4) + " y" : (d3 > 0.5D ? this.decimalFormat.format(d3) + " d" :
```
Phroa: I think nested ternaries is against style but I'm not certain.

src/main/java/org/spongepowered/common/data/type/SpongeStatisticGroup.java
```diff
@@ -0,0 +1,119 @@
...
+    public String getName() {
+        return this.id;
+    }
+
+    @Override
+    public Translation getTranslation() {
+        switch (this.id) {
+            case "GENERAL":
+            default:
+                return new FixedTranslation("General");
```
Phroa: I'd put the default case at the end but that's just me.

src/main/java/org/spongepowered/common/util/Predicates.java
```diff
@@ -0,0 +1,49 @@
...
+import net.minecraft.entity.EntityLivingBase;
+import net.minecraft.entity.monster.EntityZombie;
+import net.minecraft.entity.passive.EntityHorse;
+import org.spongepowered.api.data.DataHolder;
+
+import java.util.function.Predicate;
+
+public class Predicates {
+
+    public static Predicate<DataHolder> rangedAttributePredicate(String name) {
```
Phroa: Granted, I don't know everything about data, but the name of this method and what it actually does seems very unobvious...

Additionally, is creating a new class for one method really worth it? You only use it in one place as far as I can tell.

Rain336: I don't know. But I need a Class to put it in for the MixinRangedAttribute.

Deamon5550:
I'm not a fan of this as you are just rather hackishly implementing the catalog types without any of the related api pieces. And additionally some of these are already proposed within existing PRs.

Specifically I was looking at the StatisticFormats which are implemented in a much better way in #131. So at the very least you will need to remove the statistics related types from this PR.

src/main/java/org/spongepowered/common/data/type/SpongeEquipmentType.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+package org.spongepowered.common.data.type;
+
+import org.spongepowered.api.item.inventory.equipment.EquipmentType;
+
+public class SpongeEquipmentType implements EquipmentType {
+
+    private final String id;
+
+    public SpongeEquipmentType(String id) {
```
gabizou: Since this is controlled by us, can't you make it where the id is separate from the legible name?

gabizou:
Yeah, there are a lot of these things that can be mixed in.

gabizou:
Closing this since this is clearly not how all the `CatalogType`s need to be registered. There's a lot of them that can and should be implemented via Mixin. The attempt is appreciated, but these `CatalogType`s need to be usable by not just the API, but implementation as well (especially since they almost all have to deal with Data API in some way).

PR #163, opened 2015-09-11
Guice providers for SpongeExecutorServices

PieGuy128:
[API side][1]

This PR adds the Guice providers required for the injection of SpongeExecutorServices.

[1]: https://github.com/SpongePowered/SpongeAPI/pull/840

PR #200, opened 2015-10-11
Implement SpongeScheduledExecutor

Kiskae:
As mentioned in the API PR (https://github.com/SpongePowered/SpongeAPI/pull/871), this PR updates a commit from the old `java8` branch to the `master` branch.

The old PR for this feature can be found at #151
Binding annotations are implemented in #163 

Besides implementing the new ExecutorService interface, this commit also changes the internal time tracking mechanism from `System.currentTimeMillis()` to `System.nanoTime()` which is not affected by system clock changes.

### DIFF
--- a/src/main/java/org/spongepowered/common/guice/SpongePluginGuiceModule.java
+++ b/src/main/java/org/spongepowered/common/guice/SpongePluginGuiceModule.java
@@ -31,9 +31,14 @@ import com.google.inject.TypeLiteral;
 import ninja.leaping.configurate.commented.CommentedConfigurationNode;
 import ninja.leaping.configurate.loader.ConfigurationLoader;
 import org.slf4j.Logger;
+import org.spongepowered.api.Game;
 import org.spongepowered.api.plugin.PluginContainer;
 import org.spongepowered.api.service.config.ConfigDir;
 import org.spongepowered.api.service.config.DefaultConfig;
+import org.spongepowered.api.service.scheduler.AsynchronousExecutor;
+import org.spongepowered.api.service.scheduler.SchedulerService;
+import org.spongepowered.api.service.scheduler.SpongeExecutorService;
+import org.spongepowered.api.service.scheduler.SynchronousExecutor;
 import org.spongepowered.common.plugin.SpongePluginContainer;
 import org.spongepowered.common.service.config.SpongeConfigService;
 
@@ -75,6 +80,10 @@ public class SpongePluginGuiceModule extends AbstractModule {
         bind(new TypeLiteral<ConfigurationLoader<CommentedConfigurationNode>>() {
         }).annotatedWith(privateConfigFile)
                 .toProvider(PrivateHoconConfigProvider.class); // Loader for plugin-private directory config file
+
+        // SpongeExecutorServices
+        bind(SpongeExecutorService.class).annotatedWith(SynchronousExecutor.class).toProvider(SynchronousExecutorProvider.class);
+        bind(SpongeExecutorService.class).annotatedWith(AsynchronousExecutor.class).toProvider(AsynchronousExecutorProvider.class);
     }
 
     private static class PrivateConfigDirProvider implements Provider<File> {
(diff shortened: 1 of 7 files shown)


src/main/java/org/spongepowered/common/service/scheduler/SchedulerBase.java
```diff
@@ -52,14 +52,14 @@ protected String nextName(PluginContainer plugin) {
...
-     * sensitive to support different timestamp types i.e. wall clock and ticks.
+     * sensitive to support different timestamp types i.e. nanotime and ticks.
      *
      * @param task The task
      * @return Timestamp for the task
      */
     protected long getTimestamp(ScheduledTask task) {
-        // Supports wall clock time by default
-        return System.currentTimeMillis();
+        // Supports nanotime by default
```
Phroa: Should this be moved to the javadoc? (if not, I don't know what purpose the comment serves)

Kiskae: I've just updated the original comments, @simon816 appears to be the original author of that code so perhaps he can comment on it?

simon816: The base scheduling implementation supports wall clock time (seconds/milliseconds/nanoseconds/whatever). The synchronous scheduler adds an additional timestamp which is the number of ticks.

You don't need to change the comment, it's just saying that a system clock time is supported

Kiskae: @simon816 The issue is that `System.nanoTime()` is explicitly not tied to any system or wall clock, it is completely arbitrary and only useful for "x time has passed" checks. Leaving the comment would be misleading.

As specified in the javadocs:
> This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time. 
<..>
> The values returned by this method become meaningful only when the difference between two such values, obtained within the same instance of a Java virtual machine, is computed.

Kiskae:
I've specified what the values returned by `getTimestamp` should be expected to do.
Hopefully this is clear enough to avoid issues with future changes to the service.

Kiskae:
Rebased nanotime changes and executorservice implementation on the latest scheduler changes.
I've also fixed a small inconsistency where it was possible to make the future appear cancelled when it had already run by cancelling the associated task.

src/main/java/org/spongepowered/common/service/scheduler/SpongeScheduler.java
```diff
@@ -38,13 +39,14 @@
...
 import java.util.UUID;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public class SpongeScheduler implements SchedulerService {
 
     private static final SpongeScheduler INSTANCE = new SpongeScheduler();
-    public static final int TICK_DURATION = 50;
+    public static final int TICK_DURATION = (int) TimeUnit.NANOSECONDS.convert(50, TimeUnit.MILLISECONDS);
```
simon816: The method `getPreferredTickInterval()` needs to return milliseconds

Kiskae: I've fixed the internal representation so the API returns in milliseconds.
Since its a static field I've also appended a suffix that specifies the unit of measurement for clarity.

src/main/java/org/spongepowered/common/guice/SpongePluginGuiceModule.java
```diff
@@ -156,4 +165,40 @@ private PrivateHoconConfigProvider(PluginContainer container) {
...
+
+        @Inject
+        private SynchronousExecutorProvider(PluginContainer container, Game game) {
+            this.container = container;
+            this.schedulerService = game.getScheduler();
+        }
+
+        @Override
+        public SpongeExecutorService get() {
+            return this.schedulerService.createSyncExecutor(this.container.getInstance());
```
simon816: No need for getInstance(), the container can be passed

src/main/java/org/spongepowered/common/guice/SpongePluginGuiceModule.java
```diff
@@ -156,4 +165,40 @@ private PrivateHoconConfigProvider(PluginContainer container) {
...
+
+        @Inject
+        private AsynchronousExecutorProvider(PluginContainer container, Game game) {
+            this.container = container;
+            this.schedulerService = game.getScheduler();
+        }
+
+        @Override
+        public SpongeExecutorService get() {
+            return this.schedulerService.createAsyncExecutor(this.container.getInstance());
```
simon816: No need for getInstance(), the container can be passed

Kiskae:
I've improved the implementation of the internal Futures based on http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/util/concurrent/ScheduledThreadPoolExecutor.java#ScheduledThreadPoolExecutor.ScheduledFutureTask.compareTo%28java.util.concurrent.Delayed%29
It now uses the tasks' internal execution timestamp for comparison if applicable.

PR #209, opened 2015-10-16
Move DataBuilder code into a shared class

JBYoshi:
It seems that the `DataBuilder`s have quite a bit of code shared with their corresponding `DataProcessor`s. This PR moves the code into a shared class, `SpongeDataBuilder`, which uses reflection for `create()` and delegates `createFrom(DataHolder)` and `build(DataView)` to the `DataProcessor`. This PR also provides implementations for `ImmutableDataBuilder` via `SpongeImmutableDataBuilder`.

@gabizou Thoughts?


src/main/java/org/spongepowered/common/data/builder/manipulator/SpongeDataBuilder.java
```diff
@@ -22,47 +22,48 @@
...
+import org.spongepowered.common.util.ReflectionUtil;
 
 import java.util.Optional;
 
-public class HorseDataBuilder implements DataManipulatorBuilder<HorseData, ImmutableHorseData> {
+public class SpongeDataBuilder<M extends DataManipulator<M, I>, I extends ImmutableDataManipulator<I, M>> implements DataManipulatorBuilder<M, I> {
 
-    private HorseDataProcessor processor;
+    private final Class<? extends M> manipulatorClass;
+    private final DataProcessor<M, I> processor;
```
saladoc: This would need to support multiple `DataProcessor`s. It might be possible to achieve this using a `DataProcessorDelegate` though.

src/main/java/org/spongepowered/common/data/SpongeDataRegistry.java
```diff
@@ -206,6 +202,11 @@ public static void finalizeRegistration() {
         }
         checkArgument(!immutableProcessorList.contains(processor), "Duplicate DataProcessor Registration!");
         immutableProcessorList.add(processor);
+        
+        SpongeSerializationService service = (SpongeSerializationService) Sponge.getGame().getServiceManager().provide(SerializationService.class).get();
+        SpongeDataBuilder<T, I> builder = new SpongeDataBuilder<>(implClass, processor);
+        service.registerBuilderAndImpl(manipulatorClass, implClass, builder);
```
saladoc: Better move Builder registration to `finalizeRegistration()`. Since multiple `DataProcessor`s may be registered for any `DataManipulator` you must wait until all are registered before creating the builder from the `DataProcessorDelegate`. Otherwise, you'd be getting errors since there can only be one builder per `DataManipulator`.

Deamon5550:
So you're replacing a direct implementation with one based on reflection, you should probably consider the performance impact of all of those reflection calls, in particular I was thinking about the frequency that `create()` may be called.

saladoc:
@Deamon5550 I think that depends on whether the `DataManipulatorBuilder`s are used from Sponge itself or only from plugins using the API. In the latter case I figure the calls are very few and the reflection not that pricy. Far worse things happen underneath the Data API implementation (e.g. synchronous `GameProfile` lookups on the main thread when setting a `Skull`s owner).

Deamon5550:
Just something to keep in mind.

gabizou:
`DataManipulatorBuilder`s are used for EVERY `DataManipulator`, regardless whether it's Sponge provided or plugin provided. At the very least, I was considering trying do this very same idea, however, again, the issue is reflection.

Another thing is: Unnecessary creation of various `DataManipulator`s without validating first that the `DataContainer` has the proper data is an issue by itself. I was considering however, segmenting out `create()` and `createFrom(DAtaHolder)` such that method references could either be sent out as `Function`s or something thereof.

src/main/java/org/spongepowered/common/data/SpongeDataRegistry.java
```diff
@@ -206,6 +202,11 @@ public static void finalizeRegistration() {
         }
         checkArgument(!immutableProcessorList.contains(processor), "Duplicate DataProcessor Registration!");
         immutableProcessorList.add(processor);
+        
+        SpongeSerializationService service = (SpongeSerializationService) Sponge.getGame().getServiceManager().provide(SerializationService.class).get();
```
gabizou: This is also a no no, because it really really prevents any possibility of running `ManipulatorTest` since using anything with `Sponge` will result in complete failure.

src/main/java/org/spongepowered/common/data/builder/manipulator/SpongeDataBuilder.java
gabizou: This should never support multiple `DataProcessor`s, the point of a `DataBuilder` having a single instance is that there is only one provider of that possible existing `DataManipulator` or whatever it is, and that provider also knows the deserialization process required. Other providers shouldn't actually be providing any sort of interaction with this.

JBYoshi: @gabizou I understand for `build(DataView)`, but what about `createFrom(DataHolder)`? I believe mods are supposed to add their own `DataProcessor`s if they want Sponge compatibility.

src/main/java/org/spongepowered/common/data/SpongeDataRegistry.java
JBYoshi: Will fix this.

src/main/java/org/spongepowered/common/data/SpongeDataRegistry.java
JBYoshi: I don't see any work-around. Would moving this to `finalizeRegistration()` fix this?

JBYoshi: I'll address this by making `SpongeSerializationService` a singleton.

src/main/java/org/spongepowered/common/data/SpongeDataRegistry.java
JBYoshi: Done.

gabizou:
You've got some interesting changes that I don't quite see as to why they're resulting in entire files being deleted and rewritten. I think I have an alternative solution to reflection though that will make `DataManipulatorBuilder`s work in a factory sense.

gabizou:
So, if it is line endings issues: https://help.github.com/articles/dealing-with-line-endings/#platform-windows is a good way to fix them. Unfortunately, it means having to redo any changes of the files that git already marked as re-created.

JBYoshi:
@gabizou Thanks.

JBYoshi:
@gabizou I just thought of a way to avoid the reflection performance costs. When the `DataManipulator` is registered, I create **one and only one** instance of the `ImmutableDataManipulator` using reflection. `DataBuilder.create()` will convert it to a mutable `DataManipulator` and return that, while `ImmutableDataBuilder.createImmutable()` will return the immutable single instance.

gabizou:
@JBYoshi I was just fiddling with `DataConstants`, I realized that there would've been a big issue with the way you approached this is that serialization setup occurs during the `preinit` phase of the `SpongeGameRegistry` whereas the `CatalogType`s are all registered during the `init` phase, of which any references to the actual static values would've equated to `null` at runtime during the `preinit` phase of which you were hoping to use an `ImmutableDataManipulator` variant for the `DataManipulatorBuilder`s. The issue with this is that since you now have this `ImmutableDataManipualtor` that would've still failed to initialize due to the various null checks, it would've been impossible to have this work when in the context of actually working with the `CatalogType`s that haven't been initialized yet.

Granted, with the `DataConstants` I have to be extremely careful with the static initialization of the `DataConstants` class, so there's some delicacy to using various `CatalogType` constants throughout the Data implementation that have to be carefully used only and only when the `SpongeGameRegistry` has completed all registrations.

PR #213, opened 2015-10-18
Implement ColoredData

kenzierocks:
Implementation of `SpongeColoredData`.
- [X] Accurately used the appropriate `AbstractData` implementation

Implementation of `ImmutableSpongeColoredData`
- [X] Accurately extended the appropriate `AbstractImmutableData` implementation
- [X] Accurately instantiating final instance fields with an `ImmutableValue` counter part for any value getters (probably? cache utils)

Implementation of the `ItemColorValueProcessor`(s):
- [X] Appropriately extended `AbstractSpongeValueProcessor`
- [X] Individual processors for each source type possible (one for `ItemStack`/`TileEntity`/`Entity` as necessary).

Registration:
- [X] Registered the `Key` correctly by `color` in the `KeyRegistry`
- [X] Registered the `DataProcessor`s and `DataManipulatorBuilder` in `SpongeSerializationRegistry`

Test Plugin code:
http://sh.kenzierocks.me/coloreddataplugin

### DIFF
--- a/src/main/java/org/spongepowered/common/data/processor/data/ColoredDataProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/data/ColoredDataProcessor.java
@@ -47,16 +47,16 @@ import net.minecraft.item.ItemStack;
 public class ColoredDataProcessor extends AbstractItemSingleDataProcessor<Color, Value<Color>, ColoredData, ImmutableColoredData> {
 
     public ColoredDataProcessor() {
-        super(stack -> ColorUtil.getItemStackColor(stack).isPresent(), Keys.COLOR);
+        super(ColorUtil::hasColor, Keys.COLOR);
     }
 
     @Override
     public DataTransactionResult remove(DataHolder dataHolder) {
         if (dataHolder instanceof ItemStack) {
-            final DataTransactionBuilder builder = DataTransactionBuilder.builder();
             final Optional<ColoredData> optional = from(dataHolder);
             final ItemStack stack = (ItemStack) dataHolder;
-            if (ColorUtil.hasItemStackColor(stack) && optional.isPresent()) {
+            if (ColorUtil.hasNbtColor(stack) && optional.isPresent()) {
+                final DataTransactionBuilder builder = DataTransactionBuilder.builder();
                 try {
                     NbtDataUtil.removeColorFromNBT(stack);
                     return builder.replace(optional.get().getValues()).result(DataTransactionResult.Type.SUCCESS).build();
@@ -65,7 +65,7 @@ public class ColoredDataProcessor extends AbstractItemSingleDataProcessor<Color,
                     return builder.result(DataTransactionResult.Type.ERROR).build();
                 }
             } else {
-                return builder.result(DataTransactionResult.Type.SUCCESS).build();
+                return DataTransactionBuilder.successNoData();
             }
         }
         return DataTransactionBuilder.failNoData();
--- a/src/main/java/org/spongepowered/common/data/processor/data/ColoredDataProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/data/ColoredDataProcessor.java
@@ -28,6 +28,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import java.awt.Color;
 import java.util.Optional;
+import java.util.function.Predicate;
 
 import org.spongepowered.api.data.DataContainer;
 import org.spongepowered.api.data.DataHolder;
@@ -39,63 +40,25 @@ import org.spongepowered.api.data.manipulator.immutable.ImmutableColoredData;
 import org.spongepowered.api.data.manipulator.mutable.ColoredData;
 import org.spongepowered.api.data.merge.MergeFunction;
 import org.spongepowered.api.data.value.BaseValue;
+import org.spongepowered.api.data.value.immutable.ImmutableValue;
+import org.spongepowered.api.data.value.mutable.Value;
 import org.spongepowered.common.Sponge;
 import org.spongepowered.common.data.manipulator.immutable.ImmutableSpongeColoredData;
 import org.spongepowered.common.data.manipulator.mutable.SpongeColoredData;
+import org.spongepowered.common.data.processor.common.AbstractItemSingleDataProcessor;
 import org.spongepowered.common.data.processor.common.AbstractSpongeDataProcessor;
 import org.spongepowered.common.data.util.DataUtil;
 import org.spongepowered.common.data.util.NbtDataUtil;
 import org.spongepowered.common.util.ColorUtil;
 
+import net.minecraft.init.Items;
+import net.minecraft.item.ItemArmor;
 import net.minecraft.item.ItemStack;
 
-public class ColoredDataProcessor extends AbstractSpongeDataProcessor<ColoredData, ImmutableColoredData> {
+public class ColoredDataProcessor extends AbstractItemSingleDataProcessor<Color, Value<Color>, ColoredData, ImmutableColoredData> {
 
-    @Override
-    public boolean supports(DataHolder dataHolder) {
-        return dataHolder instanceof ItemStack;
-    }
-
-    @Override
-    public Optional<ColoredData> from(DataHolder dataHolder) {
-        if (dataHolder instanceof ItemStack) {
-            final ItemStack stack = (ItemStack) dataHolder;
-            return ColorUtil.getItemStackColor(stack).map(SpongeColoredData::new);
-        }
-        return Optional.empty();
-    }
-
-    @Override
-    public Optional<ColoredData> fill(DataHolder dataHolder, ColoredData manipulator, MergeFunction overlap) {
-        if (supports(dataHolder)) {
-            final ColoredData data = from(dataHolder).orElse(null);
-            final ColoredData newData = checkNotNull(overlap.merge(checkNotNull(manipulator), data));
-            final Color color = newData.color().get();
-            return Optional.of(manipulator.set(Keys.COLOR, color));
-        }
-        return Optional.empty();
-    }
-
-    @Override
-    public Optional<ColoredData> fill(DataContainer container, ColoredData colorData) {
-        if (container.contains(Keys.COLOR.getQuery())) {
-            final Color color = DataUtil.getData(container, Keys.COLOR, Color.class);
-            return Optional.of(colorData.set(Keys.COLOR, color));
-        }
-        return Optional.empty();
-    }
-
-    @Override
-    public DataTransactionResult set(DataHolder dataHolder, ColoredData manipulator, MergeFunction function) {
-        return DataTransactionBuilder.failResult(manipulator.getValues());
-    }
-
-    @Override
-    public Optional<ImmutableColoredData> with(Key<? extends BaseValue<?>> key, Object value, ImmutableColoredData immutable) {
-        if (key == Keys.COLOR) {
-            return Optional.of(new ImmutableSpongeColoredData((Color) value));
-        }
-        return Optional.empty();
+    protected ColoredDataProcessor() {
+        super(stack -> ColorUtil.getItemStackColor(stack).isPresent(), Keys.COLOR);
     }
 
     @Override
@@ -120,8 +83,25 @@ public class ColoredDataProcessor extends AbstractSpongeDataProcessor<ColoredDat
     }
 
     @Override
-    public Optional<ColoredData> createFrom(DataHolder dataHolder) {
-        return from(dataHolder);
+    protected boolean set(ItemStack itemStack, Color value) {
+        ColorUtil.setItemStackColor(itemStack, value);
+        return true;
+    }
+
+    @Override
+    protected Optional<Color> getVal(ItemStack itemStack) {
+        return ColorUtil.getItemStackColor(itemStack);
+    }
+
+    @Override
+    protected ImmutableValue<Color> constructImmutableValue(Color value) {
+        return null;
+    }
+
+    @Override
+    protected ColoredData createManipulator() {
+        // TODO Auto-generated method stub
+        return null;
     }
 
 }
--- a/src/main/java/org/spongepowered/common/data/processor/data/ColoredDataProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/data/ColoredDataProcessor.java
@@ -78,8 +78,11 @@ public class ColoredDataProcessor extends AbstractSpongeDataProcessor<ColoredDat
 
     @Override
     public Optional<ColoredData> fill(DataContainer container, ColoredData colorData) {
-        final Color color = DataUtil.getData(container, Keys.COLOR, Color.class);
-        return Optional.of(colorData.set(Keys.COLOR, color));
+        if (container.contains(Keys.COLOR.getQuery())) {
+            final Color color = DataUtil.getData(container, Keys.COLOR, Color.class);
+            return Optional.of(colorData.set(Keys.COLOR, color));
+        }
+        return Optional.empty();
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


src/main/java/org/spongepowered/common/data/processor/data/ColoredDataProcessor.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Override
+    public boolean supports(DataHolder dataHolder) {
+        return dataHolder instanceof EntitySheep || dataHolder instanceof ItemStack || dataHolder instanceof IWorldNameable;
+    }
+
+    @Override
+    public Optional<ColoredData> from(DataHolder dataHolder) {
+        Color color = null;
+        if (dataHolder instanceof EntitySheep) {
```
Aaron1011: You might want to register a different processor for each type of `DataHolder` that supports `ColoredData`.

src/main/java/org/spongepowered/common/data/processor/data/ColoredDataProcessor.java
```diff
@@ -0,0 +1,149 @@
...
+import net.minecraft.init.Items;
+import net.minecraft.item.EnumDyeColor;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemArmor;
+import net.minecraft.item.ItemBlock;
+import net.minecraft.item.ItemStack;
+import net.minecraft.nbt.NBTTagCompound;
+import net.minecraft.world.IWorldNameable;
+
+public class ColoredDataProcessor extends AbstractSpongeDataProcessor<ColoredData, ImmutableColoredData> {
```
gabizou: This should likely be split up into three separate processor implementations, one for `ItemStack`s, `EntitySheep`, and `IWorldNameable`.

src/main/java/org/spongepowered/common/data/processor/data/ColoredDataProcessor.java
```diff
@@ -0,0 +1,149 @@
...
+            final ColoredData newData = checkNotNull(overlap.merge(checkNotNull(manipulator), data));
+            final Color color = newData.color().get();
+            return Optional.of(manipulator.set(Keys.COLOR, color));
+        }
+        return Optional.empty();
+    }
+
+    @Override
+    public Optional<ColoredData> fill(DataContainer container, ColoredData colorData) {
+        final Color color = DataUtil.getData(container, Keys.COLOR, Color.class);
```
gabizou: Needs to handle the case that if the container doesn't have the key, return `Optional.empty()`.

src/main/java/org/spongepowered/common/data/processor/value/ColorValueProcessor.java
```diff
@@ -0,0 +1,71 @@
...
+import org.spongepowered.api.data.value.ValueContainer;
+import org.spongepowered.api.data.value.immutable.ImmutableValue;
+import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.common.data.processor.common.AbstractSpongeValueProcessor;
+import org.spongepowered.common.data.value.immutable.ImmutableSpongeValue;
+import org.spongepowered.common.data.value.mutable.SpongeValue;
+
+import java.util.Optional;
+
+public class ColorValueProcessor extends AbstractSpongeValueProcessor<Entity, Boolean, Value<Boolean>> {
```
gabizou: Rename tis to `EntityColorValueProcessor` and another one for `ItemColorValueProcessor`.

src/main/java/org/spongepowered/common/data/processor/data/ColoredDataProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+import org.spongepowered.common.data.manipulator.immutable.ImmutableSpongeColoredData;
+import org.spongepowered.common.data.manipulator.mutable.SpongeColoredData;
+import org.spongepowered.common.data.processor.common.AbstractSpongeDataProcessor;
+import org.spongepowered.common.data.util.DataUtil;
+import org.spongepowered.common.data.util.NbtDataUtil;
+import org.spongepowered.common.util.ColorUtil;
+
+import net.minecraft.item.ItemStack;
+
+public class ColoredDataProcessor extends AbstractSpongeDataProcessor<ColoredData, ImmutableColoredData> {
```
Aaron1011: You can extend `AbstractItemSingleDataProcessor`, which will make this a lot simpler.

kenzierocks:
Probably good now.

src/main/java/org/spongepowered/common/data/processor/value/item/ItemColorValueProcessor.java
```diff
@@ -0,0 +1,92 @@
...
+import org.spongepowered.common.Sponge;
+import org.spongepowered.common.data.processor.common.AbstractSpongeValueProcessor;
+import org.spongepowered.common.data.util.NbtDataUtil;
+import org.spongepowered.common.data.value.immutable.ImmutableSpongeValue;
+import org.spongepowered.common.data.value.mutable.SpongeValue;
+import org.spongepowered.common.util.ColorUtil;
+
+import net.minecraft.item.ItemStack;
+
+public class ItemColorValueProcessor extends AbstractSpongeValueProcessor<ItemStack, Color, Value<Color>> {
```
gabizou: Override `supports(ItemStack)` otherwise you're allowing anyone to set the color onto any item type.

src/main/java/org/spongepowered/common/data/util/NbtDataUtil.java
```diff
@@ -24,22 +24,25 @@
...
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import org.spongepowered.api.data.meta.ItemEnchantment;
+import org.spongepowered.api.item.Enchantment;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.common.text.SpongeTexts;
+
```
gabizou: Fix your import style.

src/main/java/org/spongepowered/common/data/processor/value/item/ItemColorValueProcessor.java
kenzierocks: So only leather armor items can be affected by color? No support for mods that may do the same?

src/main/java/org/spongepowered/common/data/util/NbtDataUtil.java
kenzierocks: Fixed.

src/main/java/org/spongepowered/common/data/util/NbtDataUtil.java
```diff
@@ -196,6 +198,31 @@ public static void setLoreToNBT(ItemStack stack, List<Text> lore) {
         stack.getSubCompound(ITEM_DISPLAY, true).setTag(ITEM_LORE, list);
     }
 
+    public static boolean hasColorFromNBT(ItemStack stack) {
+        return !stack.hasTagCompound() ||
+                !stack.getTagCompound().hasKey(ITEM_DISPLAY) ||
+                !stack.getTagCompound().getCompoundTag(ITEM_DISPLAY).hasKey(ITEM_COLOR);
+    }
```
lucdon: remove the ! and replace the || with && 
else it would have a color if the stack doesn't have a tagCompound

kenzierocks: Done.

src/main/java/org/spongepowered/common/data/processor/data/ColoredDataProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+import org.spongepowered.common.data.util.NbtDataUtil;
+import org.spongepowered.common.data.value.immutable.ImmutableSpongeValue;
+import org.spongepowered.common.util.ColorUtil;
+
+import net.minecraft.item.ItemStack;
+
+public class ColoredDataProcessor extends AbstractItemSingleDataProcessor<Color, Value<Color>, ColoredData, ImmutableColoredData> {
+
+    public ColoredDataProcessor() {
+        super(stack -> ColorUtil.getItemStackColor(stack).isPresent(), Keys.COLOR);
```
gabizou: Why not just have the util check that it supports color instead of trying to get the color twice?

kenzierocks: Done.

src/main/java/org/spongepowered/common/data/processor/value/item/ItemColorValueProcessor.java
```diff
@@ -0,0 +1,92 @@
...
+            if (ColorUtil.hasNbtColor(stack) && optional.isPresent()) {
+                try {
+                    NbtDataUtil.removeColorFromNBT(stack);
+                    return builder.replace(new ImmutableSpongeValue<>(Keys.COLOR, optional.get())).result(DataTransactionResult.Type.SUCCESS).build();
+                } catch (Exception e) {
+                    Sponge.getLogger().error("There was an issue removing the color from an itemstack!", e);
+                    return builder.result(DataTransactionResult.Type.ERROR).build();
+                }
+            } else {
+                return builder.result(DataTransactionResult.Type.SUCCESS).build();
```
kenzierocks: Not sure if this should be success or failure.

src/main/java/org/spongepowered/common/data/processor/data/ColoredDataProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+                final DataTransactionBuilder builder = DataTransactionBuilder.builder();
+                try {
+                    NbtDataUtil.removeColorFromNBT(stack);
+                    return builder.replace(optional.get().getValues()).result(DataTransactionResult.Type.SUCCESS).build();
+                } catch (Exception e) {
+                    Sponge.getLogger().error("There was an issue removing the color from an ItemStack!", e);
+                    return builder.result(DataTransactionResult.Type.ERROR).build();
+                }
+            } else {
+                return DataTransactionBuilder.successNoData();
```
kenzierocks: Same success/failure question here.

src/main/java/org/spongepowered/common/data/processor/value/item/ItemColorValueProcessor.java
lucdon: return DataTransactionBuilder.failNoData();

gabizou:
Plugin code?

src/main/java/org/spongepowered/common/data/processor/value/item/ItemColorValueProcessor.java
ZephireNZ: DataProcessors are usually designed with just vanilla in mind. A mod could potentially make an entity be colorable too - but we simply can't support that level of mod compatibility, unfortunately.

Mods can create custom `DataProcessor`s and `ValueProcessors` and register them to support their mod, as Sponge actually keeps a list of processors, and tries each to see a container contains that data/value.

Zidane: Just to continue but having the mod do it isn't required. If the mod provides a deobf jar or the BON tool can deobf the jar...one could make a plugin with a dependency on the mod and register processors that add support for that mod while keeping Vanilla support :).

kenzierocks: Ok,will do.

src/main/java/org/spongepowered/common/data/util/NbtDataUtil.java
```diff
@@ -196,6 +198,31 @@ public static void setLoreToNBT(ItemStack stack, List<Text> lore) {
         stack.getSubCompound(ITEM_DISPLAY, true).setTag(ITEM_LORE, list);
     }
 
+    public static boolean hasColorFromNBT(ItemStack stack) {
+        return stack.hasTagCompound() &&
+                stack.getTagCompound().hasKey(ITEM_DISPLAY) &&
+                stack.getTagCompound().getCompoundTag(ITEM_DISPLAY).hasKey(ITEM_COLOR);
+    }
+
+    public static Optional<Color> getColorFromNBT(NBTTagCompound subCompound) {
```
kenzierocks: This doesn't get used anymore. Keep it?

src/main/java/org/spongepowered/common/data/processor/value/item/ItemColorValueProcessor.java
kenzierocks: Done.

kenzierocks:
Please hold, writing test code and rebasing.

kenzierocks:
All good.

kenzierocks:
Added test plugin http://sh.kenzierocks.me/coloreddataplugin.

gabizou:
Can you add code to your test plugin to test serialization/deserialization?

src/main/java/org/spongepowered/common/data/util/NbtDataUtil.java
```diff
@@ -196,6 +198,31 @@ public static void setLoreToNBT(ItemStack stack, List<Text> lore) {
...
+    public static void removeColorFromNBT(ItemStack stack) {
+        if(stack.getSubCompound(ITEM_DISPLAY, false) == null) {
+            return;
+        }
+        stack.getSubCompound(ITEM_DISPLAY, false).removeTag(ITEM_COLOR);
+    }
+
+    public static void setColorToNbt(ItemStack stack, Color color) {
+        final int iColor = color.getRGB();
+        stack.getSubCompound(ITEM_DISPLAY, true).setInteger(ITEM_COLOR, iColor);
```
Aaron1011: According to the [API javadocs](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/data/manipulator/mutable/ColoredData.java#L35), you'll need to support setting the color of wool as well.

gabizou: If and only if the color is a valid `Color` matching a `DyeColor`.

JBYoshi: @Aaron1011 The API javadocs are incorrect. Wool is only 16 colors; `java.awt.Color` is 4294967296 colors. Wool colors are handled by `DyeColor`.

kenzierocks:
@gabizou I think I added deserialization to the test? If not, you'll need to be more clear, apologies.

src/main/java/org/spongepowered/common/data/util/NbtDataUtil.java
gabizou: @JBYoshi It's still possible to translate a `Color` to a `DyeColor` provided that they are matching up exactly.

ST-DDT: Then a plugin/mod/update changes and then everybody wonders about strange behaviour. Consts stored in plugins does not work anymore. Or the value from get does not match the one you set a moment ago.

kenzierocks: @ST-DDT We would get the color values from Mojang/vanilla `EnumDyeColor`, no constants on our side.

JBYoshi: @kenzierocks What about constants stored in plugins?

kenzierocks: They should be using `DyeColor.getColor` if they're trying to set colors like that.

ST-DDT: And the other way round thry should use something like `DyeCole.from(Color)` as well.
If they have to use a valid DyeColor anyway to get the attached Color they could just take the DyeColor itself and set it.
IMO using Colors here is like using magic numbers. Some might work but most others will not. You need more context to know which ones will work. However that context alone is sufficient to get and set the values itself.
Just thing about the quota 0.00...01% will work. (I dont have a time for a calculator currently)

PR #217, opened 2015-10-20
[WIP] Data API cleanup

ZephireNZ:
API side: SpongePowered/SpongeAPI#893

This cleans up some data implementations, especially where the requirements for data has been changed recently.

- Final/Private values
  * `ImmutableBreathingData`
  * `ImmutableFoodData`
  * `ImmutableMovementSpeedData`
- Fixing `Value`s being created with uninitialized value
  * `ImmutableBreathingData`
- Move `Value`s to instance fields
  * `ImmutableExperienceHolderData`
  * `ImmutableFlyingData`
  * `ImmutableFoodData`
  * `ImmutableHorseData`
  * `ImmutableIgnitableData` 
  * `ImmutableConnectedDirectionData`
  * `ImmutableWireAttachmentData`
  * `ImmutableVelocityData`
  * `ImmutableColorData`
  * `ImmutableCommandData`
  * `ImmutableDisplayNameData`
  * `ImmutableFireworkData`
  * `ImmutableMobSpawnerData`
  * `ImmutableRepresentedItemData`
  * `ImmutableRepresentedPlayerData`
  * `ImmutableTargetedLocationData`
- Caching usage
  * `ImmutableFoodData`
- Fixing of defaults and bounds
  * `ImmutableMovementSpeedData`/`MovementSpeedData`
  * `ImmutableHealthData`/`HealthData`
- Use appropriate abstract data
  * `ImmutableLoreData`
  * `ImmutablePagedData`
  * `ImmutableSignData`
  * `ImmutableExpOrbData`/`ExpOrbData`
  * `ImmutableBrewingStandData`/`BrewingStandData`
- Implement data getter/setter registration
  * `TradeOfferData`
  * `ConnectedDirectionData`
  * `ImmutableCommandData`
  * `ImmutableFireworkData`
  * `ImmutableMobSpawnerData`
- Refactor values into the Abstract class
  * `AbstractBooleanData` and its implementations
  * `AbstractSingleEnumData` and its implementations
  * `AbstractBoundedComparableValue` and its implementations

### DIFF
--- a/src/main/java/org/spongepowered/common/data/key/KeyRegistry.java
+++ b/src/main/java/org/spongepowered/common/data/key/KeyRegistry.java
@@ -144,6 +144,9 @@ public class KeyRegistry {
         keyMap.put("spawner_minimum_delay", makeSingleKey(Short.class, MutableBoundedValue.class, of("SpawnerMinimumDelay")));
         keyMap.put("connected_directions", makeSetKey(Direction.class, of("ConnectedDirections")));
         keyMap.put("connected_north", makeSingleKey(Boolean.class, Value.class, of("ConnectedNorth")));
+        keyMap.put("connected_south", makeSingleKey(Boolean.class, Value.class, of("ConnectedSouth")));
+        keyMap.put("connected_east", makeSingleKey(Boolean.class, Value.class, of("ConnectedEast")));
+        keyMap.put("connected_west", makeSingleKey(Boolean.class, Value.class, of("ConnectedWest")));
         keyMap.put("direction", makeSingleKey(Direction.class, Value.class, of("Direction")));
         keyMap.put("dirt_type", makeSingleKey(DirtType.class, Value.class, of("DirtType")));
         keyMap.put("disguised_block_type", makeSingleKey(DisguisedBlockType.class, Value.class, of("DisguisedBlockType")));
--- a/src/main/java/org/spongepowered/common/data/manipulator/immutable/ImmutableSpongeCommandData.java
+++ b/src/main/java/org/spongepowered/common/data/manipulator/immutable/ImmutableSpongeCommandData.java
@@ -29,12 +29,14 @@ import org.spongepowered.api.data.MemoryDataContainer;
 import org.spongepowered.api.data.key.Keys;
 import org.spongepowered.api.data.manipulator.immutable.ImmutableCommandData;
 import org.spongepowered.api.data.manipulator.mutable.CommandData;
+import org.spongepowered.api.data.value.immutable.ImmutableBoundedValue;
 import org.spongepowered.api.data.value.immutable.ImmutableOptionalValue;
 import org.spongepowered.api.data.value.immutable.ImmutableValue;
 import org.spongepowered.api.text.Text;
 import org.spongepowered.api.text.Texts;
 import org.spongepowered.common.data.manipulator.immutable.common.AbstractImmutableData;
 import org.spongepowered.common.data.manipulator.mutable.SpongeCommandData;
+import org.spongepowered.common.data.value.SpongeValueBuilder;
 import org.spongepowered.common.data.value.immutable.ImmutableSpongeOptionalValue;
 import org.spongepowered.common.data.value.immutable.ImmutableSpongeValue;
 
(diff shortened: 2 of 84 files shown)


src/main/java/org/spongepowered/common/data/manipulator/immutable/entity/ImmutableSpongeBreathingData.java
```diff
@@ -38,25 +38,30 @@
...
+
+    private final ImmutableBoundedValue<Integer> remainingAirValue;
+    private final ImmutableBoundedValue<Integer> maxAirValue;
 
     public ImmutableSpongeBreathingData(int maxAir, int remainingAir) {
         super(ImmutableBreathingData.class);
         this.maxAir = maxAir;
         this.remainingAir = remainingAir;
+
+        remainingAirValue = new ImmutableSpongeBoundedValue<>(Keys.REMAINING_AIR,
```
gabizou: You could just use the `SpongeValueBuilder.boundedBuilder(...`

ZephireNZ: I did not realise that existed! Thanks.

ZephireNZ: Ah, that'd be why I couldn't find it - committed just recently.

ZephireNZ:
@gabizou - thoughts on removing the final variables from immutable data, instead accessing them through the value?

eg
```java
public int getMaxAir() {
    return maxAirValue.get();
}
```

Rather than
```java
public int getMaxAir() {
    return this.maxAir;
}
````

**Comment by gabizou** [comment]
@ZephireNZ I'd rather not. It's cheaper still to use field access than to call the value object to get the value.

File: src/main/java/org/spongepowered/common/data/key/KeyRegistry.java [hunk 2]
```diff
@@ -127,6 +127,9 @@ private static void generateKeyMap() {
         keyMap.put("spawner_minimum_delay", makeSingleKey(Short.class, MutableBoundedValue.class, of("SpawnerMinimumDelay")));
         keyMap.put("connected_directions", makeSetKey(Direction.class, of("ConnectedDirections")));
         keyMap.put("connected_north", makeSingleKey(Boolean.class, Value.class, of("ConnectedNorth")));
+        keyMap.put("connected_south", makeSingleKey(Boolean.class, Value.class, of("ConnectedSouth")));
+        keyMap.put("connected_east", makeSingleKey(Boolean.class, Value.class, of("ConnectedEast")));
+        keyMap.put("connected_west", makeSingleKey(Boolean.class, Value.class, of("ConnectedWest")));
```
ZephireNZ: For serialisation this probably seems odd - maybe need to change datacontainer to only output ConnectedDirections instead of all five?

ZephireNZ:
Looks like there's some duplication between this and #234 - refactoring is gonna be soo fun...

gabizou:
@ZephireNZ Yeah.... considering that PR attempts to solve a lot of problems at the same time, it might take some time at the least. It might be divided into phases as @Mumfrey and I have pondered on some ways to resolve the issue of "stateless" `CatalogType` references while at the same time revamping the overall `SpongeGameRegistry` as a whole.

src/main/java/org/spongepowered/common/data/manipulator/immutable/ImmutableSpongeCommandData.java
```diff
@@ -47,33 +48,50 @@
...
+
+        this.storedValue = new ImmutableSpongeValue<>(Keys.COMMAND, this.storedCommand);
+        this.successValue = SpongeValueBuilder.boundedBuilder(Keys.SUCCESS_COUNT)
+                .actualValue(this.success)
+                .defaultValue(0)
+                .minimum(0)
+                .maximum(Integer.MAX_VALUE)
+                .build()
+                .asImmutable();
+        this.tracksValue = ImmutableSpongeValue.cachedOf(Keys.TRACKS_OUTPUT, this.tracks, false);
```
gabizou: The arguments are switched.

src/main/java/org/spongepowered/common/data/manipulator/immutable/entity/ImmutableSpongeBreathingData.java
```diff
@@ -33,13 +33,14 @@
 import org.spongepowered.api.data.value.immutable.ImmutableBoundedValue;
 import org.spongepowered.common.data.manipulator.immutable.common.AbstractImmutableData;
 import org.spongepowered.common.data.manipulator.mutable.entity.SpongeBreathingData;
-import org.spongepowered.common.data.util.ComparatorUtil;
-import org.spongepowered.common.data.value.immutable.ImmutableSpongeBoundedValue;
+
+import static org.spongepowered.common.data.value.SpongeValueBuilder.boundedBuilder;
```
gabizou: no.

src/main/java/org/spongepowered/common/data/manipulator/mutable/SpongeFireworkData.java
```diff
@@ -24,6 +24,9 @@
  */
 package org.spongepowered.common.data.manipulator.mutable;
 
+import static com.google.common.base.Preconditions.*;
```
gabizou: no star imports please.

src/main/java/org/spongepowered/common/data/manipulator/mutable/entity/SpongeMovementSpeedData.java
```diff
@@ -84,8 +84,8 @@ protected void registerGettersAndSetters() {
     @Override
     public MutableBoundedValue<Double> walkSpeed() {
         return SpongeValueBuilder.boundedBuilder(Keys.WALKING_SPEED)
-            .defaultValue(0.1D)
-            .minimum(Double.MIN_VALUE)
+            .defaultValue(0.699999988079071D) // Why not 0.7D? Ask Mojang
+            .minimum(0D)
```
gabizou: the minimum shouldn't be restricted in this case.

src/main/java/org/spongepowered/common/data/manipulator/mutable/entity/SpongeMovementSpeedData.java
```diff
@@ -84,8 +84,8 @@ protected void registerGettersAndSetters() {
     @Override
     public MutableBoundedValue<Double> walkSpeed() {
         return SpongeValueBuilder.boundedBuilder(Keys.WALKING_SPEED)
-            .defaultValue(0.1D)
-            .minimum(Double.MIN_VALUE)
+            .defaultValue(0.699999988079071D) // Why not 0.7D? Ask Mojang
```
gabizou: I'm pretty sure the default walking speed is not 0.7D.

gabizou:
All well and fine. Just make sure it's rebased properly on master and I'd be happy to merge.

src/main/java/org/spongepowered/common/data/manipulator/mutable/entity/SpongeMovementSpeedData.java
simon816: @ZephireNZ It's got this weird value because 0.7 can't be represented in binary without rounding, the same way 1/3 is 0.3333... in decimal.
This is what happens when decompiling the class, it is 0.7D in Mojang's code

ZephireNZ: @gabizou - The 0.7D was got from the `SharedMonsterAttributes` - as far as I can tell this is what's used in vanilla as the default movement speed, with some mobs overriding that.

PR #220, opened 2015-10-21
Fix recursive commands.

JBYoshi:
Fixes SpongePowered/SpongeForge#381.

### DIFF
--- a/src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
+++ b/src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
@@ -69,7 +69,7 @@ public class MinecraftCommandWrapper implements CommandCallable {
 
         @Override
         protected Deque<Throwable> initialValue() {
-            return new LinkedList<Throwable>();
+            return new LinkedList<>();
         }
     };
     // This differs from null in that null means "not active".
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
```diff
@@ -209,8 +218,14 @@ public Text getUsage(CommandSource source) {
     }
 
     public static void setError(Throwable error) {
-        if (commandErrors.get() == noError) {
-            commandErrors.set(error);
+        Preconditions.checkNotNull(error);
```
gabizou: Can static import this method.

JBYoshi: Done.

src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
```diff
@@ -61,7 +65,13 @@
     private static final String TRANSLATION_NO_PERMISSION = "commands.generic.permission";
     private final PluginContainer owner;
     protected final ICommand command;
-    private static final ThreadLocal<Throwable> commandErrors = new ThreadLocal<>();
+    private static final ThreadLocal<Deque<Throwable>> commandErrors = new ThreadLocal<Deque<Throwable>>() {
+
+        @Override
+        protected Deque<Throwable> initialValue() {
+            return new LinkedList<Throwable>();
```
gabizou: can't you use a lambda here? We're not in a mixin.

JBYoshi:
Okay, hold on... something's still going wrong. Don't merge yet.

src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
JBYoshi: No. `ThreadLocal` is a class, not a functional interface.

JBYoshi:
@gabizou Fixed. Ready for merge.

kenzierocks:
Yes you can use a lambda. ThreadLocal has a static method for this.

JBYoshi:
@kenzierocks I'll take care of that tomorrow.

PR #221, opened 2015-10-21
Extract various constants into a static DataConstants

gabizou:
Still a WIP, but discussion is highly suggested as this just serves as a marker discussion for a topic affecting a very large part of the implementation.

Throughout the implementation of Data API, there are various "constants" thrown around and lately, I've noticed that some of these constants aren't uniform in that they aren't the same constant used for the exam same case. While normally, constants aren't really given much care in the world, they do matter.

This brings the Data implementation up to speed with using a nice uniform `DataConstants` class that is a little sensitive with regards to the `SpongeGameRegistry` having completed registration of a few `CatalogType`s, but still nonetheless sensitive to the phase of the game. As well, there's a few places where `ImmutableDataManipulator`s are actually optimized with regards to storing `ImmutableValue`s in `private final` instance fields instead of always returning new objects, but that's not the main point of this PR, let alone this change.

Pros | Cons
--- | ---
One place to store all constants | Dependent on the `GameRegistry` having finished registration
Never mistakingly using a different constant |  
Brings into uniformity with `NbtDataUtil`'s `NbtTagCompound` keys |

While I work on this, I'm still mulling over the issue of static initialization of various constants, specifically with `CatalogType` based constants, but that issue could easily be resolved if need be.

For the time being, this is still a PR that is a WIP.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/util/DataConstants.java
+++ b/src/main/java/org/spongepowered/common/data/util/DataConstants.java
@@ -1,0 +1,108 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data.util;
+
+import net.minecraft.util.IChatComponent;
+import org.spongepowered.api.data.type.BigMushroomType;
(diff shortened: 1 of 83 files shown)


src/main/java/org/spongepowered/common/data/util/DataConstants.java
```diff
@@ -0,0 +1,108 @@
...
+ * <p><em>WARNING</em>: USAGE OF THESE CONSTANTS, DUE TO STATIC INITIALIZATION,
+ * IS ABSOLUTELY FORBIDDEN UNTIL THE GAME IS DURING THE POST-INIT PHASE DUE
+ * TO REGISTRATION OF CATALOG TYPES. UNTIL THE REGISTRATION IS HANDLED WHERE
+ * THE PROVIDED CATALOG TYPES ARE PROPERLY REGISTERED AND NOT <code>null</code>,
+ * ANY USE OF THIS CLASS WILL RESULT IN A GLORIOUS FAIL INDESCRIBABLE MAGNITUDES.
+ * </p>
+ */
+public final class DataConstants {
+
+    public static final Text EMPTY_TEXT = Texts.of();
```
simon816: `Texts.of()` already [returns a constant](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/Texts.java#L69) so I don't see an advantage of this field.

simon816:
For default values, I think there could be a `DefaultValueRegistry` or something where it stores a map of Key to Value defaults. That way you won't even need to pass the default to `SpongeValue` because it can lookup the default for that key. This also gives the advantage of having a single place to find a default value for any key

gabizou:
> For default values, I think there could be a DefaultValueRegistry or something where it stores a map of Key to value defaults.

The last thing we honestly should be doing is generating yet another `Map` and require registration galore everywhere else. The point of the `DataConstants` is just to have a single place where you visually see the constants and more than likely are inlined by the compiler (in the case of the primitive constants). 

> That way you won't even need to pass the default to SpongeValue because it can lookup the default for that key.

No.. No processing should ever be handled by `Value`s, everything is passed in to them and that's it. Nothing else. This would also fall completely apart when you have `Value`s based on defaults like `Keys.MAX_HEALTH` where the default is the current max health of an entity, and not just the constant 20.

> This also gives the advantage of having a single place to find a default value for any key

Yes, it's a single place, but it's a stupid place to provide lookups. We're not making a registry, we're just making a constants class, just like `NbtDataUtil`. You'd be over engineering the system to force registration when it really really doesn't need to be a registration system.

lucdon:
@gabizou just an idea:
is it possible to do something like this:

````java
@Inject
public static final Optional<ItemType> myItemType = new CatalogTypeRefrence<ItemType>("Iron_Sword").optional();
````
so sponge injects the `ItemTypes.Iron_Sword` into this field after the `catalogtype`s are set. so it is possible to create a final static `CatalogType` without waiting for sponge to load.

we can create a service like the event service were people register a class to the service so the service checks if the class needs `CatalogType injects`. with this people can create a static class like your `DataConstants` and let the service inject the `catalogType`s after they are set.

maybe this works as a work around of the `CatalogType`s being null in the init phase.

gabizou:
@lucdon Do elaborate as I'm planning on doing some major changes to the `SpongeGameRegistry` to avoid the complete mess that it is at the moment.

PR #228, opened 2015-10-26
Implement DurabilityData

Rain336:
Implementation of `SpongeDurabilityData`.

Implementation of `ImmutableSpongeDurabilityData`

Implementation of the `UnbreakableValueProcessor`

Implementation of the `ItemDurabilityValueProcessor`

Registration:
- [x] Registered the `Key` correctly by `FIELD_NAME` in the `KeyRegistry`

Test Plugin code:
```java
package net.rain.test;

import org.spongepowered.api.data.manipulator.mutable.item.DurabilityData;
import org.spongepowered.api.event.game.state.GamePostInitializationEvent;
import org.spongepowered.api.item.inventory.ItemStack;
import org.spongepowered.api.text.Texts;
import org.spongepowered.api.text.format.TextColors;
import org.spongepowered.api.util.command.CommandResult;
import org.spongepowered.api.util.command.args.GenericArguments;
import org.spongepowered.api.util.command.spec.CommandSpec;

@Plugin(id = "CodeTester", name = "CodeTester", version = "1.0.0")
public class CodeTester {

    @Listener
    public void postInit(GamePostInitializationEvent event) {
        DurabilityData data = event.getGame().getRegistry().getManipulatorRegistry().getBuilder(DurabilityData.class).get().create()
                .set(Keys.ITEM_DURABILITY, 10).set(Keys.UNBREAKABLE, true);

        CommandSpec testData = CommandSpec.builder()
                .description(Texts.of(TextColors.GREEN, "Tests a DataManipulator"))
                .arguments(GenericArguments.integer(Texts.of("Type")))
                .executor(((src, args) -> {
                    int i = args.<Integer>getOne("Type").get();
                    if (src instanceof Player) {
                        Player player = (Player) src;
                        if (player.getItemInHand().isPresent()) {
                            ItemStack item = player.getItemInHand().get();
                            if (i == 1) {
                                if (item.supports(Keys.ITEM_DURABILITY)) {
                                    int durability = item.get(Keys.ITEM_DURABILITY).get();
                                    player.sendMessage(Texts.of("Durability: ", durability));
                                    item.offer(Keys.ITEM_DURABILITY, durability - 5);
                                }
                            }
                            if (i == 2) {
                                if (item.supports(Keys.UNBREAKABLE)) {
                                    boolean unbreakable = item.get(Keys.UNBREAKABLE).get();
                                    player.sendMessage(Texts.of("Unbreakable: ", unbreakable));
                                    item.offer(Keys.UNBREAKABLE, unbreakable ? false : true);
                                }
                            }
                            if (i == 3) {
                                if (item.supports(data.durability()) && item.supports(data.unbreakable())) {
                                    item.offer(data);
                                }
                            }
                            player.setItemInHand(item);
                        }
                    }
                    return CommandResult.empty();
                })).build();
        event.getGame().getCommandDispatcher().register(this, testData, "datatest");
    }
}
 ```

### DIFF
--- a/src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeDurabilityData.java
+++ b/src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeDurabilityData.java
@@ -1,0 +1,98 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data.manipulator.immutable.item;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static org.spongepowered.common.data.util.ComparatorUtil.intComparator;
+
+import com.google.common.collect.ComparisonChain;
+import org.spongepowered.api.data.DataContainer;
+import org.spongepowered.api.data.MemoryDataContainer;
+import org.spongepowered.api.data.key.Keys;
+import org.spongepowered.api.data.manipulator.immutable.item.ImmutableDurabilityData;
+import org.spongepowered.api.data.manipulator.mutable.item.DurabilityData;
+import org.spongepowered.api.data.value.immutable.ImmutableBoundedValue;
+import org.spongepowered.api.data.value.immutable.ImmutableValue;
+import org.spongepowered.common.data.manipulator.immutable.common.AbstractImmutableData;
+import org.spongepowered.common.data.manipulator.mutable.item.SpongeDurabilityData;
+import org.spongepowered.common.data.value.immutable.ImmutableSpongeBoundedValue;
+import org.spongepowered.common.data.value.immutable.ImmutableSpongeValue;
+
+public class ImmutableSpongeDurabilityData extends AbstractImmutableData<ImmutableDurabilityData, DurabilityData> implements ImmutableDurabilityData {
+
+    final int durability;
+    final boolean unbreakable;
+
+    public ImmutableSpongeDurabilityData(int durability, boolean unbreakable) {
+        super(ImmutableDurabilityData.class);
+        checkArgument(durability >= 0);
+        this.durability = durability;
+        this.unbreakable = unbreakable;
+        this.registerGetters();
+    }
+
+    public ImmutableSpongeDurabilityData() {
+        this(60, false);
+    }
+
+    @Override
+    protected void registerGetters() {
+        registerFieldGetter(Keys.ITEM_DURABILITY, () -> this.durability);
+        registerKeyValue(Keys.ITEM_DURABILITY, this::durability);
+
+        registerFieldGetter(Keys.UNBREAKABLE, () -> this.unbreakable);
+        registerKeyValue(Keys.UNBREAKABLE, this::unbreakable);
+    }
+
+    @Override
+    public ImmutableBoundedValue<Integer> durability() {
+        return new ImmutableSpongeBoundedValue<Integer>(Keys.ITEM_DURABILITY, this.durability, 60, intComparator(), 0, Integer.MAX_VALUE);
+    }
+
+    @Override
+    public ImmutableValue<Boolean> unbreakable() {
+        return new ImmutableSpongeValue<Boolean>(Keys.UNBREAKABLE, false, this.unbreakable);
+    }
+
+    @Override
+    public DurabilityData asMutable() {
+        return new SpongeDurabilityData(this.durability, this.unbreakable);
+    }
+
+    @Override
+    public int compareTo(ImmutableDurabilityData o) {
+        return ComparisonChain.start()
+                .compare(this.durability, o.durability().get().intValue())
+                .compare(this.unbreakable, o.unbreakable().get())
+                .result();
+    }
+
+    @Override
+    public DataContainer toContainer() {
+        return new MemoryDataContainer()
+                .set(Keys.ITEM_DURABILITY, this.durability)
+                .set(Keys.UNBREAKABLE, this.unbreakable);
+    }
+}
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeDurabilityData.java [hunk 1]
```diff
@@ -0,0 +1,98 @@
...
+        this.registerGetters();
+    }
+
+    public ImmutableSpongeDurabilityData() {
+        this(60, false);
+    }
+
+    @Override
+    protected void registerGetters() {
+        registerFieldGetter(Keys.ITEM_DURABILITY, () -> this.durability);
```
gabizou: Should be just a method reference like the others.

src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeDurabilityData.java
```diff
@@ -0,0 +1,98 @@
...
+        registerKeyValue(Keys.ITEM_DURABILITY, this::durability);
+
+        registerFieldGetter(Keys.UNBREAKABLE, this.unbreakable::get);
+        registerKeyValue(Keys.UNBREAKABLE, this::unbreakable);
+    }
+
+    @Override
+    public ImmutableBoundedValue<Integer> durability() {
+        return new ImmutableSpongeBoundedValue<Integer>(Keys.ITEM_DURABILITY, this.durability, 60, intComparator(), 0, Integer.MAX_VALUE);
```
ZephireNZ: Use a final class variable so you don't create this object each time.

Also, use the `SpongeValueBuilder` as well.

gabizou: store to final field locally, no need to keep returning new instances. Also, I recommend using the `SpongeValueBuilder.boundedBuilder()` to make a mutable and then call `asImmutable()`.

src/main/java/org/spongepowered/common/data/manipulator/mutable/item/SpongeDurabilityData.java
```diff
@@ -0,0 +1,116 @@
...
+        registerFieldSetter(Keys.ITEM_DURABILITY, this::setDurability);
+        registerKeyValue(Keys.ITEM_DURABILITY, this::durability);
+
+        registerFieldGetter(Keys.UNBREAKABLE, () -> this.unbreakable);
+        registerFieldSetter(Keys.UNBREAKABLE, value -> this.unbreakable = value);
+        registerKeyValue(Keys.UNBREAKABLE, this::unbreakable);
+    }
+
+    public void setDurability(int durability) {
+        if (durability >= 0) {
```
gabizou: replace with `checkArgument(durability >= 0, "Durability cannot be less than zero!");`

src/main/java/org/spongepowered/common/data/manipulator/mutable/item/SpongeDurabilityData.java
```diff
@@ -0,0 +1,116 @@
...
+            throw new IllegalArgumentException("Invalid value for durability");
+        }
+    }
+
+    @Override
+    public MutableBoundedValue<Integer> durability() {
+        return SpongeValueBuilder.boundedBuilder(Keys.ITEM_DURABILITY)
+                .minimum(0)
+                .maximum(Integer.MAX_VALUE)
+                .defaultValue(60)
```
gabizou: This probably won't be the default value for many items, I believe these should be dependent on the `ItemStack` being provided to the `DataProcessor`/`ValueProcessor`.

src/main/java/org/spongepowered/common/data/processor/data/item/DurabilityDataProcessor.java
```diff
@@ -0,0 +1,99 @@
...
+    public boolean set(ItemStack itemStack, Map<Key<?>, Object> keyValues) {
+        itemStack.setItemDamage(itemStack.getMaxDamage() - (int) keyValues.get(Keys.ITEM_DURABILITY));
+        final boolean unbreakable = (boolean) keyValues.get(Keys.UNBREAKABLE);
+        if (unbreakable) {
+            itemStack.setItemDamage(0);
+        }
+        if (!itemStack.hasTagCompound()) {
+            itemStack.setTagCompound(new NBTTagCompound());
+        }
+        itemStack.getTagCompound().setBoolean("Unbreakable", unbreakable);
```
gabizou: Extract the string to `NbtDataUtil` please.

src/main/java/org/spongepowered/common/data/processor/value/item/ItemDurabilityValueProcessor.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Override
+    public boolean set(ItemStack container, Integer value) {
+        container.setItemDamage(container.getMaxDamage() - value);
+        return true;
+    }
+
+    @Override
+    public Optional<Integer> getVal(ItemStack container) {
+        return Optional.of(container.getMaxDamage() - container.getItemDamage());
```
gabizou: You're not checking if the `ItemStack.isDamageable()`.

ZephireNZ:
Should plugins be able to get/set the `DurablityData` of all items? I know that all items do technically have a durability, but the javadocs only talk about items which can be damaged (and that's really what this data is for, not setting the raw data value). Minecraft is moving away from most item damage anyway, so we should potentially keep that in mind to account for breaking changes in the future.

gabizou:
> Should plugins be able to get/set the DurablityData of all items? I know that all items do technically have a durability, but the javadocs only talk about items which can be damaged (and that's really what this data is for, not setting the raw data value). Minecraft is moving away from most item damage anyway, so we should potentially keep that in mind to account for breaking changes in the future.

No, only `Item`s that can be actually damaged, not items that have subtypes, which is why `isDamageable()` is such an important check.

src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeDurabilityData.java
JBYoshi: @gabizou What's the difference? AFAIK the compiler just converts lambdas to a method and replaces their usage with a call to the method.

gabizou: yes, but it's just cleaner.

src/main/java/org/spongepowered/common/data/manipulator/mutable/item/SpongeDurabilityData.java
```diff
@@ -0,0 +1,112 @@
...
+import org.spongepowered.api.data.value.mutable.MutableBoundedValue;
+import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.common.data.manipulator.immutable.item.ImmutableSpongeDurabilityData;
+import org.spongepowered.common.data.manipulator.mutable.common.AbstractData;
+import org.spongepowered.common.data.value.SpongeValueBuilder;
+
+public class SpongeDurabilityData extends AbstractData<DurabilityData, ImmutableDurabilityData> implements DurabilityData {
+
+    int durability;
+    boolean unbreakable;
```
Phroa: Should be private

src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeDurabilityData.java
```diff
@@ -0,0 +1,101 @@
...
+public class ImmutableSpongeDurabilityData extends AbstractImmutableData<ImmutableDurabilityData, DurabilityData> implements ImmutableDurabilityData {
+
+    final ImmutableBoundedValue<Integer> durability;
+    final ImmutableValue<Boolean> unbreakable;
+
+    public ImmutableSpongeDurabilityData(int durability, boolean unbreakable) {
+        super(ImmutableDurabilityData.class);
+        checkArgument(durability >= 0);
+        this.durability = SpongeValueBuilder.boundedBuilder(Keys.ITEM_DURABILITY)
+                .minimum(0)
```
gabizou: missing a default value.

src/main/java/org/spongepowered/common/data/manipulator/mutable/item/SpongeDurabilityData.java
```diff
@@ -0,0 +1,112 @@
...
+
+    public SpongeDurabilityData() {
+        this(60, false);
+    }
+
+    public SpongeDurabilityData(int durability, boolean unbreakable) {
```
gabizou: You should likely throw the default value in the constructor as well considering each itemstack has a custom durability setting.

src/main/java/org/spongepowered/common/data/processor/data/item/DurabilityDataProcessor.java
```diff
@@ -0,0 +1,100 @@
...
+
+public class DurabilityDataProcessor extends AbstractItemDataProcessor<DurabilityData, ImmutableDurabilityData> {
+
+    public DurabilityDataProcessor() {
+        super(input -> input.getItem().isDamageable());
+    }
+
+    @Override
+    public boolean doesDataExist(ItemStack itemStack) {
+        return true;
```
phroa: Do *all* itemstacks have some form of durability?

JBYoshi: Should be `return itemStack.isDamageable();` (haven't tested, but there should be a similar method in `ItemStack`). Plugins shouldn't need to use damage values anymore unless the item is acually damageable.

PR #233, opened 2015-10-29
Implementation of FallDistanceData

ZephireNZ:
SpongePowered/SpongeAPI#891

Implementation of `FallDistanceData`.

Implementation of `ImmutableFallDistanceData`

Implementation of the `FallDistanceValueProcessor`(s):

Registration:
- [x] Registered the `Key` correctly by `FALL_DISTANCE` in the `KeyRegistry`

Test Plugin code:

```java
@Plugin(id = "data-test", name = "DataTest")
public class DataTest {

    @Listener
    public void onDamage(DisplaceEntityEvent.Move.TargetLiving event) {
        if(entity.supports(FallDistanceData.class) && entity instanceof Player) {
            FallDistanceData data = entity.getOrCreate(FallDistanceData.class).get();
            data.set(Keys.FALL_DISTANCE, 0F);
            entity.offer(data);
        }
        if(entity instanceof Player) {
            float dist = entity.get(Keys.FALL_DISTANCE).get();
            System.out.println(((Player) entity).getProfile().getName() + " fell " + dist);
        } else {
            Optional<MutableBoundedValue<Float>> opt = entity.getValue(Keys.FALL_DISTANCE);
            if(opt.isPresent()) {
                System.out.println(entity.getUniqueId().toString() + " fell " + opt.get().get());
            }
        }
    }
}
```

Notes:

### DIFF
--- a/src/main/java/org/spongepowered/common/data/processor/data/entity/FallDistanceDataProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/data/entity/FallDistanceDataProcessor.java
@@ -1,0 +1,81 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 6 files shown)


src/main/java/org/spongepowered/common/data/processor/data/entity/FallDistanceDataProcessor.java
```diff
@@ -0,0 +1,80 @@
...
+public class FallDistanceDataProcessor extends AbstractEntitySingleDataProcessor<EntityLivingBase, Float, MutableBoundedValue<Float>,
+        FallDistanceData, ImmutableFallDistanceData> {
+
+    public FallDistanceDataProcessor() {
+        super(EntityLivingBase.class, Keys.FALL_DISTANCE);
+    }
+
+    @Override
+    protected boolean set(EntityLivingBase entity, Float value) {
+        entity.fallDistance = value;
```
gabizou: Needs some validation on the value itself.

src/main/java/org/spongepowered/common/data/processor/value/entity/FallDistanceValueProcessor.java
```diff
@@ -0,0 +1,75 @@
...
+                .actualValue(value)
+                .defaultValue(0F)
+                .minimum(0F)
+                .maximum(Float.MAX_VALUE)
+                .build();
+    }
+
+    @Override
+    protected boolean set(EntityLivingBase container, Float value) {
+        container.fallDistance = value;
```
gabizou: And here. needs some validation.

ZephireNZ:
The test code is annoying - tried it with damage event, nope doesn't seem implemented. Same goes with move even, but *only* the player one is implemented -_-

ZephireNZ:
Oopsy xD

ZephireNZ:
Looks like there's some duplication between this and #234 - refactoring is gonna be soo fun...

gabizou:
@ZephireNZ Yeah.... considering that PR attempts to solve a lot of problems at the same time, it might take some time at the least. It might be divided into phases as @Mumfrey and I have pondered on some ways to resolve the issue of "stateless" `CatalogType` references while at the same time revamping the overall `SpongeGameRegistry` as a whole.

ZephireNZ:
Wait derp, wrong PR to put that with xD

gabizou:
@ZephireNZ to save time, I'm gonna re-post the above comment in #217

PR #234, opened 2015-10-30
Revamp GameRegistry to use RegistryModules

gabizou:
To work in tandem of DataConstants, this PR will effectively provide "modules" for registering of various `CatalogType`s such that the "defaults" exposed in the API will be already set for use with (hopefully) static initialized fields for not only plugins, but also the implementation.


src/main/java/org/spongepowered/common/data/manipulator/immutable/common/AbstractImmutableBoundedComparableData.java
```diff
@@ -57,22 +56,42 @@
...
         super(immutableClass, value, usedKey);
         this.comparator = comparator;
         checkArgument(!Modifier.isAbstract(mutableClass.getModifiers()), "The immutable class cannot be abstract!");
         checkArgument(!Modifier.isInterface(mutableClass.getModifiers()), "The immutable class cannot be an interface!");
+        checkArgument(comparator.compare(defaultValue, lowerBound) >= 0 && comparator.compare(defaultValue, upperBound) <= 0 );
         this.mutableClass = mutableClass;
         this.lowerBound = lowerBound;
         this.upperBound = upperBound;
+        this.defaultValue = defaultValue;
+        if (this.comparator.compare(this.upperBound, this.lowerBound) <= ImmutableDataCachingUtil.CACHE_LIMIT_FOR_INDIVIDUAL_TYPE) {
```
ZephireNZ: Will this work for float/double values?

20F is within this bound, but has many more than 100 potential values.

src/main/java/org/spongepowered/common/registry/provider/DirectionFacingProvider.java
```diff
@@ -0,0 +1,41 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.collect.ImmutableBiMap;
+import net.minecraft.util.EnumFacing;
+import org.spongepowered.api.util.Direction;
+import org.spongepowered.common.registry.TypeProvider;
+
+import java.util.Optional;
+
+public final class DirectionFacingProvider implements TypeProvider<Direction, EnumFacing> {
```
gratimax: :+1: beautiful

src/main/java/org/spongepowered/common/registry/type/ArtRegistryModule.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Override
+    public void registerDefaults() {
+        for (EntityPainting.EnumArt art : EntityPainting.EnumArt.values()) {
+            this.artMappings.put(((Art) (Object) art).getId().toLowerCase(), (Art) (Object) art);
+        }
+    }
+
+    @AdditionalRegistration
+    public void registerAdditionals() {
```
gratimax: This method doesn't seem to accomplish much. Is it being called later on in initialization?

src/main/java/org/spongepowered/common/registry/type/ArgumentRegistryModule.java
```diff
@@ -0,0 +1,87 @@
...
+import org.spongepowered.common.text.selector.SpongeSelectorFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@RegistrationDependency({SelectorFactoryModule.class, SelectorTypeRegistryModule.class})
+public class ArgumentRegistryModule implements RegistryModule {
+
+    @RegisterCatalog(ArgumentTypes.class)
+    private final Map<String, ArgumentHolder<?>> argumentTypeMap = new HashMap<>();
```
gratimax: The annotation is neat, but perhaps all that is needed is to parameterize RegistryModule and add a method `Map<String, T> getMap()`, it would be simpler

src/main/java/org/spongepowered/common/registry/type/ArtRegistryModule.java
gabizou: @AdditionalRegistration is meant for post-init to catch any additionally registered values (since Forge has the nicely exposed `EnumHelper`) that otherwise, the default values are likely not to contain.

src/main/java/org/spongepowered/common/registry/type/ArgumentRegistryModule.java
gabizou: Probably, but would likely be better used for `CatalogRegistryModule`, though for some highly specialized modules, it won't work out well.

gratimax: What sort of highly specialized modules? I see that CatalogRegistryModule already has a type parameter.

src/main/java/org/spongepowered/common/registry/type/ArtRegistryModule.java
gratimax: Okay, makes sense. Maybe it would be a good idea to pass in a lifecycle enum into the annotation, i.e. `@AdditionalRegistration(Lifecycles.AFTER_OTHER_MODS)` just to clarify when it is being called.

src/main/java/org/spongepowered/common/registry/type/ArgumentRegistryModule.java
gabizou: https://github.com/SpongePowered/SpongeCommon/blob/feature/registry-temp/src/main/java/org/spongepowered/common/registry/type/EntityTypeRegistryModule.java#L159 

The `EntityTypeRegistryModule` by far is the biggest hack due to the nature of how it all works.

PR #235, opened 2015-10-30
Implementing CooldownData

Rain336:
Implementation of `SpongeCooldownData`.

Implementation of `ImmutableSpongeCooldownData`

Implementation of the `CooldownValueProcessor`

Registration:
- [x] Registered the `Key` correctly by `COOLDOWN` in the `KeyRegistry`

Test Plugin code:
```java
@Plugin(id = "CodeTester", name = "CodeTester", version = "1.0.0")
public class CodeTester {

    @Listener
    public void onPlayerInteractBlock(InteractBlockEvent event) {
        Optional<Player> optional = event.getCause().first(Player.class);
        if (optional.isPresent()) {
            Player player = optional.get();
            if (player.getItemInHand().isPresent()) {
                ItemStack item = player.getItemInHand().get();
                Hopper hopper = (Hopper) event.getTargetBlock().getLocation().get().getTileEntity().get();
                if (item.getItem() == ItemTypes.STICK) {
                    player.sendMessage(Texts.of("HopperTransferCooldown: ", hopper.get(Keys.COOLDOWN).get()));
                }
                if (item.getItem() == ItemTypes.BLAZE_ROD) {
                    hopper.offer(Keys.COOLDOWN, 100);
                }
            }
        }
    }
}
```


src/main/java/org/spongepowered/common/data/manipulator/immutable/ImmutableSpongeCooldownData.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public CooldownData asMutable() {
+        return new SpongeCooldownData(this.getValue());
+    }
+
+    @Override
+    public ImmutableValue<Integer> cooldown() {
+        return ImmutableSpongeValue.cachedOf(Keys.COOLDOWN, -1, this.getValue());
```
ZephireNZ: Please make this a final field

gabizou:
So, a couple of things about your test plugins:
- Your test plugin is very buggy and easily throws exceptions in many places unless you know exactly what is going on.
- You don't check any optionals, resulting in assumptions and exceptions
- You aren't testing everything as asked
  - Manipulators aren't being retrieved, you're just interacting with values directly
  - Serialization and deserialization isn't being tested.

Aside from that, I've tested it enough to know that it works, but that's only because I know the underlying code well enough, but seriously, you should consider your test code design as all corner cases should be considered.

PR #240, opened 2015-11-01
Make dirty signs update

clienthax:


Kiskae:
There is probably also a missing mark dirty @ https://github.com/clienthax/SpongeCommon/blob/patch-2/src/main/java/org/spongepowered/common/data/processor/value/tileentity/SignLinesValueProcessor.java#L97

PR #241, opened 2015-11-02
Implement NoteData

ZephireNZ:
Implementation of `NoteData`.
- [ ] Accurately used the appropriate `AbstractData` implementation

Implementation of `ImmutableNoteData`
- [ ] Accurately extended the appropriate `AbstractImmutableData` implementation

Implementation of the `NoteDataProcessor`(s):

Registration:
- [x] Registered the `Key` correctly by `NOTE_PITCH` in the `KeyRegistry`

Test Plugin code:

Wrap your plugin code with ```java
// PUT CODE IN HERE

And at the end of your plugin ```

Notes:


ZephireNZ:
Not sure how good the code is for registering the NotePitches is - Mojang (as far as I can see) has no enum for the notes, they're only ever referenced by byte.

ZephireNZ:
Additionally, the data implementations could be simpler if NoteData was a catalog type rather than a standard type

PR #249, opened 2015-11-06
Revamp GameRegistry to use RegistryModules

gabizou:
To work in tandem of DataConstants, this PR will effectively provide "modules" for registering of various `CatalogType`s such that the "defaults" exposed in the API will be already set for use with (hopefully) static initialized fields for not only plugins, but also the implementation.a


src/main/java/org/spongepowered/common/registry/AdditionalRegistration.java
```diff
@@ -0,0 +1,38 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface AdditionalRegistration {
+
+    RegistrationPhase value() default RegistrationPhase.POST_INIT;
```
gabizou: @gratimax this is that registration phase I think you were talking about before.

src/main/java/org/spongepowered/common/data/util/DataConstants.java
```diff
@@ -0,0 +1,108 @@
...
+ * <p><em>WARNING</em>: USAGE OF THESE CONSTANTS, DUE TO STATIC INITIALIZATION,
+ * IS ABSOLUTELY FORBIDDEN UNTIL THE GAME IS DURING THE POST-INIT PHASE DUE
+ * TO REGISTRATION OF CATALOG TYPES. UNTIL THE REGISTRATION IS HANDLED WHERE
+ * THE PROVIDED CATALOG TYPES ARE PROPERLY REGISTERED AND NOT <code>null</code>,
+ * ANY USE OF THIS CLASS WILL RESULT IN A GLORIOUS FAIL INDESCRIBABLE MAGNITUDES.
+ * </p>
+ */
+public final class DataConstants {
+
+    public static final Text EMPTY_TEXT = Texts.of();
```
Minecrell: I'm pretty sure someone mentioned this somewhere already, this is kind of redundant - `Texts.of()` actually already returns a constant.

src/main/java/org/spongepowered/common/data/util/DataConstants.java
```diff
@@ -0,0 +1,108 @@
...
+
+    public static final IChatComponent EMPTY_TEXT_COMPONENT = SpongeTexts.toComponent(Texts.of());
```
Minecrell: These shouldn't be stored in a constant as they're mutable internally.

gabizou: So I guess then the better option is DataConstants.emptyChatComponent() that returns a new instance each time?

Minecrell: Yes, exactly

PR #251, opened 2015-11-06
Implement launching projectiles from ProjectileSources

simon816:
Not every projectile is launched perfectly but it's good enough.
It's difficult to account for every situation, e.g if firing a large fireball from a dispenser sometimes the dispenser may blow up.

Closes SpongePowered/SpongeForge#184


simon816:
@SpongePowered/developers Any comments?

Zidane:
@simon816 

This looks good. How about getting this merged?

PR #254, opened 2015-11-08
Cache player locales

kashike:
This changes `getLocale` to use a `LoadingCache` to get `Locale`s, instead of creating a new `Locale` each call.

### DIFF
--- a/src/main/java/org/spongepowered/common/util/LanguageUtil.java
+++ b/src/main/java/org/spongepowered/common/util/LanguageUtil.java
@@ -1,0 +1,46 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


jamierocks:
:+1:

src/main/java/org/spongepowered/common/util/LanguageUtil.java
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.commons.lang3.LocaleUtils;
+import org.spongepowered.api.util.annotation.NonnullByDefault;
+
+import java.util.Locale;
+
+@NonnullByDefault
+public final class LanguageUtil {
+
+    public static final LoadingCache<String, Locale> LOCALE_CACHE = CacheBuilder.newBuilder()
+        .build(new CacheLoader<String, Locale>() {
```
gratimax: Could just be `.build(LocaleUtils::toLocale)`?

JBYoshi: @gratimax I believe `CacheLoader` is an abstract class, and method references don't work with them.

kashike: @JBYoshi is correct.

PR #261, opened 2015-11-14
Fixing some problems with blockData

lucdon:
@gabizou, @bloodmc thoughts?


gabizou:
@lucdon While it's normally inexpensive, I'd rather it's left to the block to determine whether extended properties are needed. Calling extended properties for every block is somewhat costly imo.

lucdon:
@gabizou so you want to leave it as it is now? i created this pr because @bloodmc had some issues with how it is atm.

gabizou:
@lucdon can you re-test after rebasing on master?

Aaron1011:
@lucdon: Can you rebase this against master?

lucdon:
@gabizou  @Aaron1011  bit late but done!

gabizou:
Awesome, thanks @lucdon .

PR #265, opened 2015-11-18
Fix ItemName processor nbt

clienthax:


gabizou:
Is the value processor affected as well?

clienthax:
@gabizou value processor was using ITEM_DISPLAY_NAME but the value of it was wrong, this fixs both

PR #272, opened 2015-11-19
Allow shutting down the server if the EULA is not accepted

Aaron1011:
Currently, the dedicated servers ends up looping forever in `finalTick` if the EULA was not accepted. While this has always been vanilla behavior, it's generally undesireable, since there's no point leaving the server running in a half-initialized state.

This PR adds a new config section and option to control whether the server automatically shuts down when the EULA is not accepted (which defaults to true). The only change that needs to be made in SpongeForge and SpongeVanilla is to enable the mixin config.

Note: Forge logs a message about an unexpected transition to the state `SERVER_STOPPED` when this change is enabled. This is correct - this change is functionally equivalent to manually stopping or killing server when the EULA is not accepted, which actually does prevent some initialization from running.

### DIFF
--- a/src/main/java/org/spongepowered/common/configuration/SpongeConfig.java
+++ b/src/main/java/org/spongepowered/common/configuration/SpongeConfig.java
@@ -97,6 +97,9 @@ public class SpongeConfig<T extends SpongeConfig.ConfigBase> {
     // BUNGEECORD
     public static final String BUNGEECORD_IP_FORWARDING = "ip-forwarding";
 
+    // EULA
+    public static final String EULA_SHUTDOWN_SERVER = "shutdown-server";
+
     // GENERAL
     public static final String GENERAL_DISABLE_WARNINGS = "disable-warnings";
     public static final String GENERAL_CHUNK_LOAD_OVERRIDE = "chunk-load-override";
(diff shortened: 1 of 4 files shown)


simon816:
I don't think this needs to be it's own mixin module and have a config option IMO

Aaron1011:
@simon816: The main reason I did this is because it breaks Vanilla behavior. This has always been the behavior of Vanilla since the EULA requirement was added, and I'm loath to change it without an option to disable it

JBYoshi:
Actually, why not allow the user to accept the EULA without stopping and restarting the server? Something like `Please read the EULA shown at https://account.mojang.com/documents/minecraft_eula. If you accept, type "I agree". Otherwise, type "stop".`

Aaron1011:
@JBYoshi: It would be much nicer, but I'm extremely hesitant to include something like that in Sponge because of legal issues. I want to avoid us interfering in any way with the actual accepting of the EULA, as Mojang has defined it.

JBYoshi:
@Aaron1011 Sounds good.

simon816:
I just ran into another situation when the server would hang - when it fails to bind the port. Perhaps this could be expanded for more general hanging cases?

Aaron1011:
@simon816: I considered that. I was thinking that those should probably be handled on a case-by-case basis.

Minecrell:
We can close this now, can't we? https://github.com/SpongePowered/SpongeCommon/commit/2a3058de999a11605215167d97daf439fa762b14
We'll need to revert this again for 1.8.8 because that seems to be part of Vanilla behaviour now.

Aaron1011:
@Minecrell: Right - I just added this for now, since it was annoying me :P

PR #277, opened 2015-11-21
Implement extended platform API

Minecrell:
This implements the changes made in https://github.com/SpongePowered/SpongeAPI/pull/928 together with some minor cleanup like adding `Platform` to the dependency injection.

SpongeVanilla PR: https://github.com/SpongePowered/SpongeVanilla/pull/187
SpongeForge PR: https://github.com/SpongePowered/SpongeForge/pull/423

### DIFF
--- a/src/main/java/org/spongepowered/common/SpongeVersion.java
+++ b/src/main/java/org/spongepowered/common/SpongeVersion.java
@@ -22,23 +22,30 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.common.plugin;
+package org.spongepowered.common;
 
-import com.google.common.base.Objects;
-import org.spongepowered.api.plugin.PluginContainer;
+import static com.google.common.base.Objects.firstNonNull;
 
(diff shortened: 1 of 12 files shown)


src/main/java/org/spongepowered/common/SpongeVersion.java
```diff
@@ -0,0 +1,51 @@
...
+import static com.google.common.base.Objects.firstNonNull;
+
+import org.spongepowered.api.MinecraftVersion;
+
+import java.util.Optional;
+
+public final class SpongeVersion {
+
+    // TODO: Keep up to date
+    public static final MinecraftVersion MINECRAFT_VERSION = new SpongeMinecraftVersion("1.8", 47);
```
ST-DDT: Csnt we set this using a gradle task?

Minecrell: Not the protocol version.

ST-DDT: Why not? Everything else is stored there as well.
And iirc then both (version and protocol) things will change at the same time so imo it would be a good idea to configure them in a single location. And since we need the other stuff in gradle anyway lets just move everything there.

PR #278, opened 2015-11-22
Perform IP Forwarding tasks earlier in the handshake process.

dualspiral:
In order to be able to use the latest iteration of IP Forwarding for Bungee and Forge clients and thus achieve zero configuration on the Bungee side without breaking backwards compatibility, we need to be able to read the LoginProfile properties much earlier than we do right now, so we can indicate to SpongeForge that the user is a Forge client with zero config.

This does not impact on the existing implementation, it simply moves the checks.

Context for this change: https://github.com/SpigotMC/BungeeCord/pull/1557#issuecomment-158671754
Related SpongeForge PR: https://github.com/SpongePowered/SpongeForge/pull/424

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/bungee/network/MixinNetHandlerHandshakeTCP.java
+++ b/src/main/java/org/spongepowered/common/mixin/bungee/network/MixinNetHandlerHandshakeTCP.java
@@ -27,6 +27,7 @@ package org.spongepowered.common.mixin.bungee.network;
 import com.google.gson.Gson;
 import com.mojang.authlib.properties.Property;
 import com.mojang.util.UUIDTypeAdapter;
+import net.minecraft.network.EnumConnectionState;
 import net.minecraft.network.NetworkManager;
 import net.minecraft.network.handshake.client.C00Handshake;
 import net.minecraft.network.login.server.S00PacketDisconnect;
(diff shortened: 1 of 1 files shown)


Minecrell:
Could you make sure that this is also working on SpongeVanilla? Are there additional changes necessary there?

dualspiral:
I'll check, but it should just work with no additional changes.

Minecrell:
@dualspiral You'll probably need to add the [BungeeCord mixin config](https://github.com/SpongePowered/SpongeVanilla/blob/master/src/main/java/org/spongepowered/server/launch/VanillaServerTweaker.java#L125-L128) for SpongeCommon first.

dualspiral:
Heh, forgot about that. Will test and PR any changes needed then, you'd need it for the current implementation anyway.

dualspiral:
@Minecrell It works fine with SpongeVanilla once you add the BungeeCord plugin, as you suggested - https://github.com/dualspiral/SpongeVanilla/commit/dae786e5ae58bc79f47aecefc8d7eb8583fa6504 is all it needs to work.

My test plugin to verify this can be found at https://gist.github.com/dualspiral/abc96bd5e06c717e4818, noting that Sponge could get my UUID successfully.

I won't bother making a PR for this into Vanilla as it is a one line change, unless you really want me to!

Minecrell:
@dualspiral Awesome, thanks a lot for testing. No need to make a PR, I'll pull your commit in as soon this is merged.

dualspiral:
@bloodmc This will need merging in order for the PR that was merged in https://github.com/SpongePowered/SpongeForge/pull/424 to be effective (sorry, should have made that clear in the other PR!).

PR #287, opened 2015-11-25
Make arrows consistent

clienthax:
instead of selectively removing the arrows on the first / last page,
keep the arrows there, but gray them out so the bottom line is constant across all pages

### DIFF
--- a/src/main/java/org/spongepowered/common/service/pagination/ActivePagination.java
+++ b/src/main/java/org/spongepowered/common/service/pagination/ActivePagination.java
@@ -160,6 +160,8 @@ abstract class ActivePagination {
         TextBuilder ret = Texts.builder();
         if (hasPrevious) {
             ret.append(this.prevPageText).append(DIVIDER_TEXT);
+        } else {
+            ret.append(Texts.of("«")).append(DIVIDER_TEXT);
         }
         boolean needsDiv = false;
         int totalPages = getTotalPages();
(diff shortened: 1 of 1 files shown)


hsyyid:
This bugs me, thanks for fixing it

src/main/java/org/spongepowered/common/service/pagination/ActivePagination.java
```diff
@@ -160,6 +160,8 @@ protected Text calculateFooter(int currentPage) {
         TextBuilder ret = Texts.builder();
         if (hasPrevious) {
             ret.append(this.prevPageText).append(DIVIDER_TEXT);
+        } else {
+            ret.append(Texts.of("«")).append(DIVIDER_TEXT);
```
JBYoshi: I think these are supposed to be stored in a field set in the constructor (see line 71).

src/main/java/org/spongepowered/common/service/pagination/ActivePagination.java
```diff
@@ -172,6 +174,11 @@ protected Text calculateFooter(int currentPage) {
                 ret.append(DIVIDER_TEXT);
             }
             ret.append(this.nextPageText);
+        } else {
+            if (needsDiv) {
+                ret.append(DIVIDER_TEXT);
+            }
+            ret.append(Texts.of("»"));
```
JBYoshi: Same here.

clienthax: They are private finals declared in the class themselves..

PR #289, opened 2015-11-26
Exclude common from the generated POM

kenzierocks:
Yes it's not in the deobf jar, people can use `deobfCompile` on the regular JAR instead. This makes setting up the SDK a little easier.


liach:
:+1: I hope this commit could fix the issue I commented [here](https://github.com/SpongePowered/SpongeForge/issues/377#issuecomment-149050865).

kenzierocks:
The deobf jar may or may not work.

progwml6:
we remove the deobf jar they aren't needed with FG 2.0 & 2.1 since it can deobf production jars on its own

Minecrell:
@progwml6 FG can't deobf our mixins properly, the JAR it generates wouldn't be properly runnable currently. We will either need special support in FG for this (which would be kinda hacky) or change something in Mixin to make that work properly.
The primary problem is the refmap generated by Mixin, which maps the annotation values (that aren't reobfuscated in the class itself) to the SRG names. If that is present it won't apply the mixins properly. And even if it would be removed, you would still need to use the exact same mappings version as Sponge because otherwise they would be (possibly) mapped incorrectly. And that would kind of defeat the purpose of letting FG handle the deobfuscation.

kenzierocks:
This should now be in working order. Still checking to see how hard it is to apply the ATs from Sponge during dev setup.

kenzierocks:
Pinging @Minecrell: We need to use FMLAT to make SpongeForge easy to depend on. FG looks for the FMLAT stuff to pull down ATs, and we need that in order to make it runnable at dev-time. Do you agree with changing the AT location in SpongeForge so it is compatible with FMLAT?

Minecrell:
Why do you need to change the AT location for that?

kenzierocks:
FMLAT only works for ATs in the META-INF.

bloodmc:
@Minecrell can you check this PR please. Thanks

Minecrell:
I'm closing this for now, I have merged it into a branch and will merge this to master together with a few other changes soon.

kashike:
In other words, check back in March.

Minecrell:
@kashike Exactly. :+1:

PR #290, opened 2015-11-26
Fix #288 - NbtTranslator has bad logic or formatting

### LINKED ISSUE #288: NbtTranslator has bad logic or formatting.
[This line](https://github.com/SpongePowered/SpongeCommon/blob/11a77b370af6119cb660fcb4bc7a21b84f9d8654/src/main/java/org/spongepowered/common/service/persistence/NbtTranslator.java#L88) should be `else if` or formatted properly if it should be an `if`.

CrystalMageX:


Mumfrey:
Closing because frivolous

CrystalMageX:
@Mumfrey There is a serious logical uncertainty here.

Are you sure you would like to continue running the code in **both** the 1st `if` block and the `else` statement, leading to two `setTag()` calls? I find it hard to believe that adding the value **twice** is intended.

If indeed, the double call is done on purpose, then the code in the `else` block should be merged into the 1st `if` block.

simon816:
It was fixed in https://github.com/SpongePowered/SpongeCommon/commit/9a772dae0091fa21872537367f02a256fb2de561

saladoc:
Also refer to the [Contribution Guidelines](https://docs.spongepowered.org/en/contributing/guidelines.html), namely the part about "small changes under 20 lines".

CrystalMageX:
@saladoc O.K., will remember to do that next time.

Issue #149, opened 2015-08-29
Misleading VelocityData values


RobertHerhold:
While testing out the newly implemented `VelocityData`, I found something odd: while at rest, the player's (entity's) velocity is: `(0.0, -0.0784000015258789, 0.0)`. Obviously, the player should not have a downward velocity while at rest. When I jumped, I got the following output:

```
(0.0, -0.0784000015258789, 0.0)
(0.0, 0.41999998688697815, 0.0)
(0.0, 0.24813599859094576, 0.0)
(0.0, 0.16477328182606651, 0.0)
(0.0, 0.16477328182606651, 0.0)
(0.0, 0.08307781780646721, 0.0)
(0.0, 0.0030162615090425808, 0.0)
(0.0, -0.0784000015258789, 0.0)
(0.0, -0.230527368912964, 0.0)
(0.0, -0.230527368912964, 0.0)
(0.0, -0.30431682745754424, 0.0)
(0.0, -0.0784000015258789, 0.0)
```
And while these values make sense relative to each other, the issue still persisted. Additionally, while just moving around in general, the x and z components of the velocity vector never deviated from 0. 

I used [this plugin](https://gist.github.com/RobertHerhold/a3fb64c69dbc5561565c) to test this. Is there something I'm missing, or...?

JBYoshi:
I think the y value would be due to gravity?

RobertHerhold:
But the entity is not moving down, so the velocity should not be anything but 0. I see what you're saying, but it doesn't make sense logically.

ST-DDT:
@RobertHerhold Have you tested this with non-player entities as well?

DDoS:
I think someone mistook gravity for velocity...

RobertHerhold:
@ST-DDT I just tested it with a minecart, and it's velocity is `(0.0, -0.0, 0.0)` when it is at rest

JBYoshi:
In that case, blame the movement packet. Sponge could do some custom calculations server-side though.

nickrobson:
I'm just going to go ahead and point out that you don't experience that behaviour in Vanilla or Bukkit, so it is **definitely** Sponge.

gabizou:
Is this still broken?

RobertHerhold:
@gabizou Yes. I updated the gist to the latest version of the API.

simon816:
From testing the value is always `-0.0784000015258789` at rest.

I have done some investigation and found out how this value came to be.

[...]

PR #293, opened 2015-11-27
Improve velocity behavior.

JBYoshi:
No longer requires magic values. Works properly with players, but haven't tested with non-player `EntityLivingBase`s.

See the discussion in #149.

Code:
```java
package jbyoshi.sponge.test;

import org.slf4j.Logger;
import org.spongepowered.api.event.entity.DisplaceEntityEvent;
import org.spongepowered.api.service.scheduler.Task;

import javax.inject.Inject;

@Plugin(id = "jbyoshi.sponge.test", name="JBYoshi Sponge Test")
public final class SpongeTestPlugin {
    
    @Inject private Logger logger;
    
    @Listener
    public void playerMoved(DisplaceEntityEvent.TargetPlayer event) {
        this.logger.info(event.getToTransform().getPosition().sub(event.getFromTransform().getPosition()).toString());
        event.getGame().getScheduler().createTaskBuilder().execute(() -> {
            this.logger.info("-> " + event.getTargetEntity().get(Keys.VELOCITY).get());
        }).submit(this);
    }
}
```


JBYoshi:
I'll need to check if this issue also happens with non-player `EntityLivingBase`s.

PR #297, opened 2015-11-30
Implement the affects spawning api

kashike:
API: https://github.com/SpongePowered/SpongeAPI/pull/941

Implementation of `AffectsSpawningData`.

Implementation of `ImmutableAffectsSpawningData`

Implementation of the `AffectsSpawningValueProcessor`(s):

Registration:
- [x] Registered the `Key` correctly by `affects_spawning` in the `KeyRegistry`

Test Plugin: https://gist.github.com/kashike/996c698f7933d54351b1

### DIFF
--- a/src/main/java/org/spongepowered/common/interfaces/IMixinWorld.java
+++ b/src/main/java/org/spongepowered/common/interfaces/IMixinWorld.java
@@ -27,6 +27,7 @@ package org.spongepowered.common.interfaces;
 import com.google.common.collect.ImmutableList;
 import net.minecraft.block.Block;
 import net.minecraft.block.state.IBlockState;
+import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.util.BlockPos;
 import org.spongepowered.api.block.BlockSnapshot;
 import org.spongepowered.api.block.tileentity.TileEntity;
@@ -39,6 +40,8 @@ import org.spongepowered.common.config.SpongeConfig;
 
 import java.util.List;
 import java.util.Optional;
+
+import javax.annotation.Nullable;
 
 public interface IMixinWorld {
 
@@ -99,5 +102,13 @@ public interface IMixinWorld {
     void setWeatherStartTime(long weatherStartTime);
 
     void notifyBlockOfStateChange(BlockPos notifyPos, Block sourceBlock, BlockPos sourcePos);
+
+    @Nullable
+    EntityPlayer getClosestPlayerToEntityWhoAffectsSpawning(net.minecraft.entity.Entity entity, double distance);
+
+    @Nullable
+    EntityPlayer getClosestPlayerWhoAffectsSpawning(double x, double y, double z, double distance);
+
+    boolean isAnyPlayerWithinRangeAtWhoAffectsSpawning(double x, double y, double z, double range);
 
 }
--- a/src/main/java/org/spongepowered/common/mixin/core/block/tiles/MixinMobSpawnerBaseLogic.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/block/tiles/MixinMobSpawnerBaseLogic.java
@@ -22,16 +22,21 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.common.interfaces;
+package org.spongepowered.common.mixin.core.block.tiles;
 
-import net.minecraft.util.BlockPos;
+import net.minecraft.tileentity.MobSpawnerBaseLogic;
+import net.minecraft.world.World;
+import org.spongepowered.asm.mixin.Mixin;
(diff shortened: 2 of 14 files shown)


src/main/java/org/spongepowered/common/mixin/core/block/tiles/MixinMobSpawnerBaseLogic.java
```diff
@@ -0,0 +1,45 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+import org.spongepowered.common.interfaces.IMixinWorld;
+
+@Mixin(MobSpawnerBaseLogic.class)
+public class MixinMobSpawnerBaseLogic {
+
+    @Shadow private int activatingRangeFromPlayer;
```
gabizou: Unnecessary field?

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
```diff
@@ -125,4 +126,14 @@ public boolean onIsPlayerSleeping(EntityPlayer self) {
         }
         return false;
     }
+
+    @Override
+    public boolean affectsSpawning() {
+        return this.affectsSpawning;
```
gabizou: This probably also needs to check whether the player's capabilities include that of spectator mode (tile entities don't tick last I recall).

src/main/java/org/spongepowered/common/mixin/core/block/tiles/MixinMobSpawnerBaseLogic.java
kashike: Yep, fixed.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
kashike: Agreed.

hsyyid:
Nice job

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -2507,4 +2510,49 @@ public long getWeatherStartTime() {
...
+
+            double playerDistance = player.getDistanceSq(x, y, z);
+
+            if ((distance < 0.0D || playerDistance < distance * distance) && (bestDistance == -1.0D || playerDistance < bestDistance)) {
+                bestDistance = playerDistance;
+                result = player;
+            }
+        }
+
+        return result;
```
gabizou: IF this can return null, either mark it as `@Nullable` or make it an `Optional`.

src/main/java/org/spongepowered/common/interfaces/IMixinWorld.java
```diff
@@ -100,4 +103,12 @@
 
     void notifyBlockOfStateChange(BlockPos notifyPos, Block sourceBlock, BlockPos sourcePos);
 
+    @Nullable
+    EntityPlayer getClosestPlayerToEntityWhoAffectsSpawning(net.minecraft.entity.Entity entity, double distance);
+
+    @Nullable
+    EntityPlayer getClosestPlayerWhoAffectsSpawning(double x, double y, double z, double distance);
+
+    boolean isAnyPlayerWithinRangeAtWhoAffectsSpawning(double x, double y, double z, double range);
```
phroa: is `AtWho` intentional?

kashike: It is a "who affects spawning" version of `net/minecraft/world/World.isAnyPlayerWithinRangeAt
`, so yes.

kenzierocks: Maybe `isAnySpawnAffectingPlayerWithinRangeAt`?

ST-DDT: If you want to stick as closely to the original name of that method then you could but the parts in proper order. 
`isAnyPlayerWhoAffectsSpawningWithinRangeAt`

but @kenzierocks will work as well. (I'm not a grammar expert and can't tell which version is better, but the current naming is just strage)

phroa: Okay, I understand what you're trying to get at. The range's position is the `At`, the players affecting spawning are the `Who`.
Not a fan of the name though. I'd suggest something along the lines of `isPlayerNearRangeAffectingSpawning` but even that's pretty long winded. Might just have to go with a short thing and explain in javadocs.

kashike: I'd rather stick with the current name - it's just adding two words onto the original. It's not like this is an API method.

PR #335, opened 2015-12-18
Fix affectsSpawning defaulting to false

kashike:


bloodmc:
Added in https://github.com/SpongePowered/SpongeCommon/commit/ee0b5ade8349627d4474696112cc9953d13a3ecf

Thanks.

PR #356, opened 2015-12-23
[WIP] MobSpawnerData Processor implementations

ZephireNZ:
Implementation of `REPLACE_WITH_DATA_MANIPULATOR`.
- [ ] Registration of field getters and setters
- [ ] Accurately used the appropriate `AbstractData` implementation

Implementation of `REPLACE_WITH_IMMUTABLE_MANIPULATOR`
- [ ] Accurately extended the appropriate `AbstractImmutableData` implementation
- [ ] Accurately instantiating final instance fields with an `ImmutableValue` counter part for any value getters

Implementation of the `VALUE_PROCESSOR`(s):
- [ ] Individual processors for each source type possible (one for `ItemStack`/`TileEntity`/`Entity` as necessary).

Registration:
- [ ] Registered the `Key` correctly by `FIELD_NAME` in the `KeyRegistry`
- [ ] Registered the `DataProcessor`s and `DataManipulatorBuilder` in `SpongeSerializationRegistry`

Test Plugin code:

There will be a lot of tests for such a large plugin, will post to a gist:
https://gist.github.com/ZephireNZ/187e1b6960feea9a0a80

Notes:


src/main/resources/common_at.cfg
```diff
@@ -179,8 +179,20 @@ public-f net.minecraft.server.management.UserList *
...
+public net.minecraft.tileentity.MobSpawnerBaseLogic field_98283_g # minSpawnDelay
+public net.minecraft.tileentity.MobSpawnerBaseLogic field_98293_h # maxSpawnDelay
+public net.minecraft.tileentity.MobSpawnerBaseLogic field_98294_i # spawnCount
 public net.minecraft.tileentity.MobSpawnerBaseLogic field_98292_k # maxNearbyEntities
+public net.minecraft.tileentity.MobSpawnerBaseLogic field_98289_l # activatingRangeFromPlayer
+public net.minecraft.tileentity.MobSpawnerBaseLogic field_98290_m # spawnRange
+public net.minecraft.tileentity.MobSpawnerBaseLogic func_98276_e()Ljava/lang/String; # getEntityNameToSpawn
+public net.minecraft.tileentity.MobSpawnerBaseLogic func_180613_a(Lnet/minecraft/entity/Entity;Z)Lnet/minecraft/entity/Entity; # spawnNewEntity
+public net.minecraft.tileentity.MobSpawnerBaseLogic$WeightedRandomMinecart field_98222_b # nbtData
+public net.minecraft.tileentity.MobSpawnerBaseLogic$WeightedRandomMinecart field_98223_c # entityType
```
ZephireNZ: Should I use a mixin with getters/setters for this instead?

ST-DDT: If possible use existing getters/setters (unless they have unintended side effects), but you dont have to create them if they dont exist.

gabizou: No. It's not necessary to write a mixin to just have access to these.

src/main/java/org/spongepowered/common/data/processor/value/SpawnerMaximumDelayValueProcessor.java
```diff
@@ -0,0 +1,51 @@
...
+    public SpawnerMaximumDelayValueProcessor() {
+        super(IMixinMobSpawner.class, Keys.SPAWNER_MAXIMUM_DELAY);
+    }
+
+    @Override
+    protected MutableBoundedValue<Short> constructValue(Short actualValue) {
+        return SpongeValueFactory.boundedBuilder(Keys.SPAWNER_MAXIMUM_DELAY)
+                .minimum((short) 1)
+                .maximum(Short.MAX_VALUE)
+                .defaultValue((short) 800)
```
gabizou: DataConstants please.

src/main/java/org/spongepowered/common/data/processor/value/SpawnerMaximumDelayValueProcessor.java
```diff
@@ -0,0 +1,51 @@
...
+public class SpawnerMaximumDelayValueProcessor extends AbstractSpongeValueProcessor<IMixinMobSpawner, Short, MutableBoundedValue<Short>> {
+
+    public SpawnerMaximumDelayValueProcessor() {
+        super(IMixinMobSpawner.class, Keys.SPAWNER_MAXIMUM_DELAY);
+    }
+
+    @Override
+    protected MutableBoundedValue<Short> constructValue(Short actualValue) {
+        return SpongeValueFactory.boundedBuilder(Keys.SPAWNER_MAXIMUM_DELAY)
+                .minimum((short) 1)
```
gabizou: Data Constants.

src/main/java/org/spongepowered/common/data/processor/value/SpawnerMinimumDelayValueProcessor.java
```diff
@@ -0,0 +1,51 @@
...
+public class SpawnerMinimumDelayValueProcessor extends AbstractSpongeValueProcessor<IMixinMobSpawner, Short, MutableBoundedValue<Short>> {
+
+    public SpawnerMinimumDelayValueProcessor() {
+        super(IMixinMobSpawner.class, Keys.SPAWNER_MINIMUM_DELAY);
+    }
+
+    @Override
+    protected MutableBoundedValue<Short> constructValue(Short actualValue) {
+        return SpongeValueFactory.boundedBuilder(Keys.SPAWNER_MINIMUM_DELAY)
+                .minimum((short) 0)
```
gabizou: Data Constants.

src/main/java/org/spongepowered/common/data/processor/data/MobSpawnerDataProcessor.java
```diff
@@ -0,0 +1,137 @@
...
+            weight = logic.randomEntity.itemWeight;
+        }
+
+        Class<Entity> entityClass = (Class<Entity>) EntityList.stringToClassMapping.getOrDefault(logic.getEntityNameToSpawn(), "Pig");
+        EntityType type = EntityTypeRegistryModule.getInstance().getForClass(entityClass);
+
+        Optional<Entity> nextEntity = createEntity(type, logic.getSpawnerWorld());
+        if(nextEntity.isPresent()) {
+            logic.spawnNewEntity(nextEntity.get(), false);
+            values.put(Keys.SPAWNER_NEXT_ENTITY_TO_SPAWN, new WeightedSerializableObject<>(new SpongeEntitySnapshotBuilder().from(nextEntity.get()).build(), weight));
```
ST-DDT: You could simplify this  a bit using `nextEntity.get().createSnapshot()`.

src/main/java/org/spongepowered/common/data/manipulator/immutable/ImmutableSpongeMobSpawnerData.java
```diff
@@ -25,6 +25,14 @@
 package org.spongepowered.common.data.manipulator.immutable;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static org.spongepowered.common.data.util.DataConstants.DEFAULT_SPAWNER_MAXMIMUM_NEARBY_ENTITIES;
```
gabizou: don't static import these please, just qualify them.

src/main/java/org/spongepowered/common/data/manipulator/mutable/SpongeMobSpawnerData.java
```diff
@@ -25,6 +25,7 @@
 package org.spongepowered.common.data.manipulator.mutable;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static org.spongepowered.common.data.util.DataConstants.*;
```
gabizou: star imports, and don't import the individual fields.

ZephireNZ:
@gabizou are you still willing to test this personally, or should I set up one?

ZephireNZ:
I'll rebase regardless :)

src/main/java/org/spongepowered/common/data/DataRegistrar.java
```diff
@@ -564,6 +564,9 @@ public static void setupSerialization(Game game) {
         dataManager.registerDataProcessorAndImpl(DyeableData.class, SpongeDyeableData.class, ImmutableDyeableData.class,
                 ImmutableSpongeDyeableData.class, dyeableDataProcessor);
 
+        final MobSpawnerDataProcessor mobSpawnerDataProcessor = new MobSpawnerDataProcessor();
```
simon816: Don't need to create this variable

src/main/java/org/spongepowered/common/data/processor/common/SpawnerUtils.java
```diff
@@ -0,0 +1,100 @@
+package org.spongepowered.common.data.processor.common;
```
simon816: Missing license header

src/main/java/org/spongepowered/common/data/processor/common/SpawnerUtils.java
```diff
@@ -0,0 +1,100 @@
...
+import org.spongepowered.api.util.weighted.WeightedTable;
+import org.spongepowered.common.entity.SpongeEntitySnapshot;
+import org.spongepowered.common.entity.SpongeEntitySnapshotBuilder;
+import org.spongepowered.common.registry.type.entity.EntityTypeRegistryModule;
+
+import java.util.Optional;
+
+public class SpawnerUtils {
+
+    public static Optional<Entity> createEntity(EntityType type, World world) {
```
simon816: Why not just SpongeAPI's `world.createEntity`?

src/main/java/org/spongepowered/common/data/manipulator/immutable/ImmutableSpongeMobSpawnerData.java
```diff
@@ -189,39 +188,39 @@ public ImmutableNextEntityToSpawnValue nextEntityToSpawn() {
         return this.toSpawnValue;
     }
 
-    public short getRemaining() {
+    private short getRemaining() {
```
simon816: lots of diff noise here. could be an ocd change

src/main/java/org/spongepowered/common/data/processor/data/MobSpawnerDataProcessor.java
```diff
@@ -0,0 +1,109 @@
+package org.spongepowered.common.data.processor.data;
```
simon816: license header again

src/main/java/org/spongepowered/common/data/processor/value/SpawnerMaximumNearbyEntitiesValueProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+                .minimum((short) 0)
+                .maximum(Short.MAX_VALUE)
+                .defaultValue(DataConstants.DEFAULT_SPAWNER_MAXMIMUM_NEARBY_ENTITIES)
+                .actualValue(actualValue)
+                .build();
+    }
+
+    @Override
+    protected boolean set(IMixinMobSpawner container, Short value) {
+        container.getSpawnerBaseLogic().maxNearbyEntities = Preconditions.checkNotNull(value);
```
simon816: `Preconditions` methods are imported statically

gabizou:
Still a WIP?

src/main/java/org/spongepowered/common/data/manipulator/immutable/ImmutableSpongeMobSpawnerData.java
gabizou: Nah, these could also just be removed with lambda notations: `() -> this.remaining`

src/main/java/org/spongepowered/common/data/manipulator/immutable/ImmutableSpongeMobSpawnerData.java
```diff
@@ -189,39 +188,39 @@ public ImmutableNextEntityToSpawnValue nextEntityToSpawn() {
...
     }
 
-    public WeightedSerializableObject<EntitySnapshot> getNextToSpawn() {
+    private WeightedSerializableObject<EntitySnapshot> getNextToSpawn() {
         return this.nextToSpawn;
     }
 
-    public WeightedTable<EntitySnapshot> getEntitiesToSpawn() {
+    private WeightedTable<EntitySnapshot> getEntitiesToSpawn() {
         return this.entitiesToSpawn;
```
gabizou: Note that this is leaking a mutable collection, so at least, this could be duplicated on get.

src/main/java/org/spongepowered/common/data/processor/common/SpawnerUtils.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    public static Optional<WeightedSerializableObject<EntitySnapshot>> getNextSnapshot(MobSpawnerBaseLogic logic) {
+        int weight = 1;
+        if(logic.getRandomEntity() != null) {
+            weight = logic.getRandomEntity().itemWeight;
+        }
+
+        Class<? extends Entity> entityClass = EntityList.stringToClassMapping.getOrDefault(logic.getEntityNameToSpawn(), EntityPig.class);
```
gabizou: This is going to need to be verified with Forge since I believe Forge handles entity registration differently.

src/main/java/org/spongepowered/common/data/processor/common/SpawnerUtils.java
ZephireNZ: That method requires that a location be required - is Location(0,0,0) a reasonable default? This would also be carried across to the EntitySnapshot, so it will have a potentially unreasonable default spawn.

simon816: I think that the method you've got right here will also have a default position of 0,0,0 so it's no worse. difference is that world.createEntity does validation on the position so if it returns a non-empty result you know it can spawn there

src/main/java/org/spongepowered/common/data/processor/common/SpawnerUtils.java
```diff
@@ -0,0 +1,100 @@
...
+    public static void setSnapshots(MobSpawnerBaseLogic logic, WeightedTable<EntitySnapshot> table) {
+        logic.minecartToSpawn.clear();
+        for (TableEntry<EntitySnapshot> entry : table) {
+            if(!(entry instanceof WeightedObject)) continue;
+            WeightedObject<EntitySnapshot> object = (WeightedObject<EntitySnapshot>) entry;
+
+            // Can't get NBT
+            if(!(object.get() instanceof SpongeEntitySnapshot)) continue;
+
+            String name = EntityList.classToStringMapping.get(object.get().getType().getEntityClass());
```
ZephireNZ: @gabizou Was trying to work out a way to do this using `EntityTypeRegistryModule` however it looks like it throws away the capitalized entity type. Edit SpongeEntityType, or override the MC method for forge compat?

src/main/java/org/spongepowered/common/data/processor/common/SpawnerUtils.java
ZephireNZ: I assume getting the class from `EntityTypeRegistryModule` would use the existing solution?

src/main/java/org/spongepowered/common/data/manipulator/mutable/SpongeMobSpawnerData.java
```diff
@@ -205,75 +204,7 @@ public DataContainer toContainer() {
...
-    }
-
-    public WeightedTable<EntitySnapshot> getEntities() {
-        return this.entities;
-    }
-
-    public void setEntities(WeightedTable<EntitySnapshot> entities) {
+    private void setEntities(WeightedTable<EntitySnapshot> entities) {
         checkNotNull(entities).forEach(Preconditions::checkNotNull);
         this.entities = entities;
```
gabizou: Should defensively clone the `WeightedTable`.

src/main/java/org/spongepowered/common/data/processor/value/SpawnerRequiredPlayerRangeValueProcessor.java
```diff
@@ -0,0 +1,77 @@
...
+                .minimum((short) 0)
+                .maximum(Short.MAX_VALUE)
+                .defaultValue(DataConstants.DEFAULT_SPAWNER_REQUIRED_PLAYER_RANGE)
+                .actualValue(actualValue)
+                .build();
+    }
+
+    @Override
+    protected boolean set(IMixinMobSpawner container, Short value) {
+        container.getSpawnerBaseLogic().activatingRangeFromPlayer = checkNotNull(value);
```
gabizou: No need for a checknontnull, the superclass does that for you.

src/main/java/org/spongepowered/common/data/processor/value/SpawnerSpawnCountValueProcessor.java
```diff
@@ -0,0 +1,77 @@
...
+                .minimum((short) 0)
+                .maximum(Short.MAX_VALUE)
+                .defaultValue(DataConstants.DEFAULT_SPAWNER_SPAWN_COUNT)
+                .actualValue(actualValue)
+                .build();
+    }
+
+    @Override
+    protected boolean set(IMixinMobSpawner container, Short value) {
+        container.getSpawnerBaseLogic().spawnCount = checkNotNull(value);
```
gabizou: And here.

src/main/java/org/spongepowered/common/data/processor/common/SpawnerUtils.java
gabizou: Forge compat will have to be done in the forge side. Modifying the implementation accordingly (basically, just have a higher priority for the processor).

ZephireNZ:
@gabizou, I think the method I use currently seems like it will work with the forge entity system, however I'm not experienced with it.

https://github.com/MinecraftForge/MinecraftForge/blob/6644e8c9ed4eb9db868540f8672bb83cb8aa09c0/src/main/java/net/minecraftforge/fml/common/registry/EntityRegistry.java#L178

It looks like Forge has it's own registry, in addition to registering them into Vanilla's EntityList. Only thing I can think of is if mod entities are not serialised to NBT in the format that is saved in EntityList.

gabizou:
Yeah, basically, you'll have to work with it through the forge side, where there's another `ForgeMobSpawnerDataProcessor` and likewise value processor variants on the forge end, with higher priorities (so they get called earlier).

ZephireNZ:
Just did some testing of the Mob spawners with mods to ensure that it's fine as is.

As expected - I am able to add a modded entity to the spawner, and have it properly serialise and deserialise from NBT.

However I have encountered a hiccup - generally the NBT for an entity's spawner data is very limited. With the current system, *everything* is added, including location, uuid, etc.

ZephireNZ:
It looks like the base game never has to deal with this - spawners in dungeons only ever have a type, not data when generated, and the `/blockdata` command expects players to choose the data to use.

@gabizou, thoughts on how to proceed?

ZephireNZ:
Ok, so after some more thinking:

Perhaps add to javadocs that using EntitySnapshot created directly from an entity is *bad*, unless you really really want all data copied across.

Then, for `EntitySnapshot`s created without a tag:

- If they exist, serialise location/rotation from snapshot
- For custom data manipulators, save as normal
- This is the tricky bit
  - Create an entity of the given type
  - For each non-custom manipulator, serialise to the dataholder all values
  - Compare the old and new tags, and save the difference to the container (it's more work, but stops us from saving an entire entity to the tag)

There's a high chance I'm missing an easier way haha

gabizou:
@ZephireNZ Care to update me on what's holding this up?

ZephireNZ:
Right now trying to implement what I mentioned above. Entity based on the `DataManipulators` on the snapshot, not just the raw NBT data.

What are you thoughts on what I mentioned?

src/main/java/org/spongepowered/common/data/processor/data/MobSpawnerDataProcessor.java
simon816: Should be in the `processor.multi` package

gabizou:
@ZephireNZ in working with @Deamon5550 for [`EntityArchetype`](https://github.com/SpongePowered/SpongeAPI/blob/feature/schematics/src/main/java/org/spongepowered/api/entity/EntityArchetype.java), I believe this will end up replacing the use of `EntitySnapshot`s entirely for the use case of `MobSpawnerData` and the immutable equivalent.

ZephireNZ:
In that case, will put this on ice until we have the `EntityArchetype`s, as those will be better than me hacking together the solution I described above.

ZephireNZ:
Closed by #710

PR #364, opened 2015-12-26
Implementation of Text API 3

Minecrell:
See SpongePowered/SpongeAPI#988


src/main/java/org/spongepowered/common/text/serializer/xml/TextXmlTextSerializer.java
```diff
@@ -48,8 +44,8 @@
 /**
  * Xml format serializer for Text instances.
  */
-public class TextXmlRepresentation implements TextRepresentation {
-    public static final TextXmlRepresentation INSTANCE = new TextXmlRepresentation();
+public class TextXmlTextSerializer implements TextSerializer {
```
kashike: `XmlTextSerializer`

Minecrell: no.

kashike: should probably clarify that this is "Text XML" and not "XML" then

src/main/java/org/spongepowered/common/data/manipulator/mutable/SpongeCommandData.java
```diff
@@ -155,7 +154,7 @@ public SpongeCommandData shouldTrackOutput(boolean track) {
     }
 
     public SpongeCommandData setLastOutput(Text message) {
-        if (Texts.toPlain(checkNotNull(message, "Null message! Use empty text instead!")).isEmpty()) {
```
simon816: still should keep the checkNotNull

Minecrell: Fixed.

ST-DDT:
Yesterday, somebody did complain about me not testing my proposed changes.
I created a test plugin to test all implementations of `CatalogType`s that can be obtained from the `GameRegistry`.
Here is the result: http://pastebin.com/8pC49mn6 
* There is a summary at the bottom (last 100 lines)
* I omited the stacktraces for AbstractMethodErrors and repeating errors
* tested with this branch/PR

There are round about 100 methods that do not properly implement `CatalogType`s, although most of them will be fixed if the super class is fixed. (Some don't even implement `getId()`!)

[Test-Plugin](https://dl.dropboxusercontent.com/u/16999313/Sponge/ZoneAPIChecker.jar)
[Source](https://dl.dropboxusercontent.com/u/16999313/Sponge/ZoneAPIChecker.zip)

Mumfrey:
> I created a test plugin to test all implementations of CatalogTypes that can be obtained from the GameRegistry

This was kind of unnecessary since a Mixin audit would give you exactly the same information with zero work.

ST-DDT:
@Mumfrey Then why didn't anybody do it before and uncovered those?

Shall I fix them or would you like to do it yourself?

Minecrell:
https://github.com/SpongePowered/SpongeCommon/commit/1a560d6ccdacb1d2413fba0944acee61a291d5fe

PR #368, opened 2015-12-27
Fix MixinExplosion breaking blocks

### LINKED ISSUE #324: MixinExplosion does not work on breaking blocks
Currently, MixinExplosion will return whether the explosion `isSmoking` when calling `shouldBreakBlocks()`. The boolean should be added and the explosion should actually be injected to override the blocks as long as the boolean is `true`.

AIOOB:
Fixes #324

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/MixinExplosion.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/MixinExplosion.java
@@ -35,15 +35,17 @@ import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import org.spongepowered.common.interfaces.world.IMixinExplosion;
 
 import java.util.List;
 import java.util.Optional;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/mixin/core/world/MixinExplosion.java
```diff
@@ -56,11 +56,21 @@
     @SuppressWarnings("rawtypes")
     @Shadow public List affectedBlockPositions;
 
+    private boolean shouldBreakBlocks;
```
gabizou: this should be settable from the `ExplosionBuilder` as well, not just default to whether if `isSmoking == true`.

AIOOB: It is, in `ExplosionBuilder`, `shouldBreakBlocks` is passed in in place of `isSmoking` and `isSmoking` is set separately; I could change this or comment it if you think this is a problem.

gabizou: Change it so that the explosion builder is setting isSmoking and then sets the should break blocks correctly. no need for confusion.

AIOOB: Fixed

PR #376, opened 2015-12-31
Call ChangeGameModeEvent.TargetPlayer on gamemode change

kashike:


src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -452,4 +459,35 @@ public void setVelocity(Vector3d velocity) {
...
         this.velocityOverride = velocity;
     }
+
+    /**
+     * @author kashike - 30/12/2015
+     */
+    @Overwrite
+    public void setGameType(WorldSettings.GameType gameType) {
+        // Sponge start
+        ChangeGameModeEvent.TargetPlayer event = SpongeEventFactory.createChangeGameModeEventTargetPlayer(Cause.of(NamedCause.source(this)),
```
simon816: Could make this an `@Inject` at HEAD if it's just modifying at the top of the method

kashike: It changes a local variable, this is the cleanest way imo.

gabizou: I'm fairly sure this can be done with either a `ModifyArgs` or `Inject` with capture locals.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -452,4 +459,35 @@ public void setVelocity(Vector3d velocity) {
     public void setVelocityOverride(@Nullable Vector3d velocity) {
         this.velocityOverride = velocity;
     }
+
+    /**
+     * @author kashike - 30/12/2015
```
gabizou: Add a bit more descriptive documentation of what this is doing and why, etc.

kashike: done

PR #377, opened 2015-12-31
Implement PlayerChangeClientSettingsEvent and associated data

kashike:
This implements [`SkinPart`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/data/type/SkinPart.java), [`ChatVisibility`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/chat/ChatVisibility.java),  and [`PlayerChangeClientSettingsEvent`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/event/entity/living/humanoid/player/PlayerChangeClientSettingsEvent.java).

### DIFF
--- a/src/main/java/org/spongepowered/common/data/type/SpongeSkinPart.java
+++ b/src/main/java/org/spongepowered/common/data/type/SpongeSkinPart.java
@@ -1,0 +1,52 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data.type;
+
+import com.google.common.base.Objects;
+import org.spongepowered.api.data.type.SkinPart;
+import org.spongepowered.common.SpongeCatalogType;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+public final class SpongeSkinPart extends SpongeCatalogType.Translatable implements SkinPart {
+
+    private final int ordinal;
+    private final int mask;
+
+    public SpongeSkinPart(int ordinal, String id) {
+        super("minecraft:" + id, new SpongeTranslation("options.modelPart." + id));
+        this.ordinal = ordinal;
+        this.mask = 1 << this.ordinal;
+    }
+
+    public boolean test(int flags) {
+        return (flags & this.mask) != 0;
+    }
+
+    @Override
+    protected Objects.ToStringHelper toStringHelper() {
+        return super.toStringHelper()
+                .add("ordinal", this.ordinal);
+    }
+}
--- a/src/main/java/org/spongepowered/common/mixin/core/data/types/MixinEnumChatVisibility.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/data/types/MixinEnumChatVisibility.java
@@ -1,0 +1,90 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.mixin.core.data.types;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.collect.Sets;
+import net.minecraft.entity.player.EntityPlayer;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.text.chat.ChatType;
+import org.spongepowered.api.text.chat.ChatTypes;
+import org.spongepowered.api.text.chat.ChatVisibility;
+import org.spongepowered.api.text.translation.Translation;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+import java.util.Set;
+
+@Mixin(EntityPlayer.EnumChatVisibility.class)
+public abstract class MixinEnumChatVisibility implements ChatVisibility {
+
+    @Shadow private String resourceKey;
+    private final String id = ((Enum) (Object) this).name().toLowerCase();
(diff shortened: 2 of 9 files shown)


src/main/java/org/spongepowered/common/util/SkinUtil.java
```diff
@@ -0,0 +1,48 @@
...
+
+public final class SkinUtil {
+
+    private SkinUtil() {
+    }
+
+    public static Set<SkinPart> fromFlags(int flags) {
+        Set<SkinPart> parts = Sponge.getRegistry().getAllOf(SkinPart.class).stream()
+                .filter(part -> part instanceof SpongeSkinPart)
+                .filter(part -> (flags & (1 << ((SpongeSkinPart) part).ordinal())) != 0)
```
Minecrell: `.filter(part -> (part instanceof SpongeSkinPart) && (flags [...])`?

src/main/java/org/spongepowered/common/util/SkinUtil.java
```diff
@@ -0,0 +1,48 @@
...
+                .collect(Collectors.toSet());
```
Minecrell: If you copy anyway you can just collect it to an immutable set directly (guava collectors or something in SpongeAPI).

src/main/java/org/spongepowered/common/data/type/SpongeSkinPart.java
```diff
@@ -0,0 +1,70 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data.type;
+
+import com.google.common.base.Objects;
+import org.spongepowered.api.data.type.SkinPart;
+import org.spongepowered.api.text.translation.Translation;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+public final class SpongeSkinPart implements SkinPart {
```
ST-DDT: You could extend SpongeEntityMeta? class. That class a public int type field.
And will also provide/inherit other usefull methods see TextAPI v3.

src/main/java/org/spongepowered/common/data/type/SpongeSkinPart.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public String getId() {
+        return "minecraft:" + this.id;
+    }
+
+    @Override
+    public String getName() {
+        return this.id;
```
ST-DDT: This could be getTranslation().get()

kashike: yes

src/main/java/org/spongepowered/common/util/SkinUtil.java
```diff
@@ -0,0 +1,46 @@
...
+
+public final class SkinUtil {
+
+    private SkinUtil() {
+    }
+
+    public static Set<SkinPart> fromFlags(int flags) {
+        return Sponge.getRegistry().getAllOf(SkinPart.class).stream()
+                .filter(part -> part instanceof SpongeSkinPart)
+                .filter(part -> (flags & (1 << ((SpongeSkinPart) part).ordinal)) != 0)
```
ST-DDT: Create a util method (`machesFlag(int)` or similar) in `SpongeSkinPart` to improve readability, reuseability and maintainability.

src/main/java/org/spongepowered/common/util/SkinUtil.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Set;
+
+public final class SkinUtil {
+
+    private SkinUtil() {
+    }
+
+    public static Set<SkinPart> fromFlags(int flags) {
+        return Sponge.getRegistry().getAllOf(SkinPart.class).stream()
+                .filter(part -> part instanceof SpongeSkinPart)
```
ST-DDT: Also I thing this check is not neccessary since, we use our own implementation here, without ability to register more.
just use a plain `.map(input -> (SpongeSkinPart) input)`. If there will ever be more, then we would have to add them to our registry anyway.

src/main/java/org/spongepowered/common/data/type/SpongeSkinPart.java
```diff
@@ -0,0 +1,95 @@
...
+        this.translation = new SpongeTranslation("options.modelPart." + this.id);
+    }
+
+    public boolean test(int flags) {
+        return (flags & this.mask) != 0;
+    }
+
+    @Override
+    public String getId() {
+        return "minecraft:" + this.id;
```
simon816: Don't think this prefix is nessesary

src/main/java/org/spongepowered/common/data/type/SpongeSkinPart.java
```diff
@@ -0,0 +1,95 @@
...
+    private final int ordinal;
+    private final int mask;
+    private final String id;
+    private final Translation translation;
+
+    public SpongeSkinPart(int ordinal, String id) {
+        this.ordinal = ordinal;
+        this.mask = 1 << this.ordinal;
+        this.id = id;
+        this.translation = new SpongeTranslation("options.modelPart." + this.id);
```
Aaron1011: Wheres is this string coming from?

src/main/java/org/spongepowered/common/mixin/core/data/types/MixinEnumChatVisibility.java
```diff
@@ -0,0 +1,72 @@
...
+@Mixin(EntityPlayer.EnumChatVisibility.class)
+public abstract class MixinEnumChatVisibility implements ChatVisibility {
+
+    @Shadow private String resourceKey;
+    private final String id = ((Enum) (Object) this).name().toLowerCase();
+    private final Translation translation = new SpongeTranslation(this.resourceKey);
+
+    @Override
+    public boolean isVisible(ChatType type) {
+        switch ((EntityPlayer.EnumChatVisibility) (Object) this) {
```
Aaron1011: This should be cached in the constructor.

src/main/java/org/spongepowered/common/util/SkinUtil.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.Set;
+
+public final class SkinUtil {
+
+    private SkinUtil() {
+    }
+
+    public static Set<SkinPart> fromFlags(int flags) {
+        return Sponge.getRegistry().getAllOf(SkinPart.class).stream()
```
Aaron1011: Instead of this, why not just create a mapping of each possible flags combination to skin parts? There's only 47 total (7 skin parts * 7 skin parts)

src/main/java/org/spongepowered/common/data/type/SpongeSkinPart.java
```diff
@@ -0,0 +1,95 @@
...
+package org.spongepowered.common.data.type;
+
+import com.google.common.base.Objects;
+import org.spongepowered.api.data.type.SkinPart;
+import org.spongepowered.api.text.translation.Translation;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+import javax.annotation.Nullable;
+
+public final class SpongeSkinPart implements SkinPart {
```
ST-DDT: Extend `SpongeCatalogType.Translateable`

src/main/java/org/spongepowered/common/mixin/core/data/types/MixinEnumChatVisibility.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public String getId() {
+        return "minecraft:" + this.id;
+    }
+
+    @Override
+    public String getName() {
+        return this.id;
```
ST-DDT: If you have a translation use that if possible.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -116,11 +123,15 @@
...
     @Shadow public NetHandlerPlayServer playerNetServerHandler;
+    @Shadow public MinecraftServer mcServer;
     @Shadow public ItemInWorldManager theItemInWorldManager;
     @Shadow public int lastExperience;
-    @Shadow public MinecraftServer mcServer;
     @Shadow public abstract void setSpectatingEntity(Entity entityToSpectate);
     @Shadow public abstract void sendPlayerAbilities();
+    @Shadow private EntityPlayer.EnumChatVisibility chatVisibility;
+    @Shadow private boolean chatColours;
+    private Set<SkinPart> skinParts;
```
ST-DDT: Make this a empty set or a set containing all parts to prevent returing null in the getter if called too early.

src/main/java/org/spongepowered/common/SpongeCatalogType.java
```diff
@@ -68,7 +68,7 @@ public final boolean equals(Object obj) {
     }
 
     @Override
-    public final String toString() {
```
simon816: This should stay final

src/main/java/org/spongepowered/common/data/type/SpongeSkinPart.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public boolean test(int flags) {
+        return (flags & this.mask) != 0;
+    }
+
+    @Override
+    public String toString() {
+        return this.toStringHelper()
+                .add("ordinal", this.ordinal)
```
simon816: Just override `toStringHelper`

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -171,12 +183,42 @@ public User getUserObject() {
         return this.user;
     }
 
+    @Inject(method = "handleClientSettings", at = @At("RETURN"))
```
simon816: Should be at HEAD because the event should fire _before_ the player object gets updated

src/main/java/org/spongepowered/common/util/SkinUtil.java
simon816: Don't think a mapping is necessary for this

src/main/java/org/spongepowered/common/registry/type/text/ChatVisibilityRegistryModule.java
```diff
@@ -0,0 +1,73 @@
...
+        return Optional.ofNullable(this.chatVisibilityMap.get(checkNotNull(id, "id").toLowerCase()));
+    }
+
+    @Override
+    public Collection<ChatVisibility> getAll() {
+        return ImmutableSet.copyOf((ChatVisibility[]) (Object[]) EntityPlayer.EnumChatVisibility.values());
+    }
+
+    @AdditionalRegistration
+    public void customRegistration() {
```
simon816: Why do you need additional registration here?

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
kashike: done

src/main/java/org/spongepowered/common/data/type/SpongeSkinPart.java
kashike: done

src/main/java/org/spongepowered/common/SpongeCatalogType.java
kashike: done

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -171,12 +183,45 @@ public User getUserObject() {
         return this.user;
     }
 
+    @Inject(method = "handleClientSettings", at = @At("HEAD"))
+    public void processClientSettingsEvent(C15PacketClientSettings packet, CallbackInfo ci) {
+        PlayerChangeClientSettingsEvent event = SpongeEventFactory.createPlayerChangeClientSettingsEvent(Cause.of(NamedCause.source(this)),
```
Aaron1011: The event is supposed to contain the new values, but you're creating it with all of the old ones.

Issue #394, opened 2016-01-03
ClientConnectionEvent.Join fired twice


DarkArc:
The ClientConnectionEvent.Join event is fired twice upon a player joining the server.

simon816:
Only does it once in my test (SpongeForge)
Maybe you've registered your listener twice?

DarkArc:
My bad, yup.

PR #395, opened 2016-01-03
Prevent registering an instance of a listener more than once

kashike:
This prevents a single instance of a listener from being registered many times.

https://github.com/SpongePowered/SpongeCommon/issues/394
https://github.com/SpongePowered/SpongeForge/issues/295
https://github.com/SpongePowered/SpongeForge/issues/433

### DIFF
--- a/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
@@ -33,6 +33,7 @@ import com.google.common.cache.LoadingCache;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Multimap;
+import com.google.common.collect.Sets;
 import com.google.common.reflect.TypeToken;
 import org.spongepowered.api.event.Cancellable;
 import org.spongepowered.api.event.Event;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/event/SpongeEventManager.java
```diff
@@ -140,6 +141,12 @@ public void registerListener(PluginContainer plugin, Object listenerObject) {
         checkNotNull(plugin, "plugin");
         checkNotNull(listenerObject, "listener");
 
+        if (this.registeredListeners.contains(listenerObject)) {
+            SpongeImpl.getLogger().warn("Plugin {} attempted to register an already registered listener ({})", plugin.getId(),
+                    listenerObject.getClass().getName());
```
JBYoshi: Would be nice to dump the stack trace as well. I have a private Bukkit plugin which (I believe) is registering a certain listener twice, and I don't know where the second registration is.

kashike: alright

src/main/java/org/spongepowered/common/event/SpongeEventManager.java
```diff
@@ -68,6 +68,7 @@
     private final AnnotatedEventListener.Factory handlerFactory = new ClassEventListenerFactory("org.spongepowered.common.event.listener",
             new FilterFactory("org.spongepowered.common.event.filters", classLoader), classLoader);
     private final Multimap<Class<?>, RegisteredListener<?>> handlersByEvent = HashMultimap.create();
+    private final List<Object> registeredListeners = Lists.newArrayList();
```
Aaron1011: Can you make this a `Set` instead?

kashike: done

PR #401, opened 2016-01-05
[WIP] Refactor explosion events

kashike:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1011) | [**Common**](https://github.com/SpongePowered/SpongeCommon/pull/401) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/486)

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityMinecartTNT.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityMinecartTNT.java
@@ -24,10 +24,56 @@
  */
 package org.spongepowered.common.mixin.core.entity.item;
 
+import com.google.common.collect.Lists;
+import net.minecraft.entity.Entity;
 import net.minecraft.entity.item.EntityMinecartTNT;
+import net.minecraft.entity.projectile.EntityArrow;
 import org.spongepowered.api.entity.vehicle.minecart.MinecartTNT;
+import org.spongepowered.api.event.SpongeEventFactory;
+import org.spongepowered.api.event.cause.Cause;
+import org.spongepowered.api.event.cause.NamedCause;
+import org.spongepowered.api.event.world.ExplosionEvent;
+import org.spongepowered.api.world.World;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+import org.spongepowered.common.SpongeImpl;
+
+import java.util.List;
+
+import javax.annotation.Nullable;
 
 @Mixin(EntityMinecartTNT.class)
 public abstract class MixinEntityMinecartTNT extends MixinEntityMinecart implements MinecartTNT {
+
+    @Nullable private Entity shooter;
+
+    @Inject(method = "attackEntityFrom(Lnet/minecraft/util/DamageSource;F)Z", at = @At(value = "INVOKE", target =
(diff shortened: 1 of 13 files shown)


src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -178,6 +178,8 @@
     protected abstract void shadow$setRotation(float yaw, float pitch);
     @Shadow public abstract void setSize(float width, float height);
     @Shadow public abstract boolean isSilent();
+    // NOTE: this method CANNOT be abstract
+    @Shadow(prefix = "entity$") public void entity$onUpdate() { }
```
Zidane: Elaborate

src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityMinecartTNT.java
```diff
@@ -24,10 +24,57 @@
...
+            cause.add(NamedCause.igniter(this.shooter));
+        }
+
+        cause.add(NamedCause.source(this));
+
+        // Reset state
+        this.shooter = null;
+
+        ExplosionEvent.Pre event = SpongeEventFactory.createExplosionEventPre(Cause.of(cause), (World) this.worldObj);
+        SpongeImpl.postEvent(event);
```
Zidane: postEvent returns cancellation status. You can inline it (and I'd prefer it for leaner mixins).

src/main/java/org/spongepowered/common/mixin/core/entity/monster/MixinEntityCreeper.java
```diff
@@ -52,4 +76,98 @@
...
+                this.timeSinceIgnited = 0;
+            }
+
+            if (this.timeSinceIgnited >= this.fuseTime) {
+                this.timeSinceIgnited = this.fuseTime;
+                this.explode();
+            }
+        }
+
+        super.entity$onUpdate();
```
Zidane: This seems...interesting.

@Mumfrey can you make sure this is okay to do?

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
Minecrell: Can't call super on a abstract method

simon816: Shadow methods should have the `shadow$` prefix

kashike: I set the prefix to `entity$` to make it clear where it comes from.

src/main/java/org/spongepowered/common/mixin/core/entity/monster/MixinEntityCreeper.java
kashike: I've seen no other way to accomplish this after playing around for a good half hour, but if there is indeed another way I'll be happy to change it.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
gabizou: no, this isn't even making sense.

src/main/java/org/spongepowered/common/mixin/core/entity/projectile/MixinEntityWitherSkull.java
```diff
@@ -80,4 +94,25 @@ public void writeToNbt(NBTTagCompound compound) {
         }
     }
 
+    @Nullable
+    @Redirect(method = "onImpact(Lnet/minecraft/util/MovingObjectPosition;)V", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/World;"
```
gabizou: For readability, just set the targets to private static final fields since they're inlined in the compiler anyways.

kashike:
Superseded by https://github.com/SpongePowered/SpongeCommon/pull/453

PR #407, opened 2016-01-07
Implement GriefingData

Cybermaxke:
Todo:
- [x] Also block hiding in stone for silverfish if griefing is disabled.

Test plugin code:
```java
import org.spongepowered.api.entity.Entity;
import org.spongepowered.api.event.entity.InteractEntityEvent;
import org.spongepowered.api.event.filter.cause.First;
import org.spongepowered.api.text.Text;

@Plugin(id = "test-plugin", name = "TestPlugin", version = "DEV")
public class TestPlugin {

	@Listener
	public void onPlayerInteractWithEntity(InteractEntityEvent.Secondary event, @First Player player) {
		if (entity.supports(Keys.CAN_GRIEF)) {
			boolean flag = !entity.get(Keys.CAN_GRIEF).get();
			entity.offer(Keys.CAN_GRIEF, flag);
			player.sendMessage(Text.of("Toggled griefing state of " + entity.getType().getName() +
					" to " + flag));
		}
	}
}
```
Note:
There are much more entities (almost all living ones) that are able to grief the world, mainly through destroying crops. And in the explanation of `GriefingData` is it only limited to `Enderman` and `Humanoid`s.

### DIFF
--- a/src/main/java/org/spongepowered/common/registry/type/block/BooleanTraitRegistryModule.java
+++ b/src/main/java/org/spongepowered/common/registry/type/block/BooleanTraitRegistryModule.java
@@ -30,6 +30,7 @@ import com.google.common.collect.ImmutableList;
 import org.spongepowered.api.block.BlockType;
 import org.spongepowered.api.block.trait.BooleanTrait;
 import org.spongepowered.api.block.trait.BooleanTraits;
+import org.spongepowered.api.registry.AlternateCatalogRegistryModule;
 import org.spongepowered.api.registry.util.RegisterCatalog;
 import org.spongepowered.common.registry.SpongeAdditionalCatalogRegistryModule;
 
@@ -38,7 +39,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 
-public final class BooleanTraitRegistryModule implements SpongeAdditionalCatalogRegistryModule<BooleanTrait> {
+public final class BooleanTraitRegistryModule implements SpongeAdditionalCatalogRegistryModule<BooleanTrait>, AlternateCatalogRegistryModule<BooleanTrait> {
 
     @RegisterCatalog(BooleanTraits.class)
     private Map<String, BooleanTrait> booleanTraitMap = new HashMap<>();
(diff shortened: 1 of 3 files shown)


JBYoshi:
I was going to do it myself, but it was more complex than I thought. Thanks for taking over it for me :smile:

src/main/java/org/spongepowered/common/data/manipulator/mutable/entity/SpongeGriefingData.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.spongepowered.api.data.key.Keys;
+import org.spongepowered.api.data.manipulator.immutable.entity.ImmutableGriefingData;
+import org.spongepowered.api.data.manipulator.mutable.entity.GriefingData;
+import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.common.data.manipulator.immutable.entity.ImmutableSpongeGriefingData;
+import org.spongepowered.common.data.manipulator.mutable.common.AbstractBooleanData;
+import org.spongepowered.common.data.util.ImplementationRequiredForTest;
+
+@ImplementationRequiredForTest
```
gabizou: Implementation shouldn't be required for testing a boolean based value.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -779,6 +780,9 @@ public void readFromNbt(NBTTagCompound compound) {
                 }
             }
         }
+        if (this instanceof IMixinGriefer && ((IMixinGriefer) this).isGriefer() && compound.hasKey("CanGrief")) {
```
gabizou: Store these string constants in the `NBTDataUtil` please.

Cybermaxke:
@JBYoshi Haha, no problem, there had claimed it when I looked through the list ;)

src/main/java/org/spongepowered/common/mixin/core/entity/ai/MixinEntitySilverfishAISummon.java
```diff
@@ -0,0 +1,46 @@
...
+import org.spongepowered.common.interfaces.entity.IMixinGriefer;
+
+@Mixin(EntitySilverfish.AISummonSilverfish.class)
+public abstract class MixinEntitySilverfishAISummon extends EntityAIBase {
+
+    @Shadow(aliases = "this$0") private EntitySilverfish silverfish;
+
+    @Redirect(method = "updateTask", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/GameRules;getGameRuleBooleanValue"
+            + "(Ljava/lang/String;)Z"))
+    private boolean onCanGrief(GameRules this$0, String rule) {
```
jamierocks: Why call the GameRules object this$0?

Cybermaxke: It was like that in the examples, but I should probably change it.

Minecrell: `this$0` sounds fine the redirect you have actually refers to the same object as the current mixin you're in, however if it refers to a different object like in this case you should better name it different

Aaron1011:
@Cybermaxke: Could you rebase and squash this?

jamierocks:
RIP

PR #304, opened 2015-12-04
Data & entity contexts

kashike:
[API](https://github.com/SpongePowered/SpongeAPI/pull/945) | [**Common**](https://github.com/SpongePowered/SpongeCommon/pull/304)

PR #408, opened 2016-01-08
Implement ClientConnectionEvent.Prepare

kashike:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1020) | [**Common**](https://github.com/SpongePowered/SpongeCommon/pull/408)

This implements the `ClientConnectionEvent.Prepare` event, which fires _after_ the `ClientConnectionEvent.Login` event but _before_ the `ClientConnectionEvent.Join`.

This change is required for https://github.com/SpongePowered/SpongeAPI/pull/945 (https://github.com/SpongePowered/SpongeCommon/pull/304) to work properly.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "SpongeAPI"]
 	path = SpongeAPI
-	url = https://github.com/SpongePowered/SpongeAPI.git
+	url = https://github.com/kashike/SpongeAPI.git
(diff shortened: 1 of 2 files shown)


simon816:
One problem I see that might happen is that mods expect the event to be fired _after_ these things happened (See: [`ServerConfigurationManager.java.patch`](https://github.com/MinecraftForge/MinecraftForge/blob/master/patches/minecraft/net/minecraft/server/management/ServerConfigurationManager.java.patch#L43)) so things that they expect to work may actually not be ready yet.

If you say this fires before spawn, then that's a breaking change and you will also need to change the API too (you need to change [this](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/event/network/ClientConnectionEvent.java#L55))

Also, it would be useful to know _why_ this is required for data contexts, so perhaps we can come up with a less breaking change

kashike:
A less breaking change would be to introduce _another_ event fired between `Login` and `Join` - give me a name and I'll happily do it.

kashike:
https://github.com/SpongePowered/SpongeCommon/commit/c8a0ab22f117e84522a192a29656585cb1c7aea8

PR #412, opened 2016-01-10
Quiet the watchdog on the first run

kashike:
Prevents:
```
[23:28:04] [Server thread/INFO]: Done (10.520s)! For help, type "help" or "?"
[23:28:04] [Server Watchdog/FATAL]: A single server tick took 71.62 seconds (should be max 0.05)
[23:28:04] [Server Watchdog/FATAL]: Considering it to be crashed, server will forcibly shutdown.
[23:28:04] [Server Watchdog/ERROR]: This crash report has been saved to: /home/github/SpongePowered/SpongeForge/run/./crash-reports/crash-2015-12-30_23.28.04-server.txt
[23:28:06] [Server thread/WARN]: Can't keep up! Did the system time change, or is the server overloaded? Running 2034ms behind, skipping 40 tick(s)
```


simon816:
If this happens on forge I think it would be useful to patch it on forge too

PR #413, opened 2016-01-10
Prevent creation of a size 0 itemstack when deserializing

Aaron1011:
This resolves https://github.com/SpongePowered/SpongeForge/issues/483.

When an `ItemStack` is being deserialized from NBT, it will have a size of 0 if the `Count` tag isn't present. This usually happens when using `/summon`, as in the linked issue.

Vanilla usually doesn't care about this, however, our damage event code expects that an `ItemStack` has a size of 0 when it breaks. This means that the first time an entity wearing a size-0 `ItemStack` takes damage, our code removes the `ItemStack` from its inventory.

I've added a config option, in case this causes unexpected behavior with mods and/or Forge.

### DIFF
--- a/src/main/java/org/spongepowered/common/config/SpongeConfig.java
+++ b/src/main/java/org/spongepowered/common/config/SpongeConfig.java
@@ -96,6 +96,9 @@ public class SpongeConfig<T extends SpongeConfig.ConfigBase> {
     // BUNGEECORD
     public static final String BUNGEECORD_IP_FORWARDING = "ip-forwarding";
 
+    // PATCHES
+    public static final String ITEM_STACK_PATCH = "itemstack-count-0-patch";
+
     // GENERAL
     public static final String GENERAL_DISABLE_WARNINGS = "disable-warnings";
     public static final String GENERAL_CHUNK_LOAD_OVERRIDE = "chunk-load-override";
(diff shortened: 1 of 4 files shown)


simon816:
Because this bug is specific to the way we handle damage, I think it would be safer if we just change our check for armor stack size (`DamageEventHandler.acceptArmorModifier`) instead of make a more general patch to `ItemStack` just to be safe from unforeseen consequences

Aaron1011:
@simon816: That's exactly why I made this a config option.

When an `ItemStack` runs out of durability, its stack size is decremented in `ItemStack#damageItem`. Generally, it's safe to assume that an `ItemStack` with a stack size of 0 is supposed to be removed. 

Since the aforementioned `/summon` command usage is breaking that contract, I thought it better to fix it there. It's also much simpler to do that, then to try to determine why the `ItemStack` has a size of 0 when we remove it.

If it turns out this breaks something, all the user has to do is disable the config option, and restore the normal behavior.

Aaron1011:
Pinging @SpongePowered/developers for review.

ryantheleach:
I'd test this with ghost slots / Applied Energistics / Forrestry type mods before merging. I'm not 100% sure how they work, but I would be extremely weary of fixing it like this.

kashike:
I'd fix it like @simon816 said, doing it like this seems... bad.

Aaron1011:
@kashike: In what way?

kashike:
This would break things that require the size to be `0`. It would be better, perhaps, to do as @simon816 said.

Aaron1011:
@kashike: No, it doesnt - anything that explicitly sets the size to 0 will be fine. This only changes the case where an `ItemStack` is deserialized from NBT, the size is *not* set afterwards, and the `Count` tag is 0.

PR #415, opened 2016-01-10
Add 'require' argument to injectors

Aaron1011:
SpongeCommon | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/491) | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/203)

Now that we're using Mixin 0.4.10, a new parameter `require` is available in the various injectors (`@Inject`, `@Redirect`, `@ModifyArg`). This requires the injector to find at least as many targets as are specified by `require`.

By enabling this, we get an error when an injection fails, instead of it silently failing. While the `require` parameter can be leveraged to provide soft-failure (like `@Surrogate` for local var capture), all of our current injectors are required. This allows us to easily discover if a coremod is preventing an injection from working, and add another injection point to fix it.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/gen/populators/MixinWorldGenBigMushroom.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/gen/populators/MixinWorldGenBigMushroom.java
@@ -67,7 +67,7 @@ public abstract class MixinWorldGenBigMushroom extends MixinWorldGenerator imple
     private Function<Location<Chunk>, PopulatorObject> override = null;
     private VariableAmount mushroomsPerChunk;
 
-    @Inject(method = "<init>", at = @At("RETURN") )
+    @Inject(method = "<init>", at = @At("RETURN") , require = 1)
     public void onConstructed(CallbackInfo ci) {
         this.types = new WeightedTable<>();
         this.mushroomsPerChunk = VariableAmount.fixed(1);
--- a/src/main/java/org/spongepowered/common/mixin/core/world/gen/populators/MixinWorldGenBlockBlob.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/gen/populators/MixinWorldGenBlockBlob.java
@@ -54,7 +54,7 @@ public abstract class MixinWorldGenBlockBlob implements BlockBlob {
     private VariableAmount radius;
     private VariableAmount count;
 
-    @Inject(method = "<init>(Lnet/minecraft/block/Block;I)V", at = @At("RETURN") )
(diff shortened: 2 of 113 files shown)


Aaron1011:
Pinging @SpongePowered/developers for review.

src/main/java/org/spongepowered/common/mixin/core/world/gen/populators/MixinWorldGenBigMushroom.java
```diff
@@ -67,7 +67,7 @@
     private Function<Location<Chunk>, PopulatorObject> override = null;
     private VariableAmount mushroomsPerChunk;
 
-    @Inject(method = "<init>", at = @At("RETURN") )
+    @Inject(method = "<init>", at = @At("RETURN") , require = 1)
```
kashike: Formatting is off

src/main/java/org/spongepowered/common/mixin/core/world/gen/populators/MixinWorldGenBlockBlob.java
```diff
@@ -54,7 +54,7 @@
     private VariableAmount radius;
     private VariableAmount count;
 
-    @Inject(method = "<init>(Lnet/minecraft/block/Block;I)V", at = @At("RETURN") )
+    @Inject(method = "<init>(Lnet/minecraft/block/Block;I)V", at = @At("RETURN") , require = 1)
```
kashike: Formatting is off

src/main/java/org/spongepowered/common/mixin/core/world/gen/populators/MixinWorldGenCactus.java
```diff
@@ -49,7 +49,7 @@
     private VariableAmount cactiPerChunk;
     private VariableAmount height;
 
-    @Inject(method = "<init>()V", at = @At("RETURN") )
+    @Inject(method = "<init>()V", at = @At("RETURN") , require = 1)
```
kashike: Formatting is off

src/main/java/org/spongepowered/common/mixin/core/world/gen/populators/MixinWorldGenDeadBush.java
```diff
@@ -47,7 +47,7 @@
 
     private VariableAmount count;
 
-    @Inject(method = "<init>", at = @At("RETURN") )
+    @Inject(method = "<init>", at = @At("RETURN") , require = 1)
```
kashike: Formatting is off

kashike:
Formatting is off on a lot of injections

PR #422, opened 2016-01-12
Make sure the tag compound is added back to the itemstack on creation

clienthax:


src/main/java/org/spongepowered/common/item/inventory/SpongeItemStackSnapshot.java
```diff
@@ -141,7 +141,9 @@ public int getCount() {
 
     @Override
     public ItemStack createStack() {
-        return this.privateStack.copy();
+        net.minecraft.item.ItemStack itemStack = (net.minecraft.item.ItemStack) this.privateStack.copy();
```
Mumfrey: Ought to be `ItemStackUtil.cloneDefensiveNative(this.privateStack)`

src/main/java/org/spongepowered/common/item/inventory/SpongeItemStackSnapshot.java
```diff
@@ -141,7 +141,9 @@ public int getCount() {
...
+        itemStack.setTagCompound(this.compound);
+        return (ItemStack) itemStack;
```
Mumfrey: Ought to be `ItemStackUtil.fromNative(itemStack)`

src/main/java/org/spongepowered/common/item/inventory/SpongeItemStackSnapshot.java
```diff
@@ -141,7 +141,9 @@ public int getCount() {
 
     @Override
     public ItemStack createStack() {
-        return this.privateStack.copy();
+        net.minecraft.item.ItemStack itemStack = (net.minecraft.item.ItemStack) this.privateStack.copy();
+        itemStack.setTagCompound(this.compound);
```
gabizou: If anything, the compound should be copied as well.

src/main/java/org/spongepowered/common/item/inventory/SpongeItemStackSnapshot.java
```diff
@@ -141,7 +142,11 @@ public int getCount() {
 
     @Override
     public ItemStack createStack() {
-        return this.privateStack.copy();
+        net.minecraft.item.ItemStack nativeStack = ItemStackUtil.cloneDefensiveNative(ItemStackUtil.toNative(this.privateStack.copy()));
```
Aaron1011: I don't think all of this is necessary, as `cloneDefensiveNative` calls `ItemStack#copy`

PR #473, opened 2016-02-02
Update for CompletableFuture changes

zml2008:
**[SpongeAPI PR](https://github.com/SpongePowered/SpongeAPI/pull/1058) | SpongeCommon PR**

Companion PR providing implementation


src/main/java/org/spongepowered/common/scheduler/SpongeScheduler.java
```diff
@@ -195,4 +200,7 @@ public ListeningExecutorService getListeningExecService() {
         return this.listeningExec;
```
simon816: May as well remove `getListeningExecService()`

src/main/java/org/spongepowered/common/scheduler/SpongeScheduler.java
```diff
@@ -195,4 +200,7 @@ public ListeningExecutorService getListeningExecService() {
...
     }
 
+    public <T>CompletableFuture<T> submitAsyncTask(Callable<T> callable) {
```
Minecrell: There should be a space after the `<T>`

PR #475, opened 2016-02-02
Update for Command changes

windy1:
SpongeAPI PR: https://github.com/SpongePowered/SpongeAPI/pull/1060


src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
```diff
@@ -173,6 +153,11 @@ public boolean testPermission(CommandSource source) {
     }
 
     @Override
+    public boolean testSourceType(CommandSource source) {
+        return testPermission(source);
```
kashike: What? This should just be `true` in my opinion.

windy1: I agree, fixed.

windy1:
Closing to change target branch

PR #483, opened 2016-02-06
Text Representables

ST-DDT:
Implements: https://github.com/SpongePowered/SpongeAPI/pull/1063


src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinSpongeUser.java
```diff
@@ -100,4 +105,30 @@ public String toString() {
...
+
+    @Override
+    public Text toText() {
+        Optional<Player> player = getPlayer();
+        if (player.isPresent()) {
+            return player.get().toText();
+        }
+        Text.Builder builder;
+        Optional<DisplayNameData> optName = get(DisplayNameData.class);
+        // Can this be present at all?
```
simon816: I don't think it would be for a `User`, since User is just the player data and doesn't handle representation of the player

gabizou:
Closing as the API PR has been closed.

PR #498, opened 2016-02-10
Multi-world command fixes

### LINKED ISSUE #885: Gamerules not being respected in custom worlds
When creating a custom world, calling WorldProperties.setGameRule("doMobSpawning", "false"), still spawns mobs in the generated world.
Vanilla minecraft shares the GameRules between each world. This way, sponge has no control over gamerules in other worlds.

Minecrell:
This adds an (optional) mixin plugin that is enabled by default and will patch the Vanilla commands to use the world of the sender instead of always the main overworld which is the Vanilla behaviour. This is mostly to fix issues with command blocks and weird behaviour if they're not running in the main world but using the Vanilla commands.

I've put this in an optional mixin plugin because it can _theoretically_ break existing maps if there was for example a command block in the nether that was used to set the time on the main world. So if the command changes cause problems for someone, he can still always disable them in his config.

Fixes #885, Fixes https://github.com/SpongePowered/SpongeForge/issues/727


Mumfrey:
You could also add settings for each individual command, so that if a server owner finds a compatibility problem with a specific command they could say "enable all, except *x*", or "disable all, except *y*". Since plugins have power of veto over each mixin in a config, this would be pretty easy to add.

ryantheleach:
Whilst I like the idea, I think it covers up the bigger issue of multiworld, that is different aspects of the world need to be configurable for if they have their own scoreboard, gamerules, world borders etc, and when they should share or delegate to a primary world.

Zidane:
I don't want to accept this PR unless Mumfrey's suggestion is implemented since all or nothing will not do.

Minecrell:
@Mumfrey @Zidane Sure, that should be a simple thing to add.

Minecrell:
@Mumfrey @Zidane Would you choose a list or rather a map for the configuration? E.g. there are two ways to add this to the configuration, not sure which one is better:

```
commands {
    multi-world-patch = [
        "worldborder",
        "seed",
        "..."
    ]
}
```
```
commands {
    multi-world-patch {
        "worldborder": false
        "seed": true
        "...": ...
    }
}
```
The second one might be better because you always see what can be enabled and disabled, though the first version would be a bit easier to implement (because it would be basically a simple `contains` check)

Mumfrey:
Maybe
```json
commands {
    multi-world-patch = {
        "include": [
            "*"
        ],
        "exclude": [
            "foo", "bar"
        ]
    }
}
```
giving the option for
```json
commands {
    multi-world-patch = {
        "include": [
            "foo", "bar"
        ]
    }
}
```

Minecrell:
@Mumfrey The issue I'd see with that is that the user doesn't see directly which commands are enablable or disablable. With a simple true/false map it would always stay updated and he would see which commands he can enable or disable.

Mumfrey:
But the user has to create this setting in the first place no? So he/she needs to go look up what's available at that point?

Minecrell:
@Mumfrey No, the settings would basically exist (enabled) in the configuration already and he could just set them to `false` instead of `true`. That would require a bit of manual work in the configuration loading as each command essentially need to be added manually, so basically the question is just if the _user should look which commands can be enabled or disabled_ (if so, where? our docs?), or if you need to be careful and add each command manually when making a mixin.

Mumfrey:
Oh, well in that case what stops the mixin plugin creating the setting if it doesn't exist as each query is performed? That way there's no need to manually add anything?

Minecrell:
Well you would need to save the configuration immediately after each call of the `shouldApplyMixin` method (if the command doesn't exist yet), so first run it would be saved like 7 times - not sure if that would be a good idea. (And if you don't save it again directly, when would you save it?)

Mumfrey:
If someone's computer can't handle 7 disk writes of a small text file then they probably can't run minecraft either so I would say the point is kind of moot.

Minecrell:
In that case this should do the job.

Mumfrey:
Urelated, but I noticed from your commit message there's some terminology mismatch: FYI what you're adding is a *mixin set*, the *plugin* is the class which (optionally) accompanies the set.

Minecrell:
Fixed, thanks =)

src/main/java/org/spongepowered/common/mixin/command/MixinCommandToggleDownfall.java
```diff
@@ -0,0 +1,44 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+@Mixin(CommandToggleDownfall.class)
+public abstract class MixinCommandToggleDownfall {
+
+    @Redirect(method = "processCommand", at = @At(value = "INVOKE",
+            target = "Lnet/minecraft/command/CommandToggleDownfall;toggleDownfall()V"))
+    private void onToggleDownfall(CommandToggleDownfall this$0, ICommandSender sender, String[] args) {
```
gabizou: do we really need to use `this$0` as a method argument?

stephan-gh: It's an unneeded parameter, because the context is already available as the current instance, so it could be any name. If I remember correctly we had a weird name that was normally used for these kind of parameters, couldn't remember it correctly anymore though :P

Just tell me how to name it and I'll change it, I don't care because it is never used anyway.

Mumfrey: I recommend people use `self` in this instance.

stephan-gh: Fixed :)

src/main/resources/mixins.common.command.json
```diff
@@ -0,0 +1,17 @@
+{
+    "minVersion": "0.5.1",
+    "package": "org.spongepowered.common.mixin.command",
+    "refmap": "mixins.common.refmap.json",
+    "plugin": "org.spongepowered.common.mixin.plugin.CommandPlugin",
```
kashike: This plugin name (and the mixin config name) should reflect that this is for "multi world command patches", not commands in general.

src/main/java/org/spongepowered/common/config/SpongeConfig.java
```diff
@@ -399,9 +399,18 @@ public TimingsCategory getTimings() {
         @Setting(comment = "A mapping from unqualified command alias to plugin id of the plugin that should handle a certain command")
         private Map<String, String> aliases = new HashMap<>();
 
+        @Setting(value = "multi-world-patches", comment = "Patches the specified commands to respect the world of the sender instead of applying the "
+                + "changes on the main overworld.")
+        private Map<String, Boolean> patches = new HashMap<>();
```
kashike: I'd recommend using `multi-world-command-patches`, which means this field and the associated getter should probably be named `multiWorldCommandPatches`/`getMultiWorldCommandPatches` to match.

stephan-gh: This is in the command category, so I guess that should be already quite clear. I can rename the getter though.

jk-5:
There are a few more cases not related to commands you might also want to patch

 * `EntityPlayer.sendCommandFeedback`
 * `CommandBlockLogic.sendCommandFeedback`
 * `MinecraftServer.sendCommandFeedback`
 * `ServerConfigurationManager.canSendCommands`
 * `ServerConfigurationManager.updateTimeAndWeatherForPlayer`

src/main/java/org/spongepowered/common/mixin/command/MixinCommandTime.java
```diff
@@ -0,0 +1,55 @@
...
+@Mixin(CommandTime.class)
+public abstract class MixinCommandTime {
+
+    /**
+     * @author Minecrell
+     * @reason Modify time only in the command sender's world
+     */
+    @Overwrite
+    protected void setTime(ICommandSender sender, int time) {
+        sender.getEntityWorld().setWorldTime(time);
```
simon816: Vanilla behavior here is to set the time in all worlds (rather than just overworld) which I think kind of makes sense in this case

kashike: I disagree - if I set the time to `day` in world `A`, I don't want it being changed in all other worlds as well.

Zidane: @simon816 

This PR is for making the commands respect MULTI-WORLD abilities. Time can be a part of this.

src/main/java/org/spongepowered/common/mixin/command/multiworld/MixinCommandToggleDownfall.java
```diff
@@ -0,0 +1,46 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+@Mixin(CommandToggleDownfall.class)
+public abstract class MixinCommandToggleDownfall {
+
+    @Redirect(method = "execute", at = @At(value = "INVOKE",
+            target = "Lnet/minecraft/command/CommandToggleDownfall;toggleRainfall(Lnet/minecraft/server/MinecraftServer;)V"))
+    private void onToggleDownfall(CommandToggleDownfall self, MinecraftServer server, MinecraftServer server2, ICommandSender sender, String[] args) {
+        System.out.println("LOL" + sender.getEntityWorld().getWorldInfo());
```
Zidane: LOL

src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
```diff
@@ -540,6 +543,11 @@ public EntityPlayerMP recreatePlayerEntity(EntityPlayerMP entityPlayerMP, int ta
         this.playerEntityList.add(entityPlayerMP);
         entityPlayerMP.addSelfToInternalCraftingInventory();
 
+        // Update reducedDebugInfo game rule
+        entityPlayerMP.connection.sendPacket(new SPacketEntityStatus(entityPlayerMP,
+                worldServer.getGameRules().getBoolean(DefaultGameRules.REDUCED_DEBUG_INFO) ? (byte) 22 : 23));
+        System.out.println("SENT");
```
Zidane: SENT

src/main/java/org/spongepowered/common/mixin/command/multiworld/MixinCommandSeed.java
```diff
@@ -0,0 +1,43 @@
...
+import net.minecraft.command.CommandShowSeed;
+import net.minecraft.command.ICommandSender;
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.world.World;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+@Mixin(CommandShowSeed.class)
+public abstract class MixinCommandSeed {
```
kashike: Should be noted that this shouldn't be renamed to `MixinCommandShowSeed` because of `org.spongepowered.common.mixin.plugin.MultiWorldCommandPlugin#shouldApplyMixin`.

src/main/java/org/spongepowered/common/mixin/command/vanilla/MixinCommandGameRule.java
```diff
@@ -0,0 +1,50 @@
...
+public abstract class MixinCommandGameRule {
+
+    /**
+     * @author Minecrell - September 28th, 2016
+     * @reason Manually apply game rule to all because Sponge uses separate world info
+     */
+    @Redirect(method = "execute", at = @At(value = "INVOKE",
+            target = "Lnet/minecraft/world/GameRules;setOrCreateGameRule(Ljava/lang/String;Ljava/lang/String;)V"))
+    private void onSetOrCreateGameRule(GameRules gameRules, String key, String value) {
+        for (WorldServer server : SpongeImpl.getServer().worldServers) {
```
kashike: `WorldServer` is a `world`, not a `server`.

src/main/java/org/spongepowered/common/mixin/command/multiworld/MixinCommandToggleDownfall.java
stephan-gh: LOL, fixed!

src/main/java/org/spongepowered/common/mixin/command/vanilla/MixinCommandGameRule.java
stephan-gh: Thanks, fixed

src/main/java/org/spongepowered/common/mixin/command/multiworld/MixinCommandSeed.java
stephan-gh: Fixed

src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
stephan-gh: Fix SENT to GitHub

src/main/java/org/spongepowered/common/mixin/command/multiworld/MixinCommandSeed.java
```diff
@@ -0,0 +1,46 @@
...
+
+import net.minecraft.command.CommandShowSeed;
+import net.minecraft.command.ICommandSender;
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.world.World;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+// Note: The mixin class name is used to match the mixin to the command.
```
Zidane: Can you explain this part further @Minecrell ?

src/main/java/org/spongepowered/common/mixin/command/vanilla/MixinCommandWorldBorder.java
```diff
@@ -0,0 +1,79 @@
...
+import net.minecraft.command.CommandWorldBorder;
+import net.minecraft.world.WorldServer;
+import net.minecraft.world.border.WorldBorder;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+import org.spongepowered.common.SpongeImpl;
+
+@Mixin(CommandWorldBorder.class)
+public abstract class MixinCommandWorldBorder {
```
Zidane, line 37: @Minecrell 

Make sure that my hack to fix the WorldBorder being multiplied by 6 for `WorldProviderHell` worlds is still active and working. It was at the packet level.

src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
```diff
@@ -540,6 +543,10 @@ public EntityPlayerMP recreatePlayerEntity(EntityPlayerMP entityPlayerMP, int ta
         this.playerEntityList.add(entityPlayerMP);
         entityPlayerMP.addSelfToInternalCraftingInventory();
 
+        // Update reducedDebugInfo game rule
```
Zidane, line 549: Was this missed in the update to 1.9/1.10?

stephan-gh: Probably not. It's not necessary in Vanilla because all worlds use the same game rules. On Sponge however, the worlds can have different settings for `reducedDebugInfo`.

src/main/java/org/spongepowered/common/mixin/command/multiworld/MixinCommandSeed.java
stephan-gh: Basically what @kashike mentioned above. It will strip the `MixinCommand` from the class name, convert the suffix to lower case and use that to match the mixin to the option in the configuration. That's why it must be equal to the name of the command (instead of equal to the name of the MCP class).

src/main/java/org/spongepowered/common/mixin/plugin/MultiWorldCommandPlugin.java
```diff
@@ -0,0 +1,84 @@
...
+        return null;
+    }
+
+    @Override
+    public boolean shouldApplyMixin(String targetClassName, String mixinClassName) {
+        final String name = StringUtils.substringAfter(mixinClassName, MIXIN_PREFIX).toLowerCase();
+
+        Map<String, Boolean> patches = SpongeImpl.getGlobalConfig().getConfig().getCommands().getMultiWorldCommandPatches();
+        if (!patches.containsKey(name)) {
+            patches.put(name, false);
```
Zidane: Shouldn't this be true? By default, multi-world commands should be true.

stephan-gh: Weren't we going to keep Vanilla behaviour by default?

Zidane: I'm thinking we'll need to put this one to a vote.

@SpongePowered/developers

src/main/java/org/spongepowered/common/mixin/command/multiworld/MixinCommandSeed.java
Zidane: @Minecrell 

I'd rather the config show "ShowSeed" as now I would have to special case doing a class search in my IDE to find this particular mixin. We keep this naming style for a very important reason.

If you guys do not like the name, fix it in 1.11 MCP :p.

src/main/java/org/spongepowered/common/mixin/command/vanilla/MixinCommandWorldBorder.java
stephan-gh: I'm a bit confused what the hack is supposed to do. It makes sense because normally `WorldProviderHell` converts the world border center to nether coordinates (it divides the coordinates by 8 and you multiply them again with 8 in the packet). But on the other hand, you make [`WorldProviderHell` return a normal world border](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldProviderHell.java#L35-L45)  which returns the normal coordinates.

In my tests, Vanilla always sends the overworld coordinates even to players in the nether (because it will always send the overworld world border). So what is the hack on the packet class supposed to do?

src/main/java/org/spongepowered/common/mixin/command/multiworld/MixinCommandSeed.java
stephan-gh: Fixed. Since the seed command class is the only class with that problem I've added a simple if check in the Mixin plugin that will change the option from `showseed` to `seed`.

src/main/java/org/spongepowered/common/mixin/command/vanilla/MixinCommandWorldBorder.java
Zidane: Its cause the server sends the correct coordinates down (not ones based on Nether's scaled by 8 system) and then the client's WorldProviderHell takes it and divides it by 8. We lie to the client so the WorldBorder works correctly client-side.

Now thinking about things, you are correct that I do return a `new WorldBorder()` which is meant for the `WorldProviderHell`. With that known, we don't want to apply my packet hack when SF is installed on the client as the coordinates will now be off by 8 (as WorldProviderHell will return a non-adjusting WorldBorder for the Server and Client World).

Can you address this @Minecrell ? Test it first to see.

stephan-gh: If I understand you correctly, the correct fix in this situation would be to patch the SF client to divide the coordinates by 8 and remove the packet hack from the server side, because otherwise it would send the wrong coordinates from the server?

Zidane: Correct fix would be to not manipulate the coordinates being sent when SF is installed and this is client. So basically you'll find that here:

https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java#L769

This shouldn't be applied if its client as well as we shouldn't mix into SPacketWorldBorder at all if client (as client will have a normal WorldBorder in WorldProviderHell).

stephan-gh: The packet hack should be never applied when sending the packet. If `WorldProviderHell` is using a normal world border then the coordinates are already correctly in the overworld coordinate system which is how the Vanilla client will accept them.

The only thing that is not working correctly without a packet hack is the SpongeForge client. Normally, it should scale the coordinates to the nether system locally for nether worlds, however this is not happening currently because we also modify `WorldProviderHell` on the client.

So if the above is correct, the SF client must manually convert the world border center to the nether system and the server doesn't need to do anything additionally.

Zidane: @Minecrell 

Not correct. The packet hack is needed for the Vanilla client else the numbers you set via worldborder command are correct on the server (as the server has the normal WorldBorder) but NOT correct on the client (as it is doing the scaling by 8).

When SF is installed and we're on the client, both client and server have a normal WorldBorder so no coordinate adjustments are needed.

src/main/java/org/spongepowered/common/mixin/command/multiworld/MixinCommandTime.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * @author Minecrell
+     * @reason Modify time only in the command sender's world
+     */
+    @Redirect(method = "execute", at = @At(value = "INVOKE", target = "Lnet/minecraft/command/CommandTime;incrementAllWorldTimes"
+            + "(Lnet/minecraft/server/MinecraftServer;I)V"))
+    private void onIncrementTime(CommandTime self, MinecraftServer server, int time, MinecraftServer server2, ICommandSender sender, String[] args) {
+        World world = sender.getEntityWorld();
```
Zidane: No reason to separate this.

src/main/java/org/spongepowered/common/mixin/command/multiworld/MixinCommandToggleDownfall.java
```diff
@@ -0,0 +1,45 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+@Mixin(CommandToggleDownfall.class)
+public abstract class MixinCommandToggleDownfall {
+
+    @Redirect(method = "execute", at = @At(value = "INVOKE",
+            target = "Lnet/minecraft/command/CommandToggleDownfall;toggleRainfall(Lnet/minecraft/server/MinecraftServer;)V"))
+    private void onToggleDownfall(CommandToggleDownfall self, MinecraftServer server, MinecraftServer server2, ICommandSender sender, String[] args) {
+        WorldInfo worldInfo = sender.getEntityWorld().getWorldInfo();
```
Zidane: No reason to separate this (I like smaller diffs)

src/main/resources/mixins.common.multi-world-command.json
```diff
@@ -0,0 +1,22 @@
...
+    "package": "org.spongepowered.common.mixin.command.multiworld",
+    "refmap": "mixins.common.refmap.json",
+    "plugin": "org.spongepowered.common.mixin.plugin.MultiWorldCommandPlugin",
+    "target": "@env(DEFAULT)",
+    "compatibilityLevel": "JAVA_8",
+    "mixins": [
+        "MixinCommandDefaultGameMode",
+        "MixinCommandDifficulty",
+        "MixinCommandGameRule",
+        "MixinCommandSeed",
```
Zidane: MixinCommandShowSeed

src/main/java/org/spongepowered/common/mixin/plugin/MultiWorldCommandPlugin.java
ryantheleach: Regardless of the decision, should a message be output to the command sender alerting them to which option is active, and how to swap between them?

src/main/resources/mixins.common.multi-world-command.json
stephan-gh: Fixed

src/main/java/org/spongepowered/common/mixin/plugin/MultiWorldCommandPlugin.java
Zidane: No, that would be incredibly spammy. Just needs to go on docs page.

Minecrell:
@Zidane 
Updated Mixin library and reverted back to the Mixin with multiple targets. Should be ready to merge now.

PR #502, opened 2016-02-12
Cause Tracking Refactor Phase 2: States and Phases

gabizou:
**SpongeCommon** | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/538)

Overall, Cause tracking is amazing for what it does, but it's currently delicate and difficult to manage sanely. What this branch aims to do is transform `CauseTracking` into a more state-like machine where processing is performed through the state, and validation is performed otherwise. What this intends to do is eliminate the likes of: `StaticMixinHelper` and other static states that are currently fragile.

This is Phase 2 of the `CauseTracking` refactor involves moving towards a stack based processing handling for events and causes.

Todo:
- [x] Step 1 - Introduce Stack
 - [x] Write Stack based system for handling transitions between phase
- [x] Step 2 - Introduce Phases
  - [x] Flesh out phases to properly represent the current state of the game
- [x] Step 3 - Enter and Exit Phases
 - [x] Entering and exiting phases occurs in correct order (with all side-effects of each phase being tracked appropriately)
- [x] Step 4 - Validate Phases
  - [x] Handle processing of events per phase and state
  - [x] Validate that a phase transition does not take place where it shouldn't (basically, ensure that the phases are properly being exited and "unwound")
  - [x] Validate that upon completing a "dominant" phase that at the end of that dominant phase, the entire stack is empty by termination, reducing any bleeding into the next phase in the state machine
 - [x] Optional Owner/Notifier integration to be customizable where plugins reacting based on event listeners would be reduced to having to only include the necessary `NamedCause` when spawning/setting blocks.
 - [x] Pass around any lists/maps necessary for tracking to avoid storing said lists/maps on the `CauseTracker`. Recommended to store said information in `PhaseContext`.

- [ ] Step 5 Vanilla Testing
 - [x] Test that multiple phases do not irresponsibly associate different phases incorrectly
 - [ ] Test that plugin assumptions of causes are still valid to retain contractual obligations of what CauseTracking is supposed to do
 - [ ] Expand validation to listener based side effects (plugins spawning/changing blocks during tick managed events)
- [ ] Step 6 Mod Testing
 - [ ] Test validation with mods


kinggoesgaming:
@gabizou will this be breaking the plugins using the current Cause system?

Mumfrey:
@kinggoesgaming this is a **refactor** as it says in the title.

src/main/java/org/spongepowered/common/mixin/tracking/world/MixinChunk_Tracker.java
```diff
@@ -0,0 +1,294 @@
...
+
+    @Shadow @Final private World worldObj;
+    @Shadow @Final public int xPosition;
+    @Shadow @Final public int zPosition;
+    @Shadow @Final private ExtendedBlockStorage[] storageArrays;
+    @Shadow @Final private int[] precipitationHeightMap;
+    @Shadow @Final private int[] heightMap;
+    @Shadow private boolean isModified;
+
+    public Map<Integer, PlayerTracker> trackedIntBlockPositions = Maps.newHashMap();
```
Minecrell: You could use primitive maps from Trove here

Zidane:
http://imgur.com/kFwe07J

Printout from IRC user Thutmose from a large explosion

src/main/java/org/spongepowered/common/event/EventConsumer.java
```diff
@@ -0,0 +1,128 @@
...
+    private final E event;
+
+    private EventConsumer(Builder<E> builder) {
+        this.cancelledConsumer = builder.cancelledConsumer;
+        this.nonCancelledConsumer = builder.nonCancelledConsumer;
+        this.postConsumer = builder.postConsumer;
+        this.event = builder.event;
+    }
+
+    public static final class Builder<E extends Event> {
```
Minecrell: Is there really a need for a builder of this? Couldn't you just first require an event, and then make the other fields modifyable directly? From looking at the usages I'd say they're only used temporarily, so I looks kind of redundant to construct everything with a builder first.

gabizou: It was just a quick write up, making things immutable and such. Overall, yes it could be cutting corners by directly having setters on the actual object.

src/main/java/org/spongepowered/common/mixin/tostring/world/MixinChunk_ToString.java
```diff
@@ -0,0 +1,48 @@
...
+public abstract class MixinChunk_ToString {
+
+    @Shadow @Final private World worldObj;
+    @Shadow @Final public int xPosition;
+    @Shadow @Final public int zPosition;
+
+    @Override
+    public String toString() {
+        return Objects.toStringHelper(this)
+                .add("World:", this.worldObj)
```
kashike: Should just be `world`., `ToStringHelper` automatically formats.

src/main/java/org/spongepowered/common/mixin/tostring/world/MixinWorldServer_ToString.java
```diff
@@ -0,0 +1,43 @@
...
+import org.spongepowered.common.mixin.core.world.MixinWorld;
+
+@Mixin(WorldServer.class)
+public abstract class MixinWorldServer_ToString extends MixinWorld {
+
+    @Override
+    public String toString() {
+        return Objects.toStringHelper(this)
+                .add("LevelName", this.worldInfo.getWorldName())
+                .add("DimensionId:", this.provider.getDimensionId())
```
kashike: Should just be `DimensionId`, `ToStringHelper` automatically formats.

bloodmc:
Just did some basic testing and 2 things i noticed so far testing on the following map

http://www.planetminecraft.com/project/some-redstone-creations/download/worldsave/

1. Some blocks getting their owner data set when they should have it blank

Screenshot http://i.imgur.com/zipcJlu.png

2. I'm noticing more lag as I roam. You can notice it when you spawn into the world for the first time and begin to roam, chunks will load slow. 

I'll do some more testing soon.

gabizou:
Things that don't work currently:
- [ ] Player item drops from pressing Q
- [ ] Player death drops don't actually drop, and the inventory is filled with bad items.
- [ ] Commands summoning entities don't work

To reiterate: I'd like for public testing to take place, and if you can, please report bugs to this PR, but above all, make sure the bug was either already reported or already fixed in a newer commit.

gabizou:
Just an update: Some bugs to report:

- [ ] If cancelling a block break from a player takes place, the item drops are still performed. Some additional thoughts to this are that item drops are not tracked per block position, so the items dropped per block break is not physically captured. This likely needs some extra consideration and phase plumbing.

misterfixit:
When attempting to power a minecart furnace it locks up but does play the animation.
1. Expected outcome- Minecart furnace once given coal/charcoal should go along the tracks in the direction it faces
2. Actual outcome- Minecart plays the animation as if powered but doesn't move at all and cannot be pushed by the player

gabizou:
Closing as this is going to re-target the 1.9 branch.

PR #519, opened 2016-02-19
Implement Getter event filter

Deamon5550:
This PR implements https://github.com/SpongePowered/SpongeAPI/pull/1084

See that PR for a description of this functionality.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/filter/delegate/GetterFilterSourceDelegate.java
+++ b/src/main/java/org/spongepowered/common/event/filter/delegate/GetterFilterSourceDelegate.java
@@ -28,11 +28,15 @@ import static org.objectweb.asm.Opcodes.ACONST_NULL;
 import static org.objectweb.asm.Opcodes.ALOAD;
 import static org.objectweb.asm.Opcodes.ARETURN;
 import static org.objectweb.asm.Opcodes.ASTORE;
+import static org.objectweb.asm.Opcodes.CHECKCAST;
+import static org.objectweb.asm.Opcodes.DUP;
 import static org.objectweb.asm.Opcodes.IFEQ;
 import static org.objectweb.asm.Opcodes.IFNE;
+import static org.objectweb.asm.Opcodes.ILOAD;
 import static org.objectweb.asm.Opcodes.INSTANCEOF;
 import static org.objectweb.asm.Opcodes.INVOKEINTERFACE;
 import static org.objectweb.asm.Opcodes.INVOKEVIRTUAL;
+import static org.objectweb.asm.Opcodes.ISTORE;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/event/filter/delegate/GetterFilterSourceDelegate.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public Tuple<Integer, Integer> write(ClassWriter cw, MethodVisitor mv, Method method, Parameter param, int local) {
+        Class<?> targetType = param.getType();
+        Class<?> eventClass = method.getParameterTypes()[0];
+        String targetMethod = this.anno.value();
+        Method targetMethodObj = null;
+        for (Method mth : eventClass.getMethods()) {
+            if (targetMethod.equals(mth.getName())) {
+                if (targetMethodObj != null) {
+                    throw new IllegalArgumentException("Multiple matches found for getter " + targetMethod + " in type " + eventClass.getName());
```
gabizou: I'm not sure how there can be multiple choices provided considering the event interfaces have to compile according to the JDK's discretion (which prevents ambiguous methods from compiling, but the JVM will allow ambiguous methods to exist), correct me if I'm wrong here.

Deamon5550: if its not actually a getter and takes parameters.

gabizou: Especially considering the targeted method should be discerned not only from the method name but also the return type provided by the type info of the variable provided with the annotation, much like how Mixins does for injections etc.

gabizou: > if its not actually a getter and takes parameters.

So in what case is there going to be multiple target methods found where this exception is thrown?

src/main/java/org/spongepowered/common/event/filter/delegate/GetterFilterSourceDelegate.java
```diff
@@ -0,0 +1,110 @@
...
+                }
+                if (mth.getParameterCount() != 0) {
+                    throw new IllegalArgumentException(
+                            "Method " + mth.toGenericString() + " specified by getter annotation has non-zero parameter count");
+                }
+                if (!mth.getReturnType().equals(Optional.class) && !mth.getReturnType().equals(targetType)) {
```
gabizou: What if instead of the specific `equals` for the target type, perhaps where the target type is checked to be `isAssignableFrom`? This would allow subclass checks as well.

PR #525, opened 2016-02-20
[WIP] Localization Impl

windy1:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1088) | [**Common**](https://github.com/SpongePowered/SpongeCommon/pull/525) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/549)


src/main/java/org/spongepowered/common/locale/SpongeDictionary.java
```diff
@@ -0,0 +1,62 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.util.Locale;
+
+public class SpongeDictionary extends ConfigDictionarySingle implements ResourceDictionary {
+
+    private final File file;
```
Minecrell: Why do you use `File` instead of `Path` here?

src/main/java/org/spongepowered/common/config/SpongeConfig.java
```diff
@@ -457,6 +459,8 @@ public void setDumpThreadsOnWarn(boolean dumpThreadsOnWarn) {
         @Setting(value = GENERAL_CHUNK_LOAD_OVERRIDE,
                 comment = "Forces Chunk Loading on provide requests (speedup for mods that don't check if a chunk is loaded)")
         private boolean chunkLoadOverride = false;
+        @Setting(value = GENERAL_LOCALE, comment = "IEFT localization string (e.g. \"en_US\")")
+        private String locale = "en-US";
```
kashike: Example has an underscore, this doesn't - most people working with Minecraft translations are used to underscores, from what I've seen.

src/main/java/org/spongepowered/common/config/SpongeConfig.java
```diff
@@ -473,6 +477,14 @@ public boolean chunkLoadOverride() {
         public void setChunkLoadOverride(boolean chunkLoadOverride) {
             this.chunkLoadOverride = chunkLoadOverride;
         }
+
+        public Locale getLocale() {
+            return Locale.forLanguageTag(this.locale);
```
kashike: `return LanguageUtil.LOCALE_CACHE.getUnchecked(this.locale);`

src/main/java/org/spongepowered/common/locale/SpongeGameDictionary.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.spongepowered.common.SpongeImpl;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class SpongeGameDictionary extends SpongeDictionary {
+
+    public static final Path FILE_NAME = Paths.get("dict.conf");
```
kashike: Does this need to be public?

src/main/java/org/spongepowered/common/plugin/SpongePluginContainer.java
```diff
@@ -41,6 +44,11 @@ protected SpongePluginContainer() {
     }
 
     @Override
+    public ServiceManager getServiceManager() {
+        return serviceManager;
```
kashike: `this.`

src/main/java/org/spongepowered/common/locale/SpongeDictionary.java
windy1: The way SpongeCommon is handling the implementation right now is by checking if 'dict.conf' exists in the config directory and falling back to the jar resource if it doesn't. We need a file object here to check if it exists when `getPath` is called.

src/main/java/org/spongepowered/common/locale/SpongeGameDictionary.java
windy1: Probably not, I could see it being useful to external classes in the future though.

src/main/java/org/spongepowered/common/locale/SpongeDictionary.java
Minecrell: Why do you need a `File` object oc check if it exists? There is `Files.exists(Path)`

windy1: Ah thanks, I was not familiar with that function.

src/main/java/org/spongepowered/common/config/SpongeConfig.java
windy1: Java `Locale.fromLanguageTags` accepts hyphens only. I'll fix the comment.

src/main/java/org/spongepowered/common/locale/SpongeDictionary.java
```diff
@@ -0,0 +1,62 @@
...
+        if (this.file.exists()) {
+            return this.file.toPath();
+        }
+        return this.file.toPath().getFileName();
+    }
+
+    @Override
+    public InputStream getSource(Locale locale) throws IOException {
+        if (this.file.exists()) {
+            return Files.newInputStream(file.toPath(), StandardOpenOption.READ);
```
Minecrell: The `StandardOpenOption.READ` is redundant here, this is the default when calling `Files.newInputStream`

src/main/java/org/spongepowered/common/config/SpongeConfig.java
windy1: Why should we restrict it to MC locales? This is all done on the server so there's really no need for that.

kashike: It's not restricted to Minecraft locales. All the LOCALE_CACHE does is cache the `String` to the `Locale`.

src/main/java/org/spongepowered/common/config/SpongeConfig.java
kashike: https://github.com/SpongePowered/SpongeCommon/pull/525#discussion_r53553103

src/main/java/org/spongepowered/common/locale/SpongeDictionary.java
windy1: Eh, actually we would need a File object to join the Paths anyways so using `Files.exists(Path)` might actually end up being more redundant.

src/main/java/org/spongepowered/common/config/SpongeConfig.java
```diff
@@ -473,6 +477,14 @@ public boolean chunkLoadOverride() {
...
+        }
+
+        public void setLocale(Locale locale) {
+            this.locale = locale.toLanguageTag();
```
zml2008: Locale#toString provides the standard locale representation

src/main/java/org/spongepowered/common/locale/SpongeDictionary.java
Minecrell: No, you should join paths using `Path.resolve(Path)`

windy1: That's just going to end up being more verbose though because we'll have to do it in both `getPath` and `getSource` IMO it's cleaner just to store the `File` object.

Minecrell: If you reach a point where you need the `File` object at all you're either using an older API or just not doing it correctly. The new API has almost everything `File` has. Also I'm not sure why you need to do it in `getPath` and `getSource`? 

```java
this.path = parent.resolve(path);

// ...
public Path getPath(Locale locale) {
    if (Files.exists(this.path)) {
    // ...
    // Remove all this.file.toPath() and replace with this.path
```

windy1: I need to keep the parent and child paths separate because the full path is not used for resources. We need to check in `getPath` to return the full path if the file exists, else return the child path and in `getSource` to resolve the `InputStream`.

src/main/resources/sponge.dict
```diff
@@ -0,0 +1,6 @@
+en_US {
+    test="Hello, world!"
+}
+en_GB {
+    test="'Ello world!"
```
Mumfrey: +1, as an englishman can confirm we do all indeed say *'ello*

Mumfrey:
I saw in the API that there's a `ResourceBundleDictionary` which I assume is for consuming dictionaries provided in the "default" minecraft impl format (or did I misunderstand that?) of resource -> properties file ending in `.lang` but don't see an implementation of it here.

Have I misinterpreted `ResourceBundleDictionary` or are we not going to provide support for reading from the default localisation files?

If the latter, then can I make a case for doing so on the basis of the existence of "hybrid" plugins (plugins which are also mods) it would make sense - I think - to be able to consume the existing localisations and/or *provide* our localisations into the mojang resource system.

Some info about the way localisations are stored in mc currently:

Resource bundles in mods are based at `assets/<domain>` within the mod ziptree (where `domain` is usually the modid), and localisations are stored in (properties) files named `assets/<domain>/lang/<locale>.lang`, however they should for preference be accessed via resource manager:

```java
ResourceLocation path = new ResourceLocation(domain, sprintf("lang/%s.lang", locale));
List<IResource> langResources = resourceManager.getAllResources(path);
```

Thoughts?

Zidane:
I agree with @Mumfrey here and would argue even further that we don't want to deviate on this. Our format should be .lang and the API should rest on top of the working Minecraft localization system.

kashike:
I disagree on forcing people to use Minecraft's language format - people should be allowed to use `ResourceBundle`-based dictionaries if they want to.

liach:
@kashike I think that the devs suggest that both ways would be available.

Zidane:
I guess I'd be fine with it @kashike so long as its an addon feature to the API available. This will not be merged without support for the .lang files that the base game and all mods utilize.

Mumfrey:
Yes just to clarify, I was just suggesting that existing lang bundles in mods should be usable, mainly so that hybrid mods don't have to duplicate all their localisations across two resources.

Zidane:
@windy1 

We should also add these language entries to the base game's localization cache...unless that is on the client only (but I don't think it is). It might be only used by the client though.

If so, do this in Common please.

windy1:
@Mumfrey @Zidane 

`ResourceBundleDictionary` is simply a wrapper around Java's `ResourceBundle`. The reason I don't think mixing Mojang's localization and this together is a good idea is because they are objectively achieving two different things.

This localization API is a way to allow plugin developers to define language for their plugin while also making it configurable to server administrators. If a plugin doesn't support some language, it can expose it's `.dict` file and allow server admins to localize the plugin themselves.

Resource pack translations are indeed done on the client (`{"translate": "some.minecraft.key"}`) the reason we don't want to translate this on the server is because it's *dependent on what language setting the client has*.

I think mixing these two systems will only muddy the waters in what can otherwise be achieved separately quiet easily.

Zidane:
@windy1 

What the client has its locale set to actually has little bearing on this; you can register all your language entries and the client will use what it has.

.lang files from Mojang also remain quite configurable because mods can easily load one in assets or outside, I think its highly attractive for us to support their translation keys.

If you add per-Player locales then its easy to use OUR API to lookup lang entries vs. having to use the internals of the base game. This is regardless of being a hybrid plugin or not.

windy1:
@Zidane I'm guessing you're referring to [forge's localization system](http://www.minecraftforge.net/wiki/Localizing). If you look closely you'll see this done entirely on the client and only adds new keys to MC's language file. Client-side translations are already supported in Sponge, but there is currently no way to provide server-side localization. This is what this PR aims to achieve. On the server, you *cannot* know the translation result of MC keys.

PR #540, opened 2016-02-25
Implement player tab list API

kashike:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1099) | [**Common**](https://github.com/SpongePowered/SpongeCommon/pull/540)

### DIFF
--- a/src/main/java/org/spongepowered/common/entity/player/tab/SpongeTabList.java
+++ b/src/main/java/org/spongepowered/common/entity/player/tab/SpongeTabList.java
@@ -1,0 +1,238 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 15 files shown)


src/main/java/org/spongepowered/common/entity/player/tab/SpongeTabListEntry.java
```diff
@@ -0,0 +1,151 @@
...
+
+    private SpongeTabList list;
+    private final GameProfile profile;
+    @Nullable private Text displayName;
+    private int latency;
+    private GameMode gameMode;
+    private boolean updateWithoutSend;
+
+    public SpongeTabListEntry(TabList list, GameProfile profile, @Nullable Text displayName, int latency, GameMode gameMode) {
+        checkState(list instanceof SpongeTabList, "list is not a SpongeTabList");
```
Minecrell: Not sure if this is necessary, I mean otherwise it will simply throw a class cast exception which has even a kind of more helpful error message.

src/main/java/org/spongepowered/common/entity/player/tab/SpongeTabList.java
```diff
@@ -0,0 +1,232 @@
...
+        this.header = header;
+        this.footer = footer;
+
+        this.refreshClientHeaderFooter();
+
+        return this;
+    }
+
+    private void refreshClientHeaderFooter() {
+        S47PacketPlayerListHeaderFooter packet = new S47PacketPlayerListHeaderFooter();
```
Minecrell: There is a constructor for the header, then you don't need an AT for the header field.

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -141,7 +143,43 @@ public int getPing() {
...
+     * This method wraps packets being sent to perform any additional actions,
+     * such as rewriting data in the packet.
+     *
+     * @param packetIn The original packet to be sent
+     * @return The rewritten packet if we performed any changes, or the original
+     *     packet if we did not perform any changes
+     */
+    private Packet rewritePacket(final Packet packetIn) {
+        //noinspection UnnecessaryLocalVariable
+        Packet packetOut = packetIn;
```
Minecrell: As the inspection says, what is this variable for?

src/main/java/org/spongepowered/common/entity/player/tab/SpongeTabList.java
```diff
@@ -0,0 +1,232 @@
...
+     * @param packet The packet to process
+     */
+    public void updateEntriesOnSend(S38PacketPlayerListItem packet) {
+        for (S38PacketPlayerListItem.AddPlayerData data : packet.players) {
+            if (packet.action == S38PacketPlayerListItem.Action.ADD_PLAYER) {
+                // If an entry with the same id exists nothing will be done
+                this.addEntry(data);
+            } else if (packet.action == S38PacketPlayerListItem.Action.REMOVE_PLAYER) {
+                this.removeEntry(data.getProfile().getId());
+            } else if (packet.action == S38PacketPlayerListItem.Action.UPDATE_GAME_MODE
```
Aaron1011: You shouldn't need this, since these are the only other options.

PR #543, opened 2016-02-27
Do not use the works of PermissionService

rexbut:
My function "hasPermission(String permission)" doesn't use "getActiveContexts()"

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/command/MixinSubject.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/command/MixinSubject.java
@@ -138,7 +138,8 @@ public abstract class MixinSubject implements Subject, IMixinCommandSource, IMix
 
     @Override
     public boolean hasPermission(String permission) {
-        return hasPermission(getActiveContexts(), permission);
+        Subject subj = internalSubject();
+        return subj == null ? this.permDefault(permission).asBoolean() : subj.hasPermission(permission);
     }
 
     @Override


liach:
Reason?

rexbut:
I want to use my own function "hasPermission(String permission)" and not "hasPermission(getActiveContexts(), permission)"

ryantheleach:
You are creating a pull request which contributes your code back to the main SpongeCommon repository.

The changes as-is would stop people who wish to use contexts from doing so.

Instead I recommend creating a PermissionsService plugin that ignores the current context, if you wish for permissions to be context-less.

me4502:
@ryantheleach has explained my reasons for closing well enough.

If you meant to do something else with this PR, make another comment and I'll happily reopen it.

rexbut:
My "PermissionService" I useful in some circumstances " getActiveContexts ()" . But not for "hasPermission(String permission) ".

I don't understand why this is the only feature that automatically adds "getActiveContexts()". And so have changed for the " PermissionService "

me4502:
The problem with what you've done here, is that it completely removes the concept of getActiveContexts, which is used by numerous people.

I'm not the permissions person though, @zml2008 may have more to add.

rexbut:
"MixinSubject" allows you to perform the duties of " PermissionService " So why it's impossible to use the implementation of the "hasPermission (String permission )" of "PermissionService"

rexbut:
All "PermissionService" must implement the "hasPermission (String permission)" I only wish that perform the function is that of the "PermissionService"

PermissionsEx : https://github.com/PEXPlugins/PermissionsEx/blob/master/permissionsex-sponge/src/main/java/ninja/leaping/permissionsex/sponge/PEXSubject.java

`@Override
    public boolean hasPermission(String permission) {
        return hasPermission(getActiveContexts(), permission);
    }
`

SpongeSubject : https://github.com/SpongePowered/SpongeCommon/blob/4e86b0b42e9957dca09a124920dc4ffe2ab17b6d/src/main/java/org/spongepowered/common/service/permission/base/SpongeSubject.java

`@Override
    public boolean hasPermission(String permission) {
        return hasPermission(getActiveContexts(), permission);
    }`

rexbut:
> @ryantheleach : Also I think it would go into a never ending loop and create a stack overflow, as you are calling the hasPermission method from inside the hasPermission method.

Not because I just wish that the " hasPermission(String permission)" redirects to the "hasPermission (String permission)" by the implementer "PermissionService"

zml2008:
`hasPermission(perm)` must return the same thing as hasPermission(getActiveContexts(), perm). If your permissions plugin does not do this, it is broken. I will clarify the Subject javadocs so that this is specified.

The reason that MixinSubject does not directly delegate hasPermission is because the hasPermission(Set<Context>, String) method has some additional logic to change how undefined permissions are handled (this is not ideal, but I am still toying off-and-on with how to define additional information for specific subject types).

PR #552, opened 2016-02-28
Add TextConfigSerializer to DataRegistrar

windy1:
https://github.com/SpongePowered/SpongeAPI/pull/1079


src/main/java/org/spongepowered/common/data/DataRegistrar.java
```diff
@@ -27,40 +27,346 @@
 import org.spongepowered.api.Game;
 import org.spongepowered.api.block.BlockSnapshot;
 import org.spongepowered.api.block.BlockState;
-import org.spongepowered.api.block.tileentity.*;
```
kashike: These are intentionally star imports

gabizou: Yeah.... kinda need to reinstate these star imports....

PR #553, opened 2016-02-28
Fix Total Experience Value processor

dualspiral:
Currently, when offering a "Total Experience" value to a player, the processor is incorrectly calculating the level that the player should be at.

I have written a small test plugin that provides commands to offer experience data to the player who runs the command - see https://gist.github.com/dualspiral/fcf31718d3aa17f6d1b3. When running this test plugin, I run `/explvl 15` then `/expset 315`, which should give level 15 in both cases. However, as can be seen in the screenshot below, `/expset 315` gives an incorrect result.

fail

This PR fixes that so the experience level is set correctly - offering level 15 or 315 experience will return identical results.

The processor was also always returning false, when it should have actually been returning true (data was set, but the overall `DataTransactionResult` returns "REJECTED".) This has been fixed.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/processor/value/entity/TotalExperienceValueProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/value/entity/TotalExperienceValueProcessor.java
@@ -59,17 +59,34 @@ public class TotalExperienceValueProcessor extends AbstractSpongeValueProcessor<
 
     @Override
     protected boolean set(EntityPlayer container, Integer value) {
-        int level = 0;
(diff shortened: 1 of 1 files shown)


gabizou:
Does the same fix need to be applied to the data processor?

dualspiral:
Agh, no, not as far as I can see. I just got muddled up between "Data" and "Value" processors - forgive me, it's late here! This should be sufficent as far as I can see.

gabizou:
@dualspiral mind squashing into 1 tidy commit? Then I'll merge it in cleanly (otherwise github doesn't notice this as merged and I have to just close it otherwise).

dualspiral:
@gabizou Not a problem - it's squashed.

PR #495, opened 2016-02-10
Introduce GameProfileCache

kashike:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1068) | [**Common**](https://github.com/SpongePowered/SpongeCommon/pull/495)

PR #565, opened 2016-03-06
Fix AbstractMethodError with PlayerProfileCacheEntry mixin

kashike:
Small oversight when writing #495


zml2008:
Tested, works for me.

PR #566, opened 2016-03-08
Add mixin to use wall time instead of ticks for several things.

amaranth:
Use actual time passed to emulate a constant tick rate in places that are unlikely to break things and easily noticed by players. Should reduce "block lag" while digging on a server with a low tick rate, among other things.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/realtime/mixin/MixinEntity.java
+++ b/src/main/java/org/spongepowered/common/mixin/realtime/mixin/MixinEntity.java
@@ -1,0 +1,58 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.mixin.realtime.mixin;
+
+import net.minecraft.entity.Entity;
+import net.minecraft.world.World;
+import org.spongepowered.asm.lib.Opcodes;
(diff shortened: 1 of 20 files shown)


src/main/resources/mixins.common.realtime.json
```diff
@@ -0,0 +1,28 @@
+{
+  "minVersion": "0.5.1",
```
kashike: This file looks to be using 2 spaces instead of 4.

src/main/resources/mixins.common.realtime.json
```diff
@@ -0,0 +1,28 @@
...
+  "package": "org.spongepowered.common.mixin.realtime.mixin",
```
Aaron1011: You don't need the trailing 'mixin' package.

windy1:
I'd be curious to see some benchmarks for this PR.

Aaron1011:
@windy1: From my understanding of it, this isn't really about improving performance (in terms of efficiency/CPU usage). Rather, it allows certain tileentities and entities (along with the world time) to update at their normal rate, even when the server is lagging.

One example of this is baby animals. Normally, they take 20 minutes to grow into an adult. However, if the server is lagging, each animal will receive fewer ticks, due to some being skipped. This PR updates some of their logic to use the actual elapsed wall-clock time, rather than number of ticks. The result is that even if the server is lagging, most entity/te features which are supposed to take *X* number of second will actually take that number of second to complete.

src/main/resources/mixins.common.realtime.json
amaranth: I needed to put the mixins in a different package than the interface. It was either this or make an "interface" package.

liach: Please put interfaces in `org.spongepowered.common.interfaces` package.

simon816:
Wouldn't it be better if you used `System.nanoTime()` so that changing the system time doesn't effect it?

amaranth:
This is a good point, I should be using nanoTime.

src/main/resources/mixins.common.realtime.json
Zidane: @liach 

His interface is the same name as the one in common and adds new things. His location of it is fine and won't be moved.

@Aaron1011 
You oughta know by this point that whatever is in the root or child of root mixin page isn't classloaded and therefore cannot be referenced in managed code.

Aaron1011: @Zidane: What?? I just meant that instead of "org.spongepowered.common.mixin.realtime.mixin", the package could just be "org.spongepowered.common.mixin.realtime", to match our other mixins (e.g. "org.spongepowered.mod.mixin.entityactivation").

src/main/java/org/spongepowered/common/mixin/realtime/mixin/MixinEntity.java
```diff
@@ -0,0 +1,58 @@
...
+import org.spongepowered.asm.mixin.injection.Redirect;
+import org.spongepowered.common.mixin.realtime.IMixinMinecraftServer;
+
+// TODO(amaranth): Should this handle fire ticks too? Mobs can't move fast enough
+@Mixin(Entity.class)
+public abstract class MixinEntity {
+
+    private static final String ENTITY_RIDABLE_COOLDOWN_FIELD = "Lnet/minecraft/entity/Entity;field_184245_j:I";
+    private static final String ENTITY_PORTAL_COUNTER_FIELD = "Lnet/minecraft/entity/Entity;portalCounter:I";
+    @Shadow protected int field_184245_j;
```
Minecrell: It would be nice to name this field for readability (in the future, once we update this to 1.9)

src/main/java/org/spongepowered/common/mixin/realtime/mixin/MixinMinecraftServer.java
```diff
@@ -34,17 +34,17 @@
...
-    private static long lastTickMillis = System.currentTimeMillis();
+    private static long lastTickNanos = System.nanoTime();
     private static long realTimeTicks = 1;
 
     @Inject(method = "tick", at = @At("HEAD"))
     public void tick(CallbackInfo ci) {
-        long currentMillis = System.currentTimeMillis();
-        realTimeTicks = (currentMillis - lastTickMillis) / 50;
+        long currentNanos = System.nanoTime();
+        realTimeTicks = (currentNanos - lastTickNanos) / 50000000;
```
Kiskae: Perhaps a constant initialized as `TimeUnit.SECONDS.toNanos(1) / 20` would be less magic?

PR #591, opened 2016-03-16
Block changes

kashike:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1144) | [**Common**](https://github.com/SpongePowered/SpongeCommon/pull/591)


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "SpongeAPI"]
 	path = SpongeAPI
-	url = https://github.com/SpongePowered/SpongeAPI.git
+	url = https://github.com/kashike/SpongeAPI.git
```
bloodmc: This shouldn't be here.

kashike: Yes it should.

PR #594, opened 2016-03-17
Refactor ArmorStand data

kashike:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1145) | [**Common**](https://github.com/SpongePowered/SpongeCommon/pull/594)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "SpongeAPI"]
 	path = SpongeAPI
-	url = https://github.com/SpongePowered/SpongeAPI.git
+	url = https://github.com/kashike/SpongeAPI.git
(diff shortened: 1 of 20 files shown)


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "SpongeAPI"]
 	path = SpongeAPI
-	url = https://github.com/SpongePowered/SpongeAPI.git
+	url = https://github.com/kashike/SpongeAPI.git
```
windy1: errr

ryantheleach: It's been a common pattern for pr's to point to the contributors own spongeapi branch, that way travis can build and check them correctly, however a "Do not merge this commit" in the message would have been nice.

kashike: @windy1 This is how all non-core dev PRs have to be to get travis-ci to not fail the build, as stated by @ryantheleach.

It's split into two commits.

Issue #560, opened 2016-03-04
Inventory poll does not work with certain queries


DarkArc:
Given the following code:
```java
player.getInventory().query(ItemTypes.IRON_ORE).poll(4);
```

One would expect 4 Iron Ore to be removed regardless of the number of stacks as per this documentation:
```java
     * <p>For example, assume an inventory containing 4 slots contains stacks as
     * follows:</p>
     *
     * <blockquote>
     *     <pre>[Stone x10] [Dirt x3] [Arrows x9] [Stone x32]</pre>
     * </blockquote>
     *
     * <p>Calling <code>poll(16)</code> on this inventory will consume <em>Stone
     * </em> from the Inventory (because the first stack contains stone), and
     * will then consume the remaining 6 items from the 4th slot.</p>
```

However, given an inventory which contains more than one stack of Iron ore, no action will take place.

PR #596, opened 2016-03-18
Resolved an issue with polling inventory queries

DarkArc:
The function, getLensForOrdinal(int ordinal) was always returning null on any queries which returned multiple results, resulting in no operation. This change adds the child as a spanning child, which allows getLensForOrdinal to return the correct result, and thus operations such as poll(20) work with any amount of item stacks returned from the query.

Reference #560 for an example of a situation which this fixes.


DarkArc:
@SpongePowered/developers any objections to merging?

me4502:
Probably should get @Mumfrey to comment, he's the inventory guy

PR #599, opened 2016-03-19
Upgraded Gradle wrapper from 2.11 to 2.12

vorburger:


me4502:
I agree with this change, but it'll probably need to be tested with SpongeGradle, ForgeGradle and MixinGradle

vorburger:
@me4502 is a successful build (currently running) not sufficient? How do you test it with SpongeGradle, ForgeGradle and MixinGradle?

gabizou:
Closing for the same reason as https://github.com/SpongePowered/SpongeVanilla/pull/232

PR #600, opened 2016-03-19
Changed HashSet<> to HashSet<Team> due to Eclipse (4.6.0M5) complaint

vorburger:
Says "Type mismatch: cannot convert from HashSet to Set<Team>" without
this.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/scoreboard/MixinScoreboardLogic.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/scoreboard/MixinScoreboardLogic.java
@@ -213,9 +213,8 @@ public abstract class MixinScoreboardLogic extends Scoreboard implements IMixinS
         return Optional.empty();
     }
 
-    @SuppressWarnings("unchecked")
     public Set<Team> scoreboard$getTeams() {
-        return new HashSet<>((Set) this.teams.values());
+        return this.teams.values().stream().map(team -> (Team) team).collect(Collectors.toSet());
     }
 
     public Optional<Team> scoreboard$getMemberTeam(Text member) {


me4502:
This can probably be fixed by casting it to Set<Team>, rather than just Set

gabizou:
I don't think this is valid for a PR, let alone an issue we are causing. If anything, it sounds very much that Eclipse has a bug with generic type inference in the milestone that will likely be fixed in the next milestone release. Note that I'm even more inclined closing this as Milestone 5 isn't even an actual release. It's as much of a change required if any other IDE erroneously threw a compile error on this line when the traditional JDK compiler builds this just fine (and almost every other compiler).

vorburger:
sure, you got it: https://bugs.eclipse.org/bugs/show_bug.cgi?id=490001

But what would it hurt you to just insert a <Team> there anyway??

Minecrell:
Keep in mind IDEA uses the standard Java compiler `javac` so it's actually the Eclipse compiler that is broken if that doesn't work

src/main/java/org/spongepowered/common/mixin/core/scoreboard/MixinScoreboardLogic.java
```diff
@@ -213,9 +213,8 @@ public void removeObjective(ScoreObjective objective) {
         return Optional.empty();
     }
 
-    @SuppressWarnings("unchecked")
     public Set<Team> scoreboard$getTeams() {
-        return new HashSet<>((Set) this.teams.values());
+        return this.teams.values().stream().map(team -> (Team) team).collect(Collectors.toSet());
```
Minecrell: I've seen that they've argumented this cast is never valid on the Eclipse bug tracker, however your replaced method is redundant because `ScorePlayerTeam` implements `Team` **only at runtime** due to our use of Mixin. Therefore the previous cast was actually perfectly safe and much more efficient.

PR #613, opened 2016-03-27
Add null check to DyeableDataProcessor#isDyeable

dualspiral:
When running `ItemStack.supports(Keys.DYE_COLOR)` on an item that is not destined to become a block (or is not a dye), an NPE is thrown because `Block.getBlockFromItem(item)` returns `null`. This PR adds a null check to the method so that if the item is not a block, the method correctly returns `false`.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/processor/data/DyeableDataProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/data/DyeableDataProcessor.java
@@ -59,11 +59,16 @@ public class DyeableDataProcessor extends AbstractSingleDataProcessor<DyeColor, 
     }
 
     public static boolean isDyeable(Item item) {
-        return item.equals(Items.dye)
-                || Block.getBlockFromItem(item).equals(Blocks.wool)
-                || Block.getBlockFromItem(item).equals(Blocks.stained_glass)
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/data/processor/data/DyeableDataProcessor.java
```diff
@@ -59,11 +59,16 @@ public DyeableDataProcessor() {
...
-        return item.equals(Items.dye)
-                || Block.getBlockFromItem(item).equals(Blocks.wool)
-                || Block.getBlockFromItem(item).equals(Blocks.stained_glass)
-                || Block.getBlockFromItem(item).equals(Blocks.stained_glass_pane)
-                || Block.getBlockFromItem(item).equals(Blocks.stained_hardened_clay);
+        if (item.equals(Items.dye)) {
+            return true;
+        }
+
+        Block bl = Block.getBlockFromItem(item);
```
kashike: Name it `block`? Three more characters won't hurt. :older_woman:

dualspiral: Old habits die hard! In the context of this, I can't see it mattering much (it should be obvious from the context), but that's all personal style I guess.

Sorted it.

PR #623, opened 2016-04-01
Bug Fixes with UserDiscoverer Lag

hsyyid:
Previously, it would try to get a `GameProfile`, but it would always return `null`. This would cause pretty much infinite searching of the servers dat files - causing large amounts of lag.

Through this, it gets the `GameProfile` correctly and all lag is gone. :smile:

### DIFF
--- a/src/main/java/org/spongepowered/common/service/user/UserDiscoverer.java
+++ b/src/main/java/org/spongepowered/common/service/user/UserDiscoverer.java
@@ -38,6 +38,7 @@ import net.minecraft.server.management.UserListBansEntry;
 import net.minecraft.server.management.UserListWhitelist;
 import net.minecraft.server.management.UserListWhitelistEntry;
 import net.minecraft.world.storage.SaveHandler;
+import org.spongepowered.api.Sponge;
 import org.spongepowered.api.entity.living.player.User;
 import org.spongepowered.common.entity.player.SpongeUser;
 import org.spongepowered.common.interfaces.IMixinEntityPlayerMP;
@@ -50,8 +51,11 @@ import java.io.IOException;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Locale;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 
 class UserDiscoverer {
@@ -171,7 +175,9 @@ class UserDiscoverer {
         // want the player entity to be cached.
         IMixinEntityPlayerMP player = (IMixinEntityPlayerMP) confMgr.getPlayerByUUID(uniqueId);
         if (player != null) {
-            return player.getUserObject();
+            User user = player.getUserObject();
+            userCache.put(uniqueId, user);
+            return user;
         }
         return null;
     }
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/service/user/UserDiscoverer.java
```diff
@@ -45,20 +45,19 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 
 class UserDiscoverer {
 
-    private static final Cache<UUID, User> userCache = CacheBuilder.newBuilder()
-            .expireAfterAccess(1, TimeUnit.DAYS)
-            .build();
+    private static final Cache<UUID, User> userCache = CacheBuilder.newBuilder().expireAfterAccess(1, TimeUnit.DAYS).build();
```
kashike: Why this change?

src/main/java/org/spongepowered/common/service/user/UserDiscoverer.java
```diff
@@ -182,20 +184,29 @@ private static User getFromStoredData(UUID uniqueId) {
...
         } else {
             return null;
         }
     }
 
+    private static Optional<org.spongepowered.api.profile.GameProfile> getProfileFromServer(UUID uuid) {
+        CompletableFuture<org.spongepowered.api.profile.GameProfile> gameProfile = Sponge.getServer().getGameProfileManager().get(uuid);
+
+        try {
+            if (gameProfile.get() != null) {
```
kashike: Store value of `gameProfile.get()` instead of calling it twice

src/main/java/org/spongepowered/common/service/user/UserDiscoverer.java
```diff
@@ -182,20 +184,29 @@ private static User getFromStoredData(UUID uniqueId) {
         if (dataFile == null) {
             return null;
         }
-        GameProfile profile = MinecraftServer.getServer().getPlayerProfileCache().getProfileByUUID(uniqueId);
-        if (profile != null) {
-            User user = create(profile);
-            try {
-                ((SpongeUser) user).readFromNbt(CompressedStreamTools.readCompressed(new FileInputStream(dataFile)));
```
Aaron1011: What happened to this line?

src/main/java/org/spongepowered/common/service/user/UserDiscoverer.java
```diff
@@ -77,6 +81,7 @@ static User create(GameProfile profile) {
      */
     static User findByUuid(UUID uniqueId) {
         User user = userCache.getIfPresent(uniqueId);
+
```
kashike: No need for the new whitespace

src/main/java/org/spongepowered/common/service/user/UserDiscoverer.java
```diff
@@ -196,6 +203,21 @@ private static User getFromStoredData(UUID uniqueId) {
...
+        CompletableFuture<org.spongepowered.api.profile.GameProfile> gameProfile = Sponge.getServer().getGameProfileManager().get(uuid);
+
+        try {
+            org.spongepowered.api.profile.GameProfile profile = gameProfile.get();
+            if (profile != null) {
+                return Optional.of(profile);
+            } else {
+                return Optional.empty();
+            }
+        } catch (InterruptedException | ExecutionException e) {
```
gabizou: Why are you just swallowing the exception? This needs to be logged for any reason.

kenzierocks: Isn't that happening below?

PR #636, opened 2016-04-05
Replace MixinEntityArrow impact overwrite with injects

kashike:
@bloodmc


ryantheleach:
I'm curious, why does it need separate injections for dev vs production?

kashike:
@ryantheleach The LVT is different in production and dev.

me4502:
Now I'm curious as to why the LVT would be different. It's the same code and the same mixins isn't it? Apart from the obfuscation

PR #639, opened 2016-04-06
Pad out each page with empty lines, end pages with next page with ... fixes #638

### LINKED ISSUE #638: Pagination jitters.
At the moment Pagination jitters if the next element would span multiple lines making it appear inconsistent in chat.

Perhaps the contents should be padded with lines so that way it fills the entire chat box, making the previous and next arrows of the previous text hidden.

@Zidane You asked for more justification.

image

Is how it looks without these changes, Notice that it can and will show the previous set of navigation links.

This PR adds whitespace so that this doesn't happen on the default chat width/height.

ryantheleach:
Expands out the pagination pages with empty lines, Reduces the jitter when changing pages.

Also the last padded empty line when a next page exists is replaced with "..." to show that there are additional pages.

Fixes #638 

Edit: @Zidane You asked for more justification, if you havn't already read the issue. 

image

Is how it looks without these changes, Notice that it can and will show the previous set of navigation links.

This PR adds whitespace so that this doesn't happen on the default chat width/height.

image

### DIFF
--- a/src/main/java/org/spongepowered/common/service/pagination/IterablePagination.java
+++ b/src/main/java/org/spongepowered/common/service/pagination/IterablePagination.java
@@ -30,8 +30,8 @@ import com.google.common.base.Function;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 import com.google.common.collect.PeekingIterator;
-import org.spongepowered.api.text.Text;
 import org.spongepowered.api.command.CommandException;
+import org.spongepowered.api.text.Text;
 import org.spongepowered.api.text.channel.MessageReceiver;
 
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/service/pagination/IterablePagination.java
```diff
@@ -24,21 +24,20 @@
...
+import org.spongepowered.api.text.Text;
 import org.spongepowered.api.text.channel.MessageReceiver;
 
+import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
-import javax.annotation.Nullable;
+import static org.spongepowered.common.util.SpongeCommonTranslationHelper.t;
```
kashike: Please ensure you have the proper import order, this isn't it.

src/main/java/org/spongepowered/common/service/pagination/IterablePagination.java
```diff
@@ -94,6 +95,17 @@ public Text apply(Map.Entry<Text, Integer> input) {
         return ret;
     }
 
+    private void padPage(final List<Text> currentPage, final int currentPageLines, final Text continuation) {
+        final int maxContentLinesPerPage = getMaxContentLinesPerPage();
+        for (int i = currentPageLines; i< maxContentLinesPerPage; i++){
+            if(i == (maxContentLinesPerPage - 1)) {
```
kashike: `if (`

PR #644, opened 2016-04-10
Implement WorldStorage

simon816:


Zidane:
@Minecrell 

Gonna save you a comment and tell @simon816 now to use Java 7 NIO here haha.

simon816:
@Zidane @Minecrell done

src/main/java/org/spongepowered/common/world/storage/WorldStorageUtil.java
```diff
@@ -0,0 +1,114 @@
...
+        NBTTagCompound level = data.getCompoundTag("Level");
+        if (!level.hasKey("Sections", NbtDataUtil.TAG_LIST)) {
+            return null;
+        }
+        return NbtTranslator.getInstance().translateFrom(level);
+    }
+
+    public static Iterable<Path> listRegionFiles(Path worldDir) {
+        try {
+            return Files.newDirectoryStream(worldDir.resolve("region"), "*.mca");
```
Minecrell: You need to make sure you close this stream somehow (e.g. try-with-resources, although that would require inlining it..

src/main/java/org/spongepowered/common/world/storage/WorldStorageUtil.java
```diff
@@ -0,0 +1,114 @@
...
+        });
+    }
+
+    public static DataContainer readDataFromRegion(DataInputStream stream) throws IOException {
+        if (stream == null) {
+            return null;
+        }
+        NBTTagCompound data = CompressedStreamTools.read(stream);
+
+        if (!data.hasKey("Level", NbtDataUtil.TAG_COMPOUND)) {
```
gabizou: Add some constants in NbtDataUtil please. I don't want to see any literal string values used in the implementation as it usually leads to potential upgrade issues later on.

simon816: I kept this very close to `AnvilChunkLoader#checkedReadChunkFromNBT` so moving the strings out of the code would make it harder to follow/compare but alright I'll move it

gabizou: If this is a verbatim copy from `AnvilChunkLoader`, then I'd recommend following [this](https://github.com/SpongePowered/SpongeCommon/blob/1.9/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java#L286-L461) as a guide.

bloodmc:
@simon816 Fix the conflicts then get this into master once tested.

PR #653, opened 2016-04-15
[WIP] Close and Open inventory implementation

ewized:
Contains WIP for the implementation of the inventory opening and closing API.


src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -154,18 +157,19 @@
     @Shadow public int lastExperience;
     @Shadow private EntityPlayer.EnumChatVisibility chatVisibility = EntityPlayer.EnumChatVisibility.FULL;
     @Shadow private boolean chatColours;
+    @Shadow private int currentWindowId = 0;
```
kashike: don't need an initialiser

ewized:
Closing until I have more time

PR #661, opened 2016-04-21
Added the ability to override the java version for Sponge

### LINKED ISSUE #660: Version Checker Problems
Ubuntu's OpenJDK builds say their version is the following when accessed via the ```java.version``` property:
```
1.8.0_03-Ubuntu
```

However, this is not the case, and the build is actually:
```
1.8.0_77
```

If you use ```java.runtime.version``` property you receive:
```
1.8.0_03-Ubuntu-8u77-b03-3ubuntu3-b03
```

DarkArc:
java -jar -Dsponge.override.java.version=1.8.0_40 -jar ...

Can now be used to override the java version in the version
check. This is done to allow users who have invalidly versioned
versions of Java to still run the server.

Fixes  #660

### DIFF
--- a/src/java6/java/org/spongepowered/launch/JavaVersionCheckUtils.java
+++ b/src/java6/java/org/spongepowered/launch/JavaVersionCheckUtils.java
@@ -50,6 +50,6 @@ public class JavaVersionCheckUtils {
     }
 
     private static String getCurrentVersion() {
-        return System.getProperty("java.version");
+        return System.getProperty("sponge.override.java.version", System.getProperty("java.version"));
     }
 }


Minecrell:
I personally think this should simply skip the check completely (at their own risk) instead of requiring to add a version that is checked instead.

DarkArc:
My original thought was they could specify their actual current Java version (or our bare minimum requirement) and then if there's a change that requires a new Java version you have to change your startup script again, verifying that everything is still OK in the process.

---- On Thu, 21 Apr 2016 22:00:57 -0700 notifications@github.com wrote ----

I personally think this should simply skip the check completely (at their own risk) instead of requiring to add a version that is checked instead.

DarkArc:
Closing in favour of #684

PR #667, opened 2016-04-24
[WIP] Item pickup and despawn delay

kashike:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1197) | [**Common**](https://github.com/SpongePowered/SpongeCommon/pull/667)

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityItem.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityItem.java
@@ -26,6 +26,7 @@ package org.spongepowered.common.mixin.core.entity.item;
 
 import net.minecraft.entity.item.EntityItem;
 import net.minecraft.item.ItemStack;
+import net.minecraft.nbt.NBTTagCompound;
 import org.spongepowered.api.data.key.Keys;
 import org.spongepowered.api.data.manipulator.DataManipulator;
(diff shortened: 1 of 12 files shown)


src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityItem.java
```diff
@@ -26,156 +26,47 @@
...
-            } else {
-                this.infiniteDespawnDelay = false;
-            }
-        }
+    public int getDespawnDelay() {
+        return this.age;
     }
 
     @Override
-    public void writeToNbt(NBTTagCompound compound) {
```
simon816: The data should still be persisted

kashike: The pickup delay and age are persisted by vanilla.

simon816: Oh I see you're not adding back `setInfinitePickupDelay()`, maybe you could?

kashike: There's no need for it.

simon816: The original intention was to avoid magic values
https://github.com/SpongePowered/SpongeAPI/pull/338#discussion_r22431290

PR #668, opened 2016-04-25
NBT Data Cleanup

gabizou:
With 1.9's development, it's now apparent that the cleanup of `NbtDataUtil` is required, let alone that anything remaining in the implementation NOT using `NbtDataUtil` constants should be found and refactored to use as such. This is mostly another code cleanup.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/util/NbtDataUtil.java
+++ b/src/main/java/org/spongepowered/common/data/util/NbtDataUtil.java
@@ -51,55 +51,8 @@ import java.util.Optional;
  */
 public final class NbtDataUtil {
 
-    public static final String BANNER_PATTERN_ID = "Pattern";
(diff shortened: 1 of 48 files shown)


src/main/java/org/spongepowered/common/data/util/NbtDataUtil.java
```diff
@@ -116,42 +69,132 @@ private NbtDataUtil() {
...
+            // Compatibility tags for Forge
+            public static final String FORGE_ROOT = "ForgeData";
+            public static final String FORGE_ENTITY_TYPE = "entity_name";
+        }
+
+    }
+
+    public static final class World {
+
+        public static final String DIMENSION_TYPE = "dimensionType";
```
gabizou: I was considering that these would actually go under a new class for `Sponge` such that we can understand a difference between `Vanilla` and `Sponge` added constants.

PR #669, opened 2016-04-25
Implement Chunk#getNeighbor and optimize World#isAreaLoaded

amaranth:
Implement a version of Chunk#getNeighbor that uses neighbor pointers
that are updated on chunk load/unload. We then use this implementation to
create an optimized version of World#isAreaLoaded. This removes a major
source of chunk map lookups as it is called every tick for every entity and
was doing 25 chunk map lookups for each call.


src/main/java/org/spongepowered/common/mixin/core/world/MixinChunk.java
```diff
@@ -886,4 +933,88 @@ public int getBlockDigTimeWith(int x, int y, int z, ItemStack itemStack, Cause c
...
+            case SOUTH:
+            case SOUTHEAST:
+            case SOUTHWEST:
+                return 1;
+            case EAST:
+                return 2;
+            case WEST:
+                return 3;
+            default:
+                throw new IllegalArgumentException("Unexpected direction");
```
simon816: Just thinking that `Direction.UP` and `DOWN` are not 'unexpected', but rather unsupported. Perhaps `getNeighbor` should gracefully return `Optional.empty()` if you attempt to get a chunk in the y-plane (since it's unsupported in vanilla)

amaranth: Perhaps it should return this instead to prepare for a future where it can do something useful.

PR #682, opened 2016-04-30
Key bindings implementation

Cybermaxke:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1210) | **Common** | [Forge](https://github.com/SpongePowered/SpongeForge/pull/661) | [LiteMod](https://github.com/Cybermaxke/LiteSpongeClient/releases/)

This implementation includes all the code that is needed to handle the key bindings on the server side, SpongeForge (client) implementation will follow once it's updated to 1.9

There is already a small LiteLoader mod I made for the client features, it won't work in combination with  Forge. Most of the keyboard code will be able to be reused for the Forge implementation.

Requires https://github.com/SpongePowered/SpongeCommon/pull/1011 to be fixed in order to get the inventory key context (`KeyContexts#INVENTORY`) to work.


src/main/java/org/spongepowered/common/keyboard/KeyNetworkHandler.java
```diff
@@ -0,0 +1,168 @@
...
+    @Listener
+    public void onRegisterChannel(ChannelRegistrationEvent.Register event, @First PlayerConnection connection) {
+        if (event.getChannel().equals(SpongeMessageHandler.CHANNEL_NAME)) {
+            handleKeyRegistration(connection);
+        }
+    }
+
+    public static void handleKeyRegistration(PlayerConnection connection) {
+        final Player player = connection.getPlayer();
+        // TODO: Why is the connection still null at this point, is this normal in sp (SpongeForge) ???
```
kashike, line 69: Assuming a channel registration can be done before a player has finished logging in, this is normal.

Cybermaxke: Well this is rather annoying because you cannot send packets directly to a `PlayerConnection`, you have to pass the player instance through in the `ChannelBinding`, meaning in the SpongeAPI, you cannot use the event to send packets to the client. IMO the player connection should be set for the player during this event, even if it's reset afterwards (like I did here to work around the issue).

ImMorpheus:

PR #683, opened 2016-05-01
Add profile lookups to cache

kashike:


src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerProfileCache.java
```diff
@@ -195,7 +202,16 @@ public boolean add(GameProfile profile, boolean overwrite, @Nullable Date expiry
 
         this.getServer().getGameProfileRepository().findProfilesByNames(Iterables.toArray(names, String.class), Agent.MINECRAFT, new MapProfileLookupCallback(result));
 
-        return result.isEmpty() ? ImmutableMap.of() : ImmutableMap.copyOf(result);
+        if (!result.isEmpty()) {
+            for (Map.Entry<String, Optional<GameProfile>> entry : result.entrySet()) {
```
Aaron1011: You can just iterate over the values here.

PR #684, opened 2016-05-01
Better JVM version checking

### LINKED ISSUE #660: Version Checker Problems

RobertHerhold:
This fixes #660 by more accurately determining if one version of Java is higher/lower than another. With the current setup, "1.8.0_102" would be considered lower than "1.8.0_40", which is obviously not the case.

### DIFF
--- a/src/java6/java/org/spongepowered/launch/JavaVersionCheckUtils.java
+++ b/src/java6/java/org/spongepowered/launch/JavaVersionCheckUtils.java
@@ -40,12 +40,16 @@ public class JavaVersionCheckUtils {
     public static void ensureJava8() {
         String version = getCurrentVersion();
         if (getVersionValue(version) < getVersionValue(REQUIRED_VERSION)) {
-            String error = String.format(ERROR_MESSAGE, version, REQUIRED_VERSION);
+            if(!version.startsWith("1.8")) {
+                String error = String.format(ERROR_MESSAGE, version, REQUIRED_VERSION);
 
-            if (!GraphicsEnvironment.isHeadless()) {
-                JOptionPane.showMessageDialog(null, error, "PEBKACException!", JOptionPane.ERROR_MESSAGE);
+                if (!GraphicsEnvironment.isHeadless()) {
+                    JOptionPane.showMessageDialog(null, error, "PEBKACException!", JOptionPane.ERROR_MESSAGE);
+                }
+                throw new RuntimeException(error);
+            } else {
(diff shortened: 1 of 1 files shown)


src/java6/java/org/spongepowered/launch/JavaVersionCheckUtils.java
```diff
@@ -52,4 +52,32 @@ public static void ensureJava8() {
...
+        if(version.contains("-")) {
+            version = version.substring(0, version.indexOf('-'));
+        }
+        // Replace underscores with periods for easier String splitting
+        version = version.replace('_', '.');
+        // Split the version up into parts
+        String[] versionParts = version.split("\\.");
+        double versionValue = 0;
+        for(int i = 0; i < versionParts.length; i++) {
+            int part = Integer.valueOf(versionParts[i]);
```
ewized: Should do some more strict checking here, will throw an exception if it's null or not a number. It's not a good idea to crash because a bad `java.version`

RobertHerhold: I thought of that, but the chances of that happening are almost none. I'll check for it regardless.

gabizou:
One thing to note, on OpenJDK 8, there are various malformed versions that this will still throw an exception for, and end up preventing the server from starting. In cases like this, I'd rather the last resort check be "Are we in Java 8? yes? ok, just print a warning saying that any crashes from Sponge may require an update to your Java, but we cannot determine the update version of Java you are running." Something along those lines would help server administrators at least use Sponge knowing their version of Java is up to date.

RobertHerhold:
@gabizou 
I've changed it to be more lenient on JVM version while still hopefully more accurately determining if it is outdated. Is that agreeable?

JBYoshi:
Idea: Maybe try to replicate the conditions of the crash and display an error if they fail?

gabizou:
@DarkArc Could you re-test with the latest changes?

DarkArc:
Server loads now

RobertHerhold:
Anything left to do with this?

PR #687, opened 2016-05-01
Add the SpongeAPI dependency to the annotation processor.

JBYoshi:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1214) | **SpongeCommon** | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/622) | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/238)

Allows SpongePowered/Ore#95 to be added without giving (most) plugin developers any additional work. This uses a dependency version range, so it will work with any SpongeAPI version that's guaranteed to be compatible.


me4502:
Should SpongeCommon give a warning if you use an API version outside of the bounds? It may still work, but it'd encourage the server owner to look for an updated version.

PR #695, opened 2016-05-03
Ensure empty ListPaginations don't throw

phroa:
Previously, the call to `page > this.pages.size()` would throw a
`CommandException` as soon as you try and fetch the first page of an empty
pagination.

Now, an empty List is returned when there are no items.

Tested it locally on today's freshest SF 1.8.9, works. Not that I expected this to be a breaking change.

### DIFF
--- a/src/main/java/org/spongepowered/common/service/pagination/ListPagination.java
+++ b/src/main/java/org/spongepowered/common/service/pagination/ListPagination.java
@@ -26,6 +26,7 @@ package org.spongepowered.common.service.pagination;
 
 import static org.spongepowered.common.util.SpongeCommonTranslationHelper.t;
 
+import com.google.common.collect.ImmutableList;
 import org.spongepowered.api.command.CommandException;
 import org.spongepowered.api.text.Text;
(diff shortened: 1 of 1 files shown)


phroa:
A potential alternative would be throwing `Nothing to display`, I suppose.

src/main/java/org/spongepowered/common/service/pagination/ListPagination.java
```diff
@@ -64,7 +65,9 @@ public ListPagination(MessageReceiver src, PaginationCalculator calc, List<Map.E
 
     @Override
     protected Iterable<Text> getLines(int page) throws CommandException {
-        if (page < 1) {
+        if (this.pages.size() == 0) {
+            return Lists.newArrayList();
```
kashike: Assuming this is to be used to read-only: `return ImmutableList.of()` - uses a single instance

gabizou: Or simpler: `Collections.emptyList()`.

PR #702, opened 2016-05-04
Refactor the particles implementation.

Cybermaxke:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1222) | Common


gabizou:
Remove the commit changing the API submodule destination.

Cybermaxke:
@gabizou I rebased the pr and I moved the submodule change in a separate commit.

src/main/java/org/spongepowered/common/effect/particle/SpongeParticleHelper.java
```diff
@@ -75,6 +85,40 @@ private static int getBlockState(SpongeParticleEffect effect, Optional<BlockStat
...
+            case NORTH:
+                return 7;
+            case NORTHWEST:
+                return 8;
+            default:
+                return 4;
+        }
+    }
+
+    private static final EntityFireworkRocket FIREWORK_ROCKET_DUMMY = new EntityFireworkRocket(null);
```
kashike: What are you doing here, and why?

src/main/java/org/spongepowered/common/effect/particle/SpongeParticleEffect.java
```diff
@@ -58,98 +50,30 @@ public SpongeParticleType getType() {
         return this.type;
     }
 
+    @SuppressWarnings("unchecked")
     @Override
-    public Vector3d getMotion() {
-        return this.motion;
+    public <V> Optional<V> getOption(ParticleOption<V> option) {
+        return (Optional) Optional.ofNullable(this.options.get(option));
```
liach: Why is there an extra cast?

src/main/java/org/spongepowered/common/effect/particle/SpongeParticleHelper.java
Cybermaxke: I created a dummy `EntityFireworkRocket` to easily create packets for the fireworks effect, and reusing the entity for all the effects. And the `World` of an entity can be null.

src/main/java/org/spongepowered/common/effect/particle/SpongeParticleEffect.java
Cybermaxke: @liach The `Map` has a `Object` as value type. So I had to cast it to the generic, though I could just have casted the object to the generic : `return Optional.ofNullable((V) this.options.get(option));`

src/main/java/org/spongepowered/common/effect/particle/SpongeParticleHelper.java
kashike: I understand that, but why are you caching the packets and modifying the data, and then sending? I can only see problems doing this - being modified again before being sent with different data, etc.

Cybermaxke: The `DESTROY_FIREWORK_ROCKET_DUMMY` and `FIREWORK_ROCKET_DUMMY_EFFECT` are never going to change. And the `FIREWORK_ROCKET_DUMMY` is only used to set the position, id, uniqueId in the packet (This won't cause problems unless you want to use it async). That is also the reason why I didn't use the `EntityDataManager` constructor in `SPacketEntityMetadata`, because the `DataEntry` are shared.

Cybermaxke: I guess I can remove the dummy to prevent that issue.

kashike: I can guarantee there are going to be issues, so it would be best to
> I guess I can remove the dummy to prevent that issue.

src/main/java/org/spongepowered/common/effect/particle/SpongeParticleHelper.java
```diff
@@ -111,12 +112,23 @@ private static int getDirectionData(Direction direction) {
...
-        FIREWORK_ROCKET_DUMMY.getDataManager().getDirty();
+        final EntityFireworkRocket entityFireworkRocket = new EntityFireworkRocket(null);
+        FIREWORK_ROCKET_ID = entityFireworkRocket.getEntityId();
+        FIREWORK_ROCKET_UNIQUE_ID = entityFireworkRocket.getUniqueID();
+
+        DESTROY_FIREWORK_ROCKET_DUMMY = new SPacketDestroyEntities(FIREWORK_ROCKET_ID);
+
+        FIREWORK_ROCKET_DUMMY_EFFECT = new SPacketEntityStatus();
+        FIREWORK_ROCKET_DUMMY_EFFECT.entityId = FIREWORK_ROCKET_ID;
+        FIREWORK_ROCKET_DUMMY_EFFECT.logicOpcode = 17;
```
kashike: Where does this number come from? A comment on this line would be nice for anyone looking at this in the future.

src/main/java/org/spongepowered/common/effect/particle/SpongeParticleHelper.java
```diff
@@ -139,16 +151,22 @@ private static int getDirectionData(Direction direction) {
...
                     return Collections.emptyList();
                 }
-                FIREWORK_ROCKET_DUMMY.setPosition(position.getX(), position.getY(), position.getZ());
+                final SPacketSpawnObject packetSpawnObject = new SPacketSpawnObject();
+                packetSpawnObject.entityId = FIREWORK_ROCKET_ID;
+                packetSpawnObject.uniqueId = FIREWORK_ROCKET_UNIQUE_ID;
+                packetSpawnObject.x = position.getX();
+                packetSpawnObject.y = position.getY();
+                packetSpawnObject.z = position.getZ();
+                packetSpawnObject.type = 76;
```
kashike: Where does this number come from? A comment on this line would be nice for anyone looking at this in the future.

src/main/java/org/spongepowered/common/effect/particle/SpongeParticleHelper.java
Cybermaxke: Fixed

src/main/java/org/spongepowered/common/effect/particle/SpongeParticleHelper.java
Cybermaxke: Fixed

PR #711, opened 2016-05-06
All living entities are a Projectile source

kashike:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1227) | [**Common**](https://github.com/SpongePowered/SpongeCommon/pull/711)

### DIFF
--- a/src/main/java/org/spongepowered/common/entity/projectile/DispenserSourceLogic.java
+++ b/src/main/java/org/spongepowered/common/entity/projectile/DispenserSourceLogic.java
@@ -1,0 +1,82 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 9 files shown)


src/main/java/org/spongepowered/common/entity/projectile/DispenserSourceLogic.java
```diff
@@ -0,0 +1,58 @@
+package org.spongepowered.common.entity.projectile;
```
gabizou, line 25: Missing license header

src/main/java/org/spongepowered/common/entity/projectile/SimpleEntityLaunchLogic.java
```diff
@@ -0,0 +1,38 @@
+package org.spongepowered.common.entity.projectile;
```
gabizou, line 25: Missing license header

src/main/java/org/spongepowered/common/entity/projectile/SimpleDispenserLaunchLogic.java
```diff
@@ -0,0 +1,26 @@
+package org.spongepowered.common.entity.projectile;
```
gabizou, line 25: Missing license header

Issue #586, opened 2016-03-13
Use FastUtil rather than Trove for more memory efficient collections and maps.


me4502:
Whilst Trove has been around for a while, it isn't as actively developed anymore. FastUtil is much more recent, and actively developed.

On top of this, Trove doesn't hold up in terms of performance compared to more modern libraries. Here's an example of a benchmark that pits numerous libraries together,

http://java-performance.info/hashmap-overview-jdk-fastutil-goldman-sachs-hppc-koloboke-trove-january-2015/

Another option is Koloboke, which appears to include enhanced integration with JDK8.

me4502:
Koloboke is looking to be a better option, as the FastUtil jar is 17mb.

Zidane:
@amaranth 

Your thoughts on this?

Minecrell:
Koloboke's full JAR is even bigger (about 18 MB), so if we'd choose one of them we'd need to filter the unused classes out before shading, because that would be way to big to shade completely.

amaranth:
Now that Minecraft is including a subset of fastutil it makes sense to use that in place of trove.

Minecrell:
Fixed in https://github.com/SpongePowered/SpongeCommon/commit/81e15b64da5b1c3dda4ec3cce2469cef13a4f0dd

PR #733, opened 2016-05-18
Replace usages of Trove with fastutil

Minecrell:
As discussed in #586 this replaces all usages of Trove with fastutil. A subset of the fastutil classes is already bundled with Minecraft in 1.9.3+, however for Sponge we need some additional classes.

For now I've added the additional classes to a separate sourceset on SpongeCommon, so at the end only the additional classes will be shaded. There are quite a few, so I'm not sure whether we should rather move that to a separate repo or just upload it to the Maven repository directly (the classes are all unmodified).

This saves about ~2MB of the final SpongeVanilla JAR because only ~200KB of fastutil is additionally shaded in Sponge.

**Additional SpongeVanilla branch:** [feature/fastutil](https://github.com/SpongePowered/SpongeVanilla/tree/feature/fastutil)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,6 +48,13 @@ minecraft {
     version = '1.9.4'
 }
 
+configurations {
+    fastutil
+    compile {
+        extendsFrom fastutil
+    }
+}
+
 dependencies {
     compile api
     compile('org.spongepowered:mixin:0.5.5-SNAPSHOT') {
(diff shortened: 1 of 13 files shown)


simon816:
What about including it as a dependency the normal way, then configure the gradle shade plugin to only include selected classes?

Minecrell:
@simon816 Might be a bit annoying to figure out which classes are needed but that should also work

Minecrell:
@simon816 Changed!

PR #734, opened 2016-05-19
Update Pagination to not pad if only one page exists.

dualspiral:
When #639 was merged, some of my pagination broke - see https://github.com/NucleusPowered/Nucleus/issues/158. In this particular case:

* A one page MOTD which has the potential to have more than one page now fills the entire chat box, which is a waste of chat space when there is only one page to display.
* I have code that sets the number of lines to display to the console to `Integer.MAX_VALUE`, so that the console can just see the entire MOTD. However, since merging #639, this causes an out of memory error because it's trying to create a _huge_ list. (While I accept I could lower this value, I still want to use the styling that the Pagination service provides without any padding when I do not know how many lines I have.)

I think the change provided in #639 is a real usability win, but only when there is more than one page, and this breaks what plugins were expecting before. This PR returns the previous behaviour to one page items, while retaining the new behaviour when there is more than one page.

I have a test plugin that will demonstrate the behaviour: https://gist.github.com/dualspiral/067000c4b1fc80407e79bd51cce151d8

The commands are:

* `/pg <numberoflines>`, put the word "text" into the pagination list the specified number of times, displaying 20 lines per page.
* `/pglg <numberoflines>`, put the phrase "long text text text text text text text text text text text text text text text text text text text text text text text" into the pagination list the specified number of times, 20 lines per page.
* `/pgmax <numberoflines>` is the same as `/pg`, but can have `Integer.MAX_VALUE` number of lines.

I tested with `/pg 1`, `/pg 21`, `/pglg 10` and `/pgmax 25`.

### DIFF
--- a/src/main/java/org/spongepowered/common/service/pagination/ActivePagination.java
+++ b/src/main/java/org/spongepowered/common/service/pagination/ActivePagination.java
@@ -188,25 +188,14 @@ abstract class ActivePagination {
         return ret.build();
     }
 
-    protected void padPage(final List<Text> currentPage, final int currentPageLines, final PadBehavior behavior) {
+    protected void padPage(final List<Text> currentPage, final int currentPageLines, final boolean addContinuation) {
         final int maxContentLinesPerPage = getMaxContentLinesPerPage();
         for (int i = currentPageLines; i < maxContentLinesPerPage; i++) {
-            if (i == maxContentLinesPerPage - 1 && behavior != PadBehavior.NO_CONTINUATION) {
+            if (addContinuation && i == maxContentLinesPerPage - 1) {
                 currentPage.add(CONTINUATION_TEXT);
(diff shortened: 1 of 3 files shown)


ryantheleach:
```
6:07 AM <@zml> originally console had everything output as one page
6:07 AM <@zml> iirc giving a lines per page of -1 meant unlimited length
6:07 AM <@zml> dualspiral, ryantheleach ^
6:07 AM <@zml> dunno if that's still the case
6:09 AM <ryantheleach> I suspect it might be, and even cancel the padding
```

dualspiral:
> originally console had everything output as one page

It's not, which is why I've been using `Integer.MAX_LENGTH`. Others might be too.

> iirc giving a lines per page of -1 meant unlimited length

It's not documented, so I wasn't aware of that (though I think I realised that was the case when whipping this up) - it's good to know. However, my other points do still stand.

PR #756, opened 2016-05-29
[1.9] Implement EnderDragon.getHealingCrystal().

JBYoshi:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1260) | **SpongeCommon**
I heard @gabizou say some time ago that `DataManipulator`s are no longer allowed to store `Entity`s. So instead of implementing `HealingSourceData`, I just moved it into `EnderDragon` and implemented it there.


JBYoshi:
@Zidane It looks like you merged the API changes; could you also get the implementation merged in? I fixed the merge conflicts.

Issue #739, opened 2016-05-22
Broken CatalogTypes


ST-DDT:
Based on rejected PR #387.

**Currently existing issues:**
- Existing `CatalogType`s not being accessible via `GameRegistry`
- `CatalogType`'s `getId()` method not being implemented or throwing errors
- `CatalogType`'s `getName()` method not being implemented or throwing errors
- `CatalogType`'s `getTranslation()` method not being implemented or throwing errors or returning the wrong `Translation`s (AFAICT)
- `CatalogType`'s `toString()` method only using the default implementation or throwing errors
- `GameRegistry#getAll(Class)` returning duplicate instances
- `GameRegistry`'s `getType(type, id)` returning absent or errors for valid ids
- `GameRegistry`'s `getType(type, id)` throwing an `Exception` instead of `Optional.empty()` if a key is missing/invalid (but != null)
- Classes mentioned in `@CatalogedBy` containing public static fields with `null` values

**Test-Plugin**
- [Plugin.jar](https://dl.dropboxusercontent.com/u/16999313/Sponge/ZoneAPIChecker.jar)
- [Plugin sources](https://dl.dropboxusercontent.com/u/16999313/Sponge/ZoneAPIChecker.zip)

Wait for `GameStartedServerEvent` and you will get an overview about what is and isn't working.

The plugin searches for the following error types:
- No registry module for `CatalogType` class (`GameRegistry#getAll()` fails)
- `CatalogType#getId()` not implemented, errors or returns null
- `CatalogType#getName()` not implemented, errors or returns null
- `CatalogType#getTranslation()` not implemented, errors or returns null
- `CatalogType#getTranslation()` returns a Translation that errors if used
- `CatalogType#getTranslation()` returns a Translation that has no translation on the server
- `CatalogType#toString()` errors or returns null
- `GameRegistry#getAll(Class)` returns duplicate instances
- `GameRegistry#getType(Class, String)` errors
- `GameRegistry#getType(Class, String)` returns no instance for an id that was obtained from an instance from `GameRegistry#getAll(Class)`
- `GameRegistry#getType(Class, String)` returns a different instance for an id that was obtained from an instance from `GameRegistry#getAll(Class)`
- `GameRegistry#getType(Class, String)` errors when requesting an id that does not exist
- Classes referenced by `CatalogType`s via `@CatalogedBy` containing null values in static fields

[...]

PR #772, opened 2016-06-05
Do not override the getName method of SoundCategory.

Cybermaxke:
Fixes https://github.com/SpongePowered/SpongeForge/issues/635 and https://github.com/SpongePowered/SpongeForge/issues/634
The name shouldn't contain the `minecraft:` prefix anyway.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/sound/MixinSoundCategory.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/sound/MixinSoundCategory.java
@@ -25,7 +25,9 @@
 package org.spongepowered.common.mixin.core.sound;
 
 import net.minecraft.util.SoundCategory;
-import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Implements;
+import org.spongepowered.asm.mixin.Interface;
+import org.spongepowered.asm.mixin.Intrinsic;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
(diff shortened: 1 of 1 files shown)


gabizou:
You're not implementing the interface contracted method of `getName()`.

Cybermaxke:
@gabizou There is a method in the nms `SoundCategory` with the same signature.

gabizou:
> @gabizou There is a method in the nms SoundCategory with the same signature.

That's at development time, not at runtime. Obfuscated methods are still going to be obfuscated in production, so the contract method isn't going to magically create a bridge.

ST-DDT:
@Cybermaxke https://github.com/SpongePowered/SpongeCommon/issues/739 provides a test plugin that tests whether `CatalogType`s are fully implemented at runtime. You can use it to test your changes.

Cybermaxke:
@gabizou Ah, I didn't think of that, I will fix it.

src/main/java/org/spongepowered/common/mixin/core/sound/MixinSoundCategory.java
```diff
@@ -25,30 +25,33 @@
...
     private String id;
 
     @Inject(method = "<init>", at = @At("RETURN"))
     private void onInit(String enumName, int i, String realName, CallbackInfo ci) {
         this.id = "minecraft:" + realName;
     }
 
-    @Override
-    public String getId() {
+    public String category$getId() {
```
gabizou: You can still implement the interface here.

gabizou:
@Cybermaxke mind squashing these commits into one? After that I can merge it all in :)

Cybermaxke:
@gabizou Done ;)

PR #793, opened 2016-06-12
[WIP] Bugfixes and improvements for selectors

kenzierocks:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1270) | SpongeCommon (You're here!)

Improvements and fixes for selectors based on 1.9 and my own review of the code.

Selectors now use a `List`, the resolver returns the elements ordered properly as well as the right number of elements. Removed the `force` argument as I don't remember what it was really useful for.

### DIFF
--- a/src/main/java/org/spongepowered/common/text/selector/SelectorResolver.java
+++ b/src/main/java/org/spongepowered/common/text/selector/SelectorResolver.java
@@ -63,7 +63,6 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
-import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -73,12 +72,8 @@ import javax.annotation.Nullable;
 /**
  * A resolver that acts like Vanilla Minecraft in many regards.
  */
-// TODO decide if we want selector resolvers as part of the API, ask @kenzierocks for details
 public class SelectorResolver {
 
-    private static final Function<CommandSource, String> GET_NAME = CommandSource::getName;
-    private static final Vector3d ORIGIN = new Vector3d(0, 0, 0);
-    private static final Function<Number, Double> TO_DOUBLE = Number::doubleValue;
     private static final Collection<SelectorType> INFINITE_TYPES = ImmutableSet.of(SelectorTypes.ALL_ENTITIES, SelectorTypes.ALL_PLAYERS);
 
     private static Extent extentFromSource(CommandSource origin) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/spongepowered/common/text/selector/SelectorResolver.java
```diff
@@ -324,64 +312,57 @@ private void addTypeFilters(List<Predicate<Entity>> filters) {
...
-                if (!this.selectorFilter.test(e)) {
-                    continue;
-                }
-                entities.add(e);
-                count++;
-                if (maxToSelect != 0 && count > maxToSelect) {
-                    break;
-                }
+                List<Entity> holder = entityStream.collect(Collectors.toList());
+                Collections.shuffle(holder);
```
Cybermaxke: The shuffle won't have much effect because the stream will get sorted below. I am not sure what the right behavior would be, but are the entities always supposed to be sorted by distance? It also seems a bit waste of performance if you don't actually use it.

kenzierocks: It does have an effect due to the `limit` below.

Cybermaxke: Ah yes, your right. Still, the distance sorting seems kinda optional.

kenzierocks:
Marking this as WIP since new selectors and values have been added and the API may need a redesign.

src/main/java/org/spongepowered/common/text/selector/SpongeSelectorFactory.java
```diff
@@ -222,6 +223,8 @@ public Selector parseRawSelector(String selector) {
     public <T> SpongeArgumentType.Invertible<T> createInvertibleArgumentType(
             String key, Class<T> type, String converterKey) {
         if (!this.argumentLookupMap.containsKey(key)) {
+            checkNotNull(converterKey, "converter key cannot be null");
+            checkNotNull(converterKey, "converter key cannot be null");
```
kashike: Only one `checkNotNull` is necessary.

kenzierocks: Meant to check other fields, good catch 👍

rexbut:
@kenzierocks Can you make the last modification and add new selector added in 1.12. Or I open a new push request to realize the modification ? PR https://github.com/kenzierocks/SpongeCommon/pull/2

parlough:
Due to little activity, superseded by https://github.com/SpongePowered/SpongeCommon/pull/1519

PR #811, opened 2016-06-24
Make the difficulty cataloguetype ID not include the resource location. Fixes SpongeAPI#1254

me4502:
In reference to https://github.com/SpongePowered/SpongeAPI/issues/1254

This uses just the name for the ID of Difficulty, rather than the resource location. 

It makes more sense to use just the name, as it firstly fixes the issue above, and the ID should be something easily identifiable, which options.difficulty.etc isn't,

Edit:
This was tested by running 
> System.out.println(Sponge.getGame().getRegistry().getType(Difficulty.class, "easy"));
The output was 
> Optional[EASY]

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/data/types/MixinEnumDifficulty.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/data/types/MixinEnumDifficulty.java
@@ -31,6 +31,9 @@ import org.spongepowered.api.world.difficulty.Difficulty;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/core/data/types/MixinEnumDifficulty.java
```diff
@@ -44,7 +44,7 @@
 
     @Override
     public String getId() {
-        return this.difficultyResourceKey;
+        return this.difficultyResourceKey.replace("options.difficulty.", "");
```
ZephireNZ: No need to do this every time, just make a field in the constructor.

me4502: Yep, killed two birds with one stone and moved the translation TODO.

ST-DDT:
IMO we should use the `minecraft:` prefix more  consistently.

Aaron1011:
As @ST-DDT said, could you make the ids be prefixed with `minecraft:`? Take a look at [BlockTypeRegistryModule](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/registry/type/BlockTypeRegistryModule.java#L81) for how it's done.

me4502:
@Aaron1011 Done.

PR #819, opened 2016-06-27
Make the difficulty cataloguetype ID not include the resource location. Fixes SpongeAPI-#1254

me4502:
In reference to https://github.com/SpongePowered/SpongeAPI/issues/1254

This uses just the name for the ID of Difficulty, rather than the resource location. 

It makes more sense to use just the name, as it firstly fixes the issue above, and the ID should be something easily identifiable, which options.difficulty.etc isn't,

Edit:
This was tested by running 
> System.out.println(Sponge.getGame().getRegistry().getType(Difficulty.class, "easy"));
The output was 
> Optional[EASY]

### DIFF
--- a/src/main/java/org/spongepowered/common/registry/type/world/DifficultyRegistryModule.java
+++ b/src/main/java/org/spongepowered/common/registry/type/world/DifficultyRegistryModule.java
@@ -47,6 +47,10 @@ public final class DifficultyRegistryModule implements CatalogRegistryModule<Dif
 
     @Override
     public Optional<Difficulty> getById(String id) {
+        checkNotNull(id);
+        if (!id.contains(":") && !id.equals("none")) {
+            id = "minecraft:" + id; // assume vanilla
+        }
         return Optional.ofNullable(this.difficultyMappings.get(checkNotNull(id).toLowerCase(Locale.ENGLISH)));
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/registry/type/world/DifficultyRegistryModule.java
```diff
@@ -57,10 +61,10 @@
 
     @Override
     public void registerDefaults() {
-        this.difficultyMappings.put("peaceful", (Difficulty) (Object) EnumDifficulty.PEACEFUL);
-        this.difficultyMappings.put("easy", (Difficulty) (Object) EnumDifficulty.EASY);
-        this.difficultyMappings.put("normal", (Difficulty) (Object) EnumDifficulty.NORMAL);
-        this.difficultyMappings.put("hard", (Difficulty) (Object) EnumDifficulty.HARD);
+        this.difficultyMappings.put("minecraft:peaceful", (Difficulty) (Object) EnumDifficulty.PEACEFUL);
```
Aaron1011: You'll need to override [provideCatalogMap](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/registry/type/item/EnchantmentRegistryModule.java#L62).

PR #823, opened 2016-07-01
Add message events for when server sends message to players

simon816:
When a call is made to send a message to a player, a `MessageChannelEvent` is fired with a fixed `MessageChannel`.
This allows customizing the output messages sent to players.
Closes SpongePowered/SpongeAPI#1292

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
@@ -156,6 +156,7 @@ import org.spongepowered.common.interfaces.text.IMixinTitle;
 import org.spongepowered.common.interfaces.world.IMixinWorldServer;
 import org.spongepowered.common.item.inventory.util.ItemStackUtil;
 import org.spongepowered.common.text.SpongeTexts;
+import org.spongepowered.common.text.chat.ChatUtil;
 import org.spongepowered.common.text.chat.SpongeChatType;
 import org.spongepowered.common.util.BookFaker;
 import org.spongepowered.common.util.LocaleCache;
@@ -200,7 +201,6 @@ public abstract class MixinEntityPlayerMP extends MixinEntityPlayer implements P
     @Shadow public abstract void openGuiHorseInventory(AbstractHorse horse, IInventory horseInventory);
 
     // Inventory
-    @Shadow public abstract void sendMessage(ITextComponent component);
     @Shadow public abstract void closeScreen();
     @Shadow public int currentWindowId;
     @Shadow private void getNextWindowId() { }
@@ -414,6 +414,20 @@ public abstract class MixinEntityPlayerMP extends MixinEntityPlayer implements P
         }
 
         this.connection.sendPacket(new SPacketChat(component, ((SpongeChatType) type).getByteId()));
+    }
+
+    /**
+     * @author simon816 - 14th November, 2016
+     *
+     * @reason Redirect messages sent to the player to fire a message event. Once the
+     * event is handled, it will send the message to
+     * {@link #sendMessage(ChatType, Text)}.
+     *
+     * @param component The message
+     */
+    @Overwrite
+    public void sendMessage(ITextComponent component) {
+        ChatUtil.sendMessage(component, MessageChannel.fixed(this), (CommandSource) this.mcServer, true);
     }
 
     @Override
--- a/src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
@@ -73,6 +73,7 @@ import net.minecraft.world.storage.WorldInfo;
 import org.apache.logging.log4j.Logger;
 import org.spongepowered.api.Server;
 import org.spongepowered.api.Sponge;
+import org.spongepowered.api.command.CommandSource;
 import org.spongepowered.api.entity.Transform;
 import org.spongepowered.api.entity.living.player.Player;
 import org.spongepowered.api.event.SpongeEventFactory;
@@ -119,7 +120,6 @@ import org.spongepowered.common.interfaces.entity.IMixinEntity;
 import org.spongepowered.common.interfaces.entity.player.IMixinEntityPlayer;
 import org.spongepowered.common.interfaces.entity.player.IMixinEntityPlayerMP;
 import org.spongepowered.common.interfaces.network.play.server.IMixinSPacketWorldBorder;
-import org.spongepowered.common.interfaces.world.IMixinWorldProvider;
 import org.spongepowered.common.interfaces.world.IMixinWorldServer;
 import org.spongepowered.common.mixin.core.entity.player.MixinEntityPlayer;
 import org.spongepowered.common.service.ban.SpongeIPBanList;
@@ -127,6 +127,7 @@ import org.spongepowered.common.service.ban.SpongeUserListBans;
 import org.spongepowered.common.service.permission.SpongePermissionService;
 import org.spongepowered.common.service.whitelist.SpongeUserListWhitelist;
 import org.spongepowered.common.text.SpongeTexts;
+import org.spongepowered.common.text.chat.ChatUtil;
 import org.spongepowered.common.util.VecHelper;
 import org.spongepowered.common.world.WorldManager;
 import org.spongepowered.common.world.storage.SpongePlayerDataHandler;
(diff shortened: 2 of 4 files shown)


src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java
```diff
@@ -301,6 +301,15 @@ public CommandSource asCommandSource() {
         return this;
     }
 
+    public void addChatMessage(IChatComponent component) {
+        SpongeChatHelper.sendMessage(component, MessageChannel.TO_CONSOLE, getConsole(), true);
+    }
+
+    @Override
+    public void sendMessage(Text message) {
+        logger.info(SpongeTexts.toComponent(message).getUnformattedText());
```
kashike: I thought this (sending messages to server) was already handled, with colours?

src/main/java/org/spongepowered/common/text/chat/SpongeChatHelper.java
```diff
@@ -0,0 +1,58 @@
...
+import org.spongepowered.api.event.message.MessageEvent.MessageFormatter;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.channel.MessageChannel;
+import org.spongepowered.api.text.chat.ChatTypes;
+import org.spongepowered.common.SpongeImpl;
+import org.spongepowered.common.text.SpongeTexts;
+
+import java.util.Optional;
+
+public class SpongeChatHelper {
```
gabizou: Name it `ChatUtil` to be in line with the rest of utility classes in common.

simon816:
Updated.
I removed the event when a message is sent to the server because of difficulties with infinite loops.

Wundero:
Out of curiosity, if for instance the player sends a chat message to the server (which in itself fires a messageevent), does that mean that two messageevents are fired per chat message sent? If so I think there should be some way of differentiating incoming and outgoing messages, either in subclasses or abstract methods, so as to prevent conflicts with incoming events.

ryantheleach:
@Wundero my understanding is when the player is sending a message to the server no chat event is fired. at that stage it's still a string, and I think might fall under 

https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/event/command/SendCommandEvent.java

then once the message has been processed by the server and determined to not be a command, the server then sends the message to various players using MessageEvents

Before this change, MessageEvents were only being sent when plugins sent messages to a MessageChannel or chat messages were being sent.

After this commit, MessageEvents are also sent when the base game sends chat messages to the players.

Small change, I think.

simon816:
Any further comments on this?

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -403,6 +404,11 @@ public void sendMessage(ChatType type, Text message) {
         this.playerNetServerHandler.sendPacket(new S02PacketChat(component, ((SpongeChatType) type).getByteId()));
     }
 
+    @Overwrite
```
gabizou: Add overwrite docs.

src/main/java/org/spongepowered/common/mixin/core/server/MixinServerConfigurationManager.java
```diff
@@ -758,4 +759,9 @@ private void onWritePlayerFile(EntityPlayerMP playerMP, CallbackInfo callbackInf
         SpongePlayerDataHandler.savePlayer(playerMP.getUniqueID());
     }
 
+    @Overwrite
```
gabizou: And here.

gabizou:
Please update the target branch to bleeding (Edit this PR and change the target branch) and reopen.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -382,6 +382,11 @@ public void sendMessage(ChatType type, Text message) {
         this.connection.sendPacket(new SPacketChat(component, ((SpongeChatType) type).getByteId()));
     }
 
+    @Overwrite
```
kashike: Missing overwrite documentation.

src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
```diff
@@ -873,4 +874,10 @@ private WorldInfo onGetWorldInfo(WorldServer overworld, EntityPlayerMP player) {
         // Check the world info of the current world instead of overworld world info
         return player.worldObj.getWorldInfo();
     }
+
+    @Overwrite
```
kashike: Missing overwrite documentation.

src/main/java/org/spongepowered/common/text/chat/ChatUtil.java
```diff
@@ -0,0 +1,58 @@
...
+import org.spongepowered.api.event.message.MessageEvent.MessageFormatter;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.channel.MessageChannel;
+import org.spongepowered.api.text.chat.ChatTypes;
+import org.spongepowered.common.SpongeImpl;
+import org.spongepowered.common.text.SpongeTexts;
+
+import java.util.Optional;
+
+public class ChatUtil {
```
kashike: `final`, `private` constructor

simon816:
Updated

src/main/java/org/spongepowered/common/text/chat/ChatUtil.java
```diff
@@ -0,0 +1,58 @@
...
+        MessageChannelEvent event;
+        if (isChat) {
+            event = SpongeEventFactory.createMessageChannelEventChat(cause, channel, Optional.of(channel), formatter, raw, false);
+        } else {
+            event = SpongeEventFactory.createMessageChannelEvent(cause, channel, Optional.of(channel), formatter, false);
+        }
+        if (!SpongeImpl.postEvent(event) && !event.isMessageCancelled() && event.getChannel().isPresent()) {
+            event.getChannel().get().send(source, event.getMessage(), isChat ? ChatTypes.CHAT : ChatTypes.SYSTEM);
+        }
+    }
```
liach, line 60: Newline

simon816:
I have fixed the javadoc issues mentioned

src/main/java/org/spongepowered/common/text/chat/ChatUtil.java
```diff
@@ -0,0 +1,62 @@
...
+import org.spongepowered.api.event.message.MessageEvent.MessageFormatter;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.channel.MessageChannel;
+import org.spongepowered.api.text.chat.ChatTypes;
+import org.spongepowered.common.SpongeImpl;
+import org.spongepowered.common.text.SpongeTexts;
+
+import java.util.Optional;
+
+final public class ChatUtil {
```
liach: The convention is `public final class Xxx`.

src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
```diff
@@ -873,4 +874,19 @@ private WorldInfo onGetWorldInfo(WorldServer overworld, EntityPlayerMP player) {
...
     }
+
+    /**
+     * @author simon816 - 14th November, 2016
+     *
+     * @reason Redirect chat broadcasts to fire an event for the message. Each receiver
+     * (typically a player) will handle the actual sending of the message.
+     *
+     * @param component The message
+     * @param isChat is chat
```
liach: `Whether this is a chat message`

Wundero:
Out of curiosity, why are the methods being added "addChatMessage", whereas in the current Sponge mixins, it is "sendMessage"?

simon816:
@Wundero because `addChatMessage` is the MCP name for the internal methods. I need to overwrite them to fire the event

Edit: Looks like the MCP name changed. I have updated it now

simon816:
@SpongePowered/developers any further comments? Otherwise this is good to go

amaranth:
You should change isChat to isSystem to match how MCP works and make usage of this variable less confusing (I thought you had a bug with !isSystem). Also what is up with the random extra this$ diffs? Otherwise it seems alright to me.

kashike:
> Also what is up with the random extra this$ diffs? Otherwise it seems alright to me.

Code style.

simon816:
Will merge today if there are no further outstanding issues. This should be good to go in the 5.x branch too.

PR #824, opened 2016-07-02
Add redirection to the Player instance

pie-flavor:
Currently, many methods of User are unimplemented. This checks whether the Player exists, and if so, redirects the method calls to those of the Player. If the player is offline, it remains unimplemented.

### DIFF
--- a/src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
+++ b/src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
@@ -183,79 +183,83 @@ public class SpongeUser implements ArmorEquipable, Tamer, DataSerializable, Carr
                 .set(DataQueries.USER_SPAWNS, this.spawnLocations);
     }
 
+    private Player getUserIfOnlineForInventory() {
+        return this.self.getPlayer().orElseThrow(() -> new UnsupportedOperationException("User is offline, offline inventory API yet to be implemented"));
+    }
+
     @Override
     public boolean canEquip(EquipmentType type) {
-        throw new UnsupportedOperationException(); // TODO Inventory API
+        return getUserIfOnlineForInventory().canEquip(type); // TODO Inventory API
     }
 
     @Override
     public boolean canEquip(EquipmentType type, ItemStack equipment) {
-        throw new UnsupportedOperationException(); // TODO Inventory API
+        return getUserIfOnlineForInventory().canEquip(type, equipment); // TODO Inventory API
     }
 
     @Override
     public Optional<ItemStack> getEquipped(EquipmentType type) {
-        throw new UnsupportedOperationException(); // TODO Inventory API
+        return getUserIfOnlineForInventory().getEquipped(type); // TODO Inventory API
     }
 
     @Override
     public boolean equip(EquipmentType type, ItemStack equipment) {
-        throw new UnsupportedOperationException(); // TODO Inventory API
+        return getUserIfOnlineForInventory().equip(type, equipment); // TODO Inventory API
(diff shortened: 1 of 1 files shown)


ryantheleach:
You could add a message saying "User is offline, offlineInventoryAPI yet to be implemented"

pie-flavor:
@ryantheleach Good idea. Gives me an excuse to use NotImplementedException too.

kashike:
or `UnsupportedOperationException`.

src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
```diff
@@ -185,77 +185,137 @@ public DataContainer toContainer() {
 
     @Override
     public boolean canEquip(EquipmentType type) {
-        throw new UnsupportedOperationException(); // TODO Inventory API
+        if (this.self.isOnline()) {
```
kashike: This should work, and make it cleaner:
`this.self.getPlayer().orElseThrow(() -> new UnsupportedOperationException("User is offline, offline inventory API yet to be implemented")).canEquip(type)`

pie-flavor:
@kashike Gotcha.

gabizou:
The thing is, the current player may not actually be connected any more, so still would want to check `getPlayer()`.

pie-flavor:
@gabizou ?? confused
if the player is present, it does the thing
if it's not, it throws the exception

me4502:
In this case, seeing as the offline inventory API is intended to be implemented in the future, NotImplementedException makes the most sense IMO.

pie-flavor:
anything holding this back?

src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
```diff
@@ -183,79 +184,83 @@ public DataContainer toContainer() {
                 .set(DataQueries.USER_SPAWNS, this.spawnLocations);
     }
 
+    private NotImplementedException createException() {
```
ST-DDT: Maybe this should be named createOfflinePlayerException or notImplementedException, because currently its quite hard to guess the meaning of this Exception from just its name.

pie-flavor:
i guess i screwed up the git. crud.

src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
```diff
@@ -183,79 +184,83 @@ public DataContainer toContainer() {
...
     }
 
+    private NotImplementedException createNotImplementedException() {
+        return new NotImplementedException("User is offline, offline inventory API yet to be implemented");
+    }
+
     @Override
     public boolean canEquip(EquipmentType type) {
-        throw new UnsupportedOperationException(); // TODO Inventory API
+        return this.self.getPlayer().orElseThrow(this::createNotImplementedException).canEquip(type); // TODO Inventory API
```
Minecrell: You can put the complete `this.self.getPlayer().orElseThrow(...)` part into an extra method (e.g. `getPlayerIfOnline()`, then you have even less to write for each method.

src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
```diff
@@ -183,79 +184,83 @@ public DataContainer toContainer() {
                 .set(DataQueries.USER_SPAWNS, this.spawnLocations);
     }
 
+    private Player getPlayerIfOnline() {
```
kashike: If the exception message is going to mention the inventory, this method should be named differently.

Issue #844, opened 2016-07-13
SpongeCommand Formatting [Minor]


Starza4TW:
Figured this was too small for a PR, just let you guys fix instead.

Just noticed that the /sponge command does not constantly use the newline escape character. 
![2016-07-13_16 42 00](https://cloud.githubusercontent.com/assets/2940976/16819079/c057eef0-4918-11e6-8339-3985331943c6.png)

Unless there is a lack of space, there should be a "\n" added to the end of audit and plugins descriptions in the sponge command's extended description so that it matches the other subcommands.

Also, the command itself has the description "Text description", which probably should be changed.

Class: https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/command/SpongeCommand.java

Zidane:
You're free to PR this :) if you would still like too :).

Starza4TW:
Fixed with 84aee0e1a4a27ffb60b50da75f74515975755a37

PR #845, opened 2016-07-14
Fix SpongeCommand Descriptions & Formatting [Minor]

Starza4TW:
Added newline escape characters at the end of the descriptions for two of the subcommands of the sponge command, and added an actual description to the sponge command.

Fixes Issue #844


Starza4TW:
Fixed with 84aee0e1a4a27ffb60b50da75f74515975755a37

PR #851, opened 2016-07-21
[WIP] Implement of Refactoring Block Change API

gabizou:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1315) | **SpongeCommon** | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/770)

This is a more focused PR at dealing with several issues at once, but for the sake of having not had a decent PR description in a while, I'll rehash the conversational description that we used to have a while back.

#### So what's changing?

In short, if a plugin developer is calling any API method that changes something, an event should be thrown. Traditionally, we used to have the inability to throw events for virtually everything, however, at this point in time, it's well within our means to throw events for all `setBlock`, `createEntity`, `spawnEntity`, etc. 

#### But why force plugin developers to supply a `Cause`? Don't developers know to throw events to check for things before they make a change?

Yes and no. It's been useful for developers to throw "dummy" events to check for permissions or block modifications before finalizing a change, and for other cases, to throw dummy spawn events before an entity is indeed spawned into the world. The problem with all of this is that it's entirely up to the developer whether to throw the events, or not throw the events, and other plugins are none the wiser. This is especially complicated by the fact that Sponge has a very complex tracking system that performs various actions based on pre-defined expectations. The change resolves a great many event cause tracking determining aspects, such as a synchronous task set up by plugin X to make a tree due to player Y. 

#### Could you explain why this wasn't an issue in API 4.X?

The issue lies in the API and the implementation. Originally planned, the CauseTracker in 1.8.9 does a very good job at determining what is causing X, Y, and Z happening, when it's *expected*. Plugins that call our methods to spawn entities are bypassing the CauseTracker, but only to a limited extent, there's no `Cause` required or provided, and other plugins have no knowledge of why an entity was being spawned. Or for that matter, the CauseTracker can flip out if a multitude of block changes take place in a synchronized task, but this isn't the fault of the CauseTracker, more so our contracts we set forth in the API. If I had to pinpoint why the change is necessary, it's that the CauseTracker is more of a mechanism to determine causes for which things happen.

#### So what actually is changing API wise?

In short, any method that changes a block in a `World` (or `Extent`) will require a `Cause`. Virtually anything that touches changing blocks will now provide a `Cause` to the system, and the system will be able to apply that provided `Cause` to the ensuing `ChangeBlockEvent` (unless this is a populator, which can simply provide a recycled `Cause` as shown in the skylands populators and generators). Data API has gained an overloaded method that will provide a `Cause` for the cases that the `#offer(Key,Object,Cause)` will change a block in a `World`, however, normal offers that simply change data on a `TileEntity` can still use the traditional `#offer(Key,Object)` method. It is understandable at this point as well that the other `DataHolder` implementations may throw Data related events with the provided `Cause`.

#### What else? What's this `BlockChangeFlag`?

Ah, so that is a brainchild of the various `boolean` flags that we used to require plugin developers to provide, namely to show that there are various *options* to perform a block change. The reasoning for it is that previously, we could only allow developers to *NOT* notify neighboring blocks on a block change, but we didn't allow it to *NOT* perform block physics on placement. This caused issues with some `BlockSnapshot#restore`s, since the restore process would always perform block physics due to the nature of the implementation. At this point, instead of adding yet another `boolean` flag and set of overloads, it was simpler to use an object to determine whether neighbor notifications should take place, and if block physics should take place. In the future, more flags may be added as necessary, and this time without breaking the API ;).

#### With all of the `Cause` usage, why does this look like the implementation is leaking into the API?

It is and isn't. Due to the nature of our tracking system, this is the best way that we can achieve a proper API contract of "all your changes throw events". Likewise, this eliminates the issues of being able to deterministically resolve what is performing changes to the world, without some obscure `Cause` containing a `Task` and or a `PluginContainer`. This also eliminates any sort of tracking required on our part to *determine* what caused those changes. This ends up with the API being more contractually sound with the end result that plugins no longer have to throw dummy events and the implementation has no requirement to track and determine what is going on throughout different phases.

Any further questions can be asked below.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/processor/multi/tileentity/FurnaceDataProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/multi/tileentity/FurnaceDataProcessor.java
@@ -34,7 +34,9 @@ import org.spongepowered.api.data.key.Key;
 import org.spongepowered.api.data.key.Keys;
 import org.spongepowered.api.data.manipulator.immutable.tileentity.ImmutableFurnaceData;
 import org.spongepowered.api.data.manipulator.mutable.tileentity.FurnaceData;
+import org.spongepowered.api.event.cause.Cause;
 import org.spongepowered.api.world.World;
+import org.spongepowered.common.SpongeImpl;
 import org.spongepowered.common.data.manipulator.mutable.tileentity.SpongeFurnaceData;
 import org.spongepowered.common.data.processor.common.AbstractTileEntityDataProcessor;
 
@@ -47,6 +49,8 @@ public class FurnaceDataProcessor extends AbstractTileEntityDataProcessor<TileEn
     public FurnaceDataProcessor() {
         super(TileEntityFurnace.class);
     }
+
+    private final Cause cause = Cause.source(SpongeImpl.getPlugin()).build();
 
     @Override
     protected boolean doesDataExist(TileEntityFurnace tileEntity) {
@@ -76,7 +80,7 @@ public class FurnaceDataProcessor extends AbstractTileEntityDataProcessor<TileEn
         if (needsUpdate) {
             final World world = (World) tileEntity.getWorld();
             world.setBlockType(tileEntity.getPos().getX(), tileEntity.getPos().getY(),
-                    tileEntity.getPos().getZ(), maxBurnTime > 0 ? BlockTypes.LIT_FURNACE : BlockTypes.FURNACE);
+                    tileEntity.getPos().getZ(), maxBurnTime > 0 ? BlockTypes.LIT_FURNACE : BlockTypes.FURNACE, this.cause);
             tileEntity = (TileEntityFurnace) tileEntity.getWorld().getTileEntity(tileEntity.getPos());
         }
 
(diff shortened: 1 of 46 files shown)


src/main/java/org/spongepowered/common/data/processor/multi/tileentity/FurnaceDataProcessor.java
```diff
@@ -48,6 +50,8 @@ public FurnaceDataProcessor() {
         super(TileEntityFurnace.class);
     }
 
+    private final Cause cause = Cause.source(SpongeImpl.getPlugin()).build();
```
kashike: Perhaps store this `Cause` in `SpongeImpl` instead, rather than having the same thing in many places?

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -969,6 +965,17 @@ public boolean setBlockState(BlockPos pos, IBlockState newState, int flags) {
         }
     }
 
+    @Override
+    public boolean setBlockState(BlockPos pos, IBlockState state, BlockChangeFlag flag) {
+        if (!this.isValid(pos)) {
+            return false;
+        } else if (this.worldInfo.getTerrainType() == WorldType.DEBUG_WORLD) { // isRemote is always false since this is WorldServer
+        return false;
```
kashike: bad indentation

PR #860, opened 2016-07-26
Fix MinecraftCommandWrapper to allow for custom command sources

ZephireNZ:
Based on @JBYoshi's comment [here](https://github.com/SpongePowered/SpongeCommon/commit/2a2f8184d67b64e562e6d2eb11335d2f3e0df764#commitcomment-17244727) and due to this issue cropping up [here](https://forums.spongepowered.org/t/error-with-custom-commandsource/13787).

I have changed `MinecraftCommandWrapper` so that it allows plugins to implement and pass their own sources to vanilla minecraft commands.

Test:
```java
@Plugin(id = "test-plugin", name = "Test Plugin", version = "1.0.0")
public class TestPlugin {

    @Inject
    private Logger log;

    @Inject
    private PluginContainer container;

    @Listener
    public void onStarted(GameStartedServerEvent event) {
        CommandSource custom = new CustomSource();

        Sponge.getCommandManager().process(custom, "time set night");
    }

    class CustomSource implements CommandSource {

        @Override public String getIdentifier() {
            return "custom-source";
        }

        @Override public void sendMessage(Text message) {
            log.info("[" + getName() + "] " + message.toPlain());
        }

        @Override public String getName() {
            return "CustomSource";
        }

        @Override public Set<Context> getActiveContexts() {
            return ImmutableSet.of();
        }

        @Override public Optional<CommandSource> getCommandSource() {
            return Optional.empty();
        }

        @Override public SubjectCollection getContainingCollection() {
            return null;
        }

        @Override public MessageChannel getMessageChannel() {
            return MessageChannel.TO_ALL;
        }

        @Override public SubjectData getSubjectData() {
            return null;
        }

        @Override public void setMessageChannel(MessageChannel channel) {

        }

        @Override public SubjectData getTransientSubjectData() {
            return null;
        }

        @Override public boolean hasPermission(Set<Context> contexts, String permission) {
            return true;
        }

        @Override public Tristate getPermissionValue(Set<Context> contexts, String permission) {
            return Tristate.TRUE;
        }

        @Override public boolean isChildOf(Set<Context> contexts, Subject parent) {
            return false;
        }

        @Override public List<Subject> getParents(Set<Context> contexts) {
            return ImmutableList.of();
        }

        @Override public Optional<String> getOption(Set<Context> contexts, String key) {
            return Optional.empty();
        }
    }
}
```
Result: 
```
[14:51:40] [Server thread/INFO]: [CustomSource: Set the time to 13000]
[14:51:40] [Server thread/INFO] [test-plugin]: [CustomSource] Set the time to 13000
```

### DIFF
--- a/src/main/java/org/spongepowered/common/command/WrapperICommandSender.java
+++ b/src/main/java/org/spongepowered/common/command/WrapperICommandSender.java
@@ -35,6 +35,7 @@ import net.minecraft.util.text.TextComponentString;
 import net.minecraft.world.World;
 import org.spongepowered.api.command.CommandSource;
 import org.spongepowered.api.world.Locatable;
+import org.spongepowered.common.SpongeImpl;
 import org.spongepowered.common.interfaces.IMixinCommandSource;
 import org.spongepowered.common.text.SpongeTexts;
 import org.spongepowered.common.util.VecHelper;
@@ -92,7 +93,7 @@ public class WrapperICommandSender implements ICommandSender {
         if (this.source instanceof Locatable) {
             return (World) ((Locatable) this.source).getWorld();
         }
-        return null;
+        return SpongeImpl.getServer().getEntityWorld(); // Use overworld as default
     }
 
     @Override
@@ -115,11 +116,7 @@ public class WrapperICommandSender implements ICommandSender {
 
     @Override
     public MinecraftServer getServer() {
-        final World world = getEntityWorld();
-        if (world != null) {
-            return world.getMinecraftServer();
-        }
-        return null;
+        return getEntityWorld().getMinecraftServer();
     }
 
     public static ICommandSender of(CommandSource source) {


src/main/java/org/spongepowered/common/command/WrapperICommandSender.java
```diff
@@ -119,7 +120,7 @@ public MinecraftServer getServer() {
         if (world != null) {
             return world.getMinecraftServer();
         }
-        return null;
+        return SpongeImpl.getServer();
```
ZephireNZ: Agreement on whether this should be done? If this returns null then vanilla commands cannot get the `CommandHandler` for the world and so won't work as intended. This is to provide a default server to execute the command on if a located source isn't supplied.

ZephireNZ: In 99% of cases there will only be one `MinecraftServer` anyway - right?

ZephireNZ:
Note that i haven't yet ran the test - had to run to a class!

ZephireNZ:
Test complete, and passed :)

src/main/java/org/spongepowered/common/command/WrapperICommandSender.java
gabizou: Depends on the state of the game, if there is no server, there is no server 😄

ZephireNZ: In which case a command shouldn't be run anyway, and doing so will result in an `InvalidStateException` - perfectly reasonable for that circumstance.

ryantheleach:
Make send message call the getter for the name, rather then statically send custom? maybe require plugin id?

ryantheleach:
Also how does this interact with other plugins / proxying command sources?

src/main/java/org/spongepowered/common/command/WrapperICommandSender.java
simon816: `getEntityWorld` shouldn't ever return `null` so that was wrong in the first place

ZephireNZ:
@emil18250 Best thing is to make either a new issue or a post on [the forums](https://forums.spongepowered.org) and we can take a look at that - make sure you post your log too.

This is only relevant for Minecraft 1.9+

emil18250:
@ZephireNZ Ok thanks , i post in the sponge support forum (https://forums.spongepowered.org/t/custom-command-mod-make-kick-players/14016)

ZephireNZ:
@Deamon5550 I noticed you just committed effectively the same thing with 2efbad7, only difference I can see is that I've (somewhat) fixed up the WrapperICommandSender. Maybe commit just that so this can be closed?

Zidane:
@ZephireNZ 

You meant @Minecrell

ZephireNZ:
Durr yes, oops

Minecrell:
With my changes there is always a `MinecraftServer` available. Are the changes in here actually still needed then?

JBYoshi:
@Minecrell Just the changes in `WrapperICommandSender.getServer()` so it can't return `null`.

ZephireNZ:
I'm in the process of testing new changes but CauseTracking breaks because it assumes that `NamedCause.SOURCE` is a `CommandSender`. `SpongeCommandManager` is the culprit - it passes the `CommandSource` directly to the cause. While this is fine for mixed in vanilla objects, custom objects will fail as they do not, by default, implement ICommandSender.

@gabizou You changed this section just recently, I'm not sure where to start so could you fix it up for me? :) Or tell me where to change and I can add it to this PR

ZephireNZ:
Success!
```
[14:51:40] [Server thread/INFO]: [CustomSource: Set the time to 13000]
[14:51:40] [Server thread/INFO] [test-plugin]: [CustomSource] Set the time to 13000
```

ZephireNZ:
OK @gabizou, I've updated GeneralPhase to try and get a `CommandSource` and wrap it if it's unable to find a `CommandSender` raw. Is that OK?

I assume I shouldn't wrap it at the CommandManager - if i did that, then plugins wouldn't be able to cast the source to `CommandSource`.

src/main/java/org/spongepowered/common/event/tracking/phase/GeneralPhase.java
```diff
@@ -147,7 +149,14 @@ public GeneralPhase addChild(TrackingPhase child) {
...
             final ICommandSender sender = phaseContext.getSource(ICommandSender.class)
-                    .orElseThrow(TrackingUtil.throwWithContext("Expected to be capturing a Command Sender, but none found!", phaseContext));
+                    // The sender could be a custom CommandSource, so we check for that too
+                    .orElseGet(() -> {
+                        CommandSource source = phaseContext.getSource(CommandSource.class)
+                                .orElseThrow(TrackingUtil.throwWithContext("Expected to be capturing a Command Sender, but none found!",
+                                        phaseContext));
+                        return WrapperICommandSender.of(source);
+                        }
+                    );
```
JBYoshi: I think these should be on one line, using the indentation of the second line.

`});`

src/main/java/org/spongepowered/common/event/tracking/phase/GeneralPhase.java
```diff
@@ -147,7 +149,13 @@ public GeneralPhase addChild(TrackingPhase child) {
     public void unwind(CauseTracker causeTracker, IPhaseState state, PhaseContext phaseContext) {
         if (state == State.COMMAND) {
             final ICommandSender sender = phaseContext.getSource(ICommandSender.class)
```
Minecrell: Given that the Sponge command manager will always insert the `CommandSource` as source and the code does nothing except some `instanceof` checks, it should be safe to just change this to `CommandSource.class` directly here. (Because the source will **always** be a `CommandSource` but not necessarily always a `ICommandSender`)

ZephireNZ: This was added to bleeding just recently, so I'll just drop that part of the PR now.

ZephireNZ:
@Minecrell removed the code relating to phases as that was fixed upstream. Ready to merge :)

PR #882, opened 2016-08-14
Implement Schematics API

Deamon5550:
Implements spongepowered/SpongeApi#1330

This implements an api for creating `ArchetypeVolume`s which are copies of regions of the world and may be serialized to disk in one of two schematic formats (The MCEdit format here called the legacy format, and an updated format for better mod compatability).

An example plugin using this can be found [In the cookbook](https://github.com/SpongePowered/Cookbook/blob/schematics/Plugin/CopyPasta/src/main/java/org/spongepowered/cookbook/plugin/CopyPasta.java#L47).

The gist of the usage is
```java
// copying
ArchetypeVolume volume = player.getWorld().createArchetypeVolume(min, max, player.getLocation().getPosition().toInt());

// pasting
volume.apply(player.getLocation(), BlockChangeFlag.ALL, Cause.of(NamedCause.of("plugin", this.plugin), NamedCause.source(player)));

// saving
Schematic schematic = Schematic.builder().volume(data.getClipboard()).metaValue(Schematic.METADATA_AUTHOR, player.getName()).metaValue(Schematic.METADATA_NAME, name).paletteType(PaletteTypes.LOCAL).build();
schematicData = DataTranslators.SCHEMATIC.translate(schematic);
DataFormats.NBT.writeTo(new GZIPOutputStream(new FileOutputStream(outputFile)), schematicData);
```


src/main/java/org/spongepowered/common/entity/SpongeEntityArchetypeBuilder.java
```diff
@@ -0,0 +1,177 @@
...
+            this.entityData = new MemoryDataContainer();
+        }
+        SpongeDataManager.getInstance().getRawNbtProcessor(NbtDataTypes.TILE_ENTITY, key)
+                .ifPresent(processor -> processor.offer(this.entityData, value));
+        return this;
+    }
+
+    @Override
+    public EntityArchetype build() {
+        checkState(!this.entityData.isEmpty());
```
ZephireNZ: This should be removed - it's absolutely fine for entity data to be empty. It's only during serialization to native format that it should not be empty (ie adding the `EntityType` field)

src/main/java/org/spongepowered/common/entity/SpongeEntityArchetype.java
```diff
@@ -0,0 +1,145 @@
...
+        return DataVersions.EntityArchetype.BASE_VERSION;
+    }
+
+    @Override
+    public DataContainer toContainer() {
+        return new MemoryDataContainer()
+                .set(DataQueries.EntityArchetype.ENTITY_TYPE, this.type)
+                .set(DataQueries.EntityArchetype.ENTITY_DATA, this.data)
+                ;
+    }
```
ZephireNZ: Can we get a standard way of serialising to MC format with EntityType within the data. If not I'll add it to my PR for mob spawners which use archetypes in place of located snapshots.

Deamon5550: The data in the stored nbt compound is the data in MC format. This toContainer method should probably not actually be nesting the data inside that tag honestly since there isn't really any use to doing that if this is the only place using that format. Eh @gabizou?

ZephireNZ: Yes, when converting from Entity to EntityArchetype all data including the ID is in archetype's data. However when built using a builder it is not as far as I can tell.

src/main/java/org/spongepowered/common/entity/SpongeEntitySnapshot.java
```diff
@@ -466,6 +467,11 @@ public NBTTagList newDoubleNBTList(double ... numbers) {
     }
 
     @Override
+    public EntityArchetype createArchetype() {
+        throw new UnsupportedOperationException();
```
ZephireNZ: Shouldn't this be a simple call to `EntityArchetype.builder()`?

ZephireNZ: Or should I say`new SpongeEntityArchetypeBuilder()` to avoid overhead ;)

Deamon5550: oh must have missed this

src/main/java/org/spongepowered/common/data/nbt/validation/SpongeEntityValidator.java
```diff
@@ -0,0 +1,48 @@
...
+public class SpongeEntityValidator implements RawDataValidator {
+
+    @Override
+    public ValidationType getValidationType() {
+        return Validations.ENTITY;
+    }
+
+    @Override
+    public boolean validate(NBTTagCompound view) {
+        // TODO implement this bit, the view needs various things, but it most definteily cannot
```
DDoS: Forgot about this?

src/main/java/org/spongepowered/common/world/schematic/GlobalPalette.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Optional;
+
+public class GlobalPalette implements Palette {
+
+    public static GlobalPalette instance = new GlobalPalette();
+
+    private final int length;
+
+    private GlobalPalette() {
+        // TOOD this length is incorrect
```
DDoS: Could you explain the issue here a bit more?

src/main/java/org/spongepowered/common/data/nbt/validation/SpongeEntityValidator.java
gabizou: The Data implementation for schematics of manipulating raw not is not going to be implemented just yet as it is not a requirement for merging Schematics. However, it is nonetheless a system defined to allow for further implementation once the branch is merged.

src/main/java/org/spongepowered/common/world/schematic/GlobalPalette.java
Deamon5550: oh yeah this is actually a todo that I need to fix before merging unlike most of the ones that I wrote. The number of blockstates is not actually the length because they are not packed tightly I need to somehow search through the entire block state list for the one with the highest id.

ZephireNZ:
```
Caused by: java.lang.IllegalArgumentException: Could not find a Supplier for the provided builder class: org.spongepowered.api.entity.EntityArchetype.Builder
	at org.spongepowered.common.registry.SpongeGameRegistry.createBuilder(SpongeGameRegistry.java:266) ~[SpongeGameRegistry.class:?]
	at org.spongepowered.api.entity.EntityArchetype.builder(EntityArchetype.java:45) ~[EntityArchetype.class:?]
	at org.spongepowered.api.entity.EntityArchetype.of(EntityArchetype.java:54) ~[EntityArchetype.class:?]
	at org.spongepowered.common.registry.type.world.gen.DungeonMobRegistryModule.put(DungeonMobRegistryModule.java:55) ~[DungeonMobRegistryModule.class:?]
	at org.spongepowered.common.registry.type.world.gen.DungeonMobRegistryModule.registerDefaults(DungeonMobRegistryModule.java:39) ~[DungeonMobRegistryModule.class:?]
	at org.spongepowered.common.registry.util.RegistryModuleLoader.tryModulePhaseRegistration(RegistryModuleLoader.java:61) ~[RegistryModuleLoader.class:?]
	... 15 more
```

Someone forgot to register ;)

ZephireNZ:
Mind adding this to `AbstractArchetype`? Would make things in debug a bit more readable than `SpongeEntityArchetype@deadbeef`

src/main/java/org/spongepowered/common/block/SpongeTileEntityArchetypeBuilder.java
```diff
@@ -0,0 +1,197 @@
...
+        SpongeDataManager.getInstance().getRawNbtProcessor(NbtDataTypes.TILE_ENTITY, key)
+                .ifPresent(processor -> processor.offer(this.tileData, value));
+        return this;
+    }
+
+    @Override
+    public TileEntityArchetype build() {
+        checkState(this.blockState != null, "BlockState cannot be null!");
+        checkState(this.tileEntityType != null, "TileEntityType cannot be null!");
+        if (this.tileData != null) {
```
thejohncrafter: I do not really understand why you are performing this check...
Because the problem is that this means the build Archetype will not contain the source block's tile data (i.e. a Sign won't keep its text).

So is this a choice or a mistake ? I think this is related to what @ZephireNZ said : 

> When converting from Entity to EntityArchetype all data including the ID is in archetype's data. However when built using a builder it is not as far as I can tell.

simon816: It definatelly looks like that's supposed to be a `== null` check

thejohncrafter: This part of the code looks weird, actually...
It checks that `this.tileEntityType != null` even if it is `@Nullable`. I think that they wanted to check `this.tileData` isn't null, and replace `this.tileEntityType` by a default value if it null.

src/main/java/org/spongepowered/common/world/schematic/SpongeSchematicBuilder.java
```diff
@@ -0,0 +1,158 @@
...
+    @Override
+    public Builder volume(ArchetypeVolume volume) {
+        this.volume = volume;
+        return this;
+    }
+
+    @Override
+    public Builder volume(Extent volume) {
+        this.view = volume;
+        return null;
```
thejohncrafter: Why does this method return `null` (ln. 68) ? I think this is an error as the javadoc says :

> Returns This builder, for chaining

Fortunately it's not too hard to find a work around : 
```
Builder b = Schematic.builder();
b.volume(extent);
schematic = b.build();
```
:wink:

src/main/java/org/spongepowered/common/world/schematic/SpongeSchematicBuilder.java
```diff
@@ -0,0 +1,158 @@
...
+            size = this.view.getBlockSize();
+        }
+        Map<Vector3i, TileEntityArchetype> tiles = this.volume.getTileEntityArchetypes();
+        if (this.metadata == null) {
+            this.metadata = new MemoryDataContainer();
+        }
+        for (Map.Entry<String, Object> entry : this.metaValues.entrySet()) {
+            this.metadata.set(DataQuery.of(".", entry.getKey()), entry.getValue());
+        }
+        if (this.volume == null) {
```
thejohncrafter: Another issue here... :tada: 
At line 141, there is a block handling the case of `this.volume` being null, but that's too late ! :laughing: 
Actually, `this.volume` is accessed before, at line 134, which throws a `NullPointerException`...
I think that actually line of code should be placed in a `else` statement, at line 154.

PR #887, opened 2016-08-18
Add Structure Saving Whitelist/Blacklist

gabizou:
It is possible to prevent some various structures from getting saved to file on extremely large worlds, wasting space. This was a feature initially requested by @bloodmc and to match Paper's patch.

Yes, I know this needs to be rebased and cleaned up since it hasn't been touched in several months now.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/optimization/world/gen/structure/MixinMapGenStructure_Structure_Saving.java
+++ b/src/main/java/org/spongepowered/common/mixin/optimization/world/gen/structure/MixinMapGenStructure_Structure_Saving.java
@@ -1,0 +1,38 @@
+package org.spongepowered.common.mixin.optimization.world.gen.structure;
+
+import it.unimi.dsi.fastutil.objects.Object2BooleanArrayMap;
+import it.unimi.dsi.fastutil.objects.Object2BooleanMap;
+import net.minecraft.world.World;
+import net.minecraft.world.WorldSavedData;
+import net.minecraft.world.gen.MapGenBase;
+import net.minecraft.world.gen.structure.MapGenStructure;
+import net.minecraft.world.gen.structure.MapGenStructureData;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+import org.spongepowered.common.SpongeImpl;
+import org.spongepowered.common.interfaces.world.IMixinWorldServer;
+
+@Mixin(MapGenStructure.class)
+public abstract class MixinMapGenStructure_Structure_Saving extends MapGenBase {
+
+
+    @Redirect(method = "initializeStructureData", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/World;loadItemData(Ljava/lang/Class;Ljava/lang/String;)Lnet/minecraft/world/WorldSavedData;"))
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/config/SpongeConfig.java
```diff
@@ -365,6 +365,8 @@ public void setConfigEnabled(boolean configEnabled) {
         private WorldCategory world = new WorldCategory();
         @Setting
         private TimingsCategory timings = new TimingsCategory();
+        @Setting
+        private StructureCategory structures = new StructureCategory();
```
kashike: This should be a child of the world category, in my opinion.

src/main/java/org/spongepowered/common/mixin/optimization/world/gen/structure/MixinMapGenStructure_Structure_Saving.java
```diff
@@ -0,0 +1,30 @@
...
+import org.spongepowered.asm.mixin.injection.Redirect;
+import org.spongepowered.common.interfaces.world.IMixinWorld;
+
+@Mixin(MapGenStructure.class)
+public abstract class MixinMapGenStructure_Structure_Saving extends MapGenBase {
+
+    @Redirect(method = "initializeStructureData", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/World;loadItemData(Ljava/lang/Class;Ljava/lang/String;)Lnet/minecraft/world/WorldSavedData;"))
+    private WorldSavedData checkLoadOrDefault(World world, Class<? extends WorldSavedData> structureClass, String structureName) {
+        if (world instanceof IMixinWorld) {
+            if (((IMixinWorld) world).getActiveConfig().getConfig().getStructures().structureMap.get(structureName)) {
```
kashike: This will be `false` as far as I can see - the map has no values (other than `Mineshaft`, which itself is `false`) that are true, and is never populated.

gabizou:
@bloodmc I do remember you requested this PR a while ago, if you don't mind looking this over once more before we merge it in. I'm considering chucking this into the `Optimization` category or something thereof.

src/main/java/org/spongepowered/common/mixin/optimization/world/gen/structure/MixinMapGenStructure_Structure_Saving.java
bloodmc: This value should be cached locally in this mixin and refreshed when the config gets reloaded.

src/main/java/org/spongepowered/common/config/SpongeConfig.java
```diff
@@ -675,6 +680,20 @@ public boolean isEnabled() {
     }
 
     @ConfigSerializable
+    public static class StructureCategory extends Category {
```
bloodmc: This should be moved to optimization category as suggested.

gabizou:
@bloodmc I'm unsure how to cache the value during reloads since the map get structure doesn't have hooks for reloading configs.

Deamon5550:
A better fix that we can do that I was talking about earlier in irc is to prune the nbt data for a structure when that structure is 'complete'. Just disabling saving breaks both the structure generation and the locate command, but 99% of the space in the nbt is the `children` list which can be discarded after the structure is fully formed.

gabizou:
Closing as this has been merged beforehand.

PR #888, opened 2016-08-20
New event. CommandProcessEvent

djxy:
I needed an event to handle the process of a command before and after that happen in the same tick. So I created this new event to handle that.

When some commands are fired by some players, I need to disable some functionalities. /help is one of them.

Currently there is only one event for when a command is send and it is SendCommandEvent. The commandResult isn't good.

https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/command/SpongeCommandManager.java#L258

SpongeAPI: https://github.com/SpongePowered/SpongeAPI/pull/1333

### DIFF
--- a/src/main/java/org/spongepowered/common/command/SpongeCommandManager.java
+++ b/src/main/java/org/spongepowered/common/command/SpongeCommandManager.java
@@ -24,29 +24,19 @@
  */
 package org.spongepowered.common.command;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-import static org.spongepowered.api.command.CommandMessageFormatting.error;
-import static org.spongepowered.api.util.SpongeApiTranslationHelper.t;
-
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Multimap;
 import org.slf4j.Logger;
 import org.spongepowered.api.Sponge;
-import org.spongepowered.api.command.CommandCallable;
-import org.spongepowered.api.command.CommandException;
-import org.spongepowered.api.command.CommandManager;
-import org.spongepowered.api.command.CommandMapping;
-import org.spongepowered.api.command.CommandPermissionException;
-import org.spongepowered.api.command.CommandResult;
-import org.spongepowered.api.command.CommandSource;
-import org.spongepowered.api.command.InvocationCommandException;
+import org.spongepowered.api.command.*;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/command/SpongeCommandManager.java
```diff
@@ -24,29 +24,19 @@
...
 import org.spongepowered.api.Sponge;
-import org.spongepowered.api.command.CommandCallable;
-import org.spongepowered.api.command.CommandException;
-import org.spongepowered.api.command.CommandManager;
-import org.spongepowered.api.command.CommandMapping;
-import org.spongepowered.api.command.CommandPermissionException;
-import org.spongepowered.api.command.CommandResult;
-import org.spongepowered.api.command.CommandSource;
-import org.spongepowered.api.command.InvocationCommandException;
+import org.spongepowered.api.command.*;
```
simon816: No star imports please

src/main/java/org/spongepowered/common/command/SpongeCommandManager.java
```diff
@@ -57,23 +47,17 @@
...
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 
-import javax.activation.CommandMap;
-import javax.annotation.Nullable;
-import javax.inject.Inject;
+import static com.google.common.base.Preconditions.checkNotNull;
```
simon816: Static imports should remain at the top

src/main/java/org/spongepowered/common/command/SpongeCommandManager.java
djxy: My bad, Intellij Idea did it automatically. I will change that.

src/main/java/org/spongepowered/common/command/SpongeCommandManager.java
djxy: Intellij Idea... I will look to disable that.

zml2008:
I don't see any reason to have SendCommandEvent and CommandProcessEvent.Pre be separate -- their functionality is close enough that you should be able to combine them

djxy:
Yes, I talked with @ryantheleach and we thought their functionalities were close. I will combine them.

djxy:
I removed SendCommandEvent and I replaced it by CommandProcessEvent.Pre

PR #899, opened 2016-08-23
More world fixes

rexbut:
I realized the import of many worlds with different versions of MineCraft and here are the problems I encountered :
- Incorrect name of world : 

The folder name can be different from the name of the world. This causes many problems for other functions WorldManager.

I 'm not the only one to notice the problem. ProjectWorlds directly changes the name of the world in level.dat file : https://github.com/trentech/ProjectWorlds/blob/79ef1142c180beaea39efbc42dd798a01770bd22/src/main/java/com/gmail/trentech/pjw/io/WorldData.java
- UUID : NullPointerException
  The function "registerWorldProperties" require the UUID.
- Difficulty : NullPointerException
  The difficulty of a world can be null.

### DIFF
--- a/src/main/java/org/spongepowered/common/world/WorldManager.java
+++ b/src/main/java/org/spongepowered/common/world/WorldManager.java
@@ -582,25 +582,18 @@ public final class WorldManager {
                 SpongeImpl.getLogger().error("Unable to load world [{}]. No world properties was found!", worldName);
                 return Optional.empty();
             }
-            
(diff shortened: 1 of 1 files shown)


rexbut:
Most bugs have already been fixed

ryantheleach:
>"The folder name can be different from the name of the world. "

Intended as far as I'm aware. You can create worlds in single player with duplicate names, and a - is appended to the directory without changing the name, if I recall correctly.

Issue #902, opened 2016-08-26
[1.10.2 - Beta 109] DamageTypes not initialized


lesbleu:
Hi,
When I use the DamageTypes.class during the event DestructEntityEvent, I have this error : 
http://pastebin.com/aRJrF4uc

rexbut:
@gabizou fix the problem in the commit https://github.com/SpongePowered/SpongeCommon/commit/dc02afafbc2ca753539f06214d4357182a1315b9

PR #903, opened 2016-08-26
DamageTypes not initialized

rexbut:
The creation of a "DamageSource" create a "Map" from "DamageTypes" in "DamageSourceToTypeProvider" though they are not yet initialized.

Fixe : https://github.com/SpongePowered/SpongeCommon/issues/902

### DIFF
--- a/src/main/java/org/spongepowered/common/registry/type/event/DamageSourceRegistryModule.java
+++ b/src/main/java/org/spongepowered/common/registry/type/event/DamageSourceRegistryModule.java
@@ -32,13 +32,14 @@ import org.spongepowered.api.registry.util.RegistrationDependency;
 @RegistrationDependency(DamageTypeRegistryModule.class)
 public final class DamageSourceRegistryModule implements RegistryModule {
 
-    public static final net.minecraft.util.DamageSource IGNORED_DAMAGE_SOURCE = new net.minecraft.util.DamageSource("spongespecific").setDamageBypassesArmor().setDamageAllowedInCreativeMode();
+    public static net.minecraft.util.DamageSource IGNORED_DAMAGE_SOURCE;
     public static net.minecraft.util.DamageSource DAMAGESOURCE_POISON;
     public static net.minecraft.util.DamageSource DAMAGESOURCE_MELTING;
 
     @Override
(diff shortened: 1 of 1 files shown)


gabizou:
How exactly does this fix registrations?

rexbut:
The problem is that the "DamageType" returned on death is a "[Dummy](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/event/cause/entity/damage/DamageTypes.java#L44)" : http://pastebin.com/aRJrF4uc


When you create a DamageSource it uses the class "[DamageSourceToTypeProvider](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/registry/provider/DamageSourceToTypeProvider.java#L46)" : [Line 64](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/event/cause/entity/damage/MixinDamageSource.java#L64)

But if the "[DamageSourceToTypeProvider](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/registry/provider/DamageSourceToTypeProvider.java#L46)" class is used too early, the DamageType will not yet be initialized and the "[Map\<String, DamageType\>](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/registry/provider/DamageSourceToTypeProvider.java#L46)" contain "[Dummy](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/util/generator/dummy/DummyObjectProvider.java)"

The first "DamageSource" must be created after the installation of "DamageType" : [Fix](https://github.com/rexbut/SpongeCommon/blob/b2b3a58bd30b80d9c560db8ea2a6fd4c6d15d548/src/main/java/org/spongepowered/common/registry/type/event/DamageSourceRegistryModule.java#L42)

rexbut:
@gabizou fix the problem in the commit https://github.com/SpongePowered/SpongeCommon/commit/dc02afafbc2ca753539f06214d4357182a1315b9

PR #905, opened 2016-08-26
[1.10] Refactor the SpongeTeleportHelper.

dualspiral:
The safe teleport code has been rewritten in response to SpongePowered/SpongeForge#863, as @Zidane suggested this needed a rewrite! The code now:

* Collects all the possible locations for teleport and sorts them by distance from the original target. This will strip out any targets outside of the world border.
* Iterates over all possible locations from the closest to the furthest. For each block:
 * Checks to see if the block is safe for a player's body
 * Checks if the block above is safe for the body
 * Checks if the block below is a safe floor OR safe for the body
 * If the above is safe for the body but not a floor, checks to see if the block below is floor-worthy
 * If this is all satisfactory, then the location is returned.
 * If no block satifies these constraints, return an empty optional.

To try to improve performance, blocks that have been checked get their locations cached, along with whether the block is safe for the body or as a floor. These safety checks are currently using the previous methods with minor updates, that is, using the Material enum in Minecraft.

In theory, this could be backported to 4.x too.

A quick test plugin for teleporting using the safe teleport routine can be found at https://gist.github.com/dualspiral/e505bd4ce912355252de249644e71543 - just put co-ordinates of obviously invalid blocks for teleporting!

### DIFF
--- a/src/main/java/org/spongepowered/common/world/SpongeTeleportHelper.java
+++ b/src/main/java/org/spongepowered/common/world/SpongeTeleportHelper.java
@@ -24,20 +24,23 @@
  */
 package org.spongepowered.common.world;
 
+import com.flowpowered.math.GenericMath;
 import com.flowpowered.math.vector.Vector3i;
-import net.minecraft.block.Block;
 import net.minecraft.block.material.Material;
-import net.minecraft.block.state.IBlockState;
-import org.spongepowered.api.block.BlockState;
-import org.spongepowered.api.block.BlockType;
-import org.spongepowered.api.block.BlockTypes;
+import net.minecraft.world.border.WorldBorder;
 import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.TeleportHelper;
 import org.spongepowered.api.world.World;
 import org.spongepowered.common.block.BlockUtil;
-import org.spongepowered.common.util.VecHelper;
 
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
+
+import javax.annotation.Nullable;
 
 public class SpongeTeleportHelper implements TeleportHelper {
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/world/SpongeTeleportHelper.java
```diff
@@ -47,205 +50,131 @@
...
+        }
 
-                // We'll fall onto a block, need to make sure its safe
-                if (typeBelowPos.getType() != BlockTypes.AIR && !isSafeFloorMaterial(BlockUtil.toNative(typeBelowPos).getMaterial())) {
-                    return false;
-                }
+        // Sort them according to the distance to the provided worldLocation.
+        return vectors.stream()
+                .sorted((first, second) -> Integer.compare(vectorLocation.distanceSquared(first), vectorLocation.distanceSquared(second)))
+                .collect(Collectors.toList());
```
ryantheleach: I'm slightly concerned about the amount of garbage collection required for very large search areas.
Returning the stream itself, rather then the collection would be a rather simple fix.

The other option would be to not use streams, but the for loop, and bail out early, but since it's already using streams, may as well use the stream.

In addition, the old method of iterating over the locations was able to generate them roughly in distance order to begin with.

src/main/java/org/spongepowered/common/world/SpongeTeleportHelper.java
```diff
@@ -47,205 +50,131 @@
...
+            this.isSafeBody = isSafeBodyMaterial(material);
+            this.isSafeFloor = isSafeFloorMaterial(material);
+        }
+
+        private boolean isSafeFloorMaterial(@Nullable Material material) {
+            return material != null && !(material == Material.AIR || material == Material.CACTUS || material == Material.FIRE
+                    || material == Material.LAVA);
+        }
+
+        private boolean isSafeBodyMaterial(@Nullable Material material) {
```
ryantheleach: I'm not sure this is what Zidane had in mind when he asked for a rewrite.

Is there any already written methods that see if the player is safe at a location that mods might also be able to use?

(There may not be, I didn't happen to see them the last time I fixed the safe teleport code to actually iterate correctly, but I'm by no means experienced with the codebase)

src/main/java/org/spongepowered/common/world/SpongeTeleportHelper.java
```diff
@@ -47,205 +50,131 @@
...
+            }
+        }
 
-                // We'll fall onto a block, need to make sure its safe
-                if (typeBelowPos.getType() != BlockTypes.AIR && !isSafeFloorMaterial(BlockUtil.toNative(typeBelowPos).getMaterial())) {
-                    return false;
-                }
+        // Sort them according to the distance to the provided worldLocation.
+        return vectors.stream()
+                .sorted((first, second) -> Integer.compare(vectorLocation.distanceSquared(first), vectorLocation.distanceSquared(second)))
```
ST-DDT: This should be:
````java
.sorted(Comparator.comparingInt(vectorLocation::distanceSquared))
````
as its far easier to read.

File: src/main/java/org/spongepowered/common/world/SpongeTeleportHelper.java (hunk 2 above)
Comment (dualspiral) [inline src/main/java/org/spongepowered/common/world/SpongeTeleportHelper.java]: No, I did look. The problem is that there is no easy way, as far as I can see, to tell if a block will damage you, which is the biggest issue here. I looked into seeing if there is a way to expose this, but it became apparent that the only way that I could work it out is to create a fake entity, pretend it collides with the block, then see if it damages said entity.

It _might_ be viable, but if we're going to do that, we might as well make that part of the Data API at the same time.

File: src/main/java/org/spongepowered/common/world/SpongeTeleportHelper.java (hunk 1 above)
Comment (dualspiral) [inline src/main/java/org/spongepowered/common/world/SpongeTeleportHelper.java]: > Returning the stream itself, rather then the collection would be a rather simple fix.

Yep, I'm still thinking in terms of lists, I'll do that.

> In addition, the old method of iterating over the locations was able to generate them roughly in distance order to begin with.

Going back to a modifed original method is something I could do, I just pretty much wrote my own routine and disregarded everything else because I was honestly not sure what was working and what wasn't.

> I'm also slightly worried about a spherical iteration order making it easier to clip into protected areas. Although I think it's impossible to have safe teleport, and not have the user be teleported into areas they arn't supposed to get to, unless the plugin provides some sort of predicate for testing locations.

I suppose it can't harm to add another API method to do that.

File: src/main/java/org/spongepowered/common/world/SpongeTeleportHelper.java [hunk 4]
```diff
@@ -47,205 +51,132 @@
...
-        }
-        //example at radius 2
-        //..c01
-        //.....
-        //..o..
-        //.....
-        //.....
-        // Move up to the first corner..
-        for (int i = 0; i < radius; i++) {
-            blockPos = blockPos.add(0, 0, i);
```
dualspiral: Mostly for interest, though it's good to note if it's decided that the previous way was better, this (and the other positioning for loops below) are actually broken.

The comment show what should happen, but for that to be the case, `i` should be replaced by 1. Instead, in the first iteration, `blockPos` remain in the same place, then for `i=1`, `blockPos` moves one in the `z` direction (so would end up in position 0), then it would be two, three, etc.

By setting `i` to `1` in this statement, you would get the expected behaviour.

src/main/java/org/spongepowered/common/world/SpongeTeleportHelper.java
Zidane: @ryantheleach 

The previous struct here was completely INSANE in how it determined if the place is safe. This is a huge step forward.

@dualspiral 

I think Vanilla has some code to determine if where you'll spawn is safe. See if you can track it down and make use of it + modifications (it isn't perfect, everyone knows about spawning in walls on occasion).

Huge thanks for kicking up this PR.

ryantheleach: >I think Vanilla has some code to determine if where you'll spawn is safe. See if you can track it down and make use of it + modifications (it isn't perfect, everyone knows about spawning in walls on occasion).

Is what I was meaning, sorry for not being clear.

src/main/java/org/spongepowered/common/world/SpongeTeleportHelper.java
ryantheleach: You are right, I've clearly made an error when I touched this, I'm surprised it didn't get picked up sooner.
The previous behaviour before i touched it however was even more wrong, and unclear + uncommented.

dualspiral: Happens to the best of us. I only noticed it by chance when logging things.

src/main/java/org/spongepowered/common/world/SpongeTeleportHelper.java
dualspiral: Eh, Vanilla cheats.

In `net.minecraft.world.WorldProvider#canCoordinateBeSpawn(int, int)`, they just find a grass block to spawn on (overridable by other worlds, the end just finds a solid block). In a similar vein, the `net.minecraft.entity.player.EntityPlayerMP` constructor tries to put the player at the top block for a given `x, z` co-ordinate.

So, not really much to go on.

Zidane: Yeah, went and looked...they do indeed cheat.

As for determining block damage, its a toss-up. Vanilla uses a combination of material checking + collision checks (to prevent spawning in walls). With mods, defining "Does a block hurt me" is incredibly difficult because that definition can change based on the context of the world. @dualspiral 's code works nicely here.

PR #909, opened 2016-08-30
Error : sendViewBook

rexbut:
Since 1.9, the  "sendViewBook" don't work because the "SPacketCustomPayload" require the hand. 

You need to add a varInt for the hand : 0 for the main hand, 1 for the offhand

http://wiki.vg/Plugin_channels#MC.7CBOpen

### DIFF
--- a/src/main/java/org/spongepowered/common/util/BookFaker.java
+++ b/src/main/java/org/spongepowered/common/util/BookFaker.java
@@ -58,7 +58,10 @@ public class BookFaker {
         receiver.sendPacket(new SPacketSetSlot(WINDOW_PLAYER_INVENTORY, bookSlot, ItemStackUtil.toNative(item)));
 
         // Next we tell the client to open the Book GUI
-        receiver.sendPacket(new SPacketCustomPayload("MC|BOpen", new PacketBuffer(Unpooled.buffer())));
+        // 0 for the main hand, 1 for the offhand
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/util/BookFaker.java
```diff
@@ -58,7 +58,10 @@ public static void fakeBookView(BookView bookView, Player player) {
         receiver.sendPacket(new SPacketSetSlot(WINDOW_PLAYER_INVENTORY, bookSlot, ItemStackUtil.toNative(item)));
 
         // Next we tell the client to open the Book GUI
-        receiver.sendPacket(new SPacketCustomPayload("MC|BOpen", new PacketBuffer(Unpooled.buffer())));
+        // 0 for the main hand, 1 for the offhand
+        PacketBuffer packetbuffer = new PacketBuffer(Unpooled.buffer());
+        packetbuffer.writeInt(0);
```
Minecrell: You should use `packetbuffer.writeEnumValue(EnumHand.MAIN_HAND)` here. There is no need to hard code the internal protocol representation, MC uses `writeEnumValue` for opening books.

rexbut:
Due to an error on my part, I have created a new PR : https://github.com/SpongePowered/SpongeCommon/pull/924

PR #924, opened 2016-09-06
[Update] Error : sendViewBook

rexbut:
"SPacketCustomPayload" require the hand.

Since 1.9, the "sendViewBook" don't work because the "SPacketCustomPayload" require the hand.

(Due to an error on my part, I have created a new PR)


Minecrell:
@rexbut Can you squash your commits?

PR #925, opened 2016-09-06
Fix : Enchantment#getTranslation() AbstractMethodError

rexbut:
The Enchantment#getTranslation() function was not implemented


ST-DDT:
Related issue https://github.com/SpongePowered/SpongeCommon/issues/739
Alternative solution: https://github.com/ST-DDT/SpongeCommon/blob/13eab81abd55b47dab7089dfceca22f6e4596dc0/src/main/java/org/spongepowered/common/mixin/core/item/MixinEnchantment.java

src/main/java/org/spongepowered/common/mixin/core/item/MixinEnchantment.java
```diff
@@ -113,6 +115,11 @@ public boolean isCompatibleWith(Enchantment ench) {
     public String enchantment$getName() {
         return shadow$getName();
     }
+
```
ST-DDT: Remove whitespace

Aaron1011:
@rexbut: Sorry for the delay in getting this merged.

PR #938, opened 2016-09-17
Add LaunchWrapperTestRunner that allows unit tests to use Mixins

Minecrell:
Adds a new `LaunchWrapperTestRunner` that will load the test classes using an initialized LaunchWrapper environment including the Mixin transformer. That allows the tests to use methods that are only available after the Mixins have been applied.

Basically, it works like this:
- You define a test class with the `@RunWith` annotation:
  
  ``` java
  @RunWith(LaunchWrapperTestRunner.class)
  public class ExampleTest {
  
  }
  ```
- The first time a test class with the annotation is loaded `LaunchWrapperTestRunner` will initialize a dummy LaunchWrapper environment together with the Mixin transformer.
- The test class is loaded using the `LaunchClassLoader` which will apply the Mixins to all referenced classes.

This could be also extended to call Minecraft's `Bootstrap.init()` if that is needed in a specific test.

I've added an example test for the legacy text serializer that requires the mixins to `ITextComponent` to be applied.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -68,7 +68,14 @@ dependencies {
     forgeGradleMcDeps('net.minecraft:launchwrapper:1.11') {
         transitive = false
     }
+
+    testCompile 'org.spongepowered:launchwrappertestsuite:1.0-SNAPSHOT'
 }
 
 // Include API dependencies in our POM
 ext.shadedDevProject = api
+
+test {
+    systemProperty 'lwts.tweaker', 'org.spongepowered.common.launch.TestTweaker'
+    workingDir = {test.temporaryDir}
+}
--- a/src/main/java/org/spongepowered/common/data/SpongeDataManager.java
+++ b/src/main/java/org/spongepowered/common/data/SpongeDataManager.java
@@ -148,7 +148,7 @@ public final class SpongeDataManager implements DataManager {
             }
             this.builders.put(clazz, builder);
         } else {
-            SpongeImpl.getLogger().warn("A DataBuilder has already been registered for %s. Attempted to register %s instead.%n", clazz,
+            SpongeImpl.getLogger().warn("A DataBuilder has already been registered for {}. Attempted to register {} instead.", clazz,
                     builder.getClass());
         }
     }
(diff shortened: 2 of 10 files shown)


gabizou:
Could this runner be used in conjunction with the parameterized runner for `ManipulatorTest`? Since a majority of the data manipulator implementations rely on mixins, most of them have been annotated to avoid being run due to the lack of a mixin environment.

src/test/java/org/spongepowered/common/launch/TestMain.java
```diff
@@ -0,0 +1,36 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.launch;
+
+public final class TestMain {
```
ST-DDT: Why/for what do you need this class?

src/test/java/org/spongepowered/common/launch/LaunchWrapperTestRunner.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public static Class<?> loadTestClass(Class<?> originalClass) throws InitializationError {
+        if (!initialized) {
+            initialized = true;
+            Launch.main(new String[]{"--tweakClass", "org.spongepowered.common.launch.TestTweaker"});
+        }
+
+        // JUnit attempts to lookup the @Test annotation so we need to make sure the classes are loaded
+        // using the same class loader (the main class loader)
+        Launch.classLoader.addClassLoaderExclusion("org.junit.");
```
ST-DDT: This should probably moved inside the initialization block.

src/test/java/org/spongepowered/common/launch/TestMain.java
stephan-gh: LaunchWrapper is designed to wrap a complete application, so you use `Launch` as main class and then add your original main class as launch target in your tweaker.

For the tests we don't need to do anything directly, but rather just need to load some classes because JUnit will execute the tests. So basically this class acts as a dummy main class because LaunchWrapper requires it.

Minecrell:
@gabizou 
Yes, even though it is a bit complicated (because you need to initialize the registry etc). I've implemented it here: https://github.com/SpongePowered/SpongeCommon/commits/feature/enable-all-manipulator-tests

There are still 21 tests which fail (not sure why): https://travis-ci.org/SpongePowered/SpongeCommon/jobs/160691238#L623

src/test/java/org/spongepowered/common/launch/LaunchWrapperTestRunner.java
stephan-gh: Fixed

gabizou:
> There are still 21 tests which fail (not sure why): https://travis-ci.org/SpongePowered/SpongeCommon/jobs/160691238#L623

Some of these tests are failing as expected because whoever wrote the data manipulators themselves were implemented incorrectly (referencing the API class instead of the implementation class for reflection handling). I'll check it out as time permits.

Zidane:
@Minecrell @Mumfrey 

Should this be in Mixin?

stephan-gh:
@gabizou 
I've updated this to the latest bleeding and deployed a version of LaunchWrapperTestSuite on our Maven repository. A few data tests are still failing so you will need to take a look at them before we can merge this PR.

PR #947, opened 2016-09-19
Accept bans with no reason. Fixes #836

### LINKED ISSUE #836: Null/Empty Ban Reason
So I've been having an issue with the `BanService`.  The relevant code may be found here:
https://gist.github.com/12AwsomeMan34/95f8224fc86b81540e1e1793f69d7747

Essentially, `optionalBan.get().getReason()` will always return `Optional#empty` (I am 100% certain that `getBanFor` does not return `Optional#empty`, I just removed the debug statements in the gist), despite there clearly being a reason specified, as shown in the `banned-players.json` file.

To further this, after banning the player and then restarting the server, the `getReason()` optional is instead null, and throws an NPE when attempting to call `isPresent()` on the `Optional`.

### LINKED ISSUE #929: SpongeBanBuilder should accept a ban without a reason
Creating a ban without a reason causes a precondition to fail
```java
Ban.of(player.getProfile());
```
Results in
```
java.lang.IllegalStateException: Reason cannot be null!
	at com.google.common.base.Preconditions.checkState(Preconditions.java:176) ~[guava-17.0.jar:?]
	at org.spongepowered.common.ban.SpongeBanBuilder.build(SpongeBanBuilder.java:123) ~[SpongeBanBuilder.class:?]
	at org.spongepowered.api.util.ban.Ban.of(Ban.java:61) ~[Ban.class:?]
```
Using dev version on commit 9dc74b3693f19f10eb4e841f3723da0dd38d80ca

jamierocks:


src/main/java/org/spongepowered/common/ban/SpongeBanBuilder.java
```diff
@@ -152,7 +150,7 @@ private Date toDate(Instant instant) {
             this.address = ((Ban.Ip) ban).getAddress();
         }
 
-        this.reason = ban.getReason();
+        this.reason = ban.getReason().orElseGet(null);
```
kashike: Should be `orElse(null)`.

PR #948, opened 2016-09-19
Add Type Serializers for Registered Data Serializers

m0pt0pmatt:
This PR adds TypeSerializers for all registered Data Serializers.

This is so @ConfigSerializable config classes can automatically use classes which are supported by the data translators.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/SpongeDataManager.java
+++ b/src/main/java/org/spongepowered/common/data/SpongeDataManager.java
@@ -62,6 +62,7 @@ import org.spongepowered.common.data.nbt.validation.DelegateDataValidator;
 import org.spongepowered.common.data.nbt.validation.RawDataValidator;
 import org.spongepowered.common.data.nbt.validation.ValidationType;
 import org.spongepowered.common.data.nbt.value.NbtValueProcessor;
+import org.spongepowered.common.data.persistence.DataTranslatorTypeSerializer;
 import org.spongepowered.common.data.processor.common.AbstractSingleDataSingleTargetProcessor;
 import org.spongepowered.common.data.util.ComparatorUtil;
 import org.spongepowered.common.data.util.DataFunction;
@@ -346,6 +347,9 @@ public final class SpongeDataManager implements DataManager {
             DataTranslatorRegistryModule.getInstance().registerAdditionalCatalog(translator);
         } else {
             throw new IllegalStateException("Already registered the DataTranslator for " + objectClass.getCanonicalName());
+        }
+        if (TypeSerializers.getDefaultSerializers().get(translator.getToken()) == null) {
+            TypeSerializers.getDefaultSerializers().registerType(translator.getToken(), DataTranslatorTypeSerializer.from(translator));
         }
     }
 
--- a/src/main/java/org/spongepowered/common/data/persistence/DataTranslatorTypeSerializer.java
+++ b/src/main/java/org/spongepowered/common/data/persistence/DataTranslatorTypeSerializer.java
@@ -1,0 +1,68 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data.persistence;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.reflect.TypeToken;
+import ninja.leaping.configurate.ConfigurationNode;
+import ninja.leaping.configurate.objectmapping.ObjectMappingException;
+import ninja.leaping.configurate.objectmapping.serialize.TypeSerializer;
+import org.spongepowered.api.data.persistence.DataTranslator;
+import org.spongepowered.api.data.persistence.InvalidDataException;
+
+public class DataTranslatorTypeSerializer<T> implements TypeSerializer<T> {
+
+    private final DataTranslator<T> dataTranslator;
+
+    private DataTranslatorTypeSerializer(DataTranslator<T> dataTranslator) {
+        this.dataTranslator = checkNotNull(dataTranslator, "dataTranslator");
+    }
+
+    @Override
+    public T deserialize(TypeToken<?> type, ConfigurationNode value) throws ObjectMappingException {
+        try {
+            return this.dataTranslator.translate(ConfigurateTranslator.instance().translateFrom(value));
+        } catch (InvalidDataException e) {
+            // Since the value in the config node might be null, return null if an error occurs.
+            return null;
+        }
+    }
+
+    @Override
+    public void serialize(TypeToken<?> type, T obj, ConfigurationNode value) throws ObjectMappingException {
+        try {
+            ConfigurateTranslator.instance().translateContainerToData(value, this.dataTranslator.translate(obj));
+        } catch (InvalidDataException e) {
+            throw new ObjectMappingException("Could not serialize. Data was invalid.", e);
+        }
+
+    }
+
+    public static <T> DataTranslatorTypeSerializer<T> from(DataTranslator<T> dataTranslator) {
+        return new DataTranslatorTypeSerializer<>(dataTranslator);
+    }
+
+}


gabizou:
What prevents the use of the native `DataTranslator` for `Vector#X`? Also, your formatting for imports are not in line with the code style of the project.

src/main/java/org/spongepowered/common/config/serializer/DataSerializableTypeSerializer.java
```diff
@@ -22,7 +22,7 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.common.config;
+package org.spongepowered.common.config.serializer;
```
m0pt0pmatt: Since there were now 4 TypeSerializers in the config package, I thought it would be better to put them in their own package, to not clutter the rest of the contents of the config package

m0pt0pmatt:
Object mapping does not make use of the data translators. If I want to have my config class be `@ConfigSerializable`, then TypeSerializers must be registered for the vector classes.

I can however make the type serializers use the data translators.

m0pt0pmatt:
Updated PR to fix imports and use data translators to actually serialize.

Of course, instead of just have serializers for Vector3d and Vector3i, it might make more sense to register a TypeSerializer for every DataTranslator in DataTranslators.

Cybermaxke:
It is probably easier to register a predicate (`registerPredicate`) instead, and create a `TypeSerializer` for all the objects that can be translated through a `DataTranslator`. No need to create a `TypeSerializer` for every object and you will be able to support every object type that has a registered `DataTranslator`.

ryantheleach:
TypeSerializers being registered statically for plugins seems like a thing that shouldn't be allowed honestly.
I mean, it can make sense for Types that your plugin provides, but anything that represents an already existing class from elsewhere really should be configured in your own configloader.

Configurate isn't made just for Sponge, so deprecating this isn't really an option either.

That said, I would appreciate seeing a better format for vectors(among other classes) in configs then 
```
{
x:0
y:0
z:0
ContentVersion:1
}
```

But creating plugin defined serializers just means that every plugin has to have it's own opinion on what looks clean.

Additionally, when serialising ItemStacks you end up with duplicate data,

```
ContentVersion=1
Count=1
Data=[
    {
        DataClass="org.spongepowered.common.data.manipulator.mutable.SpongeDisplayNameData"
        ManipulatorData {
            ContentVersion=1
            DisplayName="\"${pokemon.name}\""
        }
    }
]
ItemType="pixelmon:PixelmonSprite"
UnsafeDamage=0
UnsafeData {
    SpriteName="pixelmon:sprites/pokemon/${pokemon.id}"
    display {
        Name="${pokemon.name}"
    }
}
```

For instance.

This repeated information really isn't user-friendly for entering in configuration data, and just gives people more opportunity to create errors.

src/main/java/org/spongepowered/common/config/serializer/DataSerializableTypeSerializer.java
gabizou: Nah, there's a better way for handling anything that is registered for `DataTranslator`s, you can use each instance that is being registered to the `SpogneDataManager` and add a new `TypeSerializer` if applicable.

pie-flavor:
@ryantheleach Yeah, but `Data[]` is optional; it functions just fine without it.

gabizou:
All of this can be avoided by having all `DataTranslator`s registered as `TokenSerializer`s when they are registered in `SpongeDataManager`. Since you have the `TypeToken` already, it's very well possible to have them registered with Configurate just as well.

Granted, a few of them will be ignored, so just make note of that. All of that and problem solved.

m0pt0pmatt:
Anything which still needs working on here?

gabizou:
@m0pt0pmatt I'd rather this be handled by a straight injection in the `SpongeDataManager` when `registerTrranslator` is used, since the `Datatranslator` has the necessary `TypeToken`.

m0pt0pmatt:
@gabizou I've tested and this still works.
Can you take another quick look and see if there's anything else to do?
I would be very grateful we could reach a merge before the Sponge plugin competition deadline. :)

m0pt0pmatt:
Whoops, fixing license right now

src/main/java/org/spongepowered/common/data/DataTranslatorTypeSerializer.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data;
```
Deamon5550: Move this to `org.spongepowered.common.data.persistence`

src/main/java/org/spongepowered/common/data/DataTranslatorTypeSerializer.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public T deserialize(TypeToken<?> type, ConfigurationNode value) throws ObjectMappingException {
+        return dataTranslator.translate(ConfigurateTranslator.instance().translateFrom(value));
+    }
+
+    @Override
+    public void serialize(TypeToken<?> type, T obj, ConfigurationNode value) throws ObjectMappingException {
+        ConfigurateTranslator.instance().translateContainerToData(value, dataTranslator.translate(obj));
```
kashike: `this.`

src/main/java/org/spongepowered/common/data/DataTranslatorTypeSerializer.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private final DataTranslator<T> dataTranslator;
+
+    private DataTranslatorTypeSerializer(DataTranslator<T> dataTranslator) {
+        this.dataTranslator = checkNotNull(dataTranslator, "dataTranslator");
+    }
+
+    @Override
+    public T deserialize(TypeToken<?> type, ConfigurationNode value) throws ObjectMappingException {
+        return dataTranslator.translate(ConfigurateTranslator.instance().translateFrom(value));
```
kashike: `this.`

m0pt0pmatt: @kashike, I have no idea what your comments mean. Can you explain in more detail please?

src/main/java/org/spongepowered/common/data/DataTranslatorTypeSerializer.java
m0pt0pmatt: @Deamon5550 moved

src/main/java/org/spongepowered/common/data/DataTranslatorTypeSerializer.java
kashike: @m0pt0pmatt `return this.dataTranslator.translate(ConfigurateTranslator.instance().translateFrom(value));`

src/main/java/org/spongepowered/common/data/DataTranslatorTypeSerializer.java
kashike: `ConfigurateTranslator.instance().translateContainerToData(value, this.dataTranslator.translate(obj));`

src/main/java/org/spongepowered/common/data/DataTranslatorTypeSerializer.java
m0pt0pmatt: @kashike thank you. I've updated both lines, see the latest changes.

src/main/java/org/spongepowered/common/data/SpongeDataManager.java
```diff
@@ -347,6 +348,7 @@ public static void finalizeRegistration() {
         } else {
             throw new IllegalStateException("Already registered the DataTranslator for " + objectClass.getCanonicalName());
         }
+        TypeSerializers.getDefaultSerializers().registerType(translator.getToken(), DataTranslatorTypeSerializer.from(translator));
```
gabizou: You will need to check that the translator is not registering an already registered type (such as the one for `UUID` which is default registered by Configurate).

src/main/java/org/spongepowered/common/data/persistence/DataTranslatorTypeSerializer.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private final DataTranslator<T> dataTranslator;
+
+    private DataTranslatorTypeSerializer(DataTranslator<T> dataTranslator) {
+        this.dataTranslator = checkNotNull(dataTranslator, "dataTranslator");
+    }
+
+    @Override
+    public T deserialize(TypeToken<?> type, ConfigurationNode value) throws ObjectMappingException {
+        return this.dataTranslator.translate(ConfigurateTranslator.instance().translateFrom(value));
```
gabizou: Should probably surround with a `try catch` and throw an `ObjectMappingException` with the provided exception`.

src/main/java/org/spongepowered/common/data/persistence/DataTranslatorTypeSerializer.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Override
+    public void serialize(TypeToken<?> type, T obj, ConfigurationNode value) throws ObjectMappingException {
+        ConfigurateTranslator.instance().translateContainerToData(value, this.dataTranslator.translate(obj));
```
gabizou: Likewise here.

src/main/java/org/spongepowered/common/data/persistence/DataTranslatorTypeSerializer.java
m0pt0pmatt: @gabizou added try catches

src/main/java/org/spongepowered/common/data/SpongeDataManager.java
m0pt0pmatt: @gabizou check that type isn't already registered. See new files.

m0pt0pmatt:
Null values are persisted as null, but the try-catch in deserialize throws an exception. It would be better if an exception was thrown, that null was simply returned.

Otherwise null values cannot be saved to config files, and that sucks.

gabizou:
> Null values are persisted as null, but the try-catch in deserialize throws an exception. It would be better if an exception was thrown, that null was simply returned.
> 
> Otherwise null values cannot be saved to config files, and that sucks.

Yeah, no, if a null value was persisted, the data translator would sufficiently expect to receive a null value. The point of the try catch is to propagate the exception correctly.

src/main/java/org/spongepowered/common/data/persistence/DataTranslatorTypeSerializer.java
```diff
@@ -0,0 +1,67 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.reflect.TypeToken;
+import ninja.leaping.configurate.ConfigurationNode;
+import ninja.leaping.configurate.objectmapping.ObjectMappingException;
+import ninja.leaping.configurate.objectmapping.serialize.TypeSerializer;
+import org.spongepowered.api.data.persistence.DataTranslator;
+import org.spongepowered.api.data.persistence.InvalidDataException;
+
+public class DataTranslatorTypeSerializer <T> implements TypeSerializer<T> {
```
liach: `DataTranslatorTypeSerializer<T>`

src/main/java/org/spongepowered/common/data/persistence/DataTranslatorTypeSerializer.java
```diff
@@ -0,0 +1,67 @@
...
+            ConfigurateTranslator.instance().translateContainerToData(value, this.dataTranslator.translate(obj));
+        } catch (InvalidDataException e) {
+            throw new ObjectMappingException("Could not serialize. Data was invalid.", e);
+        }
+
+    }
+
+    public static <T> DataTranslatorTypeSerializer<T> from(DataTranslator<T> dataTranslator) {
+        return new DataTranslatorTypeSerializer<>(dataTranslator);
+    }
```
liach: Add a line after this bracket.

src/main/java/org/spongepowered/common/data/SpongeDataManager.java
```diff
@@ -347,6 +348,9 @@ public static void finalizeRegistration() {
         } else {
             throw new IllegalStateException("Already registered the DataTranslator for " + objectClass.getCanonicalName());
         }
+        if (TypeSerializers.getDefaultSerializers().get(translator.getToken()) == null) {
```
liach: Should this be registered just in the static block?

m0pt0pmatt: I was told by @gabizou to place it here

liach: Ok.

m0pt0pmatt:
Thank you everyone :)

PR #953, opened 2016-09-20
Implement PotionData for tipped arrows

jamierocks:


JBYoshi:
Is there already support for `PotionData` on tipped arrow `ItemStack`s? If not, could you add that as well?

jamierocks:
@JBYoshi I'll get on that when I'm back ;)

jamierocks:
That should do the trick ;)

JBYoshi:
I actually meant the tipped arrow item. It appears to use the [same NBT structure](http://minecraft.gamepedia.com/Player.dat_format#Potion_Effects) as potions though, so it shouldn't be too much work to add it in.

jamierocks:
See https://github.com/jamierocks/SpongeCommon/blob/3fe1774416c8af6f22a61942f597c3a3bb0a94a3/src/main/java/org/spongepowered/common/data/processor/data/item/ItemPotionDataProcessor.java#L54

JBYoshi:
`Items.TIPPED_ARROW`? I don't see that. I just see `Items.POTIONITEM`, `Items.SPLASH_POTION`, and `Items.LINGERING_POTION`.

jamierocks:
Ah yes - all I could remember of your original comment was item and potion. I'll get on it.

Issue #954, opened 2016-09-22
[1.8.9] Register Modded SoundTypes & Playback custom sounds.


ryantheleach:
Currently modded soundtypes are not being registered, nor does it appear to be possible to play custom sounds bundled in resource packs.

ryantheleach:
```
<Waterpicker> Sponge.getRegistry().getAllFor("pixelmon", SoundType.class); is what I used.)
```

ZephireNZ:
In my research, it looks like this is because we hooking into the registry when a sound is added for Forge. Might also need a way to hook into when resource packs get registered too.

ZephireNZ:
Actually , with resource packs at least on the server side, those sounds would never be loaded would they? The server simply gives the client the pack, never actually loading it (and the sounds) itself. Might need to add a SoundTypeBuilder to create a "dummy" sound which can be sent to the client to play.

ST-DDT:
Afaict it is possible to register catalog types already:
https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/GameRegistry.java#L194

But I'm not sure whether we expose a sound type builder or whether this is required at all.

ZephireNZ:
`if registration for that type is supported.`

`SoundRegistryModule` does not implement `AdditionalCatalogRegistryModule` and so doesn't support additional types after it's been initialized.

ryantheleach:
That or introduce the string methods for sending sounds.

ZephireNZ:
I'm going to whip up a PR to add these changes.

ZephireNZ:
Voila: https://github.com/SpongePowered/SpongeCommon/issues/954

PR #955, opened 2016-09-23
Implement SoundType changes

ZephireNZ:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1364) | **Common** | [Forge](https://github.com/SpongePowered/SpongeForge/pull/923)

This PR fixed the issues with the `SoundType` implementation seen here: #954

- Allow for building a `SoundType`
- Redirect non-registered sounds to use the custom sound packet, as they do not have an int ID
- Fix up the registry so that IDs can be fetched by "my.sound.id" instead of "my_sound_id"
- Make registry support additional sounds (which are added in the forge PR)

Test code:
```
TO COME
```

### DIFF
--- a/src/main/java/org/spongepowered/common/interfaces/world/IMixinServerWorldEventHandler.java
+++ b/src/main/java/org/spongepowered/common/interfaces/world/IMixinServerWorldEventHandler.java
@@ -1,0 +1,37 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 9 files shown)


src/main/java/org/spongepowered/common/interfaces/world/IMixinServerWorldEventHandler.java
```diff
@@ -0,0 +1,13 @@
+package org.spongepowered.common.interfaces.world;
+
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.util.SoundCategory;
+
+import javax.annotation.Nullable;
+
+public interface IMixinServerWorldEventHandler {
+
+    void playSoundToAllNearExcept(@Nullable EntityPlayer player, String soundIn, SoundCategory category, double x, double y, double z,
```
kashike: 1.10 has no `String` based sound methods.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -568,8 +570,19 @@ public void playSound(SoundType sound, SoundCategory category, Vector3d position
 
     @Override
     public void playSound(SoundType sound, SoundCategory category, Vector3d position, double volume, double pitch, double minVolume) {
-        this.connection.sendPacket(new SPacketSoundEffect(SoundEvents.getRegisteredSoundEvent(sound.getId()), (net.minecraft.util.SoundCategory) (Object) category, position.getX(), position.getY(), position.getZ(),
-                (float) Math.max(minVolume, volume), (float) pitch));
+        SoundEvent event;
+        try {
+            // Check if the event is registered (ie has an integer ID)
+            event = SoundEvents.getRegisteredSoundEvent(sound.getId());
```
kashike: Why are you doing this? All sounds should be a `SoundEvent` when properly registered through API.

src/main/java/org/spongepowered/common/interfaces/world/IMixinWorldServer.java
```diff
@@ -92,4 +96,6 @@ default boolean forceSpawnEntity(Entity entity) {
 
     void triggerInternalExplosion(Explosion explosion);
 
+    void playSound(@Nullable EntityPlayer player, double x, double y, double z, String soundIn, SoundCategory category, float volume, float pitch);
```
kashike: Again, why `String` method

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
kashike: @ZephireNZ: I suggest naming that method you add `playResourcePackSound` then, rather than `playSoundToANearExcept`

ZephireNZ:
Changed the name of the new method to `sendCustomSound` as suggested by @kashike

PR #969, opened 2016-10-08
Fire DestructEntityEvent.Death when setting a player's health to zero.

dualspiral:
When setting a player's health to zero via the Data API, the `DestructEntityEvent.Death` event is not fired, and the normal Minecraft death code does not run, for example, a player's inventory might not get dropped, and no death message is fired. This is due to the folllowing lines calling `attackEntityFrom`:
- https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/data/processor/multi/entity/HealthDataProcessor.java#L69
- https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/data/processor/value/entity/HealthValueProcessor.java#L117

The issue with this is that [there is a line in that method in EntityLivingBase](https://github.com/SpongePowered/SpongeCommon/blob/668569410220699d05eb44d68beb11ac1f7f8377/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntityLivingBase.java#L375) (and a similar method in the non-overwritten `EntityPlayer` class method) that bails out early if the health is zero - which we've already set it to! Thus, nothing happens once health is set to zero, other than the client/server just recognising that the player needs a respawn as they have no health and telling them as such.

So, there are three potential fixes, depending on what we're trying to achieve:
- In `HealthDataProcessor` and `HealthValueProcessor`, call `onDeath` instead of `attackEntityFrom`, [like it did previously](https://github.com/SpongePowered/SpongeCommon/commit/cd9655f83216f0c206543b28d58e46e607553e6c#diff-d31e4027818aa8a40fc9699a9f089481L68), and send the death sound ourselves.
- In `HealthDataProcessor` and `HealthValueProcessor`, call `attackEntityFrom` _before_ setting the health. I rejected this approach as the value processor has some try/catch block which I would then end up having to undo the "attack" if it failed. In theory, you could just perform that attack without setting the health after - but I didn't do that because I figured it wouldn't then be obvious what was going on.
- The approach in the first commit, where the `EntityPlayer#attackEntityFrom` method is overriden and the health check is bypassed in `EntityLivingBase#attackEntityFrom` if using the Sponge damage source.

If there is no real reason to use `attackEntityFrom` over `onDeath`, then I actually suggest the first approach and would be happy for the first commit to be ignored.

The second commit in this PR fixes an issue I uncovered after writing the first commit, it causes Minecraft to use the generic death message if the translation for the key provided for the death message from the `DamageSource` is just the key. Without it, when a player dies via setting health, it printed "death.attack.spongespecific" to the game, with it, it simply states "dualspiral died".

Tested on both the lastest SpongeForge and SpongeVanilla codebases.

**Test Plugin** 

Plugin: [testplugin.zip](https://github.com/SpongePowered/SpongeCommon/files/517706/testplugin.zip)
Source: https://gist.github.com/dualspiral/13392996ee94c40e6d0e9b31e4981b9c

To use, run `/death <player>`. Prior to this PR, the player dies, keeps inventory, and no `DestructEntityEvent.Death` is fired. With this PR, the player dies, standard Minecraft processing applies, and a message is printed to all players with the name and type of death of the player (which should be `custom`).

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/util/MixinDamageSource.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/util/MixinDamageSource.java
@@ -25,8 +25,11 @@
 package org.spongepowered.common.mixin.core.util;
 
 import com.google.common.base.Objects;
+import net.minecraft.entity.EntityLivingBase;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.util.EntityDamageSourceIndirect;
+import net.minecraft.util.text.ITextComponent;
+import net.minecraft.util.text.TextComponentTranslation;
 import net.minecraft.world.Explosion;
 import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.event.cause.entity.damage.DamageType;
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/mixin/core/util/MixinDamageSource.java
```diff
@@ -0,0 +1,58 @@
...
+import net.minecraft.util.text.translation.I18n;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+@Mixin(DamageSource.class)
+public class MixinDamageSource {
```
gabizou: Why is this a new mixin? There's already a [`MixinDamageSource`](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/event/cause/entity/damage/MixinDamageSource.java#L47), granted the package layout was based on the API for various reasons (mixing into both api and pseudo implementation).

dualspiral: I didn't see the other Mixin, probably because of that. I can just add it to there.

dualspiral:
I've updated this for 1.11 (bleeding). The 1.10 version is at https://github.com/SpongePowered/SpongeCommon/compare/stable-5...dualspiral:bugfix/ondeathviahealth110

PR #978, opened 2016-10-14
Force Entity Spawns to Throw Instant Events

gabizou:
This is more of a fix than anything for various mod incompatibilities that check the `boolean` flag for entity spawns. There's more to come for this fix, but I felt it necessary for some discussion potentially with some other developers to evaluate the necessity of bulk entity spawn captures.


gabizou:
Already merged through cherry-picking.

PR #980, opened 2016-10-16
Add world command

jamierocks:
From some work I did earlier this year, based on the commands from CanaryLib.

### DIFF
--- a/src/main/java/org/spongepowered/common/command/SpongeCommand.java
+++ b/src/main/java/org/spongepowered/common/command/SpongeCommand.java
@@ -24,11 +24,14 @@
  */
 package org.spongepowered.common.command;
 
+import static org.spongepowered.api.command.args.GenericArguments.catalogedElement;
 import static org.spongepowered.api.command.args.GenericArguments.choices;
 import static org.spongepowered.api.command.args.GenericArguments.dimension;
 import static org.spongepowered.api.command.args.GenericArguments.firstParsing;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/command/SpongeCommand.java
```diff
@@ -650,6 +665,173 @@ private static CommandSpec getTpsCommand() {
...
+                        )
+                        .executor((src, args) -> {
+                            final WorldProperties worldProperties = args.<WorldProperties>getOne(Text.of("world")).get();
+                            final Optional<World> world = Sponge.getServer().getWorld(worldProperties.getUniqueId());
+
+                            if (world.isPresent()) {
+                                final boolean raining = world.get().getWeather() == Weathers.RAIN;
+                                final boolean thundering = world.get().getWeather() == Weathers.THUNDER_STORM;
+                                src.sendMessage(Text.of("Chunks loaded: " + Iterables.size(world.get().getLoadedChunks())));
+                                // src.sendMessage(Text.of("Villages loaded: "));
```
jamierocks: The commented out stuff were Canary things that couldn't be implemented on Sponge.

kashike: So remove it instead of commenting it out.

Deamon5550:
We've discussed this many a time in the past and this type of functionality is not something we wish to provide with sponge but would rather leave up plugin developers who can design it with a much more rich feature set and more importantly maintain it.

PR #988, opened 2016-10-20
[WIP] Implement SimpleCustomData

ZephireNZ:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1393) | **Common**

This implements my "simple" custom data as laid out in the API PR above.

Things to note with the implementation:
- I've edited `MixinCustomDataHolder` so that the manipulator is always available to be written to. It's  a quirk of the data system that you can't write using the Key system if there is not datamanipulator present. This means the API can hide the implementation and just focus on the key-based system.
- Any data that cannot be matched to a registered key gets preserved so that it will persist even with no serialisation mechanism present.

**Tests**

```
To come later
```


ZephireNZ:
Please, no nitpicking of licenses or code styles yet - not even implemented fully :P

PR #996, opened 2016-10-23
Negative one item stack quantities

hfreeb:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1397) | Common

As discussed at https://github.com/SpongePowered/SpongeAPI/issues/1392, a -1 quantity would be useful for ignoring stack size differences with `Inventory#query(ItemStack)`.

In my implementation I have added `ItemStack.Builder#anySize` which can be used for querying inventories.

### DIFF
--- a/src/main/java/org/spongepowered/common/item/inventory/SpongeItemStackBuilder.java
+++ b/src/main/java/org/spongepowered/common/item/inventory/SpongeItemStackBuilder.java
@@ -95,6 +95,12 @@ public class SpongeItemStackBuilder implements ItemStack.Builder {
         return this;
     }
 
+    @Override public ItemStack.Builder unsafeQuantity(int quantity) {
+        checkArgument(quantity >= -1, "Quantity must be greater than or equal to -1");
+        this.quantity = quantity;
+        return this;
+    }
+
     @Override
     public <E> ItemStack.Builder keyValue(Key<? extends BaseValue<E>> key, E value) {
         if (this.keyValues == null) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/item/inventory/SpongeItemStackBuilder.java
```diff
@@ -95,6 +95,12 @@ public ItemType getCurrentItem() {
         return this;
     }
 
+    @Override public ItemStack.Builder unsafeQuantity(int quantity) {
```
kashike: Wrong formatting.

``` java
    @Override
    public ItemStack.Builder unsafeQuantity(int quantity) {
```

kashike (changes requested):

ryantheleach:
Why are you limiting this to -1 and 0 only?

hfreeb:
> Why are you limiting this to -1 and 0 only?

As they're the only quantities I have seen uses for, but after the conversation at https://github.com/SpongePowered/SpongeAPI/pull/1397 I have decided I'm only going to allow -1.

Faithcaio:
Negative or zero ItemStack quantities ALWAYS result in an "empty" Air Itemstack in 1.11:
We were able to bypass this in our mixins but this is now fixed with https://github.com/SpongePowered/SpongeCommon/commit/c2965edba2f12f63ecddac812223222541b76709.

Currently the API javadocs for Inventory#query(itemstack) are wrong: All ItemStack quantities are ignored: see matching logic in `ItemStackStrategy`

PR #1000, opened 2016-10-26
Move instance checks to direct fail fast methods for Phases and PhaseStates

gabizou:
This improves on the existing chunk flags used to make the phase checks even faster than they currently are. Please look over @bloodmc and correct me ifI'm wrong @Deamon5550 and @Mumfrey that this indeed is faster than what currently exists.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/IPhaseState.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/IPhaseState.java
@@ -110,4 +110,11 @@ public interface IPhaseState {
     }
 
 
+    default boolean shouldStateAllowChunkRequest() {
+        return false;
+    }
+
+    default boolean doesStateDenyChunkRequests() {
+        return false;
+    }
 }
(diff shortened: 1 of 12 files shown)


me4502:
I wrote a benchmark for this case, and they seem pretty close,

```
Benchmark                      Mode  Cnt        Score       Error  Units 
MethodAccessIf.measureIf      thrpt   20  1473751.627 ± 14302.940  ops/s 
MethodAccessIf.measureMethod  thrpt   20  1482203.125 ± 27944.444  ops/s
```

Note: I measured 4 ifs vs default method overrides.

Actually, I just modified the benchmark to remove a potential bottleneck, and the results are very different,

```
Benchmark                      Mode  Cnt            Score           Error  Units
MethodAccessIf.measureIf      thrpt   20  12240318655.789 ± 189356329.447  ops/s
MethodAccessIf.measureMethod  thrpt   20   4200403704.254 ± 210284612.275  ops/s
```

Error is in same magnitude, ops/s aren't. Benchmark: https://gist.github.com/me4502/5b265abe5b017988ff1899a9787feb98

Update: Even with 18 if statements, the if statement method is still 2x faster.

PR #1007, opened 2016-11-01
Implement clearPassengers, and removePassenger(Entity)

me4502:
Currently these two API methods are not implemented. This implements them.

This also includes a lot of other changes that were requested for this PR, with the [accompanying API PR](https://github.com/SpongePowered/SpongeAPI/pull/1407).

Test case,
```
    @Listener
    public void onInteract(InteractBlockEvent.Secondary event, @First Player player) {
        player.getVehicle().ifPresent((Entity::clearPassengers));
    }

    @Listener
    public void onInteract(InteractBlockEvent.Primary event, @First Player player) {
        player.getVehicle().ifPresent((entity -> entity.removePassenger(player)));
    }
```

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
@@ -221,6 +221,7 @@ public abstract class MixinEntity implements IMixinEntity {
     @Shadow public abstract net.minecraft.entity.Entity getLowestRidingEntity();
     @Shadow public abstract net.minecraft.entity.Entity getRidingEntity();
     @Shadow public abstract void dismountRidingEntity();
+    @Shadow public abstract void removePassengers();
     @Shadow public abstract void playSound(SoundEvent soundIn, float volume, float pitch);
     @Shadow public abstract boolean isEntityInvulnerable(DamageSource source);
     @Shadow public abstract boolean isSprinting();
@@ -232,6 +233,7 @@ public abstract class MixinEntity implements IMixinEntity {
     @Shadow public abstract boolean hasNoGravity();
     @Shadow public abstract void setNoGravity(boolean noGravity);
     @Shadow public abstract void setPositionAndUpdate(double x, double y, double z);
+    @Shadow protected abstract void removePassenger(net.minecraft.entity.Entity passenger);
     @Shadow protected abstract void shadow$setRotation(float yaw, float pitch);
     @Shadow protected abstract void setSize(float width, float height);
     @Shadow protected abstract void applyEnchantments(EntityLivingBase entityLivingBaseIn, net.minecraft.entity.Entity entityIn);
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -774,6 +776,36 @@ public DataTransactionResult addPassenger(Entity entity) {
     }
 
     @Override
+    public DataTransactionResult removePassenger(Entity entity) {
+        checkNotNull(entity);
+        if (!entity.getPassengers().contains(this)) {
```
GotoLink, line 777: This seems backwards.

me4502: Thanks, fixed and re-tested. My test case was only testing clearPassengers().

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -774,6 +776,34 @@ public DataTransactionResult addPassenger(Entity entity) {
     }
 
     @Override
+    public DataTransactionResult removePassenger(Entity entity) {
+        checkNotNull(entity);
+        if (entity.getPassengers().contains(this)) {
+            throw new IllegalArgumentException(String.format("Cannot remove entity %s as a passenger of %s, because the former is not a passenger "
```
kashike: Perhaps word this a bit differently?

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -774,6 +776,34 @@ public DataTransactionResult addPassenger(Entity entity) {
...
+                    + "of the latter ", entity, this));
+        }
+
+        final DataTransactionResult.Builder builder = DataTransactionResult.builder();
+        this.removePassenger((net.minecraft.entity.Entity) entity);
+
+        final ImmutableList.Builder<EntitySnapshot> passengerSnapshotsBuilder = ImmutableList.builder();
```
kashike: Intended to be empty, or have the _removed_ passenger?

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -774,6 +776,34 @@ public DataTransactionResult addPassenger(Entity entity) {
...
+
+        return builder.result(DataTransactionResult.Type.SUCCESS).success(new ImmutableSpongeListValue<>(Keys.PASSENGERS, passengerSnapshotsBuilder
+                .build())).build();
+    }
+
+    @Override
+    public DataTransactionResult clearPassengers() {
+        this.removePassengers();
+
+        final ImmutableList.Builder<EntitySnapshot> passengerSnapshotsBuilder = ImmutableList.builder(); // This is an empty list. No passengers.
```
kashike: Intended to be empty, or a list of _removed_ passengers?

kashike (dismissed):

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
me4502: From what I can tell, empty - because it's returning the data after it's been changed. I'm not sure if this is right though.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
me4502: I accidentally moved that one as well as clearPassengers, so not intended to be empty :P

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -754,42 +756,46 @@ public Entity getBaseVehicle() {
...
 
-        final DataTransactionResult.Builder builder = DataTransactionResult.builder();
-        if (!((net.minecraft.entity.Entity) entity).startRiding((net.minecraft.entity.Entity) (Object) this, true)) {
-            return builder.result(DataTransactionResult.Type.FAILURE).reject(new ImmutableSpongeListValue<>(Keys.PASSENGERS, ImmutableList.of(entity
-                    .createSnapshot()))).build();
+    @Override
+    public void removePassenger(Entity entity) {
+        checkNotNull(entity);
+        if (entity.getPassengers().contains(this)) {
+            throw new IllegalArgumentException(String.format("Cannot remove entity %s, because it is not a passenger of %s ", entity, this));
```
liach, line 778: This looks weird.

me4502: How would you suggest it? I've been fairly unsure about that line, and changed it a few times by suggestion

liach: I think it should be
```java
if (!getPassengers().contains(entity)) {
```

me4502: Ah... must've overwritten that change last time I switched the wording.. thanks

PR #1008, opened 2016-11-01
Fix: Native mappings that clashed with plugins, checked the plugin pe…

ryantheleach:
…rmission, not the native permission.

Fixes:

https://forums.spongepowered.org/t/0-13-1-nucleus-the-ultimate-essentials-plugin-mc-1-8-9-and-1-10-2/11792/745

When checking the commandMapping this was unintentionally pulling the mapping for Nucleus.

Not 100% happy with this solution.
 Should we instead have a separate registry just for the command mappings with wrapped callables?

More testing is likely needed, just wanted to get a first draft out of the way for feedback.

@JBYoshi I'd be glad if you could give it a quick look over if you have time, since you touched this area last.

### DIFF
--- a/src/main/java/org/spongepowered/common/command/CommandPermissions.java
+++ b/src/main/java/org/spongepowered/common/command/CommandPermissions.java
@@ -24,15 +24,16 @@
  */
 package org.spongepowered.common.command;
 
-import org.spongepowered.api.command.CommandCallable;
 import org.spongepowered.api.command.CommandMapping;
 import org.spongepowered.api.command.CommandSource;
 import org.spongepowered.api.service.permission.SubjectData;
 import org.spongepowered.api.util.Tristate;
 import org.spongepowered.common.SpongeImpl;
 
-import java.util.Optional;
+import java.util.List;
+import java.util.StringJoiner;
 import java.util.function.BiFunction;
+import java.util.stream.Collectors;
 
 public final class CommandPermissions {
 
@@ -53,15 +54,28 @@ public final class CommandPermissions {
         if (commandName.equals(CommandPermissions.COMMAND_BLOCK_COMMAND)) {
             return source.hasPermission(CommandPermissions.COMMAND_BLOCK_PERMISSION);
         }
-        Optional<? extends CommandMapping> mapping = SpongeImpl.getGame().getCommandManager().get(commandName);
-        if (mapping.isPresent()) {
-            CommandCallable callable = mapping.get().getCallable();
-            if (callable instanceof MinecraftCommandWrapper) {
-                return source.hasPermission(((MinecraftCommandWrapper) callable).getCommandPermission());
-            } else {
-                return callable.testPermission(source);
-            }
+        //This method is only called for Minecraft/Modded commands, so don't look at commands provided by plugins.
+        final List<? extends CommandMapping> nativeMappings = SpongeImpl.getGame().getCommandManager().getAll(commandName).stream()
+                .filter(m -> m.getCallable() instanceof MinecraftCommandWrapper)
+                .collect(Collectors.toList());
+        //Forge rules for clashing commands is whatever registers last wins, but secondary aliases never override primary.
+        //thus we never expect there to be more then one native mapping for a given alias.
+        //Possible I guess, if some plugin (not mod) intentionally re-registers a minecraft/mod command under a clashing alias...
+        if (nativeMappings.size() > 1) {
+            final int size = nativeMappings.size();
+            final StringJoiner commaSeparated = new StringJoiner(", ");
+            nativeMappings.stream()
+                    .map(m -> ((MinecraftCommandWrapper) m).getOwner().getName() + ':' + m.getPrimaryAlias())
+                    .forEach(commaSeparated::add);
+
+            throw new IllegalStateException("Only one native mapping expected for /" + commandName + "found " + size + "(" + commaSeparated + ")");
         }
+        final CommandMapping mapping = nativeMappings.isEmpty() ? null : nativeMappings.get(0);
+        if (mapping != null) {
+            MinecraftCommandWrapper callable = (MinecraftCommandWrapper) mapping.getCallable();
+            return source.hasPermission(callable.getCommandPermission());
+        }
+        //Fallback for non-command mod permissions
         return source.hasPermission(commandName);
     }
 


src/main/java/org/spongepowered/common/command/CommandPermissions.java
```diff
@@ -53,15 +53,20 @@ public static boolean testPermission(CommandSource source, String commandName) {
         if (commandName.equals(CommandPermissions.COMMAND_BLOCK_COMMAND)) {
             return source.hasPermission(CommandPermissions.COMMAND_BLOCK_PERMISSION);
         }
-        Optional<? extends CommandMapping> mapping = SpongeImpl.getGame().getCommandManager().get(commandName);
-        if (mapping.isPresent()) {
-            CommandCallable callable = mapping.get().getCallable();
-            if (callable instanceof MinecraftCommandWrapper) {
-                return source.hasPermission(((MinecraftCommandWrapper) callable).getCommandPermission());
-            } else {
-                return callable.testPermission(source);
```
ryantheleach, line 62: This else seemed to be unused / useless as this method was only ever called on native commands.

src/main/java/org/spongepowered/common/command/CommandPermissions.java
```diff
@@ -53,15 +53,20 @@ public static boolean testPermission(CommandSource source, String commandName) {
...
-            }
+        //This method is only called for Minecraft/Modded commands, so don't look at commands provided by plugins.
+        final List<? extends CommandMapping> nativeMappings = SpongeImpl.getGame().getCommandManager().getAll(commandName).stream()
+                .filter(m -> m.getCallable() instanceof MinecraftCommandWrapper)
+                .collect(Collectors.toList());
+        //Forge rules for clashing commands is whatever registers last wins, but secondary aliases never override primary.
+        //thus we never expect there to be more then one native mapping for a given alias.
+        //Possible I guess, if some plugin (not mod) intentionally re-registers a minecraft/mod command under a clashing alias...
+        if(nativeMappings.size() > 1) throw new IllegalStateException("Only one native mapping expected for "+commandName);
```
kashike: `if (` (space), missing braces

kashike (changes requested):

ryantheleach:
I've done some quick testing by loading up HermitPack in single player production environment and testing using /help which calls a permission check for each of the commands in order to display them.

Additionally I've tested this without mods in a debugger using the /forge command.

Loaded up nucleus in a production SF environment with LuckPerms, /enchant works with nucleus.enchant.base /minecraft:enchant No permission.

Seems to work nicely.

kashike (approved):

src/main/java/org/spongepowered/common/command/CommandPermissions.java
```diff
@@ -53,15 +53,22 @@ public static boolean testPermission(CommandSource source, String commandName) {
...
-            }
+        //This method is only called for Minecraft/Modded commands, so don't look at commands provided by plugins.
+        final List<? extends CommandMapping> nativeMappings = SpongeImpl.getGame().getCommandManager().getAll(commandName).stream()
+                .filter(m -> m.getCallable() instanceof MinecraftCommandWrapper)
+                .collect(Collectors.toList());
+        //Forge rules for clashing commands is whatever registers last wins, but secondary aliases never override primary.
+        //thus we never expect there to be more then one native mapping for a given alias.
+        //Possible I guess, if some plugin (not mod) intentionally re-registers a minecraft/mod command under a clashing alias...
+        if (nativeMappings.size() > 1) {
+            throw new IllegalStateException("Only one native mapping expected for " + commandName);
```
kashike: Perhaps specify how many were found.

src/main/java/org/spongepowered/common/command/CommandPermissions.java
```diff
@@ -53,15 +54,28 @@ public static boolean testPermission(CommandSource source, String commandName) {
...
-            } else {
-                return callable.testPermission(source);
-            }
+        //This method is only called for Minecraft/Modded commands, so don't look at commands provided by plugins.
+        final List<? extends CommandMapping> nativeMappings = SpongeImpl.getGame().getCommandManager().getAll(commandName).stream()
+                .filter(m -> m.getCallable() instanceof MinecraftCommandWrapper)
+                .collect(Collectors.toList());
+        //Forge rules for clashing commands is whatever registers last wins, but secondary aliases never override primary.
+        //thus we never expect there to be more then one native mapping for a given alias.
+        //Possible I guess, if some plugin (not mod) intentionally re-registers a minecraft/mod command under a clashing alias...
```
JBYoshi, line 63: I'd like to verify this; if I have time I'll check later today.

src/main/java/org/spongepowered/common/command/CommandPermissions.java
```diff
@@ -53,15 +54,28 @@ public static boolean testPermission(CommandSource source, String commandName) {
...
+        if (nativeMappings.size() > 1) {
+            final int size = nativeMappings.size();
+            final StringJoiner commaSeparated = new StringJoiner(", ");
+            nativeMappings.stream()
+                    .map(m -> ((MinecraftCommandWrapper) m).getOwner().getName() + ':' + m.getPrimaryAlias())
+                    .forEach(commaSeparated::add);
+
+            throw new IllegalStateException("Only one native mapping expected for /" + commandName+"found "+size +"("+ commaSeparated +")");
```
JBYoshi: This line needs a few formatting fixes.

JBYoshi (changes requested):

src/main/java/org/spongepowered/common/command/CommandPermissions.java
JBYoshi: There will be multiple items in `nativeMappings` if there is a name clash. The fix for this (other than preventing conflicting native commands altogether) is to somehow pass the command from `CommandBase.testPermission()` to this method. I'd use a `ThreadLocal` for this; there's some example code in `MinecraftCommandWrapper` which I used to propagate exceptions in commands through Minecraft back to Sponge if you want to do it.

Here's some test code for a conflicting command:

``` java
import net.minecraft.command.*;
import net.minecraft.server.MinecraftServer;
import net.minecraft.util.text.TextComponentString;
import net.minecraftforge.fml.common.Mod;
import net.minecraftforge.fml.common.event.FMLServerStartingEvent;
import java.util.*;

@Mod(modid="spongepermissiontestforge")
public class SpongePermissionTestForge {
    @Mod.EventHandler
    public void onFMLServerStarting(FMLServerStartingEvent event) {
        event.registerServerCommand(new CommandBase() {
            @Override public String getCommandName() {return "tp";}
            @Override public String getCommandUsage(ICommandSender sender) {return "/tp from Forge mod";}
            @Override public void execute(MinecraftServer server, ICommandSender sender, String[] args) throws CommandException {sender.addChatMessage(new TextComponentString("Forge mod command"));}
        });
    }
}
```

JBYoshi (changes requested):

src/main/java/org/spongepowered/common/command/CommandPermissions.java
ryantheleach: But just wanted to ask if you had tested this? From what I could see (without testing purposely clashing forge commands) is that forge will overwrite the tp mapping with SpongePermissionTestForge's permission mapping.

So unless this is covered by a mixin somewhere it should be fine?

Anyway I'll try to get around to compiling this test-mod at some stage to have hard imperative proof one way or another.

ryantheleach:
Closed in: https://github.com/SpongePowered/SpongeCommon/pull/1262 ?

PR #1010, opened 2016-11-02
Lens for all containers

Faithcaio:
Change LensProvider to accept Fabric allowing passing ContainerFabrics too.

New ContainerLens used for Fallback when no Lens can be found for a Container.
Use LensProviders for ContainerChest and ContainerPlayer. (more coming later)

When getting a Lens for a Container we first try to get existing Lenses from Adapter.
If the Container is no Adapter or has no Lens set yet we check for a LensProvider.
If there is still no Lens a new one will be generated for non-zero sized inventories:

For each inventory viewed in the Container we try to obtain a Lens and then wrap all of it in a ContainerLens.


src/main/java/org/spongepowered/common/item/inventory/lens/impl/minecraft/ContainerLens.java
```diff
@@ -0,0 +1,118 @@
...
+
+    public ContainerLens(Container container, InventoryAdapter<IInventory, ItemStack> adapter, SlotProvider<IInventory, ItemStack> slots) {
+        super(0, adapter.getInventory().getSize(), adapter, slots);
+        this.container = container;
+        this.init(slots);
+    }
+
+    @Override
+    protected void init(SlotProvider<IInventory, ItemStack> slots) {
+        System.out.print("### ContainerLens for: " + container.getClass().getName() + " ###\n");
```
gabizou: I'd recommend looking into using `PrettyPrinter` so you can have some pretty printed logs for debugging aid.

Mumfrey:
The original design is for `LensProvider` to do this, there shoudn't really be a "generic container Lens" because the concept (as evidenced by the complicated heuristics you've have to employ) doesn't really translate all that well to real applications without kind of dissolving in its own complexity.

Essentially an `Adapter` should be able to initialise its root lens by inspecting the fabric it receives, the idea being that if a particular fabric is actually a LensProvider, then adapter can cast down and obtain the relevant root lens from the fabric itself.

The problem with the lens having to "work out" its own internal structure is that it compromises the concern that a lens is supposed to have. The kind of details you're encoding at the lens level here are supposed to be taken care of at the fabric or adapter level (eg. either side of the lens assembly) rather than by the lens itself.

There's a simple way to determine what logic belongs where in the 3-tier system which exists:
- _Does it represent the **real** structure of an underlying inventory?_ Fabric
- _Does it represent the **logical** structure of an underlying inventory?_ Lens
- _Does it represent the logical structure of an inventory at the API level?_ Adapter

It's up to adapters to decide what _logical_ layout they are presenting, and thus the adapter builds the lens structure it requires in order to do so. Lenses never know any more than they need to do so to represent their own logical view: in a nutshell, the implementation only ever "looks through" the lens, kind of like a router. Lenses can delegate to sub-lenses where appropriate of course (eg. in the row/column sub-lenses of grid) but these are fundamental properties of the lenses themselves and are thuse encoded in the contract of those lens types.

Now in general since the Adapter _is_ the inventory in question (yay mixin) the structure is understood and the Adapter can build the lens. Revisiting our understanding of inventories from the initial case study, the Adapter is actually the Model in this instance, so it can do so with impunity.

Fabric in all cases represents a ViewModel layer, but may either delegate to a Model (an inventory) or another ViewModel (a Container) as in `DelegatingFabric`, where we essentially have a MVVMVM (!!!!) due to the the existence of two ViewModels!

Since Adapter represents a delegating View, and lenses encode the actual _structure_ of that View (and aren't expected to do anything else) it makes sense that this is normally the order of things.

Of course this comes apart with unknown containers since we have to create a View with no knowledge of the underlying Model, aside from what we can infer heuristically. What you have done here is encode the necessary heuristics, but in my view have done so at the wrong level.

Remember that an Adapter takes responsibility for building the lens structure - this is important due to the separation of concerns outlined above - but of course becomes problematic when the Adapter has no knowledge of the structure it needs to build. This is where `LensProvider` is supposed to step in.

The object "at the coal face" is the _Fabric_, and it is this which is meant to supply the information to the Adapter when the underlying model structure is unknown. The reason is that this can then support an _inheritance-based_ system for decorating certain containers with their known lens structures - which can then be consumed by a wrapping Adapter - and decentralises the logic required to obtain lenses for all types of known or unknown inventory.

The heuristics you have are therefore correct, but missing a couple of crucial implementation features.
- There should be no `ContainerLens` - the concept doesn't make sense in the intended pattern for lenses
- Any object (like a Furnace, chest, etc) which needs to _provide its own lenses_ should implement `LensProvider` and return the appropriate root lens with structure returned
- A `Fabric` which represents a container should also be a `LensProvider` and return **either** the lens from its underlying `LensProvider` **or** should fall back to heuristics only if the container is not able to return a lens for itself.

The entire point and purpose of this approach is that then _any_ inventory which wishes to encode its structure in a lens assembly is free to do so simply by returning said assembly. Heuristics occur at the `Fabric` level and are propagated up to the `Adapter` level and lenses have no intrinsic understanding of their underlying target. This maintains separation of concerns between Adapter/Lens/Fabric and ensures that all code required to build lens assemblies is kept where it belongs, with heuristics only ever becoming a fallback.

This design is the reason why:
- `initRootLens` is a separate method in `Adapter` so that adapters can initialise their own lenses when possible, but delegate at other times
- `LensProvider` exists at all, it is the stub of the functionality which wasn't completed in the initial pass

Hope that makes some sense.

src/main/java/org/spongepowered/common/mixin/core/item/inventory/MixinCustomContainer.java
```diff
@@ -79,7 +69,7 @@ private void onConstructed(EntityPlayer player, CustomInventory inventory, Callb
 
     @Override
     public Fabric<IInventory> getInventory() {
-        return inventory;
+        return fabric;
```
kashike: Missing `this.`.

kashike (changes requested):

Faithcaio:
@Mumfrey are those LensProviders looking good?

src/main/java/org/spongepowered/common/item/inventory/lens/impl/minecraft/ContainerLens.java
Faithcaio: Can I prevent it from logging twice?

`new PrettyPrinter(60).add("TEST").trace(System.out);`

```
[09:52:51] [Server thread/DEBUG] [ContainerUtil]: /****************************************************************/
[09:52:51] [Server thread/DEBUG] [ContainerUtil]: /* TEST                                                         */
[09:52:51] [Server thread/DEBUG] [ContainerUtil]: /****************************************************************/
> /****************************************************************/
/* TEST                                                         */
/****************************************************************/
```

I am using `System.out.print(...)` with `\n` because is does not print out the whole `[09:52:51] [Server thread/DEBUG] [ContainerUtil]`.

gabizou: You should use printer.trace(Level.TRACE, System.err, SpongeImpl.getLogger()) or whatever it is that I end up using. the thing is that there's the `System.err` logger and then there's the console logger. One will show up in the log files and the other only shows up in System.err.

src/main/java/org/spongepowered/common/mixin/core/item/inventory/MixinContainerChest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.spongepowered.common.item.inventory.lens.impl.minecraft.container.ContainerChestInventoryLens;
+
+@Mixin(ContainerChest.class)
+public abstract class MixinContainerChest extends MixinContainer implements LensProvider<IInventory, ItemStack> {
+
+    private ContainerChest this$ = ((ContainerChest) (Object)this);
+
+    @Override
+    public Lens<IInventory, ItemStack> getRootLens(IInventory inv, InventoryAdapter<IInventory, ItemStack> adapter) {
+        return new ContainerChestInventoryLens(adapter, inventory$getSlotProvider(), this$.numRows);
```
kashike: `this.` missing on `this$`.

kashike (changes requested):

src/main/java/org/spongepowered/common/item/inventory/lens/impl/minecraft/PlayerInventoryLens.java
```diff
@@ -62,12 +63,18 @@ protected void init(SlotProvider<IInventory, ItemStack> slots) {
         this.equipment = new EquipmentInventoryLensImpl((ArmorEquipable) player, 36, 4, 1, slots);
         this.offhand = new SlotLensImpl(37);
 
+        // TODO Hotbar in Vanilla is part of the main inventory (first 9 slots)
```
kashike: This is true, but what about it? Are you wanting to remove `Hotbar`?

Faithcaio: add a lens that wraps both maybe?

src/main/java/org/spongepowered/common/item/inventory/util/ContainerUtil.java
```diff
@@ -200,18 +219,92 @@ public static void performBlockInventoryDrops(WorldServer worldServer, double x,
...
+        // Container provides Lens?
+        if (container instanceof LensProvider) {
+            System.out.print("LensProvider found\n");
+            InventoryAdapter<IInventory, ItemStack> adapter;
+            if (container instanceof InventoryAdapter) {
+                adapter = ((InventoryAdapter) container);
+            } else {
+                adapter = new Adapter(MinecraftFabric.of(container));
+            }
+            return ((LensProvider) container).getRootLens(null, adapter); // TODO 1st param is null because we have a container and not IInventory
```
kashike: `inv` parameter of `getRootLens` should be annotated with `@Nullable` if this `null` is to remain.

src/main/java/org/spongepowered/common/mixin/core/tileentity/MixinTileEntityLockable.java
```diff
@@ -90,4 +94,16 @@ public void supplyVanillaManipulators(List<DataManipulator<?, ?>> manipulators)
         }
     }
 
+    @Override
+    public TileEntityInventory<TileEntityCarrier> getInventory() {
+        return (TileEntityInventory<TileEntityCarrier>) this;
+    }
+
+    public Optional tileentityinventory$getTileEntity() {
```
kashike: Raw type?

src/main/java/org/spongepowered/common/mixin/core/tileentity/MixinTileEntityLockable.java
```diff
@@ -90,4 +94,16 @@ public void supplyVanillaManipulators(List<DataManipulator<?, ?>> manipulators)
...
+        return Optional.of(this);
+    }
+
+    public Optional tileentityinventory$getCarrier() {
```
kashike: Raw type?

kashike (changes requested):

src/main/java/org/spongepowered/common/mixin/core/tileentity/MixinTileEntityLockable.java
Faithcaio: I cannot know the Type if the TileEntityInventory here.
I just saw that this methods are implemented 6 times with `Optional.of(this)` (and in current code still are)
and thought this could be simplyfied

src/main/java/org/spongepowered/common/item/inventory/util/ContainerUtil.java
Faithcaio: I'd like to not pass `null` but as the TODO says. I need an IInventory and have a Container.
Maybe remove the parameter or allow any Object
@Mumfrey ?

src/main/java/org/spongepowered/common/item/inventory/lens/impl/minecraft/PlayerInventoryLens.java
kashike: That could work.

Faithcaio:
@Mumfrey 
What are the Inventory methods using the Iterator (`Inventory#iterator` ; `Inventory#first` ; `Inventory#next`) supposed to return?
I think it should only be the spanning children as you can get the slots with `Inventory#slots` already.
So e.g. for (normal) Containers you would get adapter for:
 - The viewed inventory
 - The main player inventory
 - The player hotbar inventory

Is that correct? (Changes in last commit)

Currently the Iterator iterates over every child which results in iterating over the inventory multiple times.

Mumfrey:
Correct, the iterator is supposed to just iterate over the immediate (spanning) children. Basically from the original PR I mention that iteration order is via a notional spanning tree, the choice of which children are "spanning" is arbitrary but should be logical (eg. a grid could have spanning rows or spanning columns but spanning columns is more intuitive). At the "bottom" level (eg a row), the slots are the spanning children.

So yes, for a given container the spanning children are the direct sub-inventories, eg. the chest grid, player grid, player hotbar and offhand slot.

Faithcaio:
https://github.com/SpongePowered/SpongeCommon/commit/dbdea6856b9fb570f7e26b4587c3d48f1b0198dc

PR #1011, opened 2016-11-02
Fix the getOpenInventory method.

Cybermaxke:
The `Player#getOpenInventory` method returns currently always a container, even if the player inventory is closed. The reason of this is that the `openContainer` is never, but rather reset to the `inventoryContainer` when a inventory is closed.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketFunction.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketFunction.java
@@ -81,6 +81,7 @@ import org.spongepowered.common.event.tracking.ItemDropData;
 import org.spongepowered.common.event.tracking.PhaseContext;
 import org.spongepowered.common.event.tracking.TrackingUtil;
 import org.spongepowered.common.interfaces.IMixinContainer;
+import org.spongepowered.common.interfaces.entity.player.IMixinEntityPlayerMP;
 import org.spongepowered.common.interfaces.network.IMixinNetHandlerPlayServer;
 import org.spongepowered.common.interfaces.world.IMixinWorldServer;
 import org.spongepowered.common.item.inventory.adapter.impl.slots.SlotAdapter;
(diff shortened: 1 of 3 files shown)


Aaron1011:
@Cybermaxke: I'm not sure that this is correct - it leaves plugins with no way of getting the player's normal inventory container (which is of course always technically open).

Cybermaxke:
@Aaron1011 Then maybe add a different method to get that container. This currently breaks a way of knowing that the player is currently viewing a inventory and kinda breaks the point of using a `Optional` in the first place.

simon816:
Player implements `Carrier` which means it has a `getInventory()` method

bloodmc:
@Faithcaio review this please.

Faithcaio:
What do we want to do with the player container?

Cybermaxke:
@Faithcaio Only return the the player inventory container when it's actually open?

Sir-Will:
hey, are there any news to this?

Cybermaxke:
This PR will be useless in 1.12, because the `OPEN_INVENTORY` client status got removed. If this should be fixed, it will be with a custom client, but then will it not work on vanilla.

Cybermaxke:
Since the update to 1.12 is this PR broken and not fixable in vanilla mc.

MarkTwoFive:
If there's no other way of getting the players currently opened inventory in 1.12+ I'd suggest to merge this fix for <1.12 versions so they're fixed at least and add a deprecation notice for 1.12.

What do you think?

Cybermaxke:
I don't like the idea of having one method where the behavior is different on specific versions. If there is another workaround for 1.12, I would agree on this being merged, otherwise I would leave it as it is now.

MarkTwoFive:
@Cybermaxke IMO there's even less sense in having a bugged method in every version but one.
Also: That's exactly why deprecation exists in the first place.

Cybermaxke:
@MarkL4YG This should give the expected behavior, but if it works only on one version we might as well keep the old behavior to be consistent even though it might be "bugged". But I think that it should be stated in the API that this method will always return a container. If there is a specific one open, it returns that one, otherwise the player inventory container.

MarkTwoFive:
@Cybermaxke It appears that this is bugged in all versions except in 1.12 where this is no longer working at all.
Fixing the bug (which would go in favor of all usages unaware of it) or changing the purpose (which would require to change the signature/name too) and leaving a deprecation notice to tell devs that this method will no longer work in 1.12 and above seems like a reasonable thing to do.
Doing the first it's possible just to return an empty Optional for 1.12+ in order to avoid crashes.

Issue #868, opened 2016-08-02
Operator bypass whitelist with permission service


lesbleu:
Hi,
with a permissions service, operators can bypass the whitelist.
It's possible to disable this ?

https://github.com/SpongePowered/SpongeAPI/issues/1311

phit:
confirmed with 7.0.0-BETA-2741

ryantheleach:
~~~I'll take this one for a week, unless you were planning on it @phit .~~~

phit:
go for it 👍

Cybermaxke:
i have already a PR that resolves this issue. But it should be updated.

https://github.com/SpongePowered/SpongeCommon/pull/1012

ryantheleach:
@Cybermaxke would you like to update it, or would you prefer I did it?

Cybermaxke:
I can update it, I don't think that it will need a lot of changes, just some comments and checking if it still works.

PR #1012, opened 2016-11-03
Added login bypass permissions.

Cybermaxke:
Common | [Forge](https://github.com/SpongePowered/SpongeForge/pull/1023) | [Vanilla](https://github.com/SpongePowered/SpongeVanilla/pull/277)

Fixes https://github.com/SpongePowered/SpongeCommon/issues/868

This PR adds 2 permissions, one that can be used to bypass the whitelist (defaults to true when op) and a permission to bypass the player limit.

`minecraft.login.bypass-whitelist`
`minecraft.login.bypass-player-limit`

### DIFF
--- a/src/main/java/org/spongepowered/common/entity/player/LoginPermissions.java
+++ b/src/main/java/org/spongepowered/common/entity/player/LoginPermissions.java
@@ -1,0 +1,37 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.entity.player;
+
+public final class LoginPermissions {
+
+    public static final int BYPASS_WHITELIST_LEVEL = 1;
+
+    public static final String BYPASS_WHITELIST_PERMISSION = "minecraft.login.bypass-whitelist";
+
+    public static final String BYPASS_PLAYER_LIMIT_PERMISSION = "minecraft.login.bypass-player-limit";
+
+    private LoginPermissions() {
+    }
+}
--- a/src/main/java/org/spongepowered/common/mixin/core/server/MixinDedicatedPlayerList.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/server/MixinDedicatedPlayerList.java
@@ -1,0 +1,76 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.mixin.core.server;
+
+import com.mojang.authlib.GameProfile;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.service.permission.PermissionService;
+import org.spongepowered.api.service.permission.Subject;
+import org.spongepowered.api.service.permission.SubjectData;
+import org.spongepowered.api.util.Tristate;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.common.entity.player.LoginPermissions;
+import org.spongepowered.common.service.permission.SpongePermissionService;
+
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.server.dedicated.DedicatedPlayerList;
+import net.minecraft.server.management.PlayerList;
+
+@Mixin(DedicatedPlayerList.class)
+public abstract class MixinDedicatedPlayerList extends PlayerList {
+
+    public MixinDedicatedPlayerList(MinecraftServer server) {
+        super(server);
+    }
+
+    @Inject(method = "canJoin", at = @At("HEAD"), cancellable = true)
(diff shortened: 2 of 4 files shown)


src/main/java/org/spongepowered/common/service/permission/SpongePermissionService.java
```diff
@@ -180,4 +182,11 @@ public void addDescription(PermissionDescription permissionDescription) {
     public SpongeSubjectCollection getDefaultCollection() {
         return defaultCollection;
     }
+
+    public void setupDefaultPermissions() {
+        final SubjectData opGroupLevel1Data = getGroupForOpLevel(1).getSubjectData();
+        opGroupLevel1Data.setPermission(SubjectData.GLOBAL_CONTEXT, "minecraft.selector", Tristate.TRUE);
```
JBYoshi: `CommandPermissions.SELECTOR_PERMISSION` and `CommandPermissions.SELECTOR_LEVEL`

src/main/java/org/spongepowered/common/service/permission/SpongePermissionService.java
```diff
@@ -180,4 +182,11 @@ public void addDescription(PermissionDescription permissionDescription) {
...
+        opGroupLevel1Data.setPermission(SubjectData.GLOBAL_CONTEXT, "minecraft.login.bypass-whitelist", Tristate.TRUE);
+        getGroupForOpLevel(2).getSubjectData().setPermission(SubjectData.GLOBAL_CONTEXT, "minecraft.commandblock", Tristate.TRUE);
```
JBYoshi: `CommandPermissions.COMMAND_BLOCK_PERMISSION` and `CommandPermissions.COMMAND_BLOCK_LEVEL`

src/main/java/org/spongepowered/common/mixin/core/server/MixinDedicatedPlayerList.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Overwrite
+    @Override
+    public boolean canJoin(GameProfile profile) {
+        if (!isWhiteListEnabled() || getWhitelistedPlayers().isWhitelisted(profile)) {
+            return true;
+        }
+        final Subject subject = Sponge.getServiceManager().provideUnchecked(PermissionService.class)
+                .getUserSubjects().get(profile.getId().toString());
+        return subject.hasPermission("minecraft.login.bypass-whitelist");
```
JBYoshi: Could you make this a constant?

src/main/java/org/spongepowered/common/mixin/core/server/MixinDedicatedPlayerList.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Overwrite
+    @Override
+    public boolean bypassesPlayerLimit(GameProfile profile) {
+        final Subject subject = Sponge.getServiceManager().provideUnchecked(PermissionService.class)
+                .getUserSubjects().get(profile.getId().toString());
+        return subject.hasPermission("minecraft.login.bypass-player-limit");
```
JBYoshi: This too.

JBYoshi (changes requested):

src/main/java/org/spongepowered/common/mixin/core/server/MixinDedicatedPlayerList.java
```diff
@@ -0,0 +1,75 @@
...
+import org.spongepowered.common.service.permission.SpongePermissionService;
+
+@Mixin(DedicatedPlayerList.class)
+public abstract class MixinDedicatedPlayerList extends PlayerList {
+
+    public MixinDedicatedPlayerList(MinecraftServer server) {
+        super(server);
+    }
+
+    @Overwrite
```
Deamon5550: Need comments on these overwrites (Here's an [example](https://github.com/SpongePowered/SpongeCommon/blob/9e6acff2f9c25f346151420d368a26a0e3578a5b/src/main/java/org/spongepowered/common/mixin/core/world/biome/MixinBiomeDecorator.java#L39)).

Also should look if they can be injectors or redirects instead as those have much less overhead for maintenance and these look pretty basic.

Cybermaxke:
@Deamon5550 
Done.
I also tried a Redirect for the `canJoin` method, but I couldn't get it to work and it didn't give any useful exceptions to resolve the issue, any ideas?

``` java
@Redirect(method = "canJoin(Lcom/mojang/authlib/GameProfile;)Z", at = @At(value = "INVOKE",
        target = "Lnet/minecraft/server/management/PlayerList;canSendCommands(Lcom/mojang/authlib/GameProfile;)Z"))
private boolean onCanJoin(PlayerList playerList, GameProfile profile0, GameProfile profile1) {
    final Subject subject = Sponge.getServiceManager().provideUnchecked(PermissionService.class)
            .getUserSubjects().get(profile1.getId().toString());
    return subject.hasPermission(BYPASS_WHITELIST_PERMISSION);
}
```

kashike:
Did it give any exceptions, even if "not useful"?

Cybermaxke:
@kashike 
http://pastebin.com/qUKzC2kz
It just tells that there is a critical injection error, not really the cause of it.

kashike:
Have you tried:

``` java
@Redirect(method = "canJoin", at = @At(value = "INVOKE",
        target = "Lnet/minecraft/server/management/PlayerList;canSendCommands(Lcom/mojang/authlib/GameProfile;)Z"))
private boolean onCanJoin(PlayerList playerList, GameProfile profile) {
    final Subject subject = Sponge.getServiceManager().provideUnchecked(PermissionService.class)
            .getUserSubjects().get(profile.getId().toString());
    return subject.hasPermission(BYPASS_WHITELIST_PERMISSION);
}
```

Cybermaxke:
@kashike Yes, that throws the same exception.

Zidane:
@Aaron1011 

Finish up this PR.

src/main/java/org/spongepowered/common/mixin/core/command/MixinEntitySelector.java
```diff
@@ -37,6 +39,6 @@
     @Redirect(method = "matchEntities", at = @At(value = "INVOKE", target = "net.minecraft.command.ICommandSender.canCommandSenderUseCommand"
             + "(ILjava/lang/String;)Z"))
     private static boolean redirectCanUseCommand(ICommandSender self, int opLevel, String command) {
-        return WrapperCommandSource.of(self).hasPermission("minecraft.selector");
+        return WrapperCommandSource.of(self).hasPermission(SELECTOR_PERMISSION);
```
gabizou: I'd rather qualify static fields instead of importing them.

Cybermaxke:
Status?

gabizou (changes requested):
Don't static import fields like this, otherwise looks fine. I meant to comment about this a while back.

src/main/java/org/spongepowered/common/service/permission/SpongePermissionService.java
```diff
@@ -180,4 +188,13 @@ public void addDescription(PermissionDescription permissionDescription) {
     public SpongeSubjectCollection getDefaultCollection() {
         return defaultCollection;
     }
+
+    public void setupDefaultPermissions() {
+        getGroupForOpLevel(SELECTOR_LEVEL).getSubjectData()
+                .setPermission(SubjectData.GLOBAL_CONTEXT, SELECTOR_PERMISSION, Tristate.TRUE);
```
ryantheleach: I like that you have placed this into SpongeCommon, but I'd sooner see it not a method of the permissions service. 
Sponge should set up it's default permissions similar to how plugins do it, not make it optional based on what permissions service is running.

ryantheleach (changes requested):

Cybermaxke:
@gabizou I fixed the static imports.

parlough:
Would be great to see this in.

kashike:
@Cybermaxke rebase?

src/main/java/org/spongepowered/common/entity/player/LoginPermissions.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.spongepowered.common.entity.player;
+
+public final class LoginPermissions {
+
+    public static final int BYPASS_WHITELIST_LEVEL = 1;
+
+    public static final String BYPASS_WHITELIST_PERMISSION = "minecraft.login.bypass-whitelist";
+
+    public static final int BYPASS_PLAYER_LIMIT_LEVEL = 1;
```
Zidane: This isn't used.

src/main/java/org/spongepowered/common/service/permission/SpongePermissionService.java
```diff
@@ -180,4 +185,13 @@ public void addDescription(PermissionDescription permissionDescription) {
     public SpongeSubjectCollection getDefaultCollection() {
         return defaultCollection;
     }
+
+    public void setupDefaultPermissions() {
```
Zidane: This is never called.

Cybermaxke: This method is called in SpongeForge and SpongeVanilla

Zidane: Oops, I thought I merged your forge commit as well into my local bleeding but apparently forgot to do so.

gabizou: If it's being used, it should be javadoc'ed as such to avoid future confusion (I've had other methods that were made public in the past and when I tried to rename them or change them, they'd magically fail on either implementation because of the lack of documentation ;))

Cybermaxke:
I updated the PRs.

ImMorpheus:
Tested with luckperms, everything is working.

src/main/java/org/spongepowered/common/entity/player/LoginPermissions.java
```diff
@@ -0,0 +1,37 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.entity.player;
+
+public final class LoginPermissions {
+
+    public static final int BYPASS_WHITELIST_LEVEL = 1;
+
+    public static final String BYPASS_WHITELIST_PERMISSION = "minecraft.login.bypass-whitelist";
```
gabizou, line 31: Is the permission native to Minecraft or Sponge?

gabizou (commented):
Unsure about the permissions themselves, but otherwise all good.

src/main/java/org/spongepowered/common/entity/player/LoginPermissions.java
ImMorpheus: I'm pretty sure there is no such permission in Minecraft

gabizou: Then it should be sponge based.

Cybermaxke: Bypassing login and whitelist is a minecraft feature, we just modify it to be a permission. The same is done for the selector and command block permissions. This is different for the help command permission since a custom 'sponge' based help command is provided instead.

ImMorpheus: On second thought, this is consistent with minecraft.commandblock and minecraft.selector.

PR #1015, opened 2016-11-05
Implement Extent.getHighestBlockAt

Maxqia:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1416) | **Common**

EDIT: Patch now in the API PR
Double EDIT : Split into new PR


liach:
It should be Vector2i.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -355,6 +356,11 @@ public boolean setBlock(int x, int y, int z, BlockState block, Cause cause) {
 
 
     @Override
+    public Location<World> getHighestBlockAt(Vector3i position) {
+        return new Location<World>(this, position.getX(), this.func_189649_b(position.getX(), position.getZ()), position.getZ());
```
Deamon5550: This is going to return `(x, 0, z)` when called in an unloaded chunk, I don't feel like this is the best behaviour for us to return but I'm not sure what the best thing for us to do is.

Maxqia: Imma change this to load a chunk ... one day ...

Zidane (commented):
This PR needs better documentation to state that you will not get a height value besides 0 unless the chunk is loaded ahead of time.

PR #1020, opened 2016-11-11
PaginationService doesn't support \n

rexbut:
The pagination service doesn't support the character "\n". With the character " \n", the number of line is different to the real number of line.

- **Before :**

Error_1
Error_2

- **After :**

Fix_1
Fix_2

### DIFF
--- a/src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java
+++ b/src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java
@@ -161,13 +161,21 @@ public class PaginationCalculator {
                     .getBold();
 
             Integer cp;
+            boolean new_line = false;
             while(i_it.hasNext()){
                 cp = i_it.next();
                 if (cp == '\n') {
-                    total = ((int) Math.ceil((double) total / LINE_WIDTH)) * LINE_WIDTH;
+                    // if the previous character is a '\n'
+                    if (new_line) {
+                        total += LINE_WIDTH;
+                    } else {
+                        total = ((int) Math.ceil((double) total / LINE_WIDTH)) * LINE_WIDTH;
+                        new_line = true;
+                    }
                 } else {
                     int width = getWidth(cp, bold);
                     total += width;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java
```diff
@@ -163,8 +163,13 @@ int getWidth(Text text) {
             Integer cp;
             while(i_it.hasNext()){
                 cp = i_it.next();
-                int width = getWidth(cp, bold);
-                total += width;
+                //The newline character is 10
+                if (cp == 10) {
```
simon816: Would be more clear to use `'\n'` here instead of `10`

windy1 (dismissed):
Just tested this. The issue I found looks to be that you don't iterate over the text children deeply, meaning it will only pick up newlines on the top children level.

stephan-gh:
@windy1 That is weird because the code uses `Text.withChildren()` which provides an recursive iterator over all children of a Text instance. It should process all children correctly.

windy1:
I could be wrong then, I wasn't aware of that and that was really more of a (not so) educated guess, however I tested with nested text with newlines in them and it didn't work :/

rexbut:
@windy1 Didn't worked when there was /n several times.

src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java
```diff
@@ -161,10 +161,22 @@ int getWidth(Text text) {
                     .getBold();
 
             Integer cp;
+            boolean new_line = false;
```
kashike: This should be called `newLine`, not `new_line`.

kashike (changes requested):

kashike (approved):

randombyte-developer:
This also seems to fix an issue with headers(probably footers too): When having a multiline header the pages get to long. The fixed `getWidth()` should fix that.

PR #1021, opened 2016-11-12
Create method to remove custom GameRule

rexbut:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1410) | Common

I implement a new method to remove custom gamerule because there are no solution to remove custom gamerule.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/MixinGameRules.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/MixinGameRules.java
@@ -1,0 +1,72 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.mixin.core.world;
+
+import net.minecraft.world.GameRules;
+import com.google.common.collect.ImmutableSet;
+import org.spongepowered.api.world.gamerule.DefaultGameRules;
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/common/mixin/core/world/MixinGameRules.java
```diff
@@ -0,0 +1,55 @@
...
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.common.interfaces.world.IMixinGameRules;
+
+import java.util.Set;
+import java.util.TreeMap;
+
+@Mixin(GameRules.class)
+public abstract class MixinGameRules implements IMixinGameRules {
+
+    private static final Set<String> DEFAULT = Sets.newHashSet((new GameRules()).getRules());
```
kashike: Why not just create a set from the fields in https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/world/gamerule/DefaultGameRules.java?

kashike (changes requested):

src/main/java/org/spongepowered/common/mixin/core/world/MixinGameRules.java
```diff
@@ -24,20 +24,43 @@
...
 import java.util.Set;
 import java.util.TreeMap;
 
 @Mixin(GameRules.class)
 public abstract class MixinGameRules implements IMixinGameRules {
 
-    private static final Set<String> DEFAULT = Sets.newHashSet((new GameRules()).getRules());
+    private static final Set<String> DEFAULT = new HashSet<String>();
+
+    static {
```
rexbut: @kashike : Like this ?

Cybermaxke: @rexbut Use reflection to get all the default game rules, to make sure that this list is always up to date:

``` java
final ImmutableSet.Builder<String> builder = ImmutableSet.builder();
for (Field field : DefaultGameRules.class.getFields()) {
    if (Modifier.isStatic(field.getModifiers()) && field.getType() == String.class) {
        try {
            builder.add((String) field.get(null));
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
}
DEFAULT = builder.build();
```

rexbut: @Cybermaxke : Okay thank you so much 
https://github.com/SpongePowered/SpongeCommon/pull/1021/commits/1ceb19d64e3e69c1bb41081c8c9d866076ec2ce1

src/main/java/org/spongepowered/common/mixin/core/world/MixinGameRules.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Set;
+import java.util.TreeMap;
+
+@Mixin(GameRules.class)
+public abstract class MixinGameRules implements IMixinGameRules {
+
+    private static final Set<String> DEFAULT;
+
+    static {
+        final ImmutableSet.Builder<String> builder = ImmutableSet.builder();
```
gabizou: Does this static initialization even work in production? @Mumfrey I thought we could statically initialize fields, but not add to the static initializer in this way...

src/main/java/org/spongepowered/common/mixin/core/world/MixinGameRules.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Shadow @Final private TreeMap<String, ?> theGameRules;
+
+    @Shadow public abstract String getString(String name);
+
+    @Override
+    public boolean removeGameRule(String gameRule) {
+        // Cannot remove default gamerule
+        if (DEFAULT.contains(gameRule)) {
+            return false;
```
stephan-gh: Would it make sense to simply reset these game rules to the default if they're removed? (Basically behavior like if they are removed and directly added back with their default values)

src/main/java/org/spongepowered/common/mixin/core/world/MixinGameRules.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.TreeMap;
+
+@Mixin(GameRules.class)
+public abstract class MixinGameRules implements IMixinGameRules {
+
+    private static final Set<String> DEFAULT;
+
+    static {
+        final ImmutableSet.Builder<String> builder = ImmutableSet.builder();
+        for (Field field : DefaultGameRules.class.getFields()) {
```
stephan-gh: I think it would be much better to maintain a single static `GameRule` instance (which has all the default values) and check if it contains the game rule.

Basically:

```java
private static final GameRules DEFAULT_GAME_RULES = new GameRules();

// ...
if (DEFAULT_GAME_RULES.hasRule(gameRule)) {
    // ...
}
```

PR #1023, opened 2016-11-13
Fix various (extremely) minor issues with Shadows

DenWav:
Thanks to @gabizou for giving me the idea to do this...if you don't want to pull it I won't be offended. :smile: 

This was literally done entirely by my IntelliJ plugin... (https://github.com/minecraft-dev/MinecraftDev/blob/master/src/main/java/com/demonwav/mcdev/platform/mixin/util/ShadowError.java)

"fixes" some "issues" regarding Shadow access levels and in a couple cases, missing `@Final` annotations.


gabizou:
I'll get to merging this next when I have a chance.

parlough:
At this point with the amount of conflicts there are, might as well close this since it is rather minor anyway.

stephan-gh:
We should probably do this again but then merge it directly.

Issue #1030, opened 2016-11-24
Impossible to spawn ItemFrame (or probably any Hanging entity)


windy1:
There is currently no data processor for `Hanging`'s `DirectionalData` which in itself would make it impossible to spawn them since the hanging direction is initialized to `null`. I implemented a data processor for the data [here](https://github.com/windy1/SpongeCommon/commit/52762981d9c1959399cc32b58a8ae02413aa11bb) and I can verify that it's working properly however now it seems as if there's an issue with the positional data associated with the tile entity.

Consider the following example.

https://gist.github.com/windy1/b157fb9ff6ff3b13040e43798315773d

I can also verify that the spawned entity spawned *and* attached to the world, but it's location is 0. I have also used the `/summon` command with the values I'm applying programmatically and the frame is spawned without a problem.

Here's a trimmed view of the data from the example above

```
Pos=[0.0, 0.0, 0.0], 
TileY=69, TileX=98, TileZ=248,
Facing=2
```

Command I used successfully:

`/summon ItemFrame 98 69 248 {TileX:98,TileY:69,TileZ:248,Facing:2}`

These are basically the 3 key tags for spawning a hanging entity. From my testing with the `/summon` command, I believe that the `Pos` tag needs to be set correctly as well as the `TileX,Y,Z` tags. The ~~rotation~~ `ItemRotation` tag seems to be related to the rotation of the item the frame represents.

UPDATE: EntityHanging ignores position changes if the facing direction is null. Which is why it wasn't updating before.

PR #1031, opened 2016-11-24
Implement DataProcessor for Hanging entity

windy1:
Related: https://github.com/SpongePowered/SpongeCommon/issues/1030

This PR makes it possible to spawn and set the facing direction of `Hanging` entities.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/processor/data/entity/HangingDataProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/data/entity/HangingDataProcessor.java
@@ -1,0 +1,81 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/common/data/processor/data/entity/HangingDataProcessor.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    protected DirectionalData createManipulator() {
+        return new SpongeDirectionalData();
+    }
+
+    @Override
+    protected boolean set(Hanging dataHolder, Direction value) {
+        ((IMixinEntityHanging) dataHolder).setDirection(value);
```
gabizou, line 57: As I said before... this shouldn't be targeting `Hanging` directly, just have it target `IMixinEntityHanging` to avoid the unnecessary casting going on.

windy1: Whoops yeah forgot about that. Will change tomorrow.

PR #1032, opened 2016-11-25
EnderDragon phases

kashike:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1421) | [**Common**](https://github.com/SpongePowered/SpongeCommon/pull/1032)


gabizou:
@kashike I'm going to have to thoroughly go through this as I need to write handling for the CauseTracker to deal with the various phases and have direct spawning for a lot of things involved with dragon's phases. As such, this PR does have to interact with said system properly to avoid issues with tracking and capturing.

src/main/java/org/spongepowered/common/mixin/core/entity/boss/dragon/phase/MixinIPhase.java
```diff
@@ -0,0 +1,47 @@
...
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.common.util.VecHelper;
+
+import java.util.Optional;
+
+@Mixin(IPhase.class)
+public interface MixinIPhase extends EnderDragonPhase {
+
+    @Shadow Vec3d getTargetLocation();
+
```
gabizou, line 47: What about implementing `getType()`?

src/main/java/org/spongepowered/common/mixin/core/entity/boss/dragon/phase/MixinPhaseList.java
```diff
@@ -0,0 +1,34 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.mixin.core.entity.boss.dragon.phase;
+
+import net.minecraft.entity.boss.dragon.phase.PhaseList;
+import org.spongepowered.api.entity.living.complex.dragon.phase.EnderDragonPhaseType;
+import org.spongepowered.asm.mixin.Mixin;
+
+@Mixin(PhaseList.class)
+public class MixinPhaseList implements EnderDragonPhaseType {
```
gabizou, line 37: This is not implementing any of the existing methods from the type class.

gabizou (changes requested):
There is no registry module being used for the types, so as a `CatalogType`, this is failing implementation requirements.

ryantheleach:
This was stalled by an interaction with the old cause tracker, it should now be unblocked but needs to be updated for the new Phase/Cause Stack system.

ImMorpheus:
Updated.

thanks to Cyber for the missing AT

PR #1034, opened 2016-11-25
List all Moved Blocks in ChangeBlockEvent.Pre

Maxqia:


src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -223,6 +226,8 @@
     private boolean weatherThunderEnabled = true;
     private boolean weatherIceAndSnowEnabled = true;
 
+    private WorldServer this$ = (WorldServer) (Object) this;
```
gabizou: No. Get this out of my mixins.

src/main/java/org/spongepowered/common/mixin/core/block/MixinBlockPistonBase.java
```diff
@@ -75,25 +73,16 @@
...
-            if (source == null) {
-                source = ((IMixinWorldServer) worldIn).createSpongeBlockSnapshot(blockstate, blockstate, pos, 2);
-            }
-            if (!worldIn.isRemote && SpongeCommonEventFactory.handleChangeBlockEventPre((IMixinWorldServer) worldIn, targetPos, source)) {
-                ci.setReturnValue(false);
-                return;
-            }
 
+            final IBlockState currentState = worldIn.getBlockState(pos);
+            final SpongeBlockSnapshot snapshot = ((IMixinWorldServer) worldIn).createSpongeBlockSnapshot(currentState, currentState, pos, 0);
```
gabizou, line 82: The original source was intended to be pre-captured of the tile entity before the move event happens, or, creating it if it wasn't already created.

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -218,27 +218,19 @@ public static CollideEntityEvent callCollideEntityEvent(net.minecraft.world.Worl
...
         final CauseTracker causeTracker = worldIn.getCauseTracker();
-        PhaseData data = causeTracker.getCurrentPhaseData();
-        if (source == null) {
-            source = data.context.getSource(BlockSnapshot.class).orElse(null);
-            if (source == null) {
-                // safety measure
-                return false;
-            }
-        }
+        BlockSnapshot source = ((World) worldIn).createSnapshot(pos.getX(), pos.getY(), pos.getZ());
```
gabizou: The source is intended to be captured beforehand, if in any case it was captured beforehand. Creating snapshots is potentially expensive as is (for tile entities). If at all possible, retaining the previous logic is preferred. Unless @bloodmc has other opinions.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -735,6 +740,31 @@ public void onUpdateTick(Block block, net.minecraft.world.World worldIn, BlockPo
...
+            EnumFacing direction = (EnumFacing) blockstate.getValue(BlockDirectional.FACING);
+
+            BlockPistonStructureHelper movedBlocks = new BlockPistonStructureHelper(this$, pos, direction, eventId == 0);
+            List<BlockPos> blocks = new ArrayList<BlockPos>();
+            blocks.addAll(movedBlocks.getBlocksToDestroy());
+            blocks.addAll(movedBlocks.getBlocksToMove());
+
+            List<Location<org.spongepowered.api.world.World>> locations = new ArrayList<Location<org.spongepowered.api.world.World>>();
+            for (BlockPos block : blocks) {
+                locations.add(new Location<>((org.spongepowered.api.world.World) this$, block.getX(), block.getY(), block.getZ()));
```
gabizou: You're not filtering block positions that have already been added, since there can be multiple duplications of locations being added.

gabizou: Likewise, you don't need to cast to World as MixinWorld already extends World.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -827,15 +857,6 @@ public void onSaveAllChunks(boolean saveAllChunks, IProgressUpdate progressCallb
         }
     }
 
-    @Inject(method = "addBlockEvent", at = @At("HEAD"), cancellable = true)
-    public void addBlockEvent(BlockPos pos, Block blockIn, int eventID, int eventParam, CallbackInfo ci) {
-        // We fire a Pre event to make sure our captures do not get stuck in a loop.
-        // This is very common with pistons as they add block events while blocks are being notified.
-        if (SpongeCommonEventFactory.handleChangeBlockEventPre(this, pos)) {
```
gabizou: This should still be retained in the event a modded block is performing additional block changes that aren't Pistons.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
Maxqia: Well, what else should I do?

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
Maxqia: *faceplant*, I moved this event over to onAddBlockEvent to prevent event spam. (which is after the check for if it's a duplicate event)

src/main/java/org/spongepowered/common/mixin/core/block/MixinBlockPistonBase.java
Maxqia: This was just a revert of blood's commit to make sure it doesn't fire twice ...

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
Deamon5550: Shadow methods and fields that you need

Maxqia: I need an instance of the class to pass to BlockPistonStructureHelper

JBYoshi: Cast it whenever you use it. If the compile complains, first cast to `Object`, then back to whatever you need.

gabizou: If you double cast `(WorldServer) (Object) this`, mixins will strip the casts at runtime.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
Maxqia: That would be super computationally wasteful though ...

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -227,18 +227,18 @@ public static CollideEntityEvent callCollideEntityEvent(net.minecraft.world.Worl
     }
 
     public static ChangeBlockEvent.Pre callChangeBlockEventPre(IMixinWorldServer worldIn, BlockPos pos, Object source) {
+        return callChangeBlockEventPre(worldIn, pos, source, ImmutableList.of(
+                new Location<>((World) worldIn, pos.getX(), pos.getY(), pos.getZ())));
```
gabizou: Shouldn't have to cast the world.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -735,6 +738,34 @@ public void onUpdateTick(Block block, net.minecraft.world.World worldIn, BlockPo
...
+        if (blockIn instanceof BlockPistonBase) { // if it's a piston, try to simulate it
+            final IBlockState blockstate = this.getBlockState(pos);
+            EnumFacing direction = (EnumFacing) blockstate.getValue(BlockDirectional.FACING);
+
+            List<Location<org.spongepowered.api.world.World>> locations = new ArrayList<Location<org.spongepowered.api.world.World>>();
+            BlockPistonStructureHelper movedBlocks = new BlockPistonStructureHelper((WorldServer) (Object) this, pos, direction, eventId == 0);
+            if (movedBlocks.canMove()) { // calculates blocks to be moved
+                List<BlockPos> blocks = new ArrayList<BlockPos>();
+                blocks.addAll(movedBlocks.getBlocksToMove());
+                blocks.addAll(movedBlocks.getBlocksToDestroy());
```
gabizou: This needs to filter the block positions out, or provide a comment explaining that it already is filtered from the structure helper. We do not want to have duplicate positions in the event, especially since it's a simulated event, and not a final list of all blocks and snapshots changed.

gabizou (changes requested):
Just add some extra comments to avoid confusion with certain lists.

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
Maxqia: "Cannot infer type arguments for Location<>" error

Maxqia:
Currently untested, please do not merge

Dockter:
wait, makes a PR, didn't test it?

Maxqia:
No, bleeding changed since the PR, so it broke it :(
Anyways, should be fixed/working now

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -219,8 +221,10 @@ public static CollideEntityEvent callCollideEntityEvent(net.minecraft.world.Worl
         return event;
     }
 
+
```
gabizou: Unnecessary new line added.

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -281,6 +285,43 @@ public static CollideEntityEvent callCollideEntityEvent(net.minecraft.world.Worl
...
+        List<BlockPos> blocks = new ArrayList<BlockPos>();
+        blocks.add(pos);
+
+        BlockPistonStructureHelper movedBlocks = new BlockPistonStructureHelper((WorldServer) (Object) world, pos, direction, extending);
+        movedBlocks.canMove(); // calculates blocks to be moved
+
+        List<BlockPos> blocksToAdd = new ArrayList<BlockPos>();
+        blocksToAdd.addAll(movedBlocks.getBlocksToDestroy());
+        blocksToAdd.addAll(movedBlocks.getBlocksToMove());
+        for (BlockPos add : blocksToAdd) { // filter out duplicates
```
gabizou: You're doing this unnecessarily complicated filter... you can simplify this down to a simple for loop of the movedBlocks.getBlocksToMove while creating new Location objects.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -735,6 +738,18 @@ public void onUpdateTick(Block block, net.minecraft.world.World worldIn, BlockPo
 
     @Redirect(method = "addBlockEvent", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/WorldServer$ServerBlockEventList;add(Ljava/lang/Object;)Z", remap = false))
     public boolean onAddBlockEvent(WorldServer.ServerBlockEventList list, Object obj, BlockPos pos, Block blockIn, int eventId, int eventParam) {
+        // We fire a Pre event to make sure our captures do not get stuck in a loop.
+        // This is very common with pistons as they add block events while blocks are being notified.
+        if (blockIn instanceof BlockPistonBase) { // if it's a piston, try to simulate it
```
gabizou: This whole branching if else is just ugly. Simplify it down to a single if statement for lazy evaluation. 

```java
if ((blockIn instanceof BlockPistonBase && SpongeCommonEventFactory.handlePistonEvent(this, list, obj, pos, blocking, eventId, eventParam)
   || SpongeCommonEventFactory.callChangeBlockEventPre(this, pos, NamedCause.of(NamedCause.BLOCK_EVENT, this)).isCancelled()) {
  return false;
}
```

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -35,8 +35,11 @@
 import com.google.common.collect.ImmutableList;
 import com.typesafe.config.ConfigRenderOptions;
 import net.minecraft.block.Block;
+import net.minecraft.block.BlockDirectional;
 import net.minecraft.block.BlockEventData;
+import net.minecraft.block.BlockPistonBase;
 import net.minecraft.block.ITileEntityProvider;
+import net.minecraft.block.state.BlockPistonStructureHelper;
```
gabizou: Some of these imports seem to be mis-added.

gabizou (changes requested):
Clean up necessary.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
Maxqia: 0-0 woops

gabizou:
Condense your commits. No need for multiple commits.

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -281,6 +284,36 @@ public static CollideEntityEvent callCollideEntityEvent(net.minecraft.world.Worl
         return event;
     }
 
+    /**
+     * This simulates the blocks a piston moves and calls the event for saner debugging.
+     * @return if the event was cancelled
+     */
+    public static boolean handlePistonEvent(IMixinWorldServer world, WorldServer.ServerBlockEventList list, Object obj, BlockPos pos, Block blockIn, int eventId, int eventParam) {
+        boolean extending = (eventId == 0);
+        final IBlockState blockstate = ((net.minecraft.world.World)world).getBlockState(pos);
```
gabizou: Fix your spacing for casts.

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -281,6 +284,36 @@ public static CollideEntityEvent callCollideEntityEvent(net.minecraft.world.Worl
...
+        EnumFacing direction = (EnumFacing) blockstate.getValue(BlockDirectional.FACING);
+
+        List<BlockPos> blocks = new ArrayList<BlockPos>();
```
gabizou: Unnecessary list.

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -281,6 +284,36 @@ public static CollideEntityEvent callCollideEntityEvent(net.minecraft.world.Worl
...
+        blocks.add(pos);
+
+        BlockPistonStructureHelper movedBlocks = new BlockPistonStructureHelper((WorldServer) (Object) world, pos, direction, extending);
+        movedBlocks.canMove(); // calculates blocks to be moved
+        blocks.addAll(movedBlocks.getBlocksToDestroy());
```
gabizou: After removing the unnecessary list, this is not needed.

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -281,6 +284,36 @@ public static CollideEntityEvent callCollideEntityEvent(net.minecraft.world.Worl
...
+        blocks.addAll(movedBlocks.getBlocksToMove());
+
+        List<Location<org.spongepowered.api.world.World>> locations = new ArrayList<Location<org.spongepowered.api.world.World>>();
```
gabizou: And then here:
```java
for (BlockPos pos : movedBlocks.getBlocksToDestroy()) {
  locations.add(new Location<>((World) world, block.getX(), block.getY(), block.getZ());
}
for (BlockPos pos : movedBlocks.getBlocksToMove()) {
  boolean isMatched = false;
  for (Location<World> location : locations) {
    final Vector3i position = location.getBlockPosition();
    if (position.getX() == pos.getX() && position.getY() == pos.getY() && position.getZ() == pos.getZ()) {
      isMatched = true;
      break;
    }
  }
  if (!isMatched) {
    locations.add(new Location<>((World) world, pos.getX(), pos.getY(), pos.getZ());
  }
}
```

Sure, it does two iterations over, but the simple filter check on the blocks to move ensures we don't have duplicate block positions in the first place.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -735,6 +736,13 @@ public void onUpdateTick(Block block, net.minecraft.world.World worldIn, BlockPo
 
     @Redirect(method = "addBlockEvent", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/WorldServer$ServerBlockEventList;add(Ljava/lang/Object;)Z", remap = false))
     public boolean onAddBlockEvent(WorldServer.ServerBlockEventList list, Object obj, BlockPos pos, Block blockIn, int eventId, int eventParam) {
+        // We fire a Pre event to make sure our captures do not get stuck in a loop.
+        // This is very common with pistons as they add block events while blocks are being notified.
+        if ((blockIn instanceof BlockPistonBase && SpongeCommonEventFactory.handlePistonEvent(this, list, obj, pos, blockIn, eventId, eventParam))
+                || SpongeCommonEventFactory.callChangeBlockEventPre(this, pos, NamedCause.of(NamedCause.BLOCK_EVENT, this)).isCancelled()) {
+               return false;
```
gabizou: What's with your indentation here?

gabizou (changes requested):

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
Maxqia: Doesn't this do the same thing as my code? List<Location<World>>.contains() calls Location<World>.equals() which checks if the positions are equal (eg. position.getX() == pos.getX() etc...). This just makes it much more cluttered.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
Maxqia: copy pasta + eclipse auto indent didn't go well apparently...

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
gabizou: You're creating unnecessary lists and creating unnecessary objects, yes it may look a little cluttered, but it minimizes the amount of objects created.

bloodmc:
@Maxqia please update PR and fix conflicts.

Maxqia:
Yay! (that took awhile)

PR #1046, opened 2016-11-29
Spectator camera view

kashike:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1422) | [**Common**](https://github.com/SpongePowered/SpongeCommon/pull/1046)


gabizou:
And what if the target entity is vanished? What happens if you attempt to spectate a vanished entity? Does the client still need to know of the vanished entity?

kashike:
The client needs to know that the `Entity` is real, yes.

PR #1051, opened 2016-12-04
Inventory fixes

### LINKED ISSUE #1050: Inventory.offer(ItemStack) not adding item to inventory
When I offer an item to a chest/player inventory on 1.11, it doesn't add the ItemStack to the inventory.

Maxqia:
This fixes #1050

### DIFF
--- a/src/main/java/org/spongepowered/common/entity/living/human/EntityHuman.java
+++ b/src/main/java/org/spongepowered/common/entity/living/human/EntityHuman.java
@@ -507,7 +507,7 @@ public class EntityHuman extends EntityCreature implements TeamMember, IRangedAt
 
         final ItemStack itemstack = this.getHeldItem(EnumHand.OFF_HAND);
 
-        if (itemstack != null && itemstack.getItem() == Items.TIPPED_ARROW) {
+        if (!itemstack.isEmpty() && itemstack.getItem() == Items.TIPPED_ARROW) {
             entitytippedarrow.setPotionEffect(itemstack);
         }
 
--- a/src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketPhaseUtil.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketPhaseUtil.java
@@ -97,7 +97,7 @@ public final class PacketPhaseUtil {
     }
 
     public static void handlePlayerSlotRestore(EntityPlayerMP player, ItemStack itemStack, EnumHand hand) {
-        if (itemStack == null || itemStack == ItemTypeRegistryModule.NONE) {
+        if (itemStack.isEmpty() || itemStack == ItemTypeRegistryModule.NONE) {
             return;
         }
 
--- a/src/main/java/org/spongepowered/common/item/inventory/util/ItemStackUtil.java
+++ b/src/main/java/org/spongepowered/common/item/inventory/util/ItemStackUtil.java
@@ -70,7 +70,7 @@ public abstract class ItemStackUtil {
     @Nullable
     public static net.minecraft.item.ItemStack cloneDefensiveNative(@Nullable net.minecraft.item.ItemStack stack, int newSize) {
         net.minecraft.item.ItemStack clone = stack.copy();
-        if (clone != null) {
+        if (!clone.isEmpty()) {
             clone.setCount(newSize);
         }
         return clone;
@@ -104,8 +104,11 @@ public abstract class ItemStackUtil {
         return Optional.<ItemStack>ofNullable(ItemStackUtil.cloneDefensive(stack));
     }
 
+    /**
+     * This ignores item stack size on purpose
+     */
     public static boolean compare(net.minecraft.item.ItemStack stack1, net.minecraft.item.ItemStack stack2) {
-        return net.minecraft.item.ItemStack.areItemStacksEqual(stack1, stack2);
+        return stack1.isItemEqual(stack2) && net.minecraft.item.ItemStack.areItemStackTagsEqual(stack1, stack2);
     }
 
     public static boolean compare(net.minecraft.item.ItemStack stack1, ItemStack stack2) {
--- a/src/main/java/org/spongepowered/common/item/inventory/util/ItemStackUtil.java
+++ b/src/main/java/org/spongepowered/common/item/inventory/util/ItemStackUtil.java
@@ -49,14 +49,14 @@ public abstract class ItemStackUtil {
     @Nullable
     public static net.minecraft.item.ItemStack toNative(@Nullable ItemStack stack) {
         if (stack instanceof net.minecraft.item.ItemStack || stack == null) {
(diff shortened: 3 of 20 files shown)


src/main/java/org/spongepowered/common/item/inventory/util/ItemStackUtil.java
```diff
@@ -105,8 +105,8 @@ public static ItemStack cloneDefensive(@Nullable ItemStack stack, int newSize) {
     }
 
     public static boolean compare(net.minecraft.item.ItemStack stack1, net.minecraft.item.ItemStack stack2) {
-        return net.minecraft.item.ItemStack.areItemStacksEqual(stack1, stack2);
-    }
+        return stack1.isItemEqual(stack2) && net.minecraft.item.ItemStack.areItemStackTagsEqual(stack1, stack2);
+    } // This ignores stack size on purpose
```
ST-DDT: The comment should be above the code, shouldn't it?

src/main/java/org/spongepowered/common/item/inventory/util/ItemStackUtil.java
```diff
@@ -49,14 +49,14 @@ public static NBTTagCompound getTagCompound(net.minecraft.item.ItemStack itemSta
     @Nullable
     public static net.minecraft.item.ItemStack toNative(@Nullable ItemStack stack) {
         if (stack instanceof net.minecraft.item.ItemStack || stack == null) {
-            return (net.minecraft.item.ItemStack) stack;
+            return stack == null ? net.minecraft.item.ItemStack.EMPTY : (net.minecraft.item.ItemStack) stack;
         }
         throw new NativeStackException("The supplied item stack was not native to the current platform");
     }
 
     @Nullable
```
liach, line 57: `@Nonnull`?

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemBlock.java
```diff
@@ -35,7 +36,7 @@
 import java.util.Optional;
 
 @NonnullByDefault
-@Mixin(net.minecraft.item.ItemBlock.class)
+@Mixin(value = { net.minecraft.item.ItemBlock.class, net.minecraft.item.ItemAir.class })
```
Zidane, line 39: Why?

src/main/java/org/spongepowered/common/item/inventory/util/ItemStackUtil.java
Zidane: Don't change the functionality of this method and keep the name the same. If how this is used causes issues with stack size then rename this method. Your changes will lead to confusion down the line when an update ultimately has us touch this again and the comment basically says "This was changed due to reasons".

src/main/java/org/spongepowered/common/entity/living/human/EntityHuman.java
```diff
@@ -507,7 +507,7 @@ public void attackEntityWithRangedAttack(EntityLivingBase target, float p_82196_
 
         final ItemStack itemstack = this.getHeldItem(EnumHand.OFF_HAND);
 
-        if (itemstack != null && itemstack.getItem() == Items.TIPPED_ARROW) {
+        if (!itemstack.isEmpty() && itemstack.getItem() == Items.TIPPED_ARROW) {
```
Zidane, line 510: An ItemStack cannot be empty unless it's Item is ItemAir. You can remove the isEmpty altogether.

src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketPhaseUtil.java
```diff
@@ -97,7 +97,7 @@ public static void validateCapturedTransactions(int slotId, Container openContai
     }
 
     public static void handlePlayerSlotRestore(EntityPlayerMP player, ItemStack itemStack, EnumHand hand) {
-        if (itemStack == null || itemStack == ItemTypeRegistryModule.NONE) {
+        if (itemStack.isEmpty() || itemStack == ItemTypeRegistryModule.NONE) {
```
Zidane, line 100: @Faithcaio 

Since Mojang added ItemAir, I think it would make the implementation much cleaner to just outright remove our concept of a NONE stack and simply use theirs.

@bloodmc @gabizou Your comments are also appreciated.

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemBlock.java
Faithcaio: BlockTypes.AIR.getItem() no longer returns empty but an "air" item.
This changes the air itemType to return the air BlockType on getBlock() instead of empty Optional
__
nvm. I fixed the first issue in MixinBlock later...

src/main/java/org/spongepowered/common/mixin/core/block/MixinBlock.java
```diff
@@ -185,7 +186,8 @@ public BlockState getDefaultState() {
 
     @Override
     public Optional<ItemType> getItem() {
-        return Optional.ofNullable((ItemType) Item.getItemFromBlock((Block) (Object) this));
```
Zidane, line 188: Now I'm not too keen on this change. If getItem is no longer Nullable, we should change the API to return ItemType.

@Deamon5550 @gabizou

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemBlock.java
Zidane: *sigh*

ItemAir doesn't extend ItemBlock does it...

I guess we'll keep the name instead but a comment about why ItemAir is used here should be put.

src/main/java/org/spongepowered/common/mixin/core/block/MixinBlock.java
Faithcaio: The reason why I changed it this way is because I did not want to touch the API.
Previous null values are now the air ItemType (which is not present in the API)

Changing the API to allow Air ItemTypes will need a lot of changes. (InventoryAPI)

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemBlock.java
Faithcaio: exactly

src/main/java/org/spongepowered/common/item/inventory/util/ItemStackUtil.java
Maxqia: It was incorrectly changed in commit 66dddcf, this reverts that.

Faithcaio:
fyi @Maxqia I merged my commits into bleeding (not exactly the same that are in here)
and this: https://github.com/SpongePowered/SpongeCommon/tree/fix/itemstack-compare

Maxqia:
The compare change doesn't look merged yet, but it looks fine. Closing in favor of your commit

PR #1064, opened 2016-12-09
Attributes

kashike:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1428) | [**Common**](https://github.com/SpongePowered/SpongeAPI/pull/1064)


src/main/java/org/spongepowered/common/mixin/core/entity/ai/attributes/MixinAttributeModifier.java
```diff
@@ -0,0 +1,67 @@
...
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Implements;
+import org.spongepowered.asm.mixin.Interface;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.common.registry.type.entity.AttributeOperationRegistryModule;
+
+import java.util.UUID;
+
+@Implements(@Interface(iface = AttributeModifier.class, prefix = "modifier$"))
```
liach: Why this?

src/main/java/org/spongepowered/common/registry/type/entity/AttributeOperationRegistryModule.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Collection;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Optional;
+
+public class AttributeOperationRegistryModule implements CatalogRegistryModule<AttributeOperation> {
+
+    public static final AttributeOperation[] operations = new AttributeOperation[3];
+    @RegisterCatalog(AttributeOperations.class)
+    public final Map<String, AttributeOperation> operationMap = Maps.newHashMap();
```
liach: Guava suggests you to use constructor since java7.

Deamon5550: Our practice across the entire repo is to use guava, not changing now.

src/main/java/org/spongepowered/common/registry/type/entity/AttributeRegistryModule.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Optional;
+
+public class AttributeRegistryModule implements CatalogRegistryModule<AttributeType> {
+
+    private final Map<String, AttributeType> attributes = new HashMap<>();
```
liach: @Deamon5550 Then why do you use constructor here if you always uses guava?

liach: https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/Sets.java#L210-L212

src/main/java/org/spongepowered/common/mixin/core/entity/ai/attributes/MixinIAttribute.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    default Optional<AttributeType> getParent() {
+        return Optional.ofNullable((AttributeType) this.shadow$getParent());
+    }
+
+    @Intrinsic
+    @Override
+    default double getDefaultValue() {
```
gabizou: If this is intrinsic, it needs to be prefixed.

src/main/java/org/spongepowered/common/mixin/core/entity/ai/attributes/MixinIAttribute.java
```diff
@@ -0,0 +1,71 @@
...
+    @Shadow double shadow$clampValue(double value);
+
+    @Override
+    default String getId() {
+        return this.shadow$getName();
+    }
+
+    @Intrinsic
+    @Override
+    default String getName() {
```
gabizou: If this is intrinsic, needs to be prefixed.

gabizou (changes requested):
Your intrinsic methods need to have prefixes with `@Implements` annotated on the mixing class.

ryantheleach:
Kashike has some changes locally on this that need to be finished, If anyone picks this up, I'd recommend contacting them first to see what direction it was heading in.

gabizou:

PR #1071, opened 2016-12-12
Add PortionData to Double Plants.

me4502:
This adds PortionData to DoublePlants.

I'm not entirely sure if this is necessary, but already did the changes so may as well create the PR, just incase it is required.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/block/MixinBlockDoublePlant.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/block/MixinBlockDoublePlant.java
@@ -32,11 +32,15 @@ import org.spongepowered.api.data.key.Key;
 import org.spongepowered.api.data.key.Keys;
 import org.spongepowered.api.data.manipulator.ImmutableDataManipulator;
 import org.spongepowered.api.data.manipulator.immutable.block.ImmutableDoublePlantData;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/core/block/MixinBlockDoublePlant.java
```diff
@@ -45,12 +49,13 @@
 
     @Override
     public ImmutableList<ImmutableDataManipulator<?, ?>> getManipulators(IBlockState blockState) {
-        return ImmutableList.<ImmutableDataManipulator<?, ?>>of(getDoublePlantTypeFor(blockState));
+        return ImmutableList.<ImmutableDataManipulator<?, ?>>of(getDoublePlantTypeFor(blockState),
+                getPortionData(blockState));
```
kashike: This doesn't need to be on a new line.

kashike (changes requested):

src/main/java/org/spongepowered/common/mixin/core/block/MixinBlockDoublePlant.java
me4502: Fixed.

gabizou (approved):
Looks good.

me4502:
Thanks, I'll do my final testing - as I've written a fair bit of extra code to test with since I did my last test.

PR #1076, opened 2016-12-15
Implement Statistics API

windy1:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1434) | [**Common**](https://github.com/SpongePowered/SpongeCommon/pull/1076)

This PR implements the newly updated Statistics API.

Test plugin I used: https://gist.github.com/windy1/ed4bfb1d0e4d8cc8366597c305f85265


src/main/java/org/spongepowered/common/registry/type/statistic/AchievementRegistryModule.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public Map<String, Achievement> provideCatalogMap() {
+        return this.achievementMappings;
+    }
+
+    private static final class Holder {
+
+        static final AchievementRegistryModule INSTANCE = new AchievementRegistryModule();
```
liach, line 82: Probably need a non-public constructor

src/main/java/org/spongepowered/common/registry/type/statistic/StatisticRegistryModule.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public Map<String, Statistic> provideCatalogMap() {
+        return this.statisticMappings;
+    }
+
+    private static final class Holder {
+
+        static final StatisticRegistryModule INSTANCE = new StatisticRegistryModule();
```
liach, line 82: Same for constructor

src/main/java/org/spongepowered/common/registry/type/statistic/AchievementRegistryModule.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public Collection<Achievement> getAll() {
+        return ImmutableList.copyOf(this.achievementMappings.values());
+    }
+
+    @Override
+    public void registerAdditionalCatalog(Achievement achievement) {
+        this.achievementMappings.put(achievement.getId().toLowerCase(), achievement);
```
ST-DDT: This should check whether the id is already registered.
This should also define a locale for the lowercase method.

src/main/java/org/spongepowered/common/registry/type/statistic/AchievementRegistryModule.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public Optional<Achievement> getById(String id) {
+        return Optional.ofNullable(this.achievementMappings.get(checkNotNull(id).toLowerCase()));
+    }
+
+    @Override
+    public Collection<Achievement> getAll() {
+        return ImmutableList.copyOf(this.achievementMappings.values());
```
ST-DDT, line 61: This should be a `Set` to prevent duplicates in the result. And this should be cached.

src/main/java/org/spongepowered/common/registry/type/statistic/AchievementRegistryModule.java
```diff
@@ -0,0 +1,81 @@
...
+    @RegisterCatalog(Achievements.class)
+    private final Map<String, Achievement> achievementMappings = Maps.newHashMap();
+
+    public static AchievementRegistryModule getInstance() {
+        return Holder.INSTANCE;
+    }
+
+    @Override
+    public Optional<Achievement> getById(String id) {
+        return Optional.ofNullable(this.achievementMappings.get(checkNotNull(id).toLowerCase()));
```
ST-DDT: This should also define a locale for the lowercase method.

src/main/java/org/spongepowered/common/registry/type/statistic/StatisticRegistryModule.java
```diff
@@ -0,0 +1,79 @@
...
+    @RegisterCatalog(Statistics.class)
+    private final Map<String, Statistic> statisticMappings = Maps.newHashMap();
+
+    public static StatisticRegistryModule getInstance() {
+        return Holder.INSTANCE;
+    }
+
+    @Override
+    public Optional<Statistic> getById(String id) {
+        return Optional.ofNullable(this.statisticMappings.get(id.toLowerCase()));
```
ST-DDT: This should also define a locale for the lowercase method.

src/main/java/org/spongepowered/common/registry/type/statistic/StatisticRegistryModule.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public Collection<Statistic> getAll() {
+        return ImmutableList.copyOf(this.statisticMappings.values());
```
ST-DDT, line 61: This should be a `Set` to prevent duplicates in the result. And this should be cached.

src/main/java/org/spongepowered/common/registry/type/statistic/StatisticRegistryModule.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public void registerAdditionalCatalog(Statistic stat) {
+        this.statisticMappings.put(stat.getId(), stat);
```
ST-DDT: This should check whether the id is already registered and the id must be lowercased.

src/main/java/org/spongepowered/common/registry/type/statistic/StatisticRegistryModule.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public boolean allowsApiRegistration() {
+        return false;
+    }
+
+    @Override
+    public Map<String, Statistic> provideCatalogMap() {
+        return this.statisticMappings;
```
ST-DDT: This should return an immutable and cached version.

src/main/java/org/spongepowered/common/registry/type/statistic/AchievementRegistryModule.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public boolean allowsApiRegistration() {
+        return false;
+    }
+
+    @Override
+    public Map<String, Achievement> provideCatalogMap() {
+        return this.achievementMappings;
```
ST-DDT: This should return an immutable and cached version.

src/main/java/org/spongepowered/common/registry/type/statistic/StatisticTypeRegistryModule.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Optional;
+
+public final class StatisticTypeRegistryModule implements CatalogRegistryModule<StatisticType> {
+
+    @RegisterCatalog(StatisticTypes.class)
+    private final Map<String, StatisticType> universalStatisticMappings = Maps.newHashMap();
+
+    @Override
+    public Optional<StatisticType> getById(String id) {
+        return Optional.ofNullable(this.universalStatisticMappings.get(id.toLowerCase()));
```
ST-DDT: This should be to lowercase

src/main/java/org/spongepowered/common/registry/type/statistic/StatisticTypeRegistryModule.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Override
+    public Collection<StatisticType> getAll() {
+        return ImmutableList.copyOf(this.universalStatisticMappings.values());
```
ST-DDT: This should be a Set and cached.

ST-DDT (changes requested):
Here is my old PR for this: https://github.com/SpongePowered/SpongeCommon/pull/131

You can compare/copy some things from there.

src/main/java/org/spongepowered/common/registry/type/statistic/AchievementRegistryModule.java
windy1: This is how it's implemented in practically every other `RegistryModule`.

src/main/java/org/spongepowered/common/registry/type/statistic/AchievementRegistryModule.java
Deamon5550: Its a private class.

src/main/java/org/spongepowered/common/data/manipulator/mutable/entity/SpongeStatisticData.java
```diff
@@ -0,0 +1,121 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data.manipulator.mutable.entity;
+
+import com.google.common.collect.Maps;
+import org.spongepowered.api.data.DataContainer;
+import org.spongepowered.api.data.DataHolder;
+import org.spongepowered.api.data.key.Keys;
+import org.spongepowered.api.data.manipulator.immutable.entity.ImmutableStatisticData;
+import org.spongepowered.api.data.manipulator.mutable.common.AbstractMappedData;
```
gabizou: Imported the wrong `AbstractMappedData`. Use the common one.

src/main/java/org/spongepowered/common/data/manipulator/immutable/entity/ImmutableSpongeStatisticData.java
```diff
@@ -0,0 +1,58 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data.manipulator.immutable.entity;
+
+import org.spongepowered.api.data.DataContainer;
+import org.spongepowered.api.data.key.Keys;
+import org.spongepowered.api.data.manipulator.immutable.common.AbstractImmutableMappedData;
```
gabizou: Imported the wrong `AmbstractImmutableMappedData`. Use the one from common.

src/main/java/org/spongepowered/common/registry/type/statistic/AchievementRegistryModule.java
gabizou: @liach please stop commenting on this sort of access thing. There's enough reasoning that this is fine, since the `Holder` class is already private, it's not accessible by other classes. The whole point of the holder class in the first place with it's special access modifiers is to avoid creating synthetic accessor methods by the compiler, while avoiding issues with static initialization of the main class itself.

src/main/java/org/spongepowered/common/registry/type/statistic/StatisticRegistryModule.java
gabizou: No. this is fine.

src/main/java/org/spongepowered/common/registry/type/statistic/StatisticRegistryModule.java
gabizou: It cannot be cached as the lifecycle of the game doesn't permit it to be cached. Mod environments can do whatever they please, including adding new statistics.

src/main/java/org/spongepowered/common/statistic/SpongeBlockStatistic.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    public Translation getTranslation() {
+        return new SpongeTranslation(this.statId);
+    }
+
+    @Override
+    public ItemType getItemType() {
+        return Sponge.getRegistry().getType(
```
gabizou: You could bypass this and use the `ItemTypeRegistryModule` itself.

src/main/java/org/spongepowered/common/statistic/SpongeBlockStatistic.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    public void setSpongeId(String id) {
+        this.spongeId = id;
+    }
+
+    @Override
+    public String getMinecraftId() {
+        return this.statId;
+    }
+
```
gabizou, line 85: If you can, for the sake of things, add a `toString` and `equals/hashcode`.

src/main/java/org/spongepowered/common/statistic/SpongeEntityStatistic.java
```diff
@@ -0,0 +1,86 @@
...
+    private String spongeId;
+
+    public SpongeEntityStatistic(String statIdIn, ITextComponent statNameIn, String entityId) {
+        super(statIdIn, statNameIn);
+        this.entityId = entityId;
+    }
+
+    @Override
+    public EntityType getEntityType() {
+        return Sponge.getRegistry().getType(EntityType.class, this.entityId).get();
```
gabizou: You can use the `EntityTypeRegistryModule` instead here.

src/main/java/org/spongepowered/common/statistic/SpongeEntityStatistic.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    public void setSpongeId(String id) {
+        this.spongeId = id;
+    }
+
+    @Override
+    public String getMinecraftId() {
+        return this.statId;
+    }
+
```
gabizou, line 85: Overriding a `toString` and such would be nice, though not necessary.

gabizou (changes requested):
Minor Data implementation changes needed, general comments.

src/main/java/org/spongepowered/common/registry/type/statistic/AchievementRegistryModule.java
liach: God... I mean `AchievementRegistryModule` needs a non-public constructor.

Deamon5550: It has one...

src/main/java/org/spongepowered/common/statistic/SpongeBlockStatistic.java
windy1: Should we maybe just let it inherit StatBase's implementations instead?

gabizou: Perhaps might be best.

src/main/java/org/spongepowered/common/registry/type/statistic/AchievementRegistryModule.java
ST-DDT: And leads to numerous issues as explained here: https://github.com/SpongePowered/SpongeCommon/issues/739

src/main/java/org/spongepowered/common/registry/type/statistic/StatisticRegistryModule.java
ST-DDT: The cache can be reset/updated whenever a new element is added to the map.
See [this](https://github.com/SpongePowered/SpongeAPI/pull/1385/files#diff-de01d9f93ae92f897aacd14a2cd56709R60) and [this](https://github.com/SpongePowered/SpongeAPI/pull/1385/files#diff-de01d9f93ae92f897aacd14a2cd56709R105) for example.

gabizou (approved):
Looks good.

PR #1077, opened 2016-12-16
this will help fixing #1046

mcenderdragon:
THis is just to test and get the correct var that is null to fix the NPE


mcenderdragon:
https://github.com/SpongePowered/SpongeForge/issues/1046

liach:
You should build on your own and test on your own. No need for a pull really.

PR #1085, opened 2016-12-20
ClassLoader and Event Fixes

Maxqia:
The first commit is for when a plugin changes the inventory targeted during a ClickInventoryEvent. It would previously revert the slot clicked unconditionally, even if a plugin has changed it. I added a check there so that wouldn't happen.

The second commit is a fix for when a Class in a separate ClassLoader registers an event. Previously this would cause a NoClassDefFoundError due to the handler class not being in the same ClassLoader as the registered class.


Deamon5550:
Justification?

Maxqia:
Edited it to add justification, I'll make it more coherent when I can.

gabizou:
These are two separate commits that should really be two separate PR's. Do what you must, but I'm not going to consider merging this if there's two separate targets in a single PR.

PR #1086, opened 2016-12-20
Fixed 'null' item check in FireworkUtils.

me4502:
Currently it's impossible to assign a List<FireworkEffect> to a firework entity, as it needs to create an item to base the NBTTagCompound off. This change fixes the missed item null check, that prevents the item ever being set.


kashike (approved):

aromaa:
Isn't the #1081 beter?

PR #1089, opened 2016-12-21
Don't restore slot if the slot has changed

Maxqia:
Some plugins change the slot directly instead of changing the slot transaction,
this makes it so that it works, instead of not changing the slot


Zidane:
@Faithcaio @bloodmc Thoughts?

Zidane:
I'm going to close this on the grounds that it isn't by our API design to have a plugin side-step the transactions and modify the slot. I can't come up with any reason as to why you actually would. If I can be given a good reason to the contrary, I am open to reconsideration.

ryantheleach:
So is it an issue with the API design instead? Should we consider introducing some form of immutable slot or modifiable view of slot, or a way to lock the slot?

PR #1090, opened 2016-12-21
Use the classloader from the object we got to avoid NoClassDefFoundErrors

Maxqia:
When trying to register a event class from a classloader different from
Sponge's, when we try to define the class, it throws a NoClassDefFoundError
due to the class being unaccessable from the current classloader.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
@@ -32,6 +32,7 @@ import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
 import com.google.common.reflect.TypeToken;
@@ -54,6 +55,7 @@ import java.lang.reflect.Modifier;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
@@ -68,10 +70,8 @@ public class SpongeEventManager implements EventManager {
     private final Object lock = new Object();
 
     private final PluginManager pluginManager;
-    private final DefineableClassLoader classLoader = new DefineableClassLoader(getClass().getClassLoader());
-    private final AnnotatedEventListener.Factory handlerFactory = new ClassEventListenerFactory("org.spongepowered.common.event.listener",
-            new FilterFactory("org.spongepowered.common.event.filters", this.classLoader), this.classLoader);
     private final Multimap<Class<?>, RegisteredListener<?>> handlersByEvent = HashMultimap.create();
+    private final Map<ClassLoader, AnnotatedEventListener.Factory> classLoaders = Maps.newHashMap();
     private final Set<Object> registeredListeners = Sets.newHashSet();
 
     public final ListenerChecker checker = new ListenerChecker(ShouldFire.class);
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/event/SpongeEventManager.java
```diff
@@ -68,10 +70,8 @@
     private final Object lock = new Object();
 
     private final PluginManager pluginManager;
-    private final DefineableClassLoader classLoader = new DefineableClassLoader(getClass().getClassLoader());
-    private final AnnotatedEventListener.Factory handlerFactory = new ClassEventListenerFactory("org.spongepowered.common.event.listener",
-            new FilterFactory("org.spongepowered.common.event.filters", this.classLoader), this.classLoader);
     private final Multimap<Class<?>, RegisteredListener<?>> handlersByEvent = HashMultimap.create();
+    private final Map<ClassLoader, AnnotatedEventListener.Factory> classLoaders = Maps.newHashMap();
```
Deamon5550, line 74: This should be a loading cache with weak keys.

Deamon5550 (changes requested):

src/main/java/org/spongepowered/common/event/SpongeEventManager.java
Maxqia: Won't that just cause classloaders to be added like crazy?

Deamon5550: Keeps us from leaking a plugin's special classloader if they've released it to be garbage collected. Which is pretty common as when the classloader is garbage collected so is all of the classes it loaded.

Maxqia: That wouldn't change that because we'd still have a store of the object (and therefore it's classloader) in the event handler.

Deamon5550 (approved):

src/main/java/org/spongepowered/common/event/SpongeEventManager.java
```diff
@@ -239,7 +249,7 @@ private PluginContainer getPlugin(Object plugin) {
 
     @Override
     public void registerListeners(Object plugin, Object listener) {
-        registerListener(getPlugin(plugin), listener);
+        registerListeners(getPlugin(plugin), listener);
```
gabizou, line 252: How does this not cause a stack overflow?

Maxqia: Java chooses the most specific method when overloading methods

gabizou: Sure, but this is nonetheless, very very very ambiguous.

vorburger:
FYI I've done the (trivial) neccessary rebase of this PR in https://github.com/SpongePowered/SpongeCommon/compare/bleeding...vorburger:Maxqia-eventclassloader, but... it now fails the `ShouldFireTest` and `SnapshotGenerationTest` with this:

```
java.lang.IllegalArgumentException: Unknown plugin: Mock for PluginContainer, hashCode: 312890187
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:191)
	at org.spongepowered.common.event.SpongeEventManager.getPlugin(SpongeEventManager.java:296)
```

I suspect that Mockito does ClassLoader magic and because of this _fix_ it makes that fail now...

vorburger:
> now fails the ShouldFireTest and SnapshotGenerationTest with this

nah that was just because I made a mistake in the rebase (registerListener*s*), now it passes.

I cannot update this PR, so proposing as new https://github.com/SpongePowered/SpongeCommon/pull/1737/

src/main/java/org/spongepowered/common/event/SpongeEventManager.java
vorburger: @gabizou in the reincarnation over on https://github.com/SpongePowered/SpongeCommon/pull/1737/ I've undone this part, if that helps to finally get this in? That also takes care of any possible backwards compatibility concerns.

PR #562, opened 2016-03-05
Implementation of AgeableData

hsyyid:
Depends on https://github.com/SpongePowered/SpongeAPI/pull/1159

Implementation of `AgeableData`.
- [x] Registration of field getters and setters
- [x] Accurately used the appropriate `AbstractData` implementation

Implementation of `ImmutableAgeableData`
- [x] Accurately extended the appropriate `AbstractImmutableData` implementation
- [x] Accurately instantiating final instance fields with an `ImmutableValue` counter part for any value getters
- [x] If necessary, creating a new `ImmutableValue` for a value that should not be cached, or, using the existing caching utils.

Implementation of the `VALUE_PROCESSOR`(s):
- [x] Appropriately extended `AbstractSpongeValueProcessor`
- [x] Individual processors for each source type possible (one for `ItemStack`/`TileEntity`/`Entity` as necessary).

Registration:
- [x] Registered the `Key` correctly by `FIELD_NAME` in the `KeyRegistry`
- [x] Registered the `DataProcessor`s and `DataManipulatorBuilder` in `SpongeSerializationRegistry`
- [x] Registered the `ValueProcessor`s in the `SpongeSerializationRegistry`

Test Plugin code:

```java
package io.github.hsyyid.testplugin;

import com.google.inject.Inject;
import org.slf4j.Logger;
import org.spongepowered.api.Sponge;
import org.spongepowered.api.command.CommandException;
import org.spongepowered.api.command.CommandResult;
import org.spongepowered.api.command.CommandSource;
import org.spongepowered.api.command.args.CommandContext;
import org.spongepowered.api.command.args.GenericArguments;
import org.spongepowered.api.command.spec.CommandExecutor;
import org.spongepowered.api.command.spec.CommandSpec;
import org.spongepowered.api.data.key.Keys;
import org.spongepowered.api.entity.Entity;
import org.spongepowered.api.entity.EntityTypes;
import org.spongepowered.api.entity.living.player.Player;
import org.spongepowered.api.event.Listener;
import org.spongepowered.api.event.cause.Cause;
import org.spongepowered.api.event.cause.NamedCause;
import org.spongepowered.api.event.game.state.GameInitializationEvent;
import org.spongepowered.api.plugin.Plugin;
import org.spongepowered.api.text.Text;
import org.spongepowered.api.text.format.TextColors;

import java.util.Optional;

@Plugin(id = "io.github.hsyyid.testplugin", name = "TestPlugin", description = "This plugin tests stuffs.", version = "0.1")
[...]

PR #1092, opened 2016-12-23
Implementation of AgeableData

Maxqia:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1440) | **Common**
See #562 
Edited for 1.11 and for IS_ADULT to work correctly (instead of 1 tick)

### DIFF
--- a/src/main/java/org/spongepowered/common/data/manipulator/immutable/entity/ImmutableSpongeAgeableData.java
+++ b/src/main/java/org/spongepowered/common/data/manipulator/immutable/entity/ImmutableSpongeAgeableData.java
@@ -1,0 +1,90 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data.manipulator.immutable.entity;
+
+import org.spongepowered.api.data.key.Keys;
+import org.spongepowered.api.data.manipulator.immutable.entity.ImmutableAgeableData;
+import org.spongepowered.api.data.manipulator.mutable.entity.AgeableData;
+import org.spongepowered.api.data.value.immutable.ImmutableBoundedValue;
+import org.spongepowered.api.data.value.immutable.ImmutableValue;
+import org.spongepowered.common.data.manipulator.immutable.common.AbstractImmutableData;
+import org.spongepowered.common.data.manipulator.mutable.entity.SpongeAgeableData;
+import org.spongepowered.common.data.value.SpongeValueFactory;
+import org.spongepowered.common.data.value.immutable.ImmutableSpongeValue;
+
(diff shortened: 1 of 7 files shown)


me4502:
From inspection this doesn't appear like it'd support Zombies, as they don't extend EntityAgeable.

Maxqia:
That should be fixed now.

me4502 (approved):

src/main/java/org/spongepowered/common/data/processor/multi/entity/AgeableDataProcessor.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Override
+    protected boolean set(Entity entityCast, Map<Key<?>, Object> keyValues) {
+        Integer age = (Integer) keyValues.get(Keys.AGE);
+        boolean adult = (Boolean) keyValues.get(Keys.IS_ADULT);
+
+        if (entityCast instanceof EntityAgeable) {
+            EntityAgeable entity = (EntityAgeable) entityCast;
+
+            if (age != null) {
```
gabizou, line 72: In what case is the value null?

src/main/java/org/spongepowered/common/data/processor/value/entity/IsAdultValueProcessor.java
```diff
@@ -0,0 +1,86 @@
...
+        } else if (container instanceof EntityZombie) {
+            ((EntityZombie) container).setChild(!adult);
+            return true;
+        }
+        return false;
+    }
+
+    @Override
+    protected Optional<Boolean> getVal(Entity container) {
+        Boolean rtr = false;
```
gabizou, line 68: Why is this an object?

src/main/java/org/spongepowered/common/data/processor/multi/entity/AgeableDataProcessor.java
```diff
@@ -0,0 +1,123 @@
...
+    @Override
+    protected Map<Key<?>, ?> getValues(Entity entityCast) {
+        int age;
+        boolean adult;
+
+        if (entityCast instanceof EntityAgeable) {
+            EntityAgeable entity = (EntityAgeable) entityCast;
+            age = entity.getAge();
+            adult = !entity.isChild();
+        } else /*if (entityCast instanceof EntityZombie)*/ { // commented so it doesn't throw an error
```
me4502, line 101: Wait, why would this throw an error?

Maxqia: Variable isn't assigned error

me4502: entityCast isn't assigned there, but is prior?

Maxqia: For int age and boolean adult...

Deamon5550: so initialize them? or throw an error in an else that the entity is unsupported.

me4502: With the current setup, if an unsupported entity was passed through for whatever reason, it'd be cast to EntityZombie. It'd be a much better to do what Deamon has suggested.

src/main/java/org/spongepowered/common/data/manipulator/immutable/entity/ImmutableSpongeAgeableData.java
```diff
@@ -0,0 +1,90 @@
...
+import org.spongepowered.api.data.value.immutable.ImmutableBoundedValue;
+import org.spongepowered.api.data.value.immutable.ImmutableValue;
+import org.spongepowered.common.data.manipulator.immutable.common.AbstractImmutableData;
+import org.spongepowered.common.data.manipulator.mutable.entity.SpongeAgeableData;
+import org.spongepowered.common.data.value.SpongeValueFactory;
+import org.spongepowered.common.data.value.immutable.ImmutableSpongeValue;
+
+public class ImmutableSpongeAgeableData extends AbstractImmutableData<ImmutableAgeableData, AgeableData> implements ImmutableAgeableData {
+
+    private int age;
```
gabizou, line 39: Field should be final.

src/main/java/org/spongepowered/common/data/processor/multi/entity/AgeableDataProcessor.java
gabizou: Still a question to this day... it should never be null, since the retrieval method is returning an integer and not an object.

src/main/java/org/spongepowered/common/data/processor/multi/entity/AgeableDataProcessor.java
gabizou: This cannot be blindly casted. If you must cast between different entities, then perhaps you need multiple processors.

src/main/java/org/spongepowered/common/data/processor/value/entity/IsAdultValueProcessor.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    protected Value<Boolean> constructValue(Boolean actualValue) {
+        return new SpongeValue<>(Keys.IS_ADULT, true, actualValue);
+    }
+
+    @Override
+    protected boolean set(Entity container, Boolean adult) {
+        if (container instanceof EntityAgeable) {
```
gabizou, line 56: Likewise here, if you need to cast multiple times, might as well just have multiple processors.

gabizou (changes requested):
I would honestly ask for some short circuits on EntityAgeable and EntityZombie to eliminate uses of processors when able to do so.

PR #1135, opened 2017-01-15
Fix SF issues when disconnecting a player during ClientConnectionEvent#Login

dualspiral:
**SpongeCommon** | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1195) | [Original Issue](https://github.com/SpongePowered/SpongeForge/issues/885)

See SpongePowered/SpongeForge#885 for the original bug report. Whilst this PR is for SpongeCommon, it has no impact on SpongeVanilla - this is to fix an issue that is easier to fix here.

This PR does two things:

* Prevents ClientConnectionEvent#Disconnect from firing if #Login is cancelled in SpongeForge.
* Prevents disconnection messages from being broadcast to the server in SpongeForge.

I'm going to attempt to explain what I _think_ is happening, though there is every possibility that I might be wrong! However, I'm pretty sure of the solution.

Both are manifestations of how FML changes the login sequence itself - and where Sponge puts its code. As @Aaron1011 mentioned in the referenced issue, this is due to the fact that Sponge moves the ban/whitelist checks to a point where the `NetServerPlayHandler` _might_ have been created.

In SpongeVanilla, no `NetServerPlayHandler` seems to have been created (though I haven't fully looked, all I know is that one hasn't been passed to the `MixinPlayer#initializeConnectionToPlayer` method), so if a player is disconnected during login because the `ClientConnectEvent#Login` event is cancelled, then the `onDisconnect` method is not fired, no `ClientConnectionEvent#Disconnect` event is fired, and no message is returned.

In SpongeForge, FML sets up the `NetServerPlayHandler` before this and, importantly, registers this in the network pipeline. So, when a player disconnects, the `onDisconnect()` method _is_ fired, which is why the disconnect message appears. It's also why the `ClientConnectionEvent#Disconnect` event is fired, though the `Player` object is not entirely formed and calling `Player#getConnection()` returns `null`.

The fix, then, is to just check for whether `EntityPlayerMP.connection` is `null` before preceeding with Sponge's injected code - this is the only point where this is `null` and this brings SF to parity with SV, whilst allowing Minecraft/Forge to perform it's normal cleanup on the created handler, should it be required. The SpongeForge PR restores a Forge event that is prevented from being called ordinarily in this situation, more explanation is there and is not strictly required for this PR, though it is related (hence the link above).

Test code + logs from latest SF: https://gist.github.com/dualspiral/c568b2cd78162a3bbfd2673082a1bc8e

This is also a problem in 1.10.2.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
@@ -501,6 +501,12 @@ public abstract class MixinNetHandlerPlayServer implements PlayerConnection, IMi
     @Redirect(method = "onDisconnect", at = @At(value = "INVOKE",
             target = "Lnet/minecraft/server/management/PlayerList;sendMessage(Lnet/minecraft/util/text/ITextComponent;)V"))
     public void onDisconnectHandler(PlayerList this$0, ITextComponent component) {
+        // If this happens, the connection has not been fully established yet so we've kicked them during ClientConnectionEvent.Login,
+        // but FML has created this handler earlier to send their handshake. No message should be sent, no disconnection event should
+        // be fired either.
+        if (this.playerEntity.connection == null) {
+            return;
+        }
         final Player player = ((Player) this.playerEntity);
         final Text message = SpongeTexts.toText(component);
         final MessageChannel originalChannel = player.getMessageChannel();


stephan-gh:
Can you check if your changes also fix https://github.com/SpongePowered/SpongeForge/issues/527? I think it should no longer happen since you added a null check for the connection before calling the event.

dualspiral:
It does - found a random mod and just installed it serverside. It doesn't fire the `ClientConnectionEvent.Disconnect` at all in this circumstance (sample output below):

```
[18:39:11] [User Authenticator #1/INFO] [ndebug]: Auth
[18:39:11] [User Authenticator #1/INFO]: UUID of player dualspiral is dfe5c2a0-28bb-4449-ac73-a95980850077
[18:39:11] [Netty Server IO #3/INFO] [FML]: Client protocol version 2
[18:39:11] [Netty Server IO #3/INFO] [FML]: Client attempting to join with 4 mods : minecraft@1.11.2,FML@8.0.99.99,forge@13.20.0.2201,mcp@9.19
[18:39:11] [Netty Server IO #3/INFO] [FML]: Attempting connection with missing mods [spongeapi, sponge, cfm, mercurius_updater, ndebug] at CLIENT
[18:39:11] [Netty Server IO #3/INFO] [FML]: Rejecting connection CLIENT: [FMLMod:cfm{4.1.1}]
[18:39:11] [Netty Server IO #3/ERROR] [FML]: Network Disconnect: Mod rejections [FMLMod:cfm{4.1.1}]
[18:39:11] [Server thread/INFO]: dualspiral lost connection: Mod rejections [FMLMod:cfm{4.1.1}]
```

stephan-gh:
@dualspiral Awesome, thanks for testing!

kashike (approved):

kashike:
:+1:

me4502 (approved):

stephan-gh (approved):

PR #1136, opened 2017-01-17
Fixed BlockState updates not updating physics. Fixed #1054

### LINKED ISSUE #1054: Changing a block state does not trigger block notifications
Offering a powered state to a lever used to do the lever flick state noise, and update nearby redstone. This no longer occurs as of 1.11. This worked in 1.10.2.

The question is, is whether this should be considered a bug or proper behaviour. With the existence of the 'Fake Player' API that is soon to be implemented, it could make sense to use that API instead of offering the powered state here.

However, setting a powerable block to powered would generally include the intention of updating nearby redstone.

Here is gyfycat link that shows it occurring. This is using CraftBook's HiddenSwitch mechanic.
https://gfycat.com/ShabbyGlitteringKatydid

Another gfycat link, https://gfycat.com/WillingGiganticKiwi

me4502:
This fixes https://github.com/SpongePowered/SpongeCommon/issues/1054

It has been tested with CraftBook's ICs and HiddenSwitch system. 

Currently this method doesn't mirror the vanilla method; this PR fixes that.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/CauseTracker.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/CauseTracker.java
@@ -472,7 +472,7 @@ public final class CauseTracker {
         }
 
         if (flag.updateNeighbors()) { // Sponge - remove the isRemote check
-            minecraftWorld.updateObservingBlocksAt(pos, iblockstate.getBlock());
+            minecraftWorld.notifyNeighborsRespectDebug(pos, iblockstate.getBlock(), true);
 
             if (newState.hasComparatorInputOverride()) {
                 minecraftWorld.updateComparatorOutputLevel(pos, newBlock);


kashike (approved):

kashike:
This was broken in https://github.com/SpongePowered/SpongeCommon/commit/15b729d96abf10e84b8bd9427705db3b6e8b22b7#diff-e7eaa950156189fe15010abfc398db60L421:
incorrect path: `func_175722_b` (`notifyNeighborsRespectDebug`) -> `mth_000641_c` -> `func_190522_c` (`updateObservingBlocksAt`)
correct path:  `func_175722_b` (`notifyNeighborsRespectDebug`) -> `mth_000641_c` -> `func_175722_b` (`notifyNeighborsRespectDebug`)

The other occurrence ([CauseTracker#L438](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/event/tracking/CauseTracker.java#L438)), however, is correct.

From vanilla method (`World#setBlockState`):
```java
                if (!this.isRemote && (flags & 1) != 0) {
                    this.notifyNeighborsRespectDebug(pos, iblockstate.getBlock(), true);

                    if (newState.hasComparatorInputOverride()) {
                        this.updateComparatorOutputLevel(pos, block);
                    }
                } else if (!this.isRemote && (flags & 16) == 0) {
                    this.updateObservingBlocksAt(pos, block);
                }
```

PR #1138, opened 2017-01-17
Fixed Skull's not looking up Skin Data if both the UUID and Name is present. Fixes #1013

### LINKED ISSUE #1013: Setting RepresentedPlayerData on a Skull ItemStack does not set the skin.
The following code is the test case. It sets the Owner tag on the Skull, but it doesn't set the skin. The default skin is still used.

```
SkullData data = Sponge.getGame().getDataManager().getManipulatorBuilder(SkullData.class).get().create();
data.set(Keys.SKULL_TYPE, SkullTypes.PLAYER);
ItemStack stack = Sponge.getGame().getRegistry().createBuilder(ItemStack.Builder.class).itemType(ItemTypes.SKULL).itemData(data).build();
RepresentedPlayerData skinData = Sponge.getGame().getDataManager().getManipulatorBuilder(RepresentedPlayerData.class).get().create();
skinData.set(Keys.REPRESENTED_PLAYER, GameProfile.of(UUID.fromString("4c38ed11-596a-4fd4-ab1d-26f386c1cbac"), "MHF_Blaze"));
stack.offer(skinData);
```

me4502:
This fixes the issue #1013 by looking up the texture data even if both the UUID and Name are present.

Whilst I could have put an or in place in the previous lookup, the most 'proper' way is to lookup via UUID.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/processor/common/SkullUtils.java
+++ b/src/main/java/org/spongepowered/common/data/processor/common/SkullUtils.java
@@ -128,7 +128,9 @@ public class SkullUtils {
         }
         // Skulls need a name in order to properly display -> resolve if no name is contained in the given profile
         final GameProfileManager resolver = Sponge.getGame().getServer().getGameProfileManager();
-        if (!profile.getName().isPresent() || profile.getName().get().isEmpty()) {
+        if (profile.getPropertyMap().containsKey("textures")) {
+            return profile;
+        } else if (profile.getUniqueId() != null) {
             final CompletableFuture<GameProfile> future = resolver.get(profile.getUniqueId());
             try {
                 return future.get();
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/data/processor/common/SkullUtils.java
```diff
@@ -144,6 +144,14 @@ public static boolean setProfile(ItemStack skull, @Nullable GameProfile profile)
                 SpongeImpl.getLogger().debug("Exception while trying to resolve GameProfile: ", e);
                 return null;
             }
+        } else if (!profile.getPropertyMap().containsKey("textures")) {
+            final CompletableFuture<GameProfile> future = resolver.get(profile.getUniqueId());
```
kashike: Use `fill(profile.getUniqueId())`

me4502: Ah, good point. I'll fix tomorrow morning.

kashike (approved):

PR #1141, opened 2017-01-17
Fixes Array(Mutable/Immutable)BlockBuffer bugs

xcube16:
This PR fixes three bugs that effect ArrayMutableBlockBuffer and or ArrayImmutableBlockBuffer.

Bug 1:
ArrayMutableBlockBuffer#getBlock() had a broken null check. (Optional#get() would crash before it got there)

Bug 2:
ArrayMutableBlockBuffer.BackingData#get() would return negative ids. Example: ByteBackingData would only support 128 local state ids.
http://i.imgur.com/v3WkeKy.png
(To the left is the original structure containing exactly 256 different block states. To the right is what was pasted form the schematic. Ignore the "modifying it in the world" chat message)

Bug 3:
ByteBackingData can not handle more than 256 state ids. This means that when ArrayMutableBlockBuffer#setBlock() adds a new local block state id past 255 it will wrap around!
http://i.imgur.com/CDsMIwW.png
(To the left is the original structure containing exactly 256 different block states. In the middle is the pasted schematic with sponge added **after** it was pasted. To the right is the pasted schematic with sponge added **before** it was pasted, causing it to wrap around and turn to bedrock.)

To fix this bug I added PackedBackingData to replace some of the old BackingData implementations. This is sort of like how net.minecraft.world.chunk.BlockStateContainer works with some with some improvements. Example: If there are only 10 local state ids, than PackedBackingData will only use 4 bits per block.

The only other BackingData implementation I left was CharBackingData as there are a lot of places that create an ArrayMutableBlockBuffer with a char[] array.
It might be a good idea to transition away from that.

http://i.imgur.com/KtM3Kak.png
(after all fixes)

### DIFF
--- a/src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
+++ b/src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
@@ -25,6 +25,7 @@
 package org.spongepowered.common.util.gen;
 
 import com.flowpowered.math.vector.Vector3i;
+import net.minecraft.util.math.MathHelper;
 import org.spongepowered.api.block.BlockState;
 import org.spongepowered.api.block.BlockTypes;
 import org.spongepowered.api.event.cause.Cause;
@@ -39,8 +40,12 @@ import org.spongepowered.common.world.extent.MutableBlockViewDownsize;
 import org.spongepowered.common.world.extent.MutableBlockViewTransform;
 import org.spongepowered.common.world.extent.UnmodifiableBlockVolumeWrapper;
 import org.spongepowered.common.world.extent.worker.SpongeMutableBlockVolumeWorker;
+import org.spongepowered.common.world.schematic.BimapPalette;
+import org.spongepowered.common.world.schematic.GlobalPalette;
 
 public class ArrayMutableBlockBuffer extends AbstractBlockBuffer implements MutableBlockVolume {
+
+    private static final int LOCAL_PALETTE_THRESHOLD = 256;
 
     @SuppressWarnings("ConstantConditions")
     private static final BlockState AIR = BlockTypes.AIR.getDefaultState();
@@ -48,16 +53,15 @@ public class ArrayMutableBlockBuffer extends AbstractBlockBuffer implements Muta
     private final BlockPalette palette;
     private final BackingData data;
 
-    public ArrayMutableBlockBuffer(BlockPalette palette, Vector3i start, Vector3i size, BackingDataType type) {
-        super(start, size);
-        this.palette = palette;
-        this.data = type.create(size);
+    public ArrayMutableBlockBuffer(Vector3i start, Vector3i size) {
+        this(size.getX() * size.getY() * size.getZ() > LOCAL_PALETTE_THRESHOLD ?
+                new BimapPalette() : GlobalPalette.instance, start, size);
     }
 
-    public ArrayMutableBlockBuffer(BlockPalette palette, Vector3i start, Vector3i size, byte[] blocks) {
+    public ArrayMutableBlockBuffer(BlockPalette palette, Vector3i start, Vector3i size) {
         super(start, size);
         this.palette = palette;
-        this.data = new ByteBackingData(blocks);
+        this.data = new PackedBackingData(size.getX() * size.getY() * size.getZ(), palette.getHighestId());
     }
 
     public ArrayMutableBlockBuffer(BlockPalette palette, Vector3i start, Vector3i size, char[] blocks) {
@@ -66,16 +70,10 @@ public class ArrayMutableBlockBuffer extends AbstractBlockBuffer implements Muta
         this.data = new CharBackingData(blocks);
     }
 
-    public ArrayMutableBlockBuffer(BlockPalette palette, Vector3i start, Vector3i size, int[] blocks) {
-        super(start, size);
-        this.palette = palette;
-        this.data = new IntBackingData(blocks);
-    }
-
     ArrayMutableBlockBuffer(BlockPalette palette, Vector3i start, Vector3i size, BackingData blocks) {
         super(start, size);
         this.palette = palette;
-        this.data = blocks.copyOf();
+        this.data = blocks;
     }
 
     @Override
@@ -122,7 +120,7 @@ public class ArrayMutableBlockBuffer extends AbstractBlockBuffer implements Muta
     public MutableBlockVolume getBlockCopy(StorageType type) {
         switch (type) {
             case STANDARD:
-                return new ArrayMutableBlockBuffer(this.palette, this.start, this.size, this.data);
+                return new ArrayMutableBlockBuffer(this.palette, this.start, this.size, this.data.copyOf());
             case THREAD_SAFE:
             default:
                 throw new UnsupportedOperationException(type.name());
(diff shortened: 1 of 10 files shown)


src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
```diff
@@ -87,14 +85,13 @@ public BlockPalette getPalette() {
     public boolean setBlock(int x, int y, int z, BlockState block, Cause cause) {
         checkRange(x, y, z);
         this.data.set(getIndex(x, y, z), this.palette.getOrAssign(block));
-        return true;
+        return true; // -254 111 250
```
simon816: I'm guessing this is a left over comment

xcube16: Oops. Ya, I was debugging and left that there.

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
```diff
@@ -66,16 +70,10 @@ public ArrayMutableBlockBuffer(BlockPalette palette, Vector3i start, Vector3i si
...
-    public ArrayMutableBlockBuffer(BlockPalette palette, Vector3i start, Vector3i size, int[] blocks) {
-        super(start, size);
-        this.palette = palette;
-        this.data = new IntBackingData(blocks);
-    }
-
     ArrayMutableBlockBuffer(BlockPalette palette, Vector3i start, Vector3i size, BackingData blocks) {
         super(start, size);
         this.palette = palette;
-        this.data = blocks.copyOf();
```
Deamon5550, line 78: If you're going to change this then make it consistent with the immutable version (ie. remove the ctor in the immutable buffer that clones the data).

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
```diff
@@ -39,25 +40,28 @@
...
 
     private final BlockPalette palette;
     private final BackingData data;
 
-    public ArrayMutableBlockBuffer(BlockPalette palette, Vector3i start, Vector3i size, BackingDataType type) {
-        super(start, size);
-        this.palette = palette;
-        this.data = type.create(size);
+    public ArrayMutableBlockBuffer(Vector3i start, Vector3i size) {
+        this(size.getX() * size.getY() * size.getZ() > LOCAL_PALETTE_THRESHOLD ?
```
Deamon5550: This check seems backwards but what is your thinking?

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
```diff
@@ -39,25 +40,28 @@
...
+                new BimapPalette() : GlobalPalette.instance, start, size);
```
Deamon5550, line 63: If you're going to create a new BimapPalette then you need to make sure to add air to it as id 0. (I think this is the real cause of your first bug).

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
```diff
@@ -39,25 +40,28 @@
...
     }
 
-    public ArrayMutableBlockBuffer(BlockPalette palette, Vector3i start, Vector3i size, byte[] blocks) {
+    public ArrayMutableBlockBuffer(BlockPalette palette, Vector3i start, Vector3i size) {
         super(start, size);
         this.palette = palette;
```
Deamon5550, line 68: Likewise here you need to make sure that that we're setting the contents of the backing data to the id for air in the incoming palette (which may not be zero).

Deamon5550 (changes requested):
some changes needed

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
xcube16: If there is a small number of blocks, say 20 blocks for example, using the global state ids is a good idea as it does not need to allocate its own local palette.
You get more savings from using a local palette if there is a larger area as it makes up for the cost of having its own BimapPalette.

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
xcube16: I know its not the cause of bug 1. But you make a good point, I did not think about the fact that the buffer defaults to being filled with air blocks.
(In the images of my test plugin, setBlock() was called on all block postions, so there where no *undefined* blocks)

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
Deamon5550: sure until the number of bits exceeds a threshold and the size of the palette starts being too much overhead compared to just using chars. MC only bit packs the chunks if the number of bits is 8 or less for example.

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
xcube16: Already did that.

Deamon5550: I'm talking about [this one](https://github.com/xcube16/SpongeCommon/blob/7621cc4a8028c385830ecfcd8a10350434c41e10/src/main/java/org/spongepowered/common/util/gen/ArrayImmutableBlockBuffer.java#L53), I'm not sure if its even used but any usages of it should be changed for consistency.

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
xcube16: net.minecraft.world.chunk.BlockStateContainer#setBits() does have some special cases for optimizing its palette. But it still uses a BitArray (bit packing) for storage.

Should I make it so that ArrayMutableBlockBuffer can transition to a global palette when the number of different states gets heigh enough?

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
xcube16: O that one... There is only one usage, and thats [here](https://github.com/xcube16/SpongeCommon/blob/7621cc4a8028c385830ecfcd8a10350434c41e10/src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java#L132).
I like to think about it like new String(char[]), string being immutable, makes a copy of the char[] array so it cant be mutated form the outside.
However its ok if ArrayMutableBlockBuffer is changed from the outside using the char[].

This does get me thinking though... ArrayImmutableBlockBuffer() takes a palette argument. What if BlockPalette#remove() was called after ArrayImmutableBlockBuffer() was created. (Maybe this can be fixed in another PR)

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
Deamon5550: yeah that would be good I think

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
Deamon5550: mm yeah we might need an immutable palette...

xcube16: Or we could remove the BlockPalette's remove() method... I don't like that method anyway, supporting it is why BimapPalette has a BitSet. (nothing in SpongeForge or any of its subprojects use it as far as I know, so its only there for the plugins)

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
xcube16: Fixed. (this should also fix ArrayMutableBlockBuffer(start, size) as it uses ArrayMutableBlockBuffer(palette, start, size))

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
Deamon5550: Yes  that is an option and it certainly would simplify things. It also presents problems even for mutable volumes.

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
xcube16: This one is harder than I expected... Any ideas? I dont want it to be as messy as BlockStateContainer.

Deamon5550: should be as simple as iterating over the volume and translating the ids from the local palette to the global palette eh? Its more or less the same operation as your `remax` except with an added translation step.

xcube16: Right now, remax() does not need to deal with the palettes and stuff... It could use a callback and see if it should translate the ids also... But there is also a set() in flight that will also need to be translated at that point.
The way Mojang handles this is by putting a callback in the palette. (idk if I like that)

Another way to think of the problem... The local palette has gotten so full that it is almost as big as the global palette making it no longer worth having a local palette. This problem is not specific to the block buffer, but can happen to any local palette. :/
There has to be a better way to design all this, so its clean to implement.

xcube16: I think I found a solution.

src/main/java/org/spongepowered/common/util/gen/ArrayMutableBlockBuffer.java
```diff
@@ -135,140 +132,167 @@ public ImmutableBlockVolume getImmutableBlockCopy() {
...
-
-        @Override
-        public Object getBacking() {
-            return this.data;
-        }
-
         @Override
         public int get(int index) {
-            return this.data[index];
+            return this.data[index] & 0xFFFF;
```
xcube16: I think chars are unsigned, so I would not need the & 0xFFFF?

Deamon5550: yeah its not needed.

Deamon5550:
568f6ea4cddd55850f391e5228792bebad65960f

Deamon5550:
I ended up not merging your unit tests due to the amount of hacking around the other elements of gamestate, we'd like to do more unit tests but they will be done in a more controlled stable fashion.

PR #1147, opened 2017-01-20
Fixed EntityActivationRange ArmorStand cast crash

Hiroku:
EntityLivingBase does not always cast to EntityLiving, as is the case with ArmorStands to name one.
Crash that this fixes: http://pastebin.com/L2AYnA3G


kashike:
Thanks.

Issue #992, opened 2016-10-22
Chunk Generator in Sponge does not seem to read from the current regions?


StoneLegion:
I'm quite new to this stuff but I used to plugins so far nucleus and one I like better with more settings projectborders.

The issue here is if you stop or sadly crash with any of these it does not seem to read from the world/regions so it has to generate from scratch. Of course it does not chunk generate but it has to open 1000's of chunks all over again before it gets to new chunks/regions to continue to generate. 

Because of memory leaks, bugs, crashes, etc this means for someone creating a massive world it might actually be impossible to generate an entire world with this current method.

gabizou:
@KaneHart The problem would only arise in any case if the region is not _saved_ before the crash. If the server hasn't had a chance to save to disk the already generated chunks, then none of it will persist between a crash and restart.

inklit:
I tested this on 1.8.9 and can say that it is definitely going through every chunk from the beginning and reading them before actually generating anything.

When you stop the server and start the generator again, it will start from the beginning even though it should start where the last region leaves off instead of the beginning.

StoneLegion:
1.10.2 Here and I know for a fact it's going through from the start again as well. I'm actually watching it happen live. I have region files up into the 50's and I'm in the 20's right now slowly going through them. My file size of course also has not changed for over 20%.

If I stop it will do it all over again and again sadly.

StoneLegion:
Just crashed here is a prime example:

I have 130 regions and starting again with any of these mods I sadly start all the way from the start as you can see from this (last modified)

https://i.imgur.com/kb2tZIj.png

bloodmc:
@DDoS Please look into this issue.

DDoS:
Just as an update, I'm really busy right now finishing my university semester, but I should have time to work on this later this month. I'm considering one of two solutions:

- Find a way to check which chunks are generated by doing a few as possible file system operations
- Add functionality to let the plugin know which chunks have been generated, and tell the pre-generator which chunks should be skipped. Tracking and skipping of already generated chunks would then be the responsibility of the plugin

I'm leaning towards the first, although it might be quite complicated to implement.

Deamon5550:
You should only have to open up each region file and read the first 8KB from the start of them which contains the information on what chunks are present in the the region file.

PR #1149, opened 2017-01-21
Overhaul chunk pre-generation

dualspiral:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1467) | **SpongeCommon**

This PR is designed to open up the World PreGeneration code to allow plugins to effectively monitor and manipulate the pregeneration. It is also designed to try to make performance gains by skipping over chunks that have already been generated. This is based off my experience with NucleusMixins, and should allow all plugins to control the pre generation more effectively - and I'm offering this after being asked by @bloodmc to do so in https://github.com/NucleusPowered/Nucleus/issues/528

Details of the API are discussed on the API PR - most of the changes here support the API and so won't be repeated here.

**What has NOT been done**

The basic routine that generates the chunks has not changed (apart from skipping some chunks), nor has how Sponge loads chunks. The algorithm for the actual generation is exactly the same.

The system does _not_ save the chunks at any point either. This is considered to be something that the calling plugin should offer at the moment, though there probably is a valid argument to add this to the API.

The system does not offer a way to save the state of the generation, nor a way to resume it. 

**Movement of pre-gen code into its own package**

As I added a second class that only required friend access, I opted to move the code into its own package. This can be moved back if you prefer.

**Skipping chunks**

One of the biggest frustrations is when a pregeneration has stopped, it has to start from the beginning. Sponge's routine, in the past, just loaded every chunk, without caring whether or not it already exists. This wastes a lot of time if you wanted to resume a generation - when I wrote NucleusMixins, I found that checking for whether a chunk has already been generated improved performance in the early stages massively.

<s>A new internal method has been added to `MixinChunkProviderServer`, `doesChunkExistSync`.</s> Originally, I was intending to use the `doesChunkExist` method, but because I'm always needing this information synchronously, there was a lot of overhead to creating an async task, only to block on it. Thus, a custom method in the pre-generator exists - and I found through testing on my PC that I was able to generate about 21 chunks per tick. If I had used the async method, I was lucky to get 4.

The pre-gen routine loads chunks in groups of four in order to coax Minecraft into populating the chunks with structures. This PR only skips generating a chunk if _all four_ target chunks have been generated, otherwise, all four will be loaded.

**Test code & log results (same as API)**

I will try to add more later - for now, this tests the listeners work as expected - both single task listeners, and global listeners.

https://gist.github.com/dualspiral/2664950914056f5a3ce005fba78811c8


src/main/java/org/spongepowered/common/world/pregen/SpongeChunkPreGenerateTask.java
```diff
@@ -0,0 +1,470 @@
...
+        this.plugin = plugin;
+        this.world = world;
+
+        // In order to be able to check whether a chunk exists, we could use standard Sponge API methods. However,
+        // because they set up an async method which we need to get sync anyway, we just bypass it.
+        // This results in a extremely noticeable speed improvement.
+        //
+        // This also allows us to catch non Anvil file formats too.
+        if (world.getWorldStorage() instanceof IMixinAnvilChunkLoader) {
+            this.doesChunkExistCheck = v -> ((IMixinAnvilChunkLoader)world.getWorldStorage())
```
DDoS: Formatting for the cast operation.

src/main/java/org/spongepowered/common/world/pregen/SpongeChunkPreGenerateTask.java
```diff
@@ -0,0 +1,470 @@
...
+            this.currentPosition = currentPosition.get();
+        } else {
+            throw new IllegalArgumentException("Center is not a valid chunk coordinate");
+        }
+        this.currentGenCount = 4;
+        this.currentLayer = 0;
+        this.currentIndex = 0;
+        this.nextJump = 0;
+
+        this.totalChunksToGenerate = (int)Math.pow(this.chunkRadius * 2 + 1, 2);
```
DDoS: More cast operator formatting.

src/main/java/org/spongepowered/common/world/pregen/SpongeChunkPreGenerateTask.java
```diff
@@ -0,0 +1,470 @@
...
+                    this.world
+            ));
+
+            this.isCancelled = true;
+            unregisterListener();
+            task.cancel();
+        }
+    }
+
+    private boolean areAllChunksLoaded(Vector3i... chunksToCheck) {
```
DDoS: Is this only used at line 247? Because if so, it should just take 4 chunks as an argument and bypass all the object creation from using varargs and streams. It complicates things for no good reason.

src/main/java/org/spongepowered/common/mixin/core/world/gen/MixinChunkProviderServer.java
```diff
@@ -116,6 +116,15 @@ public ChunkDataStream getGeneratedChunks() {
         return WorldStorageUtil.doesChunkExist(this.world, this.chunkLoader, chunkCoords);
     }
 
+    // This method exists to allow us to speed up world pre generation.
+    @Override
+    public boolean doesChunkExistSync(Vector3i location) {
```
DDoS: Can't find where this is used.

DDoS (changes requested):

DDoS:
This [issue](https://github.com/SpongePowered/SpongeCommon/issues/992) should be closed when this is merged.

src/main/java/org/spongepowered/common/mixin/core/world/gen/MixinChunkProviderServer.java
dualspiral: <s>[It's used here, as part of a `Predicate<Vector3i>`](https://github.com/SpongePowered/SpongeCommon/pull/1149/files#diff-7db96f1d7969f596d94f9dfd4f2a99f2R118) - it's then called as part of the [`areAllChunksLoaded` routine](https://github.com/SpongePowered/SpongeCommon/pull/1149/files#diff-7db96f1d7969f596d94f9dfd4f2a99f2R297). I've done it like this to avoid repeated `instanceof` checks when we only really need one.

(You'll need to load the diff for SpongeChunkPreGenerateTask before these links work)</s>

DDoS: It isn't used by the code you linked, it calls `chunkExists`, not `doesChunkExistSync `. It actually looks like the implementation of this method is copied into the code you linked.

dualspiral: Wait, no, looking at it again, you're right, I don't use it - I'm pointing you in completely the wrong way and I'm not reading it. I'll remove it - it was there before I went for the approach I have now.

dualspiral: Yep, sorry. Removed them now.

DDoS (approved):

liach:
@bloodmc You should merge this one also.

liach (approved):

dualspiral:
@liach I'm in direct contact with @bloodmc about this and I'm updating it for the latest bleeding for quick merge.

dualspiral:
Updated to include Sponge API submodule bump.

PR #1151, opened 2017-01-23
Add 1.9+ related features into implementation

gabizou:
1.9 and up added a lot of new features that Data API was designed to be able to take advantage of. The issue is that the implementation has not been brought up to speed with all of the new features (in this case, banner data being able to be retrieved from a shield item stack). This is more of a WIP and reminder to myself to work on this and bring it into bleeding and stable-5.


src/main/java/org/spongepowered/common/data/processor/multi/item/ShieldBannerDataProcessor.java
```diff
@@ -0,0 +1,95 @@
+package org.spongepowered.common.data.processor.multi.item;
```
kashike, line 1: Missing licence header.

me4502:
One 1.9 feature that appears to be missing is the types of boats.

PR #1157, opened 2017-01-25
Implementing DestructEntityEvent for non-living entities.

me4502:
This PR implements DestructEntityEvent for non-living entities.

The following code has been used to test it,

```
@Listener
public void onBlockForm(DestructEntityEvent event) {
    System.out.println(event.getCause());
}
```

### DIFF
--- a/src/main/java/org/spongepowered/common/interfaces/entity/IMixinEntity.java
+++ b/src/main/java/org/spongepowered/common/interfaces/entity/IMixinEntity.java
@@ -192,4 +192,6 @@ public interface IMixinEntity extends org.spongepowered.api.entity.Entity, IMixi
     default void onCancelledBlockChange(EntityTickContext phaseContext) {
 
     }
+
+    void onEntityRemoved();
 }
--- a/src/main/java/org/spongepowered/common/interfaces/entity/IMixinEntity.java
+++ b/src/main/java/org/spongepowered/common/interfaces/entity/IMixinEntity.java
@@ -193,5 +193,10 @@ public interface IMixinEntity extends org.spongepowered.api.entity.Entity, IMixi
 
     }
 
+    /**
+     * Called when the entity is being removed.
+     * Usually because it was marked as dead the previous tick.
+     *
+     */
     void onEntityRemoved();
 }
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
@@ -1473,9 +1473,8 @@ public abstract class MixinEntity implements org.spongepowered.api.entity.Entity
             MessageChannel originalChannel = MessageChannel.TO_NONE;
             try (final CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
                 Object source = PhaseTracker.getInstance().getCurrentContext().getSource();
-                if (source != null) {
-                    frame.pushCause(source);
-                }
+                frame.pushCause(source);
+
                 DestructEntityEvent event = SpongeEventFactory.createDestructEntityEvent(
                         frame.getCurrentCause(), originalChannel, Optional.of(originalChannel),
                         new MessageEvent.MessageFormatter(), this, true
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
@@ -1469,7 +1469,7 @@ public abstract class MixinEntity implements org.spongepowered.api.entity.Entity
 
     @Inject(method = "setDead", at = @At(value = "RETURN"))
     private void onSetDead(CallbackInfo ci) {
-        if (ShouldFire.DESTRUCT_ENTITY_EVENT && !this.world.isRemote && !(EntityUtil.toNative(this) instanceof EntityLiving)) {
+        if (ShouldFire.DESTRUCT_ENTITY_EVENT && !((IMixinWorld) this.world).isFake() && !(EntityUtil.toNative(this) instanceof EntityLiving)) {
             MessageChannel originalChannel = MessageChannel.TO_NONE;
             try (final CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
 
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
@@ -34,6 +34,7 @@ import net.minecraft.block.Block;
 import net.minecraft.block.material.Material;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.entity.EntityList;
+import net.minecraft.entity.EntityLiving;
 import net.minecraft.entity.EntityLivingBase;
 import net.minecraft.entity.EntityTracker;
 import net.minecraft.entity.EntityTrackerEntry;
@@ -85,10 +86,13 @@ import org.spongepowered.api.event.cause.EventContextKeys;
 import org.spongepowered.api.event.cause.entity.dismount.DismountType;
 import org.spongepowered.api.event.cause.entity.dismount.DismountTypes;
 import org.spongepowered.api.event.cause.entity.teleport.TeleportTypes;
+import org.spongepowered.api.event.entity.DestructEntityEvent;
 import org.spongepowered.api.event.entity.IgniteEntityEvent;
 import org.spongepowered.api.event.entity.MoveEntityEvent;
+import org.spongepowered.api.event.message.MessageEvent;
 import org.spongepowered.api.service.user.UserStorageService;
 import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.channel.MessageChannel;
 import org.spongepowered.api.text.translation.Translation;
 import org.spongepowered.api.util.AABB;
 import org.spongepowered.api.util.Direction;
@@ -124,6 +128,7 @@ import org.spongepowered.common.event.ShouldFire;
 import org.spongepowered.common.event.SpongeCommonEventFactory;
 import org.spongepowered.common.event.damage.DamageEventHandler;
 import org.spongepowered.common.event.damage.MinecraftBlockDamageSource;
+import org.spongepowered.common.event.tracking.PhaseTracker;
 import org.spongepowered.common.event.tracking.phase.plugin.BasicPluginContext;
 import org.spongepowered.common.event.tracking.phase.plugin.PluginPhase;
 import org.spongepowered.common.interfaces.IMixinChunk;
@@ -185,6 +190,7 @@ public abstract class MixinEntity implements org.spongepowered.api.entity.Entity
     private boolean allowsBlockEventCreation = true;
     private boolean allowsEntityEventCreation = true;
     private boolean trackedInWorld = false;
+    @Nullable private DestructEntityEvent destructEvent;
 
     @Shadow public net.minecraft.entity.Entity ridingEntity;
     @Shadow @Final private List<net.minecraft.entity.Entity> riddenByEntities;
@@ -1461,4 +1467,31 @@ public abstract class MixinEntity implements org.spongepowered.api.entity.Entity
         }
     }
 
+    @Inject(method = "setDead", at = @At(value = "RETURN"))
+    private void onSetDead(CallbackInfo ci) {
+        if (ShouldFire.DESTRUCT_ENTITY_EVENT && !this.world.isRemote && !(EntityUtil.toNative(this) instanceof EntityLiving)) {
+            MessageChannel originalChannel = MessageChannel.TO_NONE;
+            try (final CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
+                frame.pushCause(this);
+                Object source = PhaseTracker.getInstance().getCurrentContext().getSource();
+                if (source != null) {
+                    frame.pushCause(source);
+                }
+                DestructEntityEvent event = SpongeEventFactory.createDestructEntityEvent(
+                        frame.getCurrentCause(), originalChannel, Optional.of(originalChannel),
+                        new MessageEvent.MessageFormatter(), this, true
+                );
+
+                this.destructEvent = event;
+            }
+        }
+    }
+
+    @Override
+    public void onEntityRemoved() {
+        if (this.destructEvent != null) {
+            SpongeImpl.postEvent(this.destructEvent);
+        }
+        this.destructEvent = null;
+    }
 }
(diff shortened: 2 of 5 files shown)


src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -462,6 +463,31 @@ public static InteractItemEvent callInteractItemEventSecondary(EntityPlayer play
         return event;
     }
 
+    public static boolean callDestructEntityEvent(Entity entity, Cause cause) {
+        if (((net.minecraft.entity.Entity) entity).isDead && (!(entity instanceof EntityLivingBase) || entity instanceof EntityArmorStand)) {
+            MessageChannel originalChannel = MessageChannel.TO_NONE;
+            Optional<User> sourceCreator = Optional.empty();
```
liach: ?

me4502: That's meant to be the creator of the entity - it's how it used to be done. The CauseTracker part of it is needed for it to be complete.

liach: Then you should add something like `//TODO waiting on cause tracker`

gabizou: @liach as the PR states, it's a WIP. don't need to comment to ingrain that every bit is a todo comment. It's useless PR noise unless warranted.

me4502:
From my testing this now works. 

It puts the Source, Source Thrower if available, Used Item, and Owner into the Cause.

liach (approved):

liach:
You need to fix conflicts in your pulls.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -308,6 +310,33 @@ public void onSetSize(float width, float height, CallbackInfo ci) {
...
+            Cause.Builder causeBuilder = Cause.builder();
+            Optional<Object> sourceOptional = phaseData.context.getSource(Object.class);
+            sourceOptional.ifPresent(source -> {
+                causeBuilder.named(NamedCause.source(source));
+                if (source instanceof Projectile) {
+                    ProjectileSource projectileSource = ((Projectile) source).getShooter();
+                    causeBuilder.named(NamedCause.of(NamedCause.THROWER, projectileSource));
+                }
+            });
+            Optional<ItemStack> usedItemOptional = phaseData.context.firstNamed("ItemUsed", ItemStack.class);
```
gabizou: Use the constants from `InternalNamedCauses`, avoids issues from refactoring changes.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -308,6 +310,33 @@ public void onSetSize(float width, float height, CallbackInfo ci) {
...
+            usedItemOptional.ifPresent(item -> causeBuilder.named(NamedCause.of("ItemUsed", item)));
```
gabizou: Make this part multi-lined to closer identify exceptions of any kind in the future.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -308,6 +310,33 @@ public void onSetSize(float width, float height, CallbackInfo ci) {
...
+
+            phaseData.context.getOwner().ifPresent(owner -> causeBuilder.named(NamedCause.owner(owner)));
```
gabizou: And here.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -689,6 +718,16 @@ public boolean isRemoved() {
     }
 
     @Override
+    public Cause getDestructCause() {
```
gabizou: It's never `null`?

gabizou (changes requested):
Minor changes

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
me4502: Which part of it do you want multiline? Nothing there should really ever throw an exception.

src/main/java/org/spongepowered/common/interfaces/entity/IMixinEntity.java
```diff
@@ -121,6 +121,8 @@ default void supplyVanillaManipulators(List<DataManipulator<?, ?>> manipulators)
 
     boolean isVanilla();
 
+    Cause getDestructCause();
```
liach: Mark `@Nullable`?

me4502: Done

liach (approved):

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -519,6 +521,13 @@ private Entity createEntity(EntityType type, Vector3d position, boolean naturall
         return Optional.empty();
     }
 
+    @Inject(method = "onEntityRemoved", at = @At(value = "HEAD"))
+    public void onEntityRemoval(net.minecraft.entity.Entity entityIn, CallbackInfo ci) {
+        if (!this.isRemote && (!(entityIn instanceof EntityLivingBase) || entityIn instanceof EntityArmorStand)) {
```
stephan-gh: The `instanceof` checks are also in `callDestructEntityEvent`, you don't need them in both methods.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -519,6 +521,13 @@ private Entity createEntity(EntityType type, Vector3d position, boolean naturall
         return Optional.empty();
     }
 
+    @Inject(method = "onEntityRemoved", at = @At(value = "HEAD"))
+    public void onEntityRemoval(net.minecraft.entity.Entity entityIn, CallbackInfo ci) {
```
stephan-gh: No need to have this public

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -343,6 +350,33 @@ public void onSetSize(float width, float height, CallbackInfo ci) {
         }
     }
 
+    @Inject(method = "setDead", at = @At("HEAD"))
+    public void onSetDead(CallbackInfo ci) {
```
stephan-gh: No need to have this public

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -473,6 +474,25 @@ public static InteractItemEvent callInteractItemEventSecondary(EntityPlayer play
...
+                    cause, originalChannel, Optional.of(originalChannel), new MessageEvent.MessageFormatter(),
+                    entity, true
+            );
+            SpongeImpl.getGame().getEventManager().post(event);
+            if (!event.isMessageCancelled()) {
+                event.getChannel().ifPresent(channel -> channel.send(entity, event.getMessage()));
+            }
+        }
+
+        return true;
```
stephan-gh: It doesn't look like the return value has any purpose, it's unused in the code that calls this method

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -519,6 +521,13 @@ private Entity createEntity(EntityType type, Vector3d position, boolean naturall
...
+        if (!this.isRemote && (!(entityIn instanceof EntityLivingBase) || entityIn instanceof EntityArmorStand)) {
+            SpongeCommonEventFactory.callDestructEntityEvent((Entity) entityIn, ((IMixinEntity) entityIn).getDestructCause());
```
stephan-gh: Is the destruct cause still needed after calling the event? If not, you should reset it (set it to null)

parlough:
@me4502 Do you think you would have a chance to update this for the conflicts soon or nah?

me4502:
Maybe this weekend

me4502:
I've partially rebased this. I added a TODO where I need a bit of help for updating to the new cause system.

ImMorpheus:
I still don't know how to get the used items, but I think this is a start.
```

            PhaseData data = PhaseTracker.getInstance().getCurrentPhaseData();

            data.context.getSource(Object.class).ifPresent(source -> {
                frame.pushCause(source);
                if (source instanceof Projectile) {
                    frame.addContext(EventContextKeys.THROWER, ((Projectile) source).getShooter());
                }
            });

            data.context.getOwner().ifPresent(owner ->
                    frame.addContext(EventContextKeys.OWNER, owner));

            this.destructCause = Sponge.getCauseStackManager().getCurrentCause();
        }
```

Zidane:
@Aaron1011 

Get this updated and merged please.

ImMorpheus:
Asked @Aaron1011 if I could take this. Not sure if I had to open a new PR.

@Zidane The old PR was storing the cause in a field. I'm storing the event. Everything else should be pretty much the same.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -1461,4 +1467,31 @@ private boolean onSpongeLightningDamage(net.minecraft.entity.Entity entity, Dama
         }
     }
 
+    @Inject(method = "setDead", at = @At(value = "RETURN"))
+    private void onSetDead(CallbackInfo ci) {
+        if (ShouldFire.DESTRUCT_ENTITY_EVENT && !this.world.isRemote && !(EntityUtil.toNative(this) instanceof EntityLiving)) {
+            MessageChannel originalChannel = MessageChannel.TO_NONE;
+            try (final CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
+                frame.pushCause(this);
```
Deamon5550: Pushing `this` as the cause of death seems useless. The event already has the entity obviously, and the entity itself in is no way 'causing' its own death here.

me4502: True

me4502: Wait it’s been modified since mine, okay

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -1461,4 +1467,30 @@ private boolean onSpongeLightningDamage(net.minecraft.entity.Entity entity, Dama
         }
     }
 
+    @Inject(method = "setDead", at = @At(value = "RETURN"))
+    private void onSetDead(CallbackInfo ci) {
+        if (ShouldFire.DESTRUCT_ENTITY_EVENT && !this.world.isRemote && !(EntityUtil.toNative(this) instanceof EntityLiving)) {
+            MessageChannel originalChannel = MessageChannel.TO_NONE;
+            try (final CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
+                Object source = PhaseTracker.getInstance().getCurrentContext().getSource();
+                if (source != null) {
```
gabizou: Source is never going to be null.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -1461,4 +1467,30 @@ private boolean onSpongeLightningDamage(net.minecraft.entity.Entity entity, Dama
...
+                        frame.getCurrentCause(), originalChannel, Optional.of(originalChannel),
+                        new MessageEvent.MessageFormatter(), this, true
+                );
+
+                this.destructEvent = event;
+            }
+        }
+    }
+
+    @Override
```
gabizou, line 1486: What is this method being called from?

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -185,6 +190,7 @@
     private boolean allowsBlockEventCreation = true;
     private boolean allowsEntityEventCreation = true;
     private boolean trackedInWorld = false;
+    @Nullable private DestructEntityEvent destructEvent;
```
gabizou, line 193: I'm hesitant to storing events entirely. I'd rather store in phase contexts where we are less likely to leak references in multiple places.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -1461,4 +1467,29 @@ private boolean onSpongeLightningDamage(net.minecraft.entity.Entity entity, Dama
         }
     }
 
+    @Inject(method = "setDead", at = @At(value = "RETURN"))
+    private void onSetDead(CallbackInfo ci) {
+        if (ShouldFire.DESTRUCT_ENTITY_EVENT && !this.world.isRemote && !(EntityUtil.toNative(this) instanceof EntityLiving)) {
+            MessageChannel originalChannel = MessageChannel.TO_NONE;
+            try (final CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
+                Object source = PhaseTracker.getInstance().getCurrentContext().getSource();
+                frame.pushCause(source);
```
gabizou: I'm almost sure that at this point the source will already be pushed to the frame stack from the PhaseStates populating lazily the cause stack. Might want to check if that is indeed the case or not here.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
gabizou: Might want to have tick checks or something to keep object references weak in some way.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
ImMorpheus: `MixinWorld#onEntityRemoved`

https://github.com/SpongePowered/SpongeCommon/pull/1157/files#diff-dcfe7e9aa63b343150e8fbb6da0669d8R1456

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
ImMorpheus: Can this be leaked ? 

`Entity#setDead` and `World#onEntityRemoved` are called on the same tick (or with 1 tick difference).
Plus we're talking about the entity removal process, if the event is leaking then the entity instance would have to be leaking as well (Kind of, see below).

EDIT: Also, the event is fired (and the field is set to null) when the world is removing 'dead' entities, so even if the entity instance is leaked the field is null.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
gabizou: Need to document it then. Since it’s added to the super interface of IMixinEntity.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
gabizou: Right, if the event is leaking and the entity is leaking, at the very least we aren’t adding to the leak by keeping it there. Mods can do ~~stupid things~~ thought provokingly questionable things with entities.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
ImMorpheus: Done. Not sure what to add, this is basically the polymorphic version of `World#onEntityRemoved`

PR #1161, opened 2017-01-25
Add corporea spark to the entity collision config

Eufranio:
Should fix https://github.com/SpongePowered/SpongeForge/issues/1214
*and yay i'm contributing to sponge!!!*


Deamon5550:
Isn't the point of configuration that you can modify it without editing the base software?

Eufranio:
Some things (like this) are better coming ready from the base software, no one will think that the problem (like the one of this issue) is this actual one. They usually think this is a bug, and report instead configuring (not even me knew that Botania sparks had collisions 😄)

PR #1183, opened 2017-02-04
Implemented TreeType data for boats.

me4502:
This allows the ability to determine and set the type of boats, using the existing TreeType data.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/processor/data/block/TreeDataProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/data/block/TreeDataProcessor.java
@@ -24,6 +24,7 @@
  */
 package org.spongepowered.common.data.processor.data.block;
 
+import com.google.common.collect.ImmutableMap;
 import net.minecraft.block.BlockPlanks;
 import net.minecraft.item.ItemStack;
 import org.spongepowered.api.data.key.Keys;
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/common/data/processor/value/entity/BoatTreeTypeValueProcessor.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
+    public DataTransactionResult removeFrom(ValueContainer<?> container) {
+        return DataTransactionResult.successNoData();
+    }
+
+    @Override
+    protected ImmutableValue<TreeType> constructImmutableValue(TreeType value) {
+        return new ImmutableSpongeValue<>(Keys.TREE_TYPE, value);
```
kashike: You can use `this.key`.

src/main/java/org/spongepowered/common/data/processor/data/block/TreeDataProcessor.java
```diff
@@ -62,8 +65,21 @@ protected TreeType getFromMeta(int meta) {
     protected Optional<TreeType> getVal(ItemStack stack) {
         if (stack.getItem() == ItemTypes.LEAVES2 || stack.getItem() == ItemTypes.LOG2) {
             return Optional.of(getFromMeta(stack.getItemDamage() + 4));
+        } else if (stack.getItem() == ItemTypes.BOAT) {
```
kashike: maybe extract `stack.getItem()` to a local variable

src/main/java/org/spongepowered/common/data/processor/data/block/TreeDataProcessor.java
```diff
@@ -45,7 +45,10 @@ public TreeDataProcessor() {
         super(Keys.TREE_TYPE, input -> input.getItem() == ItemTypes.PLANKS || input.getItem() == ItemTypes.LEAVES
                 || input.getItem() == ItemTypes.LEAVES2 || input.getItem() == ItemTypes.LOG
                 || input.getItem() == ItemTypes.LOG2 || input.getItem() == ItemTypes.SAPLING
-                || input.getItem() == ItemTypes.WOODEN_SLAB);
+                || input.getItem() == ItemTypes.WOODEN_SLAB || input.getItem() == ItemTypes.BOAT
```
kashike: maybe move the second argument of this to a private static method to make this nicer

kashike (changes requested):

src/main/java/org/spongepowered/common/data/processor/data/block/TreeDataProcessor.java
me4502: Yeah, I was thinking either that or a static final immutable set

src/main/java/org/spongepowered/common/data/processor/data/block/TreeDataProcessor.java
me4502: Hmm, fair enough.

gabizou:
Care to add a test plugin to demonstrate the usage of this?

src/main/java/org/spongepowered/common/data/processor/data/block/TreeDataProcessor.java
ST-DDT: or a `Map`, that can be used for above purposes as well (`keySet()`)

src/main/java/org/spongepowered/common/data/processor/value/entity/BoatTreeTypeValueProcessor.java
```diff
@@ -0,0 +1,110 @@
...
+                return Optional.of(TreeTypes.BIRCH);
+            case JUNGLE:
+                return Optional.of(TreeTypes.JUNGLE);
+            case ACACIA:
+                return Optional.of(TreeTypes.ACACIA);
+            case DARK_OAK:
+                return Optional.of(TreeTypes.DARK_OAK);
+        }
+
+        return Optional.empty();
```
ST-DDT, line 93: I'm not sure about this one, maybe it would be better if we throw an IAE here to faster detect new enum values. Or at lest use some logging.

parlough:
@me4502 Can you update this for mentioned comments? Would like to see this added.

me4502:
Will do when I'm both home, and have working internet.

src/main/java/org/spongepowered/common/data/processor/value/entity/BoatTreeTypeValueProcessor.java
me4502: So my issue with an IllegalArgumentException, is that if a mod adds a new type of tree the mapping won't be here.

me4502:
I've made most requested changes now.

testplugins/src/main/java/org/spongepowered/test/BoatTypeTest.java
```diff
@@ -0,0 +1,51 @@
+package org.spongepowered.test;
```
kashike, line 25: license header

kashike (changes requested):

testplugins/src/main/java/org/spongepowered/test/BoatTypeTest.java
me4502: Fixed.

gabizou (approved):

kashike (approved):

me4502:
Hmm I think I merged this wrong, but it's merged.

PR #1185, opened 2017-02-04
Improved injection support

kashike:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1499) | [**SpongeCommon**](https://github.com/SpongePowered/SpongeCommon/pull/1185) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1246) | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/295)


stephan-gh (approved):

src/main/java/org/spongepowered/common/SpongePlatform.java
```diff
@@ -42,7 +42,7 @@
 public class SpongePlatform implements Platform {
 
     private final PluginContainer api;
-    private final PluginContainer impl;
+    private PluginContainer impl;
```
stephan-gh: Add back final here (I removed it accidentally)

stephan-gh (changes requested):

src/main/java/org/spongepowered/common/inject/SpongeGuice.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.google.inject.Stage;
+
+public final class SpongeGuice {
+
+    private SpongeGuice() {
+    }
+
+    public static Stage getInjectorStage(final Stage defaultStage) {
+        return Stage.valueOf(System.getProperty("stage", defaultStage.name()).toUpperCase());
```
stephan-gh: Can you make this somewhat more Guice specific? E.g. `guice.stage` or `inject.stage` or something similar. Guice is probably not the only thing that can have a stage

stephan-gh (approved):

PR #1186, opened 2017-02-05
Move text actions to the implementation

kashike:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1477) | [**SpongeCommon**](https://github.com/SpongePowered/SpongeCommon/pull/1186)


ryantheleach:
This was reopened potentially by accident. This will be covered in the API refactor where Text is moving to implementation.

PR #1187, opened 2017-02-05
Implement stopRecord, and fix playRecord.

me4502:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1478) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1187)

This implements stopRecord, and makes playRecord actually play the record if it exists.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/tileentity/MixinTileEntityJukebox.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/tileentity/MixinTileEntityJukebox.java
@@ -44,19 +44,21 @@ import java.util.Optional;
 @Mixin(BlockJukebox.TileEntityJukebox.class)
 public abstract class MixinTileEntityJukebox extends MixinTileEntity implements Jukebox {
 
+    private static final int PLAY_RECORD_EVENT = 1010;
+
     @Shadow public abstract net.minecraft.item.ItemStack getRecord();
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/core/tileentity/MixinTileEntityJukebox.java
```diff
@@ -49,12 +49,18 @@
 
     @Override
     public void playRecord() {
-        if (getRecord() != null) {
-            this.world.playEvent(null, 1005, this.pos, Item.getIdFromItem(getRecord().getItem()));
+        if (!getRecord().isEmpty()) {
+            this.world.playEvent(null, 1010, this.pos, Item.getIdFromItem(getRecord().getItem()));
```
ST-DDT: Looks like a magic number that you use in more than a single location. Maybe we could move it to a constant.

kashike (approved):

src/main/java/org/spongepowered/common/mixin/core/tileentity/MixinTileEntityJukebox.java
ST-DDT: ... a second too slow...

me4502: Yeah, just beat you on that one :P

PR #1191, opened 2017-02-06
Restore default state for blocks with invalid metadata instead of removing the block

Niko-sk2x:
Some worlds (at least old bukkit worlds with some plugins) have blocks with invalid metadata and loading these worlds in any MC version above 1.7.10 makes these blocks disappear. This resets them to default state instead. 

This shouldn't affect vanilla behavior for worlds with valid blocks.

Should I make it configurable and move it out of core package? And should I use overwrite here instead of redirect? It would make the code a bit simpler.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/chunk/MixinBlockStateContainer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/chunk/MixinBlockStateContainer.java
@@ -1,0 +1,73 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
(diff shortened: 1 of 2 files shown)


Deamon5550:
I don't think it needs to be configurable as there is zero reason to disable it and the default behavior can be considered 'broken'.

Niko-sk2x:
It failed to build on travis because of some random gradle issue, not because the code is wrong.

Niko-sk2x:
Is there something wrong with this PR is is there some other reason it's not merged?

kashike (approved):

PR #1192, opened 2017-02-07
Add tracking phase to piston movements.

JBYoshi:
Fixes SpongePowered#1167.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/block/MixinBlockPistonBase.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/block/MixinBlockPistonBase.java
@@ -27,6 +27,9 @@ package org.spongepowered.common.mixin.core.block;
 import com.google.common.collect.ImmutableList;
 import net.minecraft.block.BlockPistonBase;
 import net.minecraft.block.state.IBlockState;
+import net.minecraft.util.EnumFacing;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
 import org.spongepowered.api.block.BlockState;
 import org.spongepowered.api.data.key.Key;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/core/block/MixinBlockPistonBase.java
```diff
@@ -88,4 +103,35 @@ private ImmutableDirectionalData getDirectionalData(IBlockState blockState) {
         return ImmutableDataCachingUtil.getManipulator(ImmutableSpongeDirectionalData.class,
                 DirectionResolver.getFor(blockState.getValue(BlockPistonBase.FACING)));
     }
+
+    @Inject(method = "doMove", at = @At("HEAD"))
+    public void onDoMoveHead(World worldIn, BlockPos pos, EnumFacing direction, boolean extending, CallbackInfoReturnable<Boolean> cir) {
```
kashike, line 108: I personally prefer methods named more descriptively than `onDoMoveHead`. `startTrackingPhase` and `completeTrackingPhase` would be great.

gabizou:
@JBYoshi The reason that the piston moving state was removed from use was that it was potentially causing issues with cancelling events and whatnot. I'd also look into modifying the current phase state to throw individualized block events with item spawns per block change as ExplosionStates do. @bloodmc would have to weigh in again about the validity of why we removed it in the first place ([in this commit](https://github.com/SpongePowered/SpongeCommon/commit/1c3042af83a885b9c75aa41cdc571778c7f2f920)), you may be able to bypass this by still capture/track individual blocks changing and entity spawns.

JBYoshi:
Closing. #1210 fixes the issue that led to this without the havoc of re-enabling piston tracking.

PR #1205, opened 2017-02-14
Implement ItemStack isEmpty

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1487) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1205)

Adds an isEmpty Method to ItemStacks.

Also ItemTypes.NONE is now the same as ItemTypes.AIR.

### DIFF
--- a/src/main/java/org/spongepowered/common/item/inventory/SpongeItemStackBuilder.java
+++ b/src/main/java/org/spongepowered/common/item/inventory/SpongeItemStackBuilder.java
@@ -90,7 +90,7 @@ public class SpongeItemStackBuilder implements ItemStack.Builder {
 
     @Override
     public ItemStack.Builder quantity(int quantity) throws IllegalArgumentException {
-        checkArgument(quantity > 0, "Quantity must be greater than 0");
+        checkArgument(quantity >= 0, "Quantity must not be smaller than 0");
         this.quantity = quantity;
         return this;
     }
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/common/item/inventory/SpongeItemStackBuilder.java
```diff
@@ -297,6 +297,11 @@ public ItemType getCurrentItem() {
     public ItemStack build() throws IllegalStateException {
         checkState(this.type != null, "Item type has not been set");
         int max = this.type.getMaxStackQuantity();
+
+        if (type == ItemTypes.AIR || quantity <= 0) {
```
gabizou: Do not forget `ItemTypes.NONE`.

src/main/java/org/spongepowered/common/item/inventory/SpongeItemStackSnapshot.java
```diff
@@ -143,6 +144,11 @@ public int getCount() {
     }
 
     @Override
+    public boolean isEmpty() {
+        return this.itemType == ItemTypes.AIR || this.count <= 0;
```
gabizou: Are you specifically allowing air itemstacks to be created now? There is a way to validate that the `ItemStack` itself is empty (hint, look at the fields in this class).

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemStack.java
```diff
@@ -239,6 +240,11 @@ public boolean equalTo(ItemStack that) {
     }
 
     @Override
+    public boolean isEmpty() {
```
gabizou: Needs to be `@Implements` with an `@Intrinsic`

src/main/java/org/spongepowered/common/item/inventory/SpongeItemStackSnapshot.java
Faithcaio: found the privateStack. :)

PR #1207, opened 2017-02-15
Actually track entity spawns during inventory clicks.

JBYoshi:
Fixes SpongePowered#1193.

Test code:
```java
package jbyoshi.sponge.test;

import org.spongepowered.api.event.item.inventory.ClickInventoryEvent;
import org.spongepowered.api.item.ItemTypes;
import org.spongepowered.api.item.inventory.transaction.SlotTransaction;

@Plugin(id = "testspongecommon1193")
public class TestSpongeCommon1193 {

    @Listener
    public void onClickInventory(ClickInventoryEvent event) {
        if (event.getTargetInventory().getClass().getName().endsWith(".ContainerFurnace")) {
            for (SlotTransaction transaction : event.getTransactions()) {
                if (transaction.getOriginal().getType() == ItemTypes.GLASS) {
                    event.setCancelled(true);
                    break;
                }
            }
        }
    }
}
```

To test, smelt some sand and try to remove the glass. Be aware that smelting glass naturally does not always give XP; try removing it repeatedly.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/phase/packet/BasicInventoryPacketState.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/phase/packet/BasicInventoryPacketState.java
@@ -100,5 +100,10 @@ public class BasicInventoryPacketState extends BasicPacketState {
                .addEntityDropCaptures();
     }
 
+    @Override
+    public boolean shouldCaptureEntity() {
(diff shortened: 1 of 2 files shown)


JBYoshi:
Fixed a bug where XP would not spawn at all.

PR #1233, opened 2017-03-04
Actually find and warn on private @Listener methods.

JBYoshi:
https://forums.spongepowered.org/t/plugin-logger-does-not-do-anything/17805?u=jbyoshi

### DIFF
--- a/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
@@ -51,14 +51,18 @@ import org.spongepowered.common.event.gen.DefineableClassLoader;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/event/SpongeEventManager.java
```diff
@@ -211,12 +213,26 @@ public void registerListener(PluginContainer plugin, Object listenerObject) {
...
+        for (Class<?> handleParent = handle; handleParent != Object.class; handleParent = handle.getSuperclass()) {
+            for (Method method : handleParent.getDeclaredMethods()) {
+                if (method.getAnnotation(Listener.class) != null && !isValidHandler(method)) {
+                    invalidMethods.add(method);
+                }
+            }
+        }
+
+        for (Method method : invalidMethods) {
+            SpongeImpl.getLogger().warn("The method {} on {} has @{} but has the wrong signature", method,
```
Deamon5550: Seeing as you're changing this, it would probably be good to give a more helpful error message (eg. if its private warn specifically that event listeners cannot be private).

JBYoshi: @Deamon5550 Fixed.

ST-DDT:
Nice idea. As an addition/alternative I would suggest using an annotation processor, that checks those things at compile time.
(I might be able to help with that if needed)

JBYoshi:
@ST-DDT https://github.com/SpongePowered/SpongeAPI/pull/1513

PR #1249, opened 2017-03-11
Fix incorrect command permission checks in some situations.

### LINKED ISSUE #1243: minecraft:command calls plugin permissions
Issue is described here in more detail
https://github.com/NucleusPowered/Nucleus/issues/739

, but short story:
Install Nucleus (or probably any other plugin that overrides vanilla mc commands)
Give user `nucleus.time.base` perm
Player should be ONLY have perms to do `/time` to get the time (no perms to change it)
Use `/minecraft:time set xx`
Output:
server changes time
Permission called `nucleus.time.base`

What should happen: You should get a no permission message

Setup: 
Nucleus 25.1
SF 1.10.2 2223
Luckperms 2.17.71

dualspiral:
When a Sponge plugin command takes a command alias that either Vanilla Minecraft or a Forge mod tries to take later, any permission checks on for the vanilla/mod command are directed to the Sponge plugin's command instead. This commit fixes this by:

* Adding updateNamespacedAlias(String ownerId) method to the CommandBase (via `IMixinCommandBase`), which is called on instantiation of `MinecraftCommandWrapper` (and, by inheritance, `ForgeMinecraftCommandWrapper` in SF), prefixing the output of `getName()` with the owning plugin's ID followed by a colon.
* Injecting into the checkPermission call to use this namespaced alias, if it exists.

This then ensures that the permission check goes to the correct alias. I've opted to do it this way, rather than just change the `MinecraftCommandWrapper`, in case something calls `CommandBase#checkPermission` directly.

As an aside, I did want to use a `@Redirect` on `this.getName()`, but I couldn't get it to work. [I tried this](https://gist.github.com/dualspiral/bea485e2382e0c5e4d47c3e6e0786283), if someone would rather have a redirect and can point out what I've done wrong - I'll update it. Note that `Lnet/minecraft/command/CommandBase;getName()Ljava/lang/String;` matches what I get when I look at the bytecode...

**Test Plugin:** [something like this should suffice](https://gist.github.com/dualspiral/103cfe174610e0de4fe940f6368f1efd). Run `/tp` and `/minecraft:tp`. You should only get the permission check message for `/tp`, but unpatched, you get the permission check message for `/minecraft:tp` too. Patched, you do not.

This fixes #1243

### DIFF
--- a/src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
+++ b/src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
@@ -47,11 +47,11 @@ import org.spongepowered.api.text.translation.Translation;
 import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
 import org.spongepowered.common.SpongeImpl;
+import org.spongepowered.common.interfaces.command.IMixinCommandBase;
 import org.spongepowered.common.interfaces.command.IMixinCommandHandler;
 import org.spongepowered.common.text.translation.SpongeTranslation;
 import org.spongepowered.common.util.VecHelper;
 
-import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -61,10 +61,13 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Optional;
 
+import javax.annotation.Nullable;
+
 /**
  * Wrapper around ICommands so they fit into the Sponge command system.
  */
 public class MinecraftCommandWrapper implements CommandCallable {
+
     private static final String TRANSLATION_NO_PERMISSION = "commands.generic.permission";
     private final PluginContainer owner;
     protected final ICommand command;
@@ -81,6 +84,11 @@ public class MinecraftCommandWrapper implements CommandCallable {
     public MinecraftCommandWrapper(final PluginContainer owner, final ICommand command) {
         this.owner = owner;
         this.command = command;
+
+        // Add the namespaced alias to the wrapped command so permission checks are sent to the right command.
+        if (this.command instanceof IMixinCommandBase) {
+            ((IMixinCommandBase) this.command).updateNamespacedAlias(this.owner.getId());
+        }
     }
 
     private String[] splitArgs(String arguments) {
--- a/src/main/java/org/spongepowered/common/interfaces/command/IMixinCommandBase.java
+++ b/src/main/java/org/spongepowered/common/interfaces/command/IMixinCommandBase.java
@@ -30,4 +30,6 @@ public interface IMixinCommandBase {
 
     void setExpandedSelector(boolean expandedSelector);
 
+    void updateNamespacedAlias(String ownerId);
+
 }
--- a/src/main/java/org/spongepowered/common/mixin/core/command/MixinCommandBase.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/command/MixinCommandBase.java
@@ -25,13 +25,20 @@
 package org.spongepowered.common.mixin.core.command;
 
 import net.minecraft.command.CommandBase;
+import net.minecraft.command.ICommand;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
 import org.spongepowered.common.interfaces.command.IMixinCommandBase;
 
+import javax.annotation.Nullable;
+
 @Mixin(CommandBase.class)
-public abstract class MixinCommandBase implements IMixinCommandBase {
+public abstract class MixinCommandBase implements IMixinCommandBase, ICommand {
 
     private boolean expandedSelector;
+    @Nullable private String namespacedAlias = null;
 
     @Override
     public boolean isExpandedSelector() {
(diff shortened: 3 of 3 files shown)


stephan-gh:
The `@Redirect` you posted works fine for me: https://gist.github.com/Minecrell/6659c542f8ecc4de60dfdca7b814574b#file-java-L104-L106

dualspiral:
Huh, interesting. I kept getting this: https://gist.github.com/dualspiral/f89650348912b3d73f7f7a731aa59922

stephan-gh:
@dualspiral Can you try again? I used [this mixin](https://gist.github.com/Minecrell/03879edeb8cd1256058a223a96cb742a) and it seems to work fine.

Edit: Nevermind, seems to be a bug in the Mixin annotation processor. The injection isn't re-obfuscated properly.

dualspiral:
@Minecrell Ahh, right! Glad I'm not going mad. I can update this PR if you like, pending any changes to the AP.

I imagine that you know what the issue is now anyway, but if it makes any difference, I'm using IntelliJ.

stephan-gh:
@dualspiral The injection should work if you run in IntelliJ IDEA directly (in the dev environment). It will only fail in production.

dualspiral:
Right, I see, I tend to just build a production jar because I don't open up SF/SV in my IDE that often and... well, haven't set it up as a dev environment yet.

Let me know if/when you want me to revisit anything on this PR. I'll leave it like this for the moment unless you want me to change it now.

src/main/java/org/spongepowered/common/mixin/core/command/MixinCommandBase.java
```diff
@@ -43,4 +55,15 @@ public void setExpandedSelector(boolean expandedSelector) {
...
+        this.namespacedAlias = ownerId + ":" + this.getName();
+    }
+
+    @Inject(method = "checkPermission", at = @At("HEAD"), cancellable = true)
+    public void onCheckPermission(MinecraftServer server, ICommandSender commandSender, CallbackInfoReturnable<Boolean> cir) {
+        if (this.namespacedAlias != null) {
+            cir.setReturnValue(commandSender.canUseCommand(this.getRequiredPermissionLevel(), this.namespacedAlias));
+        }
+    }
 }
```
liach: Needs a new line.

src/main/java/org/spongepowered/common/interfaces/command/IMixinCommandBase.java
```diff
@@ -30,4 +30,5 @@
 
     void setExpandedSelector(boolean expandedSelector);
 
+    void updateNamespacedAlias(String ownerId);
 }
```
liach: Needs a new line.

src/main/java/org/spongepowered/common/mixin/core/command/MixinCommandBase.java
```diff
@@ -43,4 +55,15 @@ public void setExpandedSelector(boolean expandedSelector) {
...
 
+    @Override
+    public void updateNamespacedAlias(String ownerId) {
+        this.namespacedAlias = ownerId + ":" + this.getName();
+    }
+
+    @Inject(method = "checkPermission", at = @At("HEAD"), cancellable = true)
+    public void onCheckPermission(MinecraftServer server, ICommandSender commandSender, CallbackInfoReturnable<Boolean> cir) {
+        if (this.namespacedAlias != null) {
+            cir.setReturnValue(commandSender.canUseCommand(this.getRequiredPermissionLevel(), this.namespacedAlias));
```
liach: You don't really need `this` for instance method calls.

src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
```diff
@@ -81,6 +82,11 @@
     public MinecraftCommandWrapper(final PluginContainer owner, final ICommand command) {
         this.owner = owner;
         this.command = command;
+
+        // Add the namespaced alias to the wrapped command so permission checks are sent to the right command.
+        if (this.command instanceof IMixinCommandBase) {
+            ((IMixinCommandBase)this.command).updateNamespacedAlias(this.owner.getId());
```
liach:
```java
((IMixinCommandBase) this.command).updateNamespacedAlias(this.owner.getId());
```
Spacing?

stephan-gh:
@dualspiral Please try using a `@Redirect` now (after a Gradle refresh). It should be fixed in https://github.com/SpongePowered/Mixin/commit/4b7b5a05f4b155e99502dce250ee77a9f7a658b0

dualspiral:
@Minecrell Will do once I'm home from work, thanks for letting me know.

dualspiral:
@Minecrell Yep, this works fine now, thanks.

The patch has been tested on API 6, but should work on API 5.2 too.

src/main/java/org/spongepowered/common/mixin/core/command/MixinCommandBase.java
```diff
@@ -43,4 +50,15 @@ public void setExpandedSelector(boolean expandedSelector) {
         this.expandedSelector = expandedSelector;
     }
 
+    @Override
+    public void updateNamespacedAlias(String ownerId) {
+        this.namespacedAlias = ownerId + ":" + getName();
+    }
+
+    @Redirect(method = "checkPermission(Lnet/minecraft/server/MinecraftServer;Lnet/minecraft/command/ICommandSender;)Z",
```
stephan-gh: Is the full method descriptor necessary here? Usually you can omit the `(Lnet/minecraft/server/MinecraftServer;Lnet/minecraft/command/ICommandSender;)Z` if only one method with the same name exists in the target class (which is the case here I think?).

dualspiral: Yeah, fair point. I did this whilst trying to get it to work before, didn't think about it after. Will test and then commit.

dualspiral: @Minecrell Yep, you were right - works. Updated.

stephan-gh (approved):

kashike:
Thanks.

dualspiral:
Is there a possiblity of getting this in `stable-5` too? #1243 was reported from a 1.10.2 server.

I can recreate this as a PR for 1.10.2 too if you want - note that it will require bumping the Mixins dep to 0.6.8-SNAPSHOT too.

kashike:
PR welcome! :smile_cat:

dualspiral:
Will do tomorrow then!

dualspiral:
#1262

PR #1250, opened 2017-03-12
Uses pagination to display plugin list for players in game.

### LINKED ISSUE #1239: Sponge Plugins list String too big
running `/sponge plugins` on big packs can kick you from the server

`[16:01:13] [Netty Server IO #1/ERROR]: io.netty.handler.codec.EncoderException: String too big (was 36841 bytes encoded, max 32767)`

old issue https://github.com/SpongePowered/SpongeForge/issues/918

zachoooo:
Fixes #1239. Plugin lists can get extremely long. It is basically useless to have a massively long list of plugins displayed to you. Pagination gives you much more useful snippets of information. In addition, sometimes plugin lists would get so long that they would cause a crash when the server attempted to send a packet with a string length that was too long. This resolves that issue. I have attached a screenshot that shows how the PR works from a players view and from the console view.
screenshot from 2017-03-12 01-45-44

### DIFF
--- a/src/main/java/org/spongepowered/common/command/SpongeCommand.java
+++ b/src/main/java/org/spongepowered/common/command/SpongeCommand.java
@@ -96,6 +96,7 @@ import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Function;
@@ -490,7 +491,8 @@ public class SpongeCommand {
                             src.sendMessage(builder.build());
                         }
                     } else {
-                        Collection<PluginContainer> plugins = SpongeImpl.getGame().getPluginManager().getPlugins();
+                        List<PluginContainer> plugins = new ArrayList<PluginContainer>(SpongeImpl.getGame().getPluginManager().getPlugins());
+                        plugins.sort(Comparator.comparing(PluginContainer::getName));
                         if (src instanceof Player) {
                             List<Text> pluginList = new ArrayList<Text>();
                             PaginationList.Builder builder = PaginationList.builder();
--- a/src/main/java/org/spongepowered/common/command/SpongeCommand.java
+++ b/src/main/java/org/spongepowered/common/command/SpongeCommand.java
@@ -494,7 +494,7 @@ public class SpongeCommand {
                         if (src instanceof Player) {
                             List<Text> pluginList = new ArrayList<Text>();
                             PaginationList.Builder builder = PaginationList.builder();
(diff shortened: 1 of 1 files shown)


Deamon5550:
I think we should also be filtering this to only be a list of actual plugins, and not include all forge mods in here as well.

ST-DDT:
@Deamon5550 I would suggest an option flag for that.
Another useful flag would be "sorted". Otherwise you have to search on x pages for an entry you might just have not seen on the previous one.

kashike:
> I think we should also be filtering this to only be a list of actual plugins, and not include all forge mods in here as well.

I disagree. If anything, it should be a flag as @ST-DDT stated.

zachoooo:
@ST-DDT How exactly do you think they should be sorted?

ST-DDT:
@zachoooo Sort them by (display-)name?!

Or are you referring to the implementation?

Just wrap the list in a new `(Array)List` and then
`plugins.sort(Comparators.comparing(PluginContainer::getName))`

Sorry, I don't get your question.

zachoooo:
Done! @ST-DDT

Deamon5550:
@kashike @ST-DDT sure a flag is fine to optionally also display mods, but by default a command returning a list of sponge plugins should only return things that are actually sponge plugins.

ST-DDT:
@Deamon5550 If the distinction between Sponge plugins and mods in general is that important. Why do we detect mods in the first place and why is there nothing in the API that allows devs to do the same distinction?

IMO we should keep this PR to a single purpose one (fix client crash) and move the mod filter feature to a different/later PR.

ST-DDT (approved):

gabizou:
What is the new usage of the command with available flags? As of course, we'll need to update SpongePowered/SpongeDocs for it, tagging @Inscrutable and @SpongePowered/docs to update the docs for this.

zachoooo:
@gabizou My changes didn't affect the usage of the command. The pagination is done based off of whether or not the sender is a player. For non players, it sends them the original list.

ryantheleach:
"For non players, it sends them the original list."
Paginations should already handle this gracefully, if they do not, that's a separate bug.

As for mods vs plugins, I'd suggest sorting by Plugin/Mod then name, so you end up with.


Plugins:
  CreeperHeal
  Essentials
  FlardShack
Mods:
  IndustrialCraft
  SleepFixer
  SpongeForge

I'm not sure if Pagination are advanced enough to have "chapter headings" but for now just having separate formatting styles for plugins / mods should make it pretty apparent, especially if in the pagination header you have ((green) Plugin) / ((red) Mod) or something.

Zidane:
@Grinch is going to take it over and add his own enhancements. Credit to the original author will be retained and we seriously appreciate this. Its one of those "we should have done this from the start.."

kashike:
Superseded by #1481.

PR #1255, opened 2017-03-14
Feed animal event

hfreeb:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1522) | **SpongeCommon**

This is a response to the feature request #1382, adding an event that is fired when a player feeds an animal.


src/main/java/org/spongepowered/common/mixin/core/entity/passive/MixinEntityAnimal.java
```diff
@@ -25,11 +25,38 @@
...
+    }
+
+    @Inject(method = "processInteract", locals = LocalCapture.CAPTURE_FAILSOFT,
+            at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/passive/EntityAnimal;ageUp(IZ)V"))
+    private void onAgeFeed(EntityPlayer player, EnumHand hand, CallbackInfoReturnable<Boolean> cir, ItemStack itemStack) {
+        SpongeImpl.postEvent(SpongeEventFactory.createFeedAnimalEventAging(Cause.of(NamedCause.source(player)),
+                Optional.of(this.getLocation()), itemStack.getItem(), this));
+    }
+
+}
```
kashike: Missing newline.

kashike (changes requested):

src/main/java/org/spongepowered/common/mixin/core/entity/passive/MixinEntityOcelot.java
```diff
@@ -80,6 +83,14 @@ public int onTame(Random rand, int bound, EntityPlayer player, EnumHand hand) {
         return 1;
     }
 
+    @Inject(method = "processInteract", locals = LocalCapture.CAPTURE_FAILHARD,
+            at = @At(value = "FIELD", target = "Lnet/minecraft/entity/player/PlayerCapabilities;isCreativeMode:Z"))
+    public void onHealFeed(EntityPlayer player, EnumHand hand, CallbackInfoReturnable<Boolean> cir,
+            org.spongepowered.api.item.inventory.ItemStack itemStack) {
```
gabizou: wait, what method is this that it's using an API item stack? That doesn't make sense....

src/main/java/org/spongepowered/common/mixin/core/entity/passive/MixinEntityAnimal.java
```diff
@@ -25,11 +25,38 @@
...
 import org.spongepowered.common.mixin.core.entity.MixinEntityAgeable;
 
+import java.util.Optional;
+
 @Mixin(EntityAnimal.class)
 public abstract class MixinEntityAnimal extends MixinEntityAgeable implements Animal {
 
+    @Inject(method = "processInteract", locals = LocalCapture.CAPTURE_FAILSOFT, at = @At(value = "INVOKE", ordinal = 0,
+            target = "Lnet/minecraft/entity/passive/EntityAnimal;consumeItemFromStack(Lnet/minecraft/entity/player/EntityPlayer;Lnet/minecraft/item/ItemStack;)V"))
+    private void onBreedFeed(EntityPlayer player, EnumHand hand, CallbackInfoReturnable<Boolean> cir, ItemStack itemStack) {
```
gabizou: Fairly sure that none of these injections can be using API itemstacks since they are all `net.minecraft.item.ItemStack`s.

gabizou (changes requested):
Verify the injections succeed at runtime and in development. I don't believe they will succeed due to the classes used for the injections.

src/main/java/org/spongepowered/common/mixin/core/entity/passive/MixinEntityAnimal.java
hfreeb: Oops that was a mistake, but... I created a plugin to test it and the itemstacks all look correct? Is that meant to work?

gabizou: Did you compile and run this in production? because again, I'm pretty sure this will fail.

hfreeb: It does compile and it does run in production, I even tested it with a plugin and it worked correctly. Although, nonetheless, I have changed it.

src/main/java/org/spongepowered/common/mixin/core/entity/passive/MixinEntityWolf.java
```diff
@@ -90,4 +95,19 @@ public void supplyVanillaManipulators(List<DataManipulator<?, ?>> manipulators)
...
+    @Inject(method = "processInteract", locals = LocalCapture.CAPTURE_FAILHARD,
+                at = @At(value = "FIELD", ordinal = 0, target = "Lnet/minecraft/entity/player/PlayerCapabilities;isCreativeMode:Z"))
+    public void onHealFeed(EntityPlayer player, EnumHand hand, CallbackInfoReturnable<Boolean> cir, ItemStack itemStack) {
+        SpongeImpl.postEvent(SpongeEventFactory.createFeedAnimalEventHealing(Cause.of(NamedCause.source(player)),
+                Optional.of(this.getLocation()), ((org.spongepowered.api.item.inventory.ItemStack) itemStack).createSnapshot(), this));
+    }
+
+    @Inject(method = "processInteract", locals = LocalCapture.CAPTURE_FAILHARD,
+            at = @At(value = "FIELD", ordinal = 2, target = "Lnet/minecraft/entity/player/PlayerCapabilities;isCreativeMode:Z"))
+    public void onTameFeed(EntityPlayer player, EnumHand hand, CallbackInfoReturnable<Boolean> cir, ItemStack itemStack) {
```
stephan-gh: Can be private

src/main/java/org/spongepowered/common/mixin/core/entity/passive/MixinEntityWolf.java
```diff
@@ -90,4 +95,19 @@ public void supplyVanillaManipulators(List<DataManipulator<?, ?>> manipulators)
         manipulators.add(get(SittingData.class).get());
         manipulators.add(new SpongeAggressiveData(this.isAngry()));
     }
+
+    @Inject(method = "processInteract", locals = LocalCapture.CAPTURE_FAILHARD,
+                at = @At(value = "FIELD", ordinal = 0, target = "Lnet/minecraft/entity/player/PlayerCapabilities;isCreativeMode:Z"))
+    public void onHealFeed(EntityPlayer player, EnumHand hand, CallbackInfoReturnable<Boolean> cir, ItemStack itemStack) {
```
stephan-gh: Can be private

src/main/java/org/spongepowered/common/mixin/core/entity/passive/MixinEntityOcelot.java
```diff
@@ -80,6 +83,13 @@ public int onTame(Random rand, int bound, EntityPlayer player, EnumHand hand) {
         return 1;
     }
 
+    @Inject(method = "processInteract", locals = LocalCapture.CAPTURE_FAILHARD,
+            at = @At(value = "FIELD", target = "Lnet/minecraft/entity/player/PlayerCapabilities;isCreativeMode:Z"))
+    public void onHealFeed(EntityPlayer player, EnumHand hand, CallbackInfoReturnable<Boolean> cir, ItemStack itemStack) {
```
stephan-gh: Can be private

src/main/java/org/spongepowered/common/mixin/core/entity/passive/MixinEntityAnimal.java
```diff
@@ -25,11 +25,38 @@
...
+        SpongeImpl.postEvent(SpongeEventFactory.createFeedAnimalEventLove(Cause.of(NamedCause.source(player)),
+                Optional.of(this.getLocation()), ((org.spongepowered.api.item.inventory.ItemStack) itemStack).createSnapshot(), this));
+    }
+
+    @Inject(method = "processInteract", locals = LocalCapture.CAPTURE_FAILSOFT, at = @At(value = "INVOKE", ordinal = 1,
```
stephan-gh: Is there a special reason for using `FAILSOFT` in this class? Usually it's better to use `FAILHARD` to ensure it applies correctly or fails to start.

src/main/java/org/spongepowered/common/mixin/core/entity/passive/MixinEntityOcelot.java
jamierocks: Seems everything has always used public.

stephan-gh: Yeah, but since you don't want to override or access it anywhere except in this class there is no reason to keep it public. I'm not entirely sure if Mixin renames injection methods now but a while back this could actually cause really weird bugs if you had something like `onInit` in multiple mixins and it was actually overriding the injection method in super classes. (so only one got called)

gabizou: Use private if it's not accessed anywhere else.

parlough:
Don't these need to account for the events being cancelled?

hfreeb:
Ah, yes, I'll implement that.

parlough:
@Desetude Do you plan to finish this? If not, I should be able to take care of it.

src/main/java/org/spongepowered/common/mixin/core/entity/passive/MixinEntityWolf.java
```diff
@@ -90,4 +95,19 @@ public void supplyVanillaManipulators(List<DataManipulator<?, ?>> manipulators)
         manipulators.add(get(SittingData.class).get());
         manipulators.add(new SpongeAggressiveData(this.isAngry()));
     }
+
+    @Inject(method = "processInteract", locals = LocalCapture.CAPTURE_FAILHARD,
+                at = @At(value = "FIELD", ordinal = 0, target = "Lnet/minecraft/entity/player/PlayerCapabilities;isCreativeMode:Z"))
```
gabizou: If you're using an ordinal, I'd rather you used a Slice, like from [here](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java#L1010-L1028).

gabizou (changes requested):

Dockter:
Any chance of this hitting API 7?

parlough:
@Dockter The API nor implemention of this is ready yet. I'll give a full review later and explain why some larger changes may be needed.

hfreeb:
I have now implemented the cancellation checks. Although it seems that MixinEntityAnimal#onAgeFeed does not return out of processInteract with `cir.setReturnValue(true)` but the other injections do, which is very odd.

liach:
Does this one have more issues? I think it should be good.

ImMorpheus:
@Desetude Can you update this ?

gabizou:

PR #1258, opened 2017-03-17
Add a global cause manager to apply global phases to all worlds.

### LINKED ISSUE #1256: java.lang.Exception: null
http://pastebin.com/hkASXX8n

```
[20:56:17] [Server thread/ERROR] [Sponge]: /**************************************************************************************************************************/
[20:56:17] [Server thread/ERROR] [Sponge]: /*                                                      Unexpected                                                        */
[20:56:17] [Server thread/ERROR] [Sponge]: /**************************************************************************************************************************/
[20:56:17] [Server thread/ERROR] [Sponge]: /* Sponge's tracking system is very dependent on                                                                          */
[20:56:17] [Server thread/ERROR] [Sponge]: /* knowing whena change to any world takes place,                                                                         */
[20:56:17] [Server thread/ERROR] [Sponge]: /* however, we have been toldto complete a "phase"                                                                        */
[20:56:17] [Server thread/ERROR] [Sponge]: /* without having entered any phases.This is an error                                                                     */
[20:56:17] [Server thread/ERROR] [Sponge]: /* usually on Sponge's part, so a reportis required                                                                       */
[20:56:17] [Server thread/ERROR] [Sponge]: /* on the issue tracker on GitHub.                                                                                        */
[20:56:17] [Server thread/ERROR] [Sponge]: /**************************************************************************************************************************/
[20:56:17] [Server thread/ERROR] [Sponge]: /*   Affected World : WorldServer{LevelName=Submersed, DimensionId=0, DimensionType=minecraft:overworld}                  */
[20:56:17] [Server thread/ERROR] [Sponge]: /* StackTrace:                                                                                                            */
[20:56:17] [Server thread/ERROR] [Sponge]: /* java.lang.Exception: null                                                                                              */
[20:56:17] [Server thread/ERROR] [Sponge]: /*     org.spongepowered.common.event.tracking.CauseTracker.printEmptyStackOnCompletion(CauseTracker.java:288)            */
[20:56:17] [Server thread/ERROR] [Sponge]: /*     org.spongepowered.common.event.tracking.CauseTracker.completePhase(CauseTracker.java:186)                          */
[20:56:17] [Server thread/ERROR] [Sponge]: /*     org.spongepowered.common.command.SpongeCommandManager.lambda$completeCommandPhase$1(SpongeCommandManager.java:403) */
[20:56:17] [Server thread/ERROR] [Sponge]: /*     java.lang.Iterable.forEach(Unknown Source)                                                                         */
[20:56:17] [Server thread/ERROR] [Sponge]: /*     java.util.Collections$UnmodifiableCollection.forEach(Unknown Source)                                               */
[20:56:17] [Server thread/ERROR] [Sponge]: /*     org.spongepowered.common.command.SpongeCommandManager.completeCommandPhase(SpongeCommandManager.java:400)          */
[20:56:17] [Server thread/ERROR] [Sponge]: /*     org.spongepowered.common.command.SpongeCommandManager.process(SpongeCommandManager.java:296)                       */
[20:56:17] [Server thread/ERROR] [Sponge]: /*     net.minecraft.command.ServerCommandManager.func_71556_a(SourceFile:1083)                                           */
[20:56:17] [Server thread/ERROR] [Sponge]: /*     net.minecraft.server.dedicated.DedicatedServer.func_71333_ah(DedicatedServer.java:420)                             */
[20:56:17] [Server thread/ERROR] [Sponge]: /*     net.minecraft.server.dedicated.DedicatedServer.func_71190_q(DedicatedServer.java:385)                              */
[20:56:17] [Server thread/ERROR] [Sponge]: /*     net.minecraft.server.MinecraftServer.func_71217_p(MinecraftServer.java:624)                                        */
[20:56:17] [Server thread/ERROR] [Sponge]: /*     net.minecraft.server.MinecraftServer.run(MinecraftServer.java:482)                                                 */
[20:56:17] [Server thread/ERROR] [Sponge]: /*     java.lang.Thread.run(Unknown Source)                                                                               */
[20:56:17] [Server thread/ERROR] [Sponge]: /*                                                                                                                        */
[20:56:17] [Server thread/ERROR] [Sponge]: /* Minecraft : 1.11.2                                                                                                     */
[20:56:17] [Server thread/ERROR] [Sponge]: /* SpongeAPI : 6.0.0-SNAPSHOT-43b286d                                                                                     */
[20:56:17] [Server thread/ERROR] [Sponge]: /* SpongeForge : 1.11.2-2227-6.0.0-BETA-2232                                                                              */
[20:56:17] [Server thread/ERROR] [Sponge]: /* Minecraft Forge : 13.20.0.2227                                                                                         */
[20:56:17] [Server thread/ERROR] [Sponge]: /**************************************************************************************************************************/
```

```
AntiWDL-1.1-all.jar
AutoBroadcasterSponge-0.2.1.jar
CommandSigns-API-6.0.0-1.3.1.jar
CustomPlayerCount.-.1.1.jar
EasyScoreboards.jar
EconomyLite-5.0.0-2.10.2.jar
FlexibleLogin.jar
Graveyards-v2.2.0.jar
griefprevention-1.11.2-3.1.1.270.jar
helpme-5.0.0-0.2.2.jar
IPLog-0.1.1.jar
Launchpad-v1.0-FIXED.jar
LuckPerms-Sponge-3.0.14.jar
MiscStuff-1.0-SNAPSHOT.jar
Nucleus-0.26.1-S6.0-plugin.jar
projectportals-5.1.0-0.13.6.jar
spongeforge-1.11.2-2227-6.0.0-BETA-2232.jar
TitleMOTD.-.1.0.jar
TreeDestroyage-0.11-DEV-API5.0.0-0-all.jar
treepers-5.1.3.jar
twentyticks.jar
walletdrop-5.1.0-0.2.1.jar
worldedit-sponge-6.1.7-SNAPSHOT-dist.jar
```

JBYoshi:
`CauseTracker`s are specific to each world. This doesn't change that.

However, sometimes some code needs to track not just one world, but all worlds. Previously, it would look like this:
```java
for (WorldServer world : WorldManager.getWorlds()) {
    ((IMixinWorldServer) world).getCauseTracker().switchToPhase(/* ... */);
}

// code to track goes here

for (WorldServer world : WorldManager.getWorlds()) {
    ((IMixinWorldServer) world).getCauseTracker().completePhase(/* ... */);
}
```

But what happens when, during one of these states, a plugin loads a new world?
```java
Sponge.getServer().loadWorld("testcaseworld");
```

The server crashes.

The initial code tracks all worlds by looping through them before and after. But if a world is loaded in the meantime, then it's never tracked. Even worse, when the "global phase" ends, the tracking code tries to complete it in all worlds... even the new world in which it was never started.

`GlobalCauseTracker` fixes that. It manages its own phase stack (by using common code moved from `CauseTracker` into a new base class, `BaseCauseTracker`) and applies phase changes to all worlds. In addition, whenever a new world is loaded, the global phases are *automatically applied to that world.* It currently doesn't stop tracking a world when that world is unloaded, but that won't cause a crash, and I can add behavior for that fairly quickly if necessary.

Fixes #1256.

### DIFF
--- a/src/main/java/org/spongepowered/common/SpongeImplHooks.java
+++ b/src/main/java/org/spongepowered/common/SpongeImplHooks.java
@@ -38,11 +38,9 @@ import net.minecraft.entity.player.EntityPlayerMP;
 import net.minecraft.init.Blocks;
 import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.EnumFacing;
-import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.ReportedException;
+import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.Util;
-import net.minecraft.util.ReportedException;
-import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.MathHelper;
 import net.minecraft.world.GameType;
@@ -55,13 +53,10 @@ import net.minecraft.world.chunk.Chunk;
 import net.minecraft.world.storage.MapStorage;
 import org.apache.logging.log4j.Logger;
 import org.spongepowered.api.event.cause.NamedCause;
-import org.spongepowered.common.event.tracking.CauseTracker;
+import org.spongepowered.common.event.tracking.GlobalCauseTracker;
 import org.spongepowered.common.event.tracking.ItemDropData;
 import org.spongepowered.common.event.tracking.PhaseContext;
 import org.spongepowered.common.event.tracking.phase.plugin.PluginPhase;
-import org.spongepowered.common.interfaces.world.IMixinWorldServer;
-import org.spongepowered.common.mixin.core.world.MixinWorldType;
-import org.spongepowered.common.world.WorldManager;
 
 import java.util.Collection;
 import java.util.Iterator;
@@ -246,18 +241,13 @@ public final class SpongeImplHooks {
     }
 
     public static Object onUtilRunTask(FutureTask<?> task, Logger logger) {
-        for (WorldServer worldServer : WorldManager.getWorlds()) {
-            final CauseTracker otherCauseTracker = ((IMixinWorldServer) worldServer).getCauseTracker();
-            otherCauseTracker.switchToPhase(PluginPhase.State.SCHEDULED_TASK, PhaseContext.start()
-                    .add(NamedCause.source(task))
-                    .addCaptures()
-                    .complete()
-            );
-        }
+        GlobalCauseTracker.getInstance().switchToPhase(PluginPhase.State.SCHEDULED_TASK, PhaseContext.start()
+                .add(NamedCause.source(task))
+                .addCaptures()
+                .complete()
+        );
         Object value = Util.runTask(task, logger);
-        for (WorldServer worldServer : WorldManager.getWorlds()) {
-            ((IMixinWorldServer) worldServer).getCauseTracker().completePhase(PluginPhase.State.SCHEDULED_TASK);
-        }
+        GlobalCauseTracker.getInstance().completePhase(PluginPhase.State.SCHEDULED_TASK);
         return value;
     }
 
--- a/src/main/java/org/spongepowered/common/event/tracking/GlobalCauseTracker.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/GlobalCauseTracker.java
@@ -1,0 +1,80 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.event.tracking;
+
+import net.minecraft.world.WorldServer;
+import org.spongepowered.common.interfaces.world.IMixinWorldServer;
+import org.spongepowered.common.world.WorldManager;
+
+public final class GlobalCauseTracker extends BaseCauseTracker {
+
+    private static final GlobalCauseTracker INSTANCE = new GlobalCauseTracker();
+
+    private GlobalCauseTracker() {
+    }
+
+    public static GlobalCauseTracker getInstance() {
+        return INSTANCE;
+    }
+
+    @Override
+    public void switchToPhase(IPhaseState state, PhaseContext phaseContext) {
+        super.switchToPhase(state, phaseContext);
+        for (WorldServer world : WorldManager.getWorlds()) {
+            ((IMixinWorldServer) world).getCauseTracker().switchToPhase(state, phaseContext.copy());
+        }
+    }
+
+    @Override
+    protected void doCompletePhase(PhaseData currentPhaseData) {
+        for (WorldServer world : WorldManager.getWorlds()) {
+            ((IMixinWorldServer) world).getCauseTracker().completePhase(currentPhaseData.state);
+        }
+    }
+
+    @Override
+    protected String getTrackerDescription() {
+        return GlobalCauseTracker.class.getSimpleName();
+    }
+
+    @Override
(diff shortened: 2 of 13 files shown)


Deamon5550:
@gabizou and I were talking about changing the per-world cause trackers to all be global as I don't believe there are actually times when we need to interlace states such that two worlds need to be in two different states. Its a larger change so probably better to do this and then that but it bears discussion.

src/main/java/org/spongepowered/common/SpongeImplHooks.java
```diff
@@ -238,18 +234,13 @@ public static boolean canEntitySpawnHere(EntityLiving entityLiving, IEntityLivin
...
     public static Object onUtilRunTask(FutureTask<?> task, Logger logger) {
-        for (WorldServer worldServer : WorldManager.getWorlds()) {
-            final CauseTracker otherCauseTracker = ((IMixinWorldServer) worldServer).getCauseTracker();
-            otherCauseTracker.switchToPhase(PluginPhase.State.SCHEDULED_TASK, PhaseContext.start()
-                    .add(NamedCause.source(task))
-                    .addCaptures()
-                    .complete()
-            );
-        }
+        GlobalCauseTracker.INSTANCE.switchToPhase(PluginPhase.State.SCHEDULED_TASK, PhaseContext.start()
```
gabizou: If we can avoid using static fields, that'd be nicer (use a static method that sanity checks itself).

src/main/java/org/spongepowered/common/event/tracking/GlobalCauseTracker.java
```diff
@@ -0,0 +1,76 @@
...
+    public static final GlobalCauseTracker INSTANCE = new GlobalCauseTracker();
+
+    private GlobalCauseTracker() {
+    }
+
+    @Override
+    public void switchToPhase(IPhaseState state, PhaseContext phaseContext) {
+        super.switchToPhase(state, phaseContext);
+        for (WorldServer world : WorldManager.getWorlds()) {
+            ((IMixinWorldServer) world).getCauseTracker().switchToPhase(state, phaseContext.copy());
```
gabizou: If you are going for a global cause tracker, there can't be phase switches on the worlds themselves. You'd need to start tracking all the changes on the global cause tracker and executing the changes from the global cause tracker.

src/main/java/org/spongepowered/common/event/tracking/GlobalCauseTracker.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+    }
+
+    @Override
+    protected void doCompletePhase(PhaseData currentPhaseData) {
+        for (WorldServer world : WorldManager.getWorlds()) {
+            ((IMixinWorldServer) world).getCauseTracker().completePhase(currentPhaseData.state);
```
gabizou: This is where we have current issues with the cause tracker, since new worlds can be created without our knowledge, we can't feasibly say for that world to complete when it has never entered a phase to begin with.

gabizou (changes requested):
I'd much rather be able to track the changes on a per world basis without having per world `CauseTracker` objects. Yes, you'll have to modify about 50 method signatures to take a world in, but it will resolve the issue of multi-worlds being out of sync with phase states. Keep it simple, 1 object per game, resolve block changes and entity changes based on the world objects they utilized during their creation (block snapshots use this already), and pass those changes into the global phase contexts.

src/main/java/org/spongepowered/common/SpongeImplHooks.java
JBYoshi: I updated it to use a getter method, but what do you mean by "sanity check"? I don't believe it can return null as the instance is automatically initialized.

src/main/java/org/spongepowered/common/event/tracking/GlobalCauseTracker.java
JBYoshi: `GlobalCauseTracker` is not truly a cause tracker in and of itself. A better way to describe it would be a "global phase manager". It's meant to delegate phase changes into multiple worlds' `CauseTracker`s.

src/main/java/org/spongepowered/common/event/tracking/GlobalCauseTracker.java
JBYoshi: > since new worlds can be created without our knowledge

Actually, they can't. Access to the list of worlds is encapsulated in `WorldManager`. Whenever the `WorldManager` is asked to add a world to the list (as in, by loading it), it tells `GlobalCauseTracker` to apply all global phases to it.

src/main/java/org/spongepowered/common/event/tracking/PhaseContext.java
```diff
@@ -702,4 +703,51 @@ public int hashCode() {
...
+            to.blockEntitySpawnSupplier = new CapturedBlockEntitySpawnSupplier();
+        }
+        if (this.captureBlockPos != null) {
+            to.captureBlockPos = new CaptureBlockPos(this.captureBlockPos.pos);
+        }
+        to.owner = this.owner;
+        to.notifier = this.notifier;
+        to.source = this.source;
+    }
 }
```
liach: new line

src/main/java/org/spongepowered/common/event/tracking/UnwindingPhaseContext.java
```diff
@@ -92,4 +92,11 @@ static PhaseContext unwind(IPhaseState state, PhaseContext context) {
...
         return Optional.ofNullable(super.getSource(sourceClass).orElseGet(() -> this.unwindingContext.getSource(sourceClass).orElse(null)));
     }
+
+    @Override
+    public UnwindingPhaseContext copy() {
+        UnwindingPhaseContext to = new UnwindingPhaseContext(this.unwindingState, this.unwindingContext);
+        copy(to);
+        return to;
+    }
 }
```
liach: new line

src/main/java/org/spongepowered/common/event/tracking/PhaseContext.java
JBYoshi: Fixed.

src/main/java/org/spongepowered/common/event/tracking/UnwindingPhaseContext.java
JBYoshi: Fixed.

Just a tip for next time: You can add multiple comments to a single review instead of splitting them among multiple reviews. Just make a comment after you click "Start review" but before you click "Finish your review" or "Submit review".

liach: I just like to comment separately. Will comment together.

PR #1260, opened 2017-03-20
Implement per-player world borders

ghost:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1521) | **SpongeCommon**

This pull request is for the implementation of per-player world borders in SpongeCommon.
[Test plugin is here](https://pastebin.com/6JbCwxaB)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,3 +75,4 @@ nb-configuration.xml
 *.iws
 /.apt_generated/
 *.launch
+/.nb-gradle/
--- a/src/main/java/org/spongepowered/common/interfaces/entity/player/IMixinEntityPlayerMP.java
+++ b/src/main/java/org/spongepowered/common/interfaces/entity/player/IMixinEntityPlayerMP.java
@@ -65,8 +65,6 @@ public interface IMixinEntityPlayerMP extends IMixinEntityPlayer {
 
     void refreshExp();
 
-    Optional<WorldBorder> getWorldBorder();
-
     PlayerOwnBorderListener getWorldBorderListener();
 
 }
--- a/src/main/java/org/spongepowered/common/interfaces/entity/player/IMixinEntityPlayerMP.java
+++ b/src/main/java/org/spongepowered/common/interfaces/entity/player/IMixinEntityPlayerMP.java
@@ -31,7 +31,11 @@ import net.minecraft.util.EnumHand;
 import net.minecraft.util.math.BlockPos;
 import org.spongepowered.api.entity.living.player.User;
 import org.spongepowered.api.text.channel.MessageChannel;
+import org.spongepowered.api.world.WorldBorder;
 import org.spongepowered.common.interfaces.world.IMixinWorldServer;
+import org.spongepowered.common.world.border.PlayerOwnBorderListener;
+
+import java.util.Optional;
 
 import javax.annotation.Nullable;
 
@@ -60,4 +64,9 @@ public interface IMixinEntityPlayerMP extends IMixinEntityPlayer {
     void setPacketItem(ItemStack itemstack);
 
     void refreshExp();
+
+    Optional<WorldBorder> getWorldBorder();
+
+    PlayerOwnBorderListener getWorldBorderListener();
+
 }
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
@@ -230,7 +230,7 @@ public abstract class MixinEntityPlayerMP extends MixinEntityPlayer implements P
     @Nullable private Vector3d velocityOverride = null;
 
     @Nullable private WorldBorder worldBorder;
-    public final PlayerOwnBorderListener borderListener = new PlayerOwnBorderListener((EntityPlayerMP) (Object) this);
+    private final PlayerOwnBorderListener borderListener = new PlayerOwnBorderListener((EntityPlayerMP) (Object) this);
 
     @Inject(method = "removeEntity", at = @At(value = "INVOKE",
             target = "Lnet/minecraft/network/NetHandlerPlayServer;sendPacket(Lnet/minecraft/network/Packet;)V"))
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
@@ -932,9 +932,7 @@ public abstract class MixinEntityPlayerMP extends MixinEntityPlayer implements P
             }
             this.worldBorder = border;
             if (this.worldBorder != null) {
-                if (!((net.minecraft.world.border.WorldBorder) this.worldBorder).listeners.contains(this.borderListener)) {
-                    ((net.minecraft.world.border.WorldBorder) this.worldBorder).addListener(this.borderListener);
-                }
+                ((net.minecraft.world.border.WorldBorder) this.worldBorder).addListener(this.borderListener);
                 this.connection.sendPacket(new SPacketWorldBorder((net.minecraft.world.border.WorldBorder) (Object) this.worldBorder, SPacketWorldBorder.Action.INITIALIZE));
             } else { //unset the border if null
                 this.connection.sendPacket(new SPacketWorldBorder(this.world.getWorldBorder(), SPacketWorldBorder.Action.INITIALIZE));
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
@@ -34,6 +34,7 @@ import net.minecraft.command.ICommandSender;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.EntityList;
 import net.minecraft.entity.EntityLivingBase;
+import net.minecraft.entity.ai.attributes.BaseAttribute;
 import net.minecraft.entity.item.EntityItem;
 import net.minecraft.entity.passive.AbstractHorse;
 import net.minecraft.entity.player.EntityPlayer;
@@ -55,6 +56,7 @@ import net.minecraft.network.play.server.SPacketResourcePackSend;
 import net.minecraft.network.play.server.SPacketSetSlot;
 import net.minecraft.network.play.server.SPacketSoundEffect;
 import net.minecraft.network.play.server.SPacketSpawnPosition;
+import net.minecraft.network.play.server.SPacketWorldBorder;
 import net.minecraft.scoreboard.IScoreCriteria;
 import net.minecraft.scoreboard.Score;
 import net.minecraft.scoreboard.ScoreObjective;
@@ -116,6 +118,7 @@ import org.spongepowered.api.text.chat.ChatTypes;
 import org.spongepowered.api.text.chat.ChatVisibility;
 import org.spongepowered.api.text.title.Title;
 import org.spongepowered.api.util.Tristate;
+import org.spongepowered.api.world.WorldBorder;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Overwrite;
@@ -164,6 +167,7 @@ import org.spongepowered.common.util.LocaleCache;
 import org.spongepowered.common.util.NetworkUtil;
 import org.spongepowered.common.util.SkinUtil;
 import org.spongepowered.common.util.VecHelper;
+import org.spongepowered.common.world.border.PlayerOwnBorderListener;
 import org.spongepowered.common.world.storage.SpongePlayerDataHandler;
 
 import java.time.Instant;
@@ -224,6 +228,9 @@ public abstract class MixinEntityPlayerMP extends MixinEntityPlayer implements P
     private Scoreboard spongeScoreboard = Sponge.getGame().getServer().getServerScoreboard().get();
 
     @Nullable private Vector3d velocityOverride = null;
+
+    @Nullable private WorldBorder worldBorder;
+    public final PlayerOwnBorderListener borderListener = new PlayerOwnBorderListener((EntityPlayerMP) (Object) this);
 
     @Inject(method = "removeEntity", at = @At(value = "INVOKE",
             target = "Lnet/minecraft/network/NetHandlerPlayServer;sendPacket(Lnet/minecraft/network/Packet;)V"))
@@ -907,6 +914,37 @@ public abstract class MixinEntityPlayerMP extends MixinEntityPlayer implements P
             event.getChannel().ifPresent(channel -> channel.send(this, event.getMessage(), ChatTypes.CHAT));
         }
         return event;
+    }
+
+    @Override
+    public Optional<WorldBorder> getWorldBorder() {
+        return Optional.ofNullable(this.worldBorder);
+    }
+
+    @Override
+    public void setWorldBorder(@Nullable WorldBorder border, Cause cause) {
+        if (this.worldBorder == border) {
+            return; //do not fire an event since nothing would have changed
+        }
+        if (!SpongeImpl.postEvent(SpongeEventFactory.createChangeWorldBorderEventTargetPlayer(cause, Optional.ofNullable(this.worldBorder), Optional.ofNullable(border), this))) {
+            if (this.worldBorder != null) { //is the world border about to be unset?
+                ((net.minecraft.world.border.WorldBorder) this.worldBorder).listeners.remove(this.borderListener); //remove the listener, if so
+            }
+            this.worldBorder = border;
+            if (this.worldBorder != null) {
+                if (!((net.minecraft.world.border.WorldBorder) this.worldBorder).listeners.contains(this.borderListener)) {
+                    ((net.minecraft.world.border.WorldBorder) this.worldBorder).addListener(this.borderListener);
+                }
+                this.connection.sendPacket(new SPacketWorldBorder((net.minecraft.world.border.WorldBorder) (Object) this.worldBorder, SPacketWorldBorder.Action.INITIALIZE));
+            } else { //unset the border if null
+                this.connection.sendPacket(new SPacketWorldBorder(this.world.getWorldBorder(), SPacketWorldBorder.Action.INITIALIZE));
+            }
+        }
+    }
+
+    @Override
+    public PlayerOwnBorderListener getWorldBorderListener() {
+        return this.borderListener;
     }
 
 }
--- a/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
@@ -110,6 +110,7 @@ import org.spongepowered.common.interfaces.IMixinNetworkManager;
 import org.spongepowered.common.interfaces.entity.player.IMixinEntityPlayerMP;
 import org.spongepowered.common.interfaces.network.IMixinNetHandlerPlayServer;
 import org.spongepowered.common.interfaces.world.IMixinWorldServer;
+import org.spongepowered.common.mixin.core.entity.player.MixinEntityPlayerMP;
 import org.spongepowered.common.text.SpongeTexts;
 import org.spongepowered.common.util.VecHelper;
 
@@ -549,6 +550,7 @@ public abstract class MixinNetHandlerPlayServer implements PlayerConnection, IMi
         if (!event.isMessageCancelled()) {
             event.getChannel().ifPresent(channel -> channel.send(player, event.getMessage()));
         }
+        ((IMixinEntityPlayerMP) this.player).getWorldBorderListener().onPlayerDisconnect();
     }
 
     @Redirect(method = "processTryUseItemOnBlock", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/management/PlayerInteractionManager;processRightClickBlock(Lnet/minecraft/entity/player/EntityPlayer;Lnet/minecraft/world/World;Lnet/minecraft/item/ItemStack;Lnet/minecraft/util/EnumHand;Lnet/minecraft/util/math/BlockPos;Lnet/minecraft/util/EnumFacing;FFF)Lnet/minecraft/util/EnumActionResult;"))
--- a/src/main/java/org/spongepowered/common/world/border/PlayerOwnBorderListener.java
+++ b/src/main/java/org/spongepowered/common/world/border/PlayerOwnBorderListener.java
@@ -29,7 +29,7 @@ import net.minecraft.network.Packet;
 import net.minecraft.network.play.server.SPacketWorldBorder;
 import net.minecraft.world.border.IBorderListener;
 import net.minecraft.world.border.WorldBorder;
-import org.spongepowered.common.interfaces.entity.player.IMixinEntityPlayerMP;
+import org.spongepowered.api.entity.living.player.Player;
 
 public class PlayerOwnBorderListener implements IBorderListener {
 
@@ -76,7 +76,7 @@ public class PlayerOwnBorderListener implements IBorderListener {
      * This method is for cleaning up the player reference once they disconnect.
      */
     public void onPlayerDisconnect() {
-        ((IMixinEntityPlayerMP) this.player).getWorldBorder().ifPresent(border -> ((WorldBorder) border).listeners.remove(this));
+        ((Player) this.player).getWorldBorder().ifPresent(border -> ((WorldBorder) border).listeners.remove(this));
     }
 
     private void sendBorderPacket(Packet<?> packet) {
--- a/src/main/java/org/spongepowered/common/world/border/PlayerOwnBorderListener.java
+++ b/src/main/java/org/spongepowered/common/world/border/PlayerOwnBorderListener.java
@@ -27,19 +27,16 @@ package org.spongepowered.common.world.border;
 import net.minecraft.entity.player.EntityPlayerMP;
 import net.minecraft.network.Packet;
 import net.minecraft.network.play.server.SPacketWorldBorder;
-import net.minecraft.server.MinecraftServer;
 import net.minecraft.world.border.IBorderListener;
 import net.minecraft.world.border.WorldBorder;
-import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.common.interfaces.entity.player.IMixinEntityPlayerMP;
 
-public final class PlayerBorderListener implements IBorderListener {
+public class PlayerOwnBorderListener implements IBorderListener {
 
-    private final int dimensionId;
-    private final MinecraftServer server;
+    private EntityPlayerMP player;
 
-    public PlayerBorderListener(MinecraftServer server, int dimensionId) {
-        this.server = server;
-        this.dimensionId = dimensionId;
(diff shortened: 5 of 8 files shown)


src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -897,4 +902,34 @@ public void setSpectatorTarget(@Nullable org.spongepowered.api.entity.Entity ent
         return event;
     }
 
+    @Override
+    public WorldBorder getWorldBorder() {
+        if (this.worldBorder == null) {
+            this.worldBorder = cloneBorder((WorldBorder) this.world.worldBorder);
```
kashike: :-1:
Instead default to the `World`'s world border, like we do with scoreboards.

kashike (changes requested):

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "SpongeAPI"]
 	path = SpongeAPI
-	url = https://github.com/SpongePowered/SpongeAPI.git
-	branch = bleeding
+	url = https://github.com/sanman00/SpongeAPI.git
```
gabizou: No need to change the submodule since we have to update the submodule anyways when merging.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
ghost: The main reason why I do not default to the `World`'s world border is because if a plugin modifies the `Player`'s world border without having set it first, all of the players in the world would get the changes rather than just the specified player. Although, if similar behaviour occurs with scoreboards (which may not be relevant depending on how they are implemented in the game), then I will certainly change it (I will still change it regardless though).

sanman00: The main reason why I do not default to the `World`'s world border is because if a plugin modifies the `Player`'s world border without having set it first, all of the players in the world would get the changes rather than just the specified player. Although, if similar behaviour occurs with scoreboards (which may not be relevant depending on how they are implemented in the game), then I will certainly change it.

kashike: > Although, if similar behaviour occurs with scoreboards (which may not be relevant depending on how they are implemented in the game), then I will certainly change it (I will still change it regardless though).

This is the case.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -897,4 +902,34 @@ public void setSpectatorTarget(@Nullable org.spongepowered.api.entity.Entity ent
         return event;
     }
 
+    @Override
+    public WorldBorder getWorldBorder() {
+        if (this.worldBorder == null) {
+            this.worldBorder = (WorldBorder) this.world.worldBorder;
```
simon816: Use `world.getWorldBorder()`

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -897,4 +902,34 @@ public void setSpectatorTarget(@Nullable org.spongepowered.api.entity.Entity ent
...
+        }
+        return this.worldBorder;
+    }
+
+    @Override
+    public void setWorldBorder(WorldBorder border, Cause cause) {
+        WorldBorder originalBorder = this.worldBorder;
+        this.worldBorder = border;
+        this.connection.sendPacket(new SPacketWorldBorder((net.minecraft.world.border.WorldBorder) (Object) border, SPacketWorldBorder.Action.INITIALIZE));
+        SpongeEventFactory.createChangeWorldBorderEventTargetPlayer(cause, originalBorder, border, this);
```
simon816: Event never fired

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -897,4 +902,34 @@ public void setSpectatorTarget(@Nullable org.spongepowered.api.entity.Entity ent
...
+    }
+
+    public static WorldBorder cloneBorder(WorldBorder border) {
```
simon816: Method not used

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -496,6 +497,18 @@ public WorldBorder getWorldBorder() {
...
     @Override
+    public void setWorldBorder(WorldBorder border, Cause cause) {
+        WorldBorder originalBorder = (WorldBorder) this.worldBorder;
+        this.worldBorder.setCenter(border.getCenter().getX(), border.getCenter().getZ());
+        this.worldBorder.setDamageAmount(border.getDamageAmount());
+        this.worldBorder.setDamageBuffer(border.getDamageThreshold());
+        this.worldBorder.setTransition(border.getDiameter(), border.getNewDiameter(), border.getTimeRemaining());
+        this.worldBorder.setWarningDistance(border.getWarningDistance());
+        this.worldBorder.setWarningTime(border.getWarningTime());
+        SpongeEventFactory.createChangeWorldBorderEventTargetWorld(cause, originalBorder, border, this);
```
simon816: Event never fired

simon816 (changes requested):

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -897,4 +902,20 @@ public void setSpectatorTarget(@Nullable org.spongepowered.api.entity.Entity ent
...
+        }
+        return this.worldBorder;
+    }
+
+    @Override
+    public void setWorldBorder(WorldBorder border, Cause cause) {
+        WorldBorder originalBorder = this.worldBorder;
+        this.worldBorder = border;
+        this.connection.sendPacket(new SPacketWorldBorder((net.minecraft.world.border.WorldBorder) (Object) border, SPacketWorldBorder.Action.INITIALIZE));
+        SpongeImpl.postEvent(SpongeEventFactory.createChangeWorldBorderEventTargetPlayer(cause, originalBorder, border, this));
```
simon816: The event is cancellable but you don't handle the case where it's cancelled. same goes for below too.

simon816 (changes requested):

simon816 (approved):

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -897,4 +902,24 @@ public void setSpectatorTarget(@Nullable org.spongepowered.api.entity.Entity ent
...
+            this.worldBorder = (WorldBorder) this.world.getWorldBorder();
+        }
+        return this.worldBorder;
+    }
+
+    @Override
+    public void setWorldBorder(WorldBorder border, Cause cause) {
+        WorldBorder originalBorder = this.worldBorder;
+        if (!SpongeImpl.postEvent(SpongeEventFactory.createChangeWorldBorderEventTargetPlayer(cause, originalBorder, border, this))) {
+            this.worldBorder = originalBorder;
```
simon816: This should be the other way around. `postEvent` returns whether the event was cancelled.

Also, pretty sure you don't need to send `SPacketWorldBorder` if it wasn't changed

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -497,6 +498,20 @@ public WorldBorder getWorldBorder() {
     }
 
     @Override
+    public void setWorldBorder(WorldBorder border, Cause cause) {
+        WorldBorder originalBorder = (WorldBorder) this.worldBorder;
+        
+        if (SpongeImpl.postEvent(SpongeEventFactory.createChangeWorldBorderEventTargetWorld(cause, originalBorder, border, this))) {
```
simon816: Should be other way around, `if(!SpongeImpl.postEvent(...))`

simon816 (changes requested):

simon816 (approved):

.gitignore
```diff
@@ -75,3 +75,4 @@ nb-configuration.xml
 *.iws
 /.apt_generated/
 *.launch
+/.nb-gradle/
```
Minecrell: Missing new line at the end of the file

stephan-gh (changes requested):
Internally, the server tracks and sends changes to clients using an `IBorderListener`. In SpongeCommon this is implemented to send the changes to all players in the same world using [`PlayerBorderListener`](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/world/border/PlayerBorderListener.java). However, with your changes the players with a custom `WorldBorder` will still receive the changes from the main world border (you should be able to reproduce this by modifying the main world border using the `/worldborder` command).

You need to modify `PlayerBorderListener` to exclude all players with a custom world border. Additionally, if a new `WorldBorder` is assigned to a player, you need to make sure it has a `IBorderListener` assigned that notifies the client about changes.

Regarding your test plugin: By default you return the main world border when the player has no custom world border set. That means you are actually modifying the main world border in your test plugin and not a player-specific one.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -225,6 +228,8 @@
 
     @Nullable private Vector3d velocityOverride = null;
 
+    private WorldBorder worldBorder;
```
kashike: `@Nullable`

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -897,4 +902,18 @@ public void setSpectatorTarget(@Nullable org.spongepowered.api.entity.Entity ent
...
     }
 
+    @Override
+    public Optional<WorldBorder> getWorldBorder() {
+        return Optional.ofNullable(this.worldBorder);
+    }
+
+    @Override
+    public void setWorldBorder(WorldBorder border, Cause cause) {
+        WorldBorder originalBorder = this.worldBorder;
```
kashike: Unnecessary local variable.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -92,6 +92,7 @@
 import org.spongepowered.api.entity.living.player.Player;
 import org.spongepowered.api.entity.projectile.EnderPearl;
 import org.spongepowered.api.entity.projectile.source.ProjectileSource;
+import org.spongepowered.api.event.SpongeEventFactory;
```
kashike: Unnecessary import.

src/main/java/org/spongepowered/common/world/border/SpongeWorldBorderBuilder.java
```diff
@@ -0,0 +1,89 @@
+package org.spongepowered.common.world.border;
```
kashike, line 25: ~~Missing licence header.~~

src/main/java/org/spongepowered/common/world/border/SpongeWorldBorderBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public WorldBorder.Builder damageAmount(double damage) {
+        this.damageAmount = damage;
+        return this;
+    }
+
+    @Override
+    public WorldBorder build() throws IllegalStateException {
+        checkNotNull(this.centre, "A centre needs to be specified");
```
kashike: `checkState(this.centre != null, "A centre needs to be specified");`

kashike (changes requested):

stephan-gh:
The changes you made for the global `WorldBorder` listener look good, however you'll need to add a separate listener to all custom world borders that are created using the builder. It would need to track the players which have the WorldBorder set and send the appropriate changes only to them.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -897,4 +903,26 @@ public void setSpectatorTarget(@Nullable org.spongepowered.api.entity.Entity ent
         return event;
     }
 
+    @Override
+    public Optional<WorldBorder> getWorldBorder() {
+        return Optional.ofNullable(this.worldBorder);
+    }
+
+    @Override
+    public void setWorldBorder(WorldBorder border, Cause cause) {
```
kashike: `@Nullable WorldBorder border`

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -897,4 +903,26 @@ public void setSpectatorTarget(@Nullable org.spongepowered.api.entity.Entity ent
...
+        if (this.worldBorder == border) {
+            return; //do not fire an event since nothing would have changed
+        }
+        if (!SpongeImpl.postEvent(SpongeEventFactory.createChangeWorldBorderEventTargetPlayer(cause, Optional.ofNullable(this.worldBorder), Optional.ofNullable(border), this))) {
+            this.worldBorder = border;
+            if (border != null) {
+                ((net.minecraft.world.border.WorldBorder) this.worldBorder).addListener(new PlayerOwnBorderListener((EntityPlayerMP) (Object) this));
```
kashike: `this.worldBorder` is nullable.

src/main/java/org/spongepowered/common/world/border/PlayerOwnBorderListener.java
```diff
@@ -0,0 +1,82 @@
...
+    public void onDamageAmountChanged(WorldBorder border, double newAmount) {
+    }
+
+    @Override
+    public void onDamageBufferChanged(WorldBorder border, double newSize) {
+    }
+
+    private void sendBorderPacket(Packet<?> packet, WorldBorder border) {
+        Player spongePlayer = (Player) this.player;
+        if (spongePlayer.getWorldBorder().orElse(null) == border && border != null) {
```
kashike: `((Player) this.player).getWorldBorder()`, no need for the local variable.

kashike (changes requested):

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
ghost: What should be put/checked here since the field is marked `@Nullable`?

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -897,4 +903,26 @@ public void setSpectatorTarget(@Nullable org.spongepowered.api.entity.Entity ent
...
+        if (this.worldBorder == border) {
+            return; //do not fire an event since nothing would have changed
+        }
+        if (!SpongeImpl.postEvent(SpongeEventFactory.createChangeWorldBorderEventTargetPlayer(cause, Optional.ofNullable(this.worldBorder), Optional.ofNullable(border), this))) {
+            this.worldBorder = border;
+            if (this.worldBorder != null) {
+                ((net.minecraft.world.border.WorldBorder) this.worldBorder).addListener(new PlayerOwnBorderListener((EntityPlayerMP) (Object) this));
+                this.connection.sendPacket(new SPacketWorldBorder((net.minecraft.world.border.WorldBorder) (Object) this.worldBorder, SPacketWorldBorder.Action.INITIALIZE));
+            }
+            else { //unset the border if null
```
stephan-gh: Keep `}` and `else` on the same line

src/main/java/org/spongepowered/common/world/border/PlayerOwnBorderListener.java
```diff
@@ -0,0 +1,81 @@
...
+    @Override
+    public void onDamageAmountChanged(WorldBorder border, double newAmount) {
+    }
+
+    @Override
+    public void onDamageBufferChanged(WorldBorder border, double newSize) {
+    }
+
+    private void sendBorderPacket(Packet<?> packet, WorldBorder border) {
+        if (((Player) player).getWorldBorder().orElse(null) == border && border != null) {
```
stephan-gh: This will work but I'm worried that this class will leak the `player`  instance if the `WorldBorder` persists after the player leaves the game (because you never remove the listener from the world border). I think the best solution for this is to track the current players in the `WorldBorder` (e.g. in a list) and remove the players when they leave the server.

ghost: What about if I remove the listener once the player disconnects (storing the listener as a field).

stephan-gh: That should work too - several ways are possible, and most of them are fine as long as you make sure there is no reference to the player instance after they have disconnected. :)

ghost:
@Minecrell @kashike @simon816 @gabizou Is there anything else that I have to do on this pull request? I want to complete what you want done and then resolve the conflicts with the bleeding branch ~~as the final thing in this pull request prior to merging~~.

stephan-gh:
@sanman00 Thanks, I'll look through it again in the next days.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -897,4 +904,27 @@ public void setSpectatorTarget(@Nullable org.spongepowered.api.entity.Entity ent
...
+    @Override
+    public void setWorldBorder(@Nullable WorldBorder border, Cause cause) {
+        if (this.worldBorder == border) {
+            return; //do not fire an event since nothing would have changed
+        }
+        if (!SpongeImpl.postEvent(SpongeEventFactory.createChangeWorldBorderEventTargetPlayer(cause, Optional.ofNullable(this.worldBorder), Optional.ofNullable(border), this))) {
+            this.worldBorder = border;
+            if (this.worldBorder != null) {
+                if (!((net.minecraft.world.border.WorldBorder) this.worldBorder).listeners.contains(this.borderListener)) {
+                    ((net.minecraft.world.border.WorldBorder) this.worldBorder).addListener(new PlayerOwnBorderListener((EntityPlayerMP) (Object) this));
```
stephan-gh: I think you want to add `this.borderListener` instead of a new one here

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -897,4 +904,27 @@ public void setSpectatorTarget(@Nullable org.spongepowered.api.entity.Entity ent
...
+                }
+                this.connection.sendPacket(new SPacketWorldBorder((net.minecraft.world.border.WorldBorder) (Object) this.worldBorder, SPacketWorldBorder.Action.INITIALIZE));
+            } else { //unset the border if null
+                this.connection.sendPacket(new SPacketWorldBorder(new net.minecraft.world.border.WorldBorder(), SPacketWorldBorder.Action.INITIALIZE));
```
stephan-gh: In this case you should initialize the world's default world border again, not an empty one.

src/main/java/org/spongepowered/common/world/border/PlayerOwnBorderListener.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Override
+    public void onDamageBufferChanged(WorldBorder border, double newSize) {
+    }
+    
+    /**
+     * This method is for cleaning up the player reference once they disconnect.
+     */
+    public void onPlayerDisconnect() {
+        this.player = null;
```
stephan-gh: This works but since you have access to the world border's listeners you could just remove the listener entirely instead of clearing the reference to the player. The same should happen when you replace the world border on the player with a different one, then you no longer need the check below. (because the listener will be only present on the border that is currently set for the player)

src/main/java/org/spongepowered/common/world/border/SpongeWorldBorderBuilder.java
```diff
@@ -0,0 +1,108 @@
...
+ */
+package org.spongepowered.common.world.border;
+
+import com.flowpowered.math.vector.Vector3d;
+import org.spongepowered.api.world.WorldBorder;
+
+public class SpongeWorldBorderBuilder implements WorldBorder.Builder {
+
+    private double diameter;
+    private Vector3d centre = Vector3d.ZERO; //use a default value otherwise null is used
```
stephan-gh: For consistency, use `center` for the field as well (the getter/setter is also called `center`)

stephan-gh (changes requested):

src/main/java/org/spongepowered/common/world/border/PlayerOwnBorderListener.java
ghost: Are you saying that I should remove the world border listener from the player's world border instead of removing the reference to the player?

stephan-gh: Yes.

sanman00:
@Minecrell Is this everything that needs to be done?

src/main/java/org/spongepowered/common/world/border/PlayerOwnBorderListener.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Override
+    public void onDamageBufferChanged(WorldBorder border, double newSize) {
+    }
+    
+    /**
+     * This method is for cleaning up the player reference once they disconnect.
+     */
+    public void onPlayerDisconnect() {
+        ((net.minecraft.world.border.WorldBorder) ((MixinEntityPlayerMP) (Object) this.player).getWorldBorder().get()).listeners.remove(this);
```
stephan-gh: Have you tested this? You shouldn't be able to cast to a mixin class because they do not exist at runtime. Also you should make sure that the world border is actually still set to prevent errors (just replace `.get()` with `ifPresent` or with an additional `isPresent()` check)

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -524,6 +525,7 @@ public void onDisconnectHandler(PlayerList this$0, ITextComponent component) {
         if (!event.isMessageCancelled()) {
             event.getChannel().ifPresent(channel -> channel.send(player, event.getMessage()));
         }
+        ((MixinEntityPlayerMP) (Object) this.player).borderListener.onPlayerDisconnect();
```
stephan-gh: Same here, this cast shouldn't work. You'll need to expose it using an additional interface (check `org.spongepowered.common.interfaces` for some examples)

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -897,4 +904,30 @@ public void setSpectatorTarget(@Nullable org.spongepowered.api.entity.Entity ent
...
+        return Optional.ofNullable(this.worldBorder);
+    }
+
+    @Override
+    public void setWorldBorder(@Nullable WorldBorder border, Cause cause) {
+        if (this.worldBorder == border) {
+            return; //do not fire an event since nothing would have changed
+        }
+        if (!SpongeImpl.postEvent(SpongeEventFactory.createChangeWorldBorderEventTargetPlayer(cause, Optional.ofNullable(this.worldBorder), Optional.ofNullable(border), this))) {
+            if (border == null) { //is the list about to be unset?
```
stephan-gh: You need to remove the border listener if the player currently has a custom world border. That means that you shouldn't check if reverting to the world's world border but rather something like `this.worldBorder != null`

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -897,4 +904,30 @@ public void setSpectatorTarget(@Nullable org.spongepowered.api.entity.Entity ent
...
+                ((net.minecraft.world.border.WorldBorder) this.worldBorder).listeners.remove(this.borderListener); //remove the listener, if so
+            }
+            this.worldBorder = border;
+            if (this.worldBorder != null) {
+                if (!((net.minecraft.world.border.WorldBorder) this.worldBorder).listeners.contains(this.borderListener)) {
```
stephan-gh: This check shouldn't be necessary because you return from the method immediately if the given world border is the same one that is currently set.

stephan-gh (changes requested):

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
ghost: The check is necessary as the `border` variable represents a different `WorldBorder` object (if it were the same object, the listener would already be set and the method would have already exited by this point) so the listener may not be added to the new border. This check makes it so that a border listener is always present on the world border.

stephan-gh: The point is that the listener should be never already in the list because it is removed when the world border is unset again.

src/main/java/org/spongepowered/common/interfaces/entity/player/IMixinEntityPlayerMP.java
```diff
@@ -60,4 +64,9 @@ default boolean usesCustomClient() {
     void setPacketItem(ItemStack itemstack);
 
     void refreshExp();
+
+    Optional<WorldBorder> getWorldBorder();
```
stephan-gh: You can access this method using the `Player` interface, there is no need to have it here.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -225,6 +229,9 @@
 
     @Nullable private Vector3d velocityOverride = null;
 
+    @Nullable private WorldBorder worldBorder;
+    public final PlayerOwnBorderListener borderListener = new PlayerOwnBorderListener((EntityPlayerMP) (Object) this);
```
stephan-gh: Can be private

src/main/java/org/spongepowered/common/world/border/PlayerOwnBorderListener.java
```diff
@@ -0,0 +1,85 @@
...
+import net.minecraft.entity.player.EntityPlayerMP;
+import net.minecraft.network.Packet;
+import net.minecraft.network.play.server.SPacketWorldBorder;
+import net.minecraft.world.border.IBorderListener;
+import net.minecraft.world.border.WorldBorder;
+import org.spongepowered.common.interfaces.entity.player.IMixinEntityPlayerMP;
+
+public class PlayerOwnBorderListener implements IBorderListener {
+
+    private EntityPlayerMP player;
```
stephan-gh: Can be `final`

stephan-gh (changes requested):

PR #1262, opened 2017-03-21
[API 5.2] Fix incorrect command permission checks in some situations.

dualspiral:
[Original Issue](https://github.com/SpongePowered/SpongeCommon/issues/1243) | [Original PR on bleeding](https://github.com/SpongePowered/SpongeCommon/pull/1249)

For context, please see the linked issue and PR for bleeding above. The second commit is a cherry pick from bleeding, with method names updated to fit with 1.10.2. It has been tested to ensure it works.

This PR also contains a commit to bump Mixin to 0.6.8-SNAPSHOT, required for my `@Redirect`. The commit fixes the visibility of methods that Mixin now checks for, ensuring that Sponge will still compile.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -50,7 +50,7 @@ ext.apiVersion = version
 
 dependencies {
     compile api
-    compile('org.spongepowered:mixin:0.6.6-SNAPSHOT') {
+    compile('org.spongepowered:mixin:0.6.8-SNAPSHOT') {
         exclude module: 'launchwrapper'
     }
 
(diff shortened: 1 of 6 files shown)


ryantheleach:
I suspect that this renders PR https://github.com/SpongePowered/SpongeCommon/pull/1008 obsolete.

Assuming that your solution doesn't break existing mods, it looks to be far simpler. +1

dualspiral:
Yes, it works for existing mods. Sponge takes over all command registration and makes sure every command is namespaced, so I simply exploit that fact. Forge commands are wrapped in a `ForgeMinecraftCommandWrapper` which extends the `MinecraftCommandWrapper` - doing the same thing here.

I'd completely forgotten about your PR, and to be honest, am surprised the issue itself wasn't actually patched sooner.

ryantheleach:
I let it age due to circumstances in my life changing. I hope to get back to Sponge development soon.

PR #1277, opened 2017-03-30
Remove requirement for events to extend AbstractEvent

simon816:
Can also go into `stable-5`

### DIFF
--- a/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
@@ -27,7 +27,6 @@ package org.spongepowered.common.event;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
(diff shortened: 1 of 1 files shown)


kashike (approved):

liach:
Does this hurt existing behavior?

ST-DDT:
What are the implications of not extending AbstractEvent? Why was that needed in the first place?

kashike:
Requiring extension of `AbstractEvent` was only introduced when https://github.com/SpongePowered/SpongeCommon/commit/4d8439796cdd4b76569a97286ef2e04cf07976c3 was pushed, and only because events are not checked for `instanceof AbstractEvent` before casted.

stephan-gh (approved):

kashike:
related: https://github.com/SpongePowered/SpongeCommon/commit/bc8c8840c6a0ffc25688c4a61221c9a78359d327

liach:
Congratulations for that commit, thanks kashike!

PR #387, opened 2016-01-02
Fix CatalogTypes

ST-DDT:
This PR fixes (almost) all (currently) broken CatalogTypes.

**These fixes contain:**
* Existing `CatalogType`s not being accessible via `GameRegistry`
* `CatalogType`'s `getId()` method not being implemented or throwing errors
* `CatalogType`'s `getName()` method not being implemented or throwing errors
* `CatalogType`'s `getTranslation()` method not being implemented or throwing errors or returning the wrong `Translation`s (AFAICT)
* `CatalogType`'s `toString()` method only using the default implementation or throwing errors
* `GameRegistry#getAll(Class)` returning duplicate instances
* `GameRegistry`'s `getType(type, id)` returning absent or errors for valid ids
* `GameRegistry`'s `getType(type, id)` throwing an `Exception` instead of `Optional.empty()` if a key is missing/invalid (but != null)
* Classes mentioned in `@CatalogedBy` containing public static fields with `null` values

**This PR does not contain:**
* Any new `CatalogType` or change of a `CatalogType`'s usage, but including some id fixes or additions of aliases.
* Changes for classes that are not accessible via `CatalogTypes.XXX` and `GameRegistry.getAllOf(X)`

**(Im-)MutableCatalogRegistryModule**

I have added `MutableCatalogRegistryModule` and `ImmutableCatalogRegistryModule` to fix and prevent the following errors/uglinesses:
* `getById(String)` not returning values for none lowercase `id`s
* `getById(String)` throwing errors if the `id` was not found
* `getAll()` returning duplicates
* double casts (`Bar bar = (Bar) (Object) foo;`) for `CatalogType`s affected by mixins
* `CatalogType`s being registered with their `name` and not their `id`
* Forgotten `@RegisterCatalog` annotations
* some code duplication (~15 - 20 lines / module)

at least it improves code readbility

**Test-Plugin**
(Tested on sponge forge BETA release and fixed version)
* [Plugin.jar](https://dl.dropboxusercontent.com/u/16999313/Sponge/ZoneAPIChecker.jar)
* [Plugin sources](https://dl.dropboxusercontent.com/u/16999313/Sponge/ZoneAPIChecker.zip)

Wait for `GameStartedServerEvent` and you will get an overview about what is and isn't working.

The plugin searches for the following error types:
* No registry module for `CatalogType` class (`GameRegistry#getAll()` fails)
* `CatalogType#getId()` not implemented, errors or returns null
* `CatalogType#getName()` not implemented, errors or returns null
[...]

PR #1283, opened 2017-04-04
Add tests for CatalogType methods

ST-DDT:
Added a test that checks whether CatalogType methods are properly implemented.

I tried fixing the broken methods as far as possible and set the remaining to ignored to avoid failing builds for the other developers. The main purpose is to uncover existing failures and prevent future implementation issues.
I won't fix the remaining (ignored) failures as this would probably explode this PR. 
(Except for simple fixes)

Based on https://github.com/SpongePowered/SpongeCommon/issues/739#issuecomment-290988331

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/block/MixinBlock.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/block/MixinBlock.java
@@ -114,6 +114,7 @@ public abstract class MixinBlock implements BlockType, IMixinBlock {
     @Shadow public abstract String getUnlocalizedName();
     @Shadow public abstract Material getMaterial(IBlockState state);
     @Shadow public abstract IBlockState shadow$getDefaultState();
+    @Shadow public abstract boolean shadow$getTickRandomly();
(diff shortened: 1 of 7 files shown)


stephan-gh:
@ST-DDT What's wrong with all the `getTranslation()` methods? Are these catalog types that simply don't have any translation we could return or do they return the wrong translations?

ST-DDT:
Most of those methods are not implemented and thus throw an AbstractMethodError. See last commit for all AME candidates.

stephan-gh:
@ST-DDT As far as I can see you already had the implementation for most of these in #387. It would be nice if you could open a new PR with the implemented `getTranslation()` methods.

ST-DDT:
I'll start with only those "not yet implemented" `getTranslation()` methods.

ST-DDT:
I implemented the missing methods in #1306 if that PR is merged then this unit test will become green. (and all AbstractMethodErrors can be removed from the ignore list)

ST-DDT:
Updated to latest bleeding. Tests still works as expected.

ST-DDT:
Updated to latest bleeding. Tests still works as expected. They will get green when #1306 is merged. Someone willing to merge this in?

PR #1290, opened 2017-04-07
Made URLs in plugin descriptions clickable.

mattmess1221:
They will not be clickable if the url is invalid.

### DIFF
--- a/src/main/java/org/spongepowered/common/command/SpongeCommand.java
+++ b/src/main/java/org/spongepowered/common/command/SpongeCommand.java
@@ -64,6 +64,7 @@ import org.spongepowered.api.event.cause.Cause;
 import org.spongepowered.api.event.cause.NamedCause;
 import org.spongepowered.api.plugin.PluginContainer;
 import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.action.ClickAction;
 import org.spongepowered.api.text.action.TextActions;
 import org.spongepowered.api.text.format.TextColors;
 import org.spongepowered.api.text.format.TextStyles;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/command/SpongeCommand.java
```diff
@@ -476,8 +479,16 @@ private static CommandSpec getPluginsCommand() {
 
                             appendPluginMeta(builder, "ID", container.getId());
                             appendPluginMeta(builder, "Description", container.getDescription());
-                            appendPluginMeta(builder, "URL", container.getUrl());
-
+                            appendPluginMeta(builder, "URL", container.getUrl().map(url -> {
+                                ClickAction.OpenUrl action = null;
+                                try {
+                                    // make the url clickable
+                                    action = TextActions.openUrl(URI.create(url).toURL());
```
simon816: Why not `new URL(url)`?

mattmess1221: I've always just created them from a `URI`. Should I change it to `URL`?

src/main/java/org/spongepowered/common/command/SpongeCommand.java
```diff
@@ -476,8 +479,16 @@ private static CommandSpec getPluginsCommand() {
...
-
+                            appendPluginMeta(builder, "URL", container.getUrl().map(url -> {
+                                ClickAction.OpenUrl action = null;
+                                try {
+                                    // make the url clickable
+                                    action = TextActions.openUrl(new URL(url));
+                                } catch (MalformedURLException e) {
+                                    // or not
+                                }
+                                return Text.builder(url).onClick(action);
```
simon816, line 490: Could even make it underlined and blue to make it clear it's a link

kashike: :-1: for underlining, it makes it hard to read with a lot of resource packs, and partly with vanilla.

stephan-gh (approved):

PR #1294, opened 2017-04-09
[WIP] Conversation API Implementation

parlough:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1532) | **SpongeCommon** | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/310) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1427)

The work in progress SpongeCommon implementation of my Conversation API.


Meronat:
I cannot figure out what I did wrong with the submodules for this and the Vanilla and Forge PRs which seems to be causing the build to fail. Could someone provide some assistance there?

src/main/java/org/spongepowered/common/conversation/SpongeConversationManager.java
```diff
@@ -0,0 +1,192 @@
...
+    private final Logger logger;
+
+    @Inject
+    public SpongeConversationManager(Logger logger) {
+        this.logger = logger;
+    }
+
+    @Override
+    public Collection<Conversation> getConversation(PluginContainer plugin, String id) {
+        return conversations.get(plugin.getId() + ":" + id.toLowerCase());
```
kashike: `this.`

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -404,10 +407,21 @@ public boolean isChatColorsEnabled() {
...
     public void sendMessage(ChatType type, Text message) {
         checkNotNull(type, "type");
         checkNotNull(message, "message");
 
+        if (this.conversation != null && this.conversation.getChatHandler(this).isPresent()
+            && !this.conversation.getChatHandler(this).get().process(message)) {
+            return;
+        }
+
         ITextComponent component = SpongeTexts.toComponent(message);
```
kashike: You should move this logic to a method like `sendMessage0` and call it from `sendMessage` and `sendThroughMessage`.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -404,10 +407,21 @@ public boolean isChatColorsEnabled() {
     }
 
     @Override
+    public void sendThroughMessage(Text message) {
+        checkNotNull(message, "The message text cannot be null!");
+        this.connection.sendPacket(new SPacketChat(SpongeTexts.toComponent(message)));
```
kashike: See message regarding `sendMessage`.

src/main/java/org/spongepowered/common/event/conversation/SpongeConversationEndCause.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hashCode(this.endType);
+    }
+
+    @Override
+    public String toString() {
+        return Objects.toStringHelper("ConversationEndCause")
```
kashike: `toStringHelper(this)`

kashike (changes requested):

src/main/java/org/spongepowered/common/conversation/SpongeConversation.java
```diff
@@ -0,0 +1,219 @@
...
+            getConversants().forEach(c -> c.sendThroughMessage(question.getPrompt()));
+        }
+    }
+
+    @Override
+    public boolean end(ConversationEndType endType, Cause cause) {
+        cause = cause.with(NamedCause.of(endType.getName(), ConversationEndCause.builder().type(endType).build()));
+        final ConversationCloseEvent.Ending endingEvent = SpongeEventFactory.createConversationCloseEventEnding(cause, this);
+
+        if (SpongeImpl.postEvent(endingEvent) || endingEvent.isCancelled()) {
```
stephan-gh: `SpongeImpl.postEvent` returns whether the event was (not?) cancelled so the extra check isn't necessary.

src/main/java/org/spongepowered/common/conversation/SpongeConversationManager.java
```diff
@@ -0,0 +1,192 @@
...
+                                // Should not occur
+                                SpongeImpl.getLogger().error("For some reason the question result type did not match any.");
+                                conversation.end(ConversationEndTypes.ERROR, Cause.of(NamedCause.source(SpongeImpl.getPlugin())));
+                        }
+                    } catch (ArgumentParseException e) {
+                        conversant.sendThroughMessage(Text.of("Failed to parse your specified arguments. Restarting question."));
+                        conversation.setQuestion(question);
+                        SpongeImpl.getLogger().error("Failed to parse the arguments passed into " + conversant.getName() +
+                            "'s conversation." + " Sending them the same question again.");
+                        e.printStackTrace();
```
stephan-gh: You should give the exception to the logger message above and use placeholders instead of building the message manually. Basically like:

```java
SpongeImpl.getLogger().error("Failed to parse the arguments passed into {}'s conversation. Sending them the same question again.", conversant.getName(), e);
```

src/main/java/org/spongepowered/common/conversation/SpongeConversationManager.java
```diff
@@ -0,0 +1,192 @@
...
+                                break;
+                            case REPEAT:
+                                conversation.setQuestion(question);
+                                break;
+                            case END:
+                                conversation.end(ConversationEndTypes.FINISHED, Cause.of(NamedCause.source(conversation.getCreator())));
+                                break;
+                            default:
+                                // Should not occur
+                                SpongeImpl.getLogger().error("For some reason the question result type did not match any.");
```
stephan-gh: `this.logger`

src/main/java/org/spongepowered/common/conversation/SpongeQuestion.java
```diff
@@ -0,0 +1,77 @@
...
+    private final CommandElement arguments;
+
+    SpongeQuestion(String id, Text prompt, AnswerHandler handler, CommandElement arguments) {
+        this.id = id;
+        this.prompt = prompt;
+        this.handler = handler;
+        this.arguments = arguments;
+    }
+
+    @Nonnull
```
stephan-gh: This should be already handled by the `@NonnullByDefault` annotation in the package.

parlough: Yeah I noticed I left quite a lot of these around, I'll remove them next commit.

src/main/java/org/spongepowered/common/conversation/SpongeConversationArchetype.java
```diff
@@ -0,0 +1,166 @@
...
+        this.endingHandlers = endingHandlers;
+        this.catchesOutput = catchesOutput;
+        this.allowCommands = allowCommands;
+        this.defaultChatHandler = defaultHandler;
+        this.startingMessage = startingMessage;
+        this.id = id.toLowerCase();
+        this.exit = exit.toLowerCase();
+        this.title = title;
+        this.padding = padding;
+        if (title != null && padding != null) {
```
kashike: Header should be controllable by the plugin.

parlough: A header is generally composed of a title and it's padding like pagination but I like this idea. If a header is set then that would override those so it could be whatever they so desire.

parlough: Or do you think the header should be something separate from the title and padding such as in pagination?

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
parlough: So I don't want this logic within `sendThroughMessage` actually, but I would like some input.

The `sendThroughMessage` method was created, because you can set a conversation to catch `Conversant` input, which would include the messages you are trying to send to them while asking questions and similar. I couldn't think of the greatest way to accomplish besides making a new method which bypasses this catching. 

I would be happy to have some input on this.

src/main/java/org/spongepowered/common/event/conversation/SpongeConversationEndCauseBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+import org.spongepowered.api.event.cause.conversation.ConversationEndCause;
+import org.spongepowered.api.event.cause.conversation.ConversationEndCause.Builder;
+import org.spongepowered.api.event.cause.conversation.common.AbstractConversationEndCauseBuilder;
+
+import javax.annotation.Nonnull;
+
+public class SpongeConversationEndCauseBuilder extends AbstractConversationEndCauseBuilder<ConversationEndCause, Builder>
+    implements ConversationEndCause.Builder {
+
+    @Nonnull
```
stephan-gh: Unneeded, should be handled by `@NonnullByDefault`

src/main/java/org/spongepowered/common/text/conversation/SpongeConversation.java
```diff
@@ -0,0 +1,256 @@
...
+        this.externalChatHandlers = Collections.synchronizedMap(new WeakHashMap<>());
+
+        conversants.forEach(c -> this.externalChatHandlers.put(c, archetype.getDefaultChatHandler()));
+
+        Task.builder()
+                .async()
+                .interval(5, TimeUnit.MINUTES)
+                .execute(() -> {
+                    if (this.externalChatHandlers.size() == 0) {
+                        SpongeImpl.getLogger().error("A " + creator.getId() + ":" + getId().toLowerCase() + " conversation has ended due to being "
```
stephan-gh: Add a locale for the call to `toLowerCase()` (usually `Locale.ROOT` is appropriate for these)

parlough:
To-do for myself:

- [x] Fix compilation errors

- [x] Switch to Guava's MapMaker instead of my stupid Synchronized map usage

- [x] Add tab completion support

- [ ] Add console as a conversant

- [x] Allow optional header setting

- [x] Remove unnecessary annotations

- [x] Allow setting input tokenizer on question builder

- [x] Allow changing the no command usage message

- [x] Move ExternalChatHandlers to some sort of catalog type

src/main/java/org/spongepowered/common/command/SpongeCommandManager.java
```diff
@@ -266,6 +266,16 @@ public boolean containsMapping(CommandMapping mapping) {
 
     @Override
     public CommandResult process(CommandSource source, String commandLine) {
+        if (source instanceof Conversant) {
+            final Optional<Conversation> optionalConversation = ((Conversant) source).getConversation();
+            if (optionalConversation.isPresent()) {
+                final Conversation conversation = optionalConversation.get();
+                if (!conversation.allowsCommands()) {
+                    ((Conversant) source).sendThroughMessage(Text.of("You must exit this conversation before you can use commands!"));
```
kashike: I'd prefer if the conversation/plugin can handle this, to allow this message to be customised.

kashike (changes requested):

src/main/java/org/spongepowered/common/command/SpongeCommandManager.java
parlough: Will do :)

src/main/java/org/spongepowered/common/command/conversation/SpongeQuestion.java
```diff
@@ -0,0 +1,87 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.command.conversation;
+
+import org.spongepowered.api.command.args.CommandElement;
+import org.spongepowered.api.command.args.parsing.InputTokenizer;
+import org.spongepowered.api.command.conversation.AnswerHandler;
+import org.spongepowered.api.command.conversation.PromptHandler;
+import org.spongepowered.api.command.conversation.Question;
+public class SpongeQuestion implements Question {
```
kashike: Missing new line

kashike (changes requested):

parlough:
If anyone could provide some insight, my latest commit does not build due to the following log: https://gist.github.com/Meronat/056ed67062bdf549610843b7d977a0e0

Seems to be an issue with Guice and my SpongeConversationManager, mainly around this piece of the log:

```
1) Error injecting constructor, java.lang.IllegalArgumentException: The provided plugin object does not have an associated plugin container (in other words, is 'plugin' actually your plugin object?
      at org.spongepowered.common.command.conversation.SpongeConversationManager.<init>(SpongeConversationManager.java:86)
      at org.spongepowered.common.command.conversation.SpongeConversationManager.class(SpongeConversationManager.java:86)
      while locating org.spongepowered.common.command.conversation.SpongeConversationManager
      while locating org.spongepowered.api.command.conversation.ConversationManager
        for field at org.spongepowered.api.Sponge.conversationManager(Sponge.java:52)
```

This does not really make sense to me anyway since there is no plugin being passed into the line there(constructor), simply a logger.

kashike:
It sounds as if you're registering a command before the plugin container for the implementation has been created and is available.

parlough:
@kashike That makes quite a lot of sense actually. Thanks!  Will fix that soon.

src/main/java/org/spongepowered/common/command/conversation/SpongeConversationManager.java
```diff
@@ -0,0 +1,282 @@
...
+    @Override
+    public void endAll(ConversationEndType endType, Cause cause) {
+        synchronized (this.conversations) {
+            for (Conversation conversation : this.conversations.values()) {
+                conversation.end(endType, cause);
+            }
+        }
+    }
+
+    public static CommandSpec getCommand() {
```
kashike: `createCommand`

kashike (changes requested):

ImMorpheus:

PR #952, opened 2016-09-20
Implement JSON DataFormat

jamierocks:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1363) | **Common**

PR #1299, opened 2017-04-12
Implement JSON data format

stephan-gh:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1363) | **Common**

Replaces #952 and implements a direct serializer/parser for JSON into a `DataContainer` without going through configurate.


src/main/java/org/spongepowered/common/data/persistence/DataViewJsonWriter.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+        @Override public void flush() throws IOException {
+            throw new AssertionError();
+        }
+        @Override public void close() throws IOException {
+            throw new AssertionError();
+        }
+    };
+
+    private final List<Object> stack = new ArrayList<>();
```
jamierocks, line 66: why not use an actual Stack?

stephan-gh: I use the same as Gson itself is using. I don't really see a reason for using an actual `Stack`, the `ArrayList` will do the job just fine :P

PR #1303, opened 2017-04-13
Remove the not-so-helpful help command

### LINKED ISSUE #1242: Help command doesn't follow Vanilla usage
In Minecraft we have the help command which follows this syntax for page traversal `/help [page]`. So `/help 2` works fine and as expected.

If you try to do this in Sponge you are met with a "Unknown command. Try /help for a list of commands". Instead you must do `/page 2` to traverse the pages. This completely breaks what people would expect from this command.

kashike:
Resolves #1242


ryantheleach:
Sponge Help could be so much more with a little refactoring. it's better then minecraft's default, it's just a little rough around the edges in regards to the /help 2 to navigate to page 2.

image

parlough:
Yeah I agree it offers a lot over Minecraft's default. With a method for specifying the page with the command and in my opinion removing the horribly cluttering underlines would make the command perfectly fine. 

More on the underline. I know it is there to show it has a click action, but the color should be enough. The underline clutters the screen, sometimes makes the commands harder to read, and with some resource packs makes the text completely unreadable. Plus it is simply kind of ugly.

kashike:
My problems with our help command:
- duplication - `/achievement /achievement <give|take> <name|*> [player]` - `/achievement` is duplicated, taking up more room than necessary
- underlining of the first occurrence of the command makes it hard to read
- useless hover event to display text which is already displayed in chat directly

ryantheleach:
1. The duplication is due to the usage text containing the primary alias. We can change this easily enough for the sponge commands, but would involve rewriting all the translations for the minecraft usage text.

The other option would be to not display the primary alias.

2. is a contentious styling issue, it's probably about time we changed it honestly. people have gotten more used to clickable chat in minecraft and we can offer tips instead.

3. the hover text is used for the long description, vs the short description listed.

I'm happy to improve it, just wondering on how you want the problem solved.

kashike:
> The other option would be to not display the primary alias.

I'm not sure of the best way to address this. Changing the usage text isn't a good idea.

> is a contentious styling issue, it's probably about time we changed it honestly. people have gotten more used to clickable chat in minecraft and we can offer tips instead.

:+1: 

> the hover text is used for the long description, vs the short description listed.

Don't set hover if short == long

VapidLinus:
> Don't set hover if short == long

It might not be entirely obvious to users why hovering over some commands shows a description and some do not. Maybe better to just show it even if it matches the short one because at least there's the feedback for when you hover over it and you realise right away that there's no further description. More consistent.

PR #1304, opened 2017-04-14
Support for resources

mattmess1221:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1537) | [**SpongeCommon**](https://github.com/SpongePowered/SpongeCommon/pull/1304)

Only resources so far


Zidane:
It will never be intended for SpongeCommon to ever implement the client API and any invocations of it would never make sense for SpongeVanilla. This needs to be closed and kept to SpongeForge.

mattmess1221:
It's been discussed that resource packs on the server would be useful, so I figured I'd open this. Resources aren't necessarily constrained to the client.

PR #1306, opened 2017-04-14
Fix some AbstractMethodErrors

ST-DDT:
Fix some AbstractMethodErrors as requested in https://github.com/SpongePowered/SpongeCommon/pull/1283#issuecomment-293851180

* `Translatable#getTranslation()`
* `ArmorType#getRepairItemType()`

Also tried to fix some existing `getTranslation()` methods.

The methods can be tested with the unit test provided by https://github.com/SpongePowered/SpongeCommon/pull/1283.

However there are some limitations to it:
* `StatisticType`s don't provide proper translations by vanilla mc and thus custom ones needs to be added to the impls or the interface has to be changed in the API. (For now I changed them to one that makes it obvious and testable.)
* `PotionEffect#getPotionTranslation()` does not provide proper values for all potions, this needs to be changed in the API (`Optional`s) or the impls have to provide them or some entries have to be special handled. (Changed from no valid translations at all to mostly working)

Here is a list of catalog types that currently fail to provide proper translations:

CatalogType | Returned Failures | Comment
------|--|------------
BlockType | 25 | Should be fixable
EntityType | 18 | Some internal ones might not have translations
GameMode | 1 | `GameMode.NOT_SET`
ItemType | 11 | Should be fixable
~~PlantType~~ | 1 | `tile.flower1.dandelion.name` instead of `tile.flower2.dandelion.name` (Fixed)
PotionEffectType | ~~18~~ 13 | Needs API changes (suggesting removal)
Statistic | 1736 | All non-`basic`/`general`
StatisticType | * | Probably needs API changes

and including the indirect ones:

CatalogType | Accessible Failures | Comment
------|--|------------
BlockState | 979 | Should be (mostly) fixable 
BlockType | 31 | Should be fixable
EntityType | 18 | Some internal ones might not have translations
~~EnumTrait~~ | 2 | `Dandellion` (Fixed)
GameMode | 1 | `GameMode.NOT_SET`
ItemType | 29 | Should be fixable
~~PlantType~~ | 1 | `Dandellion` (Fixed)
PotionEffectType | ~~18~~ 13 | Needs API changes (suggesting removal)
Statistic | 1789 | All non-`basic`/`general`
StatisticType | * | Probably needs API changes

A full list can be found [here](https://gist.github.com/ST-DDT/4076d6a28339529a094853584657b9c8)

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/data/types/MixinBlockStoneSlabNewEnumType.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/data/types/MixinBlockStoneSlabNewEnumType.java
@@ -26,11 +26,15 @@ package org.spongepowered.common.mixin.core.data.types;
 
 import net.minecraft.block.BlockStoneSlabNew;
 import org.spongepowered.api.data.type.SlabType;
+import org.spongepowered.api.text.translation.Translation;
 import org.spongepowered.asm.mixin.Implements;
 import org.spongepowered.asm.mixin.Interface;
 import org.spongepowered.asm.mixin.Intrinsic;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+import javax.annotation.Nullable;
 
 @Mixin(BlockStoneSlabNew.EnumType.class)
 @Implements(@Interface(iface = SlabType.class, prefix = "slab$"))
@@ -38,6 +42,8 @@ public abstract class MixinBlockStoneSlabNewEnumType {
 
     @Shadow public abstract String shadow$getName();
     @Shadow public abstract String shadow$getUnlocalizedName();
+
+    @Nullable private Translation translation;
 
     public String slab$getId() {
         return "minecraft:" + shadow$getName();
@@ -47,4 +53,12 @@ public abstract class MixinBlockStoneSlabNewEnumType {
     public String slab$getName() {
         return shadow$getUnlocalizedName();
     }
+
+    public Translation slab$getTranslation() {
+        if (this.translation == null) {
+            this.translation = new SpongeTranslation("tile.stoneSlab2." + shadow$getUnlocalizedName() + ".name");
+        }
+        return this.translation;
+    }
+
 }
--- a/src/main/java/org/spongepowered/common/mixin/core/item/MixinItemArmorMaterial.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/item/MixinItemArmorMaterial.java
@@ -38,6 +38,8 @@ import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
 import java.util.Optional;
 
+import javax.annotation.Nullable;
+
 @Mixin(ItemArmor.ArmorMaterial.class)
 public abstract class MixinItemArmorMaterial implements ArmorType {
 
@@ -47,6 +49,9 @@ public abstract class MixinItemArmorMaterial implements ArmorType {
     // at dev time. Since it's capitalized, the client becomes unable to retrieve
     // the texture, as the resource location is wrong.
     private String capitalizedName;
+
+    @Nullable
+    private Optional<ItemType> repairItemType;
 
     @Inject(method = "<init>", at = @At("RETURN"))
     public void onConstruct(CallbackInfo ci) {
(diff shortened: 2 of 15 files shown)


src/main/java/org/spongepowered/common/mixin/core/item/MixinItemArmorMaterial.java
```diff
@@ -58,4 +62,21 @@ public String getId() {
     public String getName() {
         return this.name;
     }
+
+    @Override
+    public Optional<ItemType> getRepairItemType() {
+        switch (name) {
```
gabizou: Instead of a switch, use lazy evaluating fields.

gabizou (changes requested):

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemArmorMaterial.java
ST-DDT: Done.

stephan-gh (approved):

stephan-gh:
@ST-DDT Sorry for the delay. Would you mind checking if any changes are necessary for 1.12?

ST-DDT:
@Minecrell I totally forgot about this one. I'll habe a look this WE.

ST-DDT:
@Minecrell AFAICT this still works. If I combine it with #1283 the tests still pass.

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemArmorMaterial.java
```diff
@@ -58,4 +67,23 @@ public String getId() {
     public String getName() {
         return this.name;
     }
+
+    @Override
+    public Optional<ItemType> getRepairItemType() {
```
Zidane, line 73: Is this changed in SpongeForge? Shouldn't be hardcoded like this.

src/main/java/org/spongepowered/common/mixin/core/data/types/MixinBlockStoneSlabNewEnumType.java
```diff
@@ -47,4 +53,12 @@
     public String slab$getName() {
         return shadow$getUnlocalizedName();
     }
+
+    public Translation slab$getTranslation() {
```
Zidane, line 57: This and all the others really should be done at the end of the constructor and not lazily created like this imo.

ST-DDT: @gabizou requested me to use lazy evaluating fields [here](https://github.com/SpongePowered/SpongeCommon/pull/1306#discussion_r111633951). Shall I change it? (Some values are set after the constructor)

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemArmorMaterial.java
ST-DDT: IIRC by the time this was implemented, there was no useful method in vanilla mc. Now there is the `shadow$getRepairItem()` that can be used for that. Fixed in latest commit. 

But it does not really matter since it is an enum.

src/main/java/org/spongepowered/common/mixin/core/data/types/MixinBlockStoneSlabNewEnumType.java
gabizou: In my opinion, unless they're used, they should be lazily evaluated.

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemArmorMaterial.java
gabizou: It is a forge related method now, since the item can be constructed differently. What we should technically do is make the breaking change to use `ItemStackSnapshot`s instead of `ItemType`s. But I did come across this being unimplemented the other day and had planned on implementing it in a hot fix commit.

ST-DDT: I cannot find the relevant code in forge can you link it here please?

IMO we should first fix it here and we can also backport this as is and then decide how to implement it in forge differently.

ST-DDT:
Updated to latest bleeding. Someone willing to merge this in?

Deamon5550 (approved):
Check that theres an issue for handling `getRepairItemType` if it needs more work that you're not handling with this PR w.r.t forge.

PR #1313, opened 2017-04-19
Bugfix: getting registered DimensionType by type id.

radfast:
See https://github.com/SpongePowered/SpongeForge/issues/1429

This helps compatibility with Forge mods using vanilla DimensionType and Forge's DimensionManager.

I've done a [search](https://github.com/search?q=org%3ASpongePowered+getRegisteredDimensionIds&type=Code) and the only place ```WorldManager.getRegisteredDimensionIds()``` is referenced in Sponge code is from Forge's DimensionManager.


gabizou:
Many thanks for looking into this! Big kudos to you.

radfast:
You're very welcome!  Sponge is impressive, and has my full support.

Issue #1153, opened 2017-01-23
Cannot control pagination padding color properly


parlough:
I am trying to create a paginated list and attempting to set the padding to gray equal signs but for some reason the color I set is not what appears. The code is for the list called "IPLog Help" in the screenshot. 

![image](https://cloud.githubusercontent.com/assets/18372958/22198008/2f6f5bae-e11a-11e6-88fa-217a6a04b128.png)

parlough:
So it seems the padding's style is set to the style of the "title". I'm not necessarily sure why, but instead maybe we could keep coloring of the padding and just remove the style such as bold, italic, etc. 

The affected line is in the [pagination calculator](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java#L217). 

So instead the `withStyle` setting the color and style to the what the title has, let's just set the style.

The original would go from -

```java
private Text withStyle(Text text, Text styled) {
    return text.toBuilder().color(styled.getColor()).style(styled.getStyle()).build();
}
```

to -

```java
private Text withStyle(Text text, Text styled) {
    return text.toBuilder().style(styled.getStyle()).build();
}
```

I'm not sure if there is an unapparent reason the color is reset though, so hoping for some clarification.

PR #1332, opened 2017-05-03
Fix pagination styling issue and clean up class a little

parlough:
This pull request accomplishes a few things. First of all it fixes the issue where the color of the title/header/footer overrides the color the padding. So a paginated list will go from this: 

image
to this as it should:
image

To accomplish this and keep the arrows(« 1/2 »)  colored, I used the padding color instead of the title color which seems to work better actually anyway as they reside right next to the padding. Doing this made me think in the API we should allow this color to be specified on its own.

I had to modify Sponge's help command as well so it would keep its padding color. 

I also cleaned up some the pagination calculator in general, and while I was at it cleaned up some of the Javadocs and added some other ones.

Fixes https://github.com/SpongePowered/SpongeCommon/issues/1153

### DIFF
--- a/src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java
+++ b/src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java
@@ -46,9 +46,11 @@ import java.util.List;
 import java.util.PrimitiveIterator;
 
 /**
- * Pagination calculator for players.
+ * Pagination calculator for players. Handles calculation of text widths,
+ * centering text, adding padding, adding spacing, and more.
  */
-public class PaginationCalculator {
+class PaginationCalculator {
+
     private static final String NON_UNICODE_CHARS;
     private static final int[] NON_UNICODE_CHAR_WIDTHS;
     private static final byte[] UNICODE_CHAR_WIDTHS;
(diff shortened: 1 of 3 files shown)


kashike:
Could you provide a before and after of the output of `/help`?

parlough:
Sure. "/help" appearance stays the same with my changes though:

Original:
image

New:
image

parlough:
Unless you guys have any requests, this PR should be good to go.

kashike (approved):

kashike:
Thanks!

src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java
```diff
@@ -184,45 +200,49 @@ int getWidth(Text text) {
...
             return text;
         }
-        Text styledSpace = this.withStyle(LiteralText.of(" "), text);
-        Text textWithSpaces = this.addSpaces(styledSpace, text);
+        final Text textWithSpaces = addSpaces(LiteralText.of(" "), text);
 
         //Minecraft breaks lines when the next character would be > then LINE_WIDTH
         boolean addSpaces = getWidth(textWithSpaces) <= LINE_WIDTH;
 
         //TODO: suspect, why are we changing the style of the padding, they may want different styles on the padding.
```
ryantheleach, line 227: This TODO could probably be removed now.

ryantheleach:
A tiny nitpick is that the numbers coloring blends in with the padding, when they have an action associated with clicking on them.

Sir-Will:
Is this not in any build yet or why is this not working in 1.10.2 build 2454?

PR #1343, opened 2017-05-09
[WIP] General SpongeCommon Cleanup

parlough:
I'm not sure if a pull request like this is desired, but as I was looking through SpongeCommon and expanding my knowledge of different parts I began to clean up some small things then I went ahead and started to fix a lot of checkstyle warnings. So I'm fine if this is closed or I have to work to change a lot.

There's still quite a few checkstyle warnings, mostly due to registry classes and mixins, but I will do my best to continue to minimize them while still being reasonable. Since I began though there are 100s less.

I still have a bit to do, especially with finishing comments here and there but I'm not necessarily sure how to describe all of them so I'll do all of them that I can. I also plan to clean up some error messages that aren't too descriptive.

I tested this for around 30 minutes with a few plugins and many different actions to guarantee I didn't mess anything up.

I'll keep this updated with changes to bleeding until I finish as well of course.

The first commit message got messed up due to my failure at rebasing but it doesn't affect anything at least :P


parlough:
The submodule is messed up due to there not being a submodule update on SpongeCommon recently.

src/main/java/org/spongepowered/common/block/SpongeTileEntityArchetype.java
```diff
@@ -134,7 +134,7 @@ protected ValidationType getValidationType() {
     }
 
     @Override
-    public TileEntityArchetype copy() {
+    public org.spongepowered.api.block.tileentity.TileEntityArchetype copy() {
```
Deamon5550, line 137: Don't change this

src/main/java/org/spongepowered/common/block/SpongeTileEntityArchetypeBuilder.java
```diff
@@ -49,41 +48,42 @@
...
+        implements org.spongepowered.api.block.tileentity.TileEntityArchetype.Builder {
 
     BlockState     blockState;    // -These two fields can never be null
     @Nullable
     TileEntityType tileEntityType;
     DataContainer  tileData;      // This can be empty, but cannot be null.
 
     public SpongeTileEntityArchetypeBuilder() {
-        super(TileEntityArchetype.class, DataVersions.TileEntitArchetype.BASE_VERSION);
+        super(org.spongepowered.api.block.tileentity.TileEntityArchetype.class, TileEntityArchetype.BASE_VERSION);
```
Deamon5550, line 66: Or this

src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeAuthorData.java
```diff
@@ -37,7 +37,7 @@
 
 public class ImmutableSpongeAuthorData extends AbstractImmutableSingleData<Text, ImmutableAuthorData, AuthorData> implements ImmutableAuthorData {
 
-    final ImmutableSpongeValue<Text> author;
+    private final ImmutableSpongeValue<Text> author;
```
Deamon5550, line 40: Make sure this isn't used anywhere that would create a synthetic accessor with this change. I don't think many people set their IDEs to warn about those but most things set to package-private is to avoid the extra synthetic.

src/main/java/org/spongepowered/common/entity/ai/target/SpongeTargetAIBuilder.java
```diff
@@ -28,8 +28,10 @@
 
 public abstract class SpongeTargetAIBuilder<A extends TargetAITask<A>, B extends TargetAITask.Builder<A, B>> implements TargetAITask.Builder<A, B> {
 
-    protected boolean checkSight, onlyNearby;
-    protected int searchDelay, interruptTargetUnseenTicks;
+    boolean checkSight;
```
Deamon5550, line 31: protected to package-private? why?

Deamon5550 (approved):
Most of the line wrapping I'd say could be left out, but whatever

src/main/java/org/spongepowered/common/entity/ai/target/SpongeTargetAIBuilder.java
parlough: This one was an accident as I moved them to their own lines. Will fix.

src/main/java/org/spongepowered/common/block/SpongeTileEntityArchetype.java
parlough: Didn't mean to do this will do :)

src/main/java/org/spongepowered/common/SpongeCatalogType.java
```diff
@@ -78,7 +78,7 @@ protected ToStringHelper toStringHelper() {
                 .add("name", getName());
     }
 
-    public static abstract class Translatable extends SpongeCatalogType implements org.spongepowered.api.text.translation.Translatable {
+    public abstract static class Translatable extends SpongeCatalogType implements org.spongepowered.api.text.translation.Translatable {
```
kashike, line 81: undo

src/main/java/org/spongepowered/common/profile/query/package-info.java
```diff
@@ -22,5 +22,4 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-@org.spongepowered.api.util.annotation.NonnullByDefault
-package org.spongepowered.common.profile.query;
+@org.spongepowered.api.util.annotation.NonnullByDefault package org.spongepowered.common.profile.query;
```
kashike, line 25: undo these changes (keep annotation on separate line), applies to all files changed

kashike (changes requested):

src/main/java/org/spongepowered/common/profile/query/package-info.java
parlough: Will do for all packages info files ✔

src/main/java/org/spongepowered/common/SpongeCatalogType.java
parlough: Okay. I will undo these then and I think remove the JLS order check style altogether as modifier order does not matter at all and it is hard to guarantee people ordering in the same way anyway. Will just get messed up again.

parlough:
For now invalidated by a huge number of changes with 1.12 and 74080b0c75e59075f164f5151b08be395e2d17fc

Zidane:
@Meronat 

Issues pointed out aside, we don't refuse these kinds of things BUT its very difficult for us to just up and pull it due to the sheer amount of files changed.

One way you can help us out is to break it up in segments...so it can be easily digested.

Regardless, thanks a lot...grunt work is annoying so I can imagine this wasn't fun.

parlough:
@Zidane That's completely understandable, and this just went along with my traversal of Sponge code while learning it so it wasn't too bad :)

There are some cleaning and unifying changes I'd still be willing to make, so I'll make sure to split them up next time.

Issue #1322, opened 2017-04-25
When giving Levels directly (with /xp <level>L) player.get(Keys.TOTAL_EXPERIENCE) isn't updated


BrainStone:
When players recieve levels directly via the `/xp <level>L [player]` command the value of `player.get(Keys.TOTAL_EXPERIENCE)` remains unchanged. However the value is correct when reciving XP orbs or using the `/xp <experience> [player]` command.

**I am currently running...**

- SpongeVanilla version: 1.11.2-6.0.0-BETA-253, 1.11.2-6.0.0-BETA-252 and 1.11.2-6.0.0-BETA-243
- Java version: 1.8.0_121
- Operating System:

 - [x] Linux (Debian)
 - [ ] macOS
 - [x] Windows

**What is the expected result?**
That the value of total experience is increased when giving levels directly.

**What is the current result?**
The value remains unchanged.

See also: SpongePowered/SpongeVanilla#312

BrainStone:
Also I noticed that when enchanting or repairing the value doesn't decrease.

ryantheleach:
Does receiving experience by an orb, after gaining it with the command, give you the correct total?

BrainStone:
@ryantheleach no it does not.

stephan-gh:
When I tried to reproduce @BrainStone's bug report on SpongeVanilla I noticed that the total experience and the experience levels are tracked separately (and also saved separately) in the internal Minecraft server. I'm not entirely sure in which cases it converts between these values (I haven't checked yet).

BrainStone:
Any updates so far?

dualspiral:
I know what the issue is, I'm just unsure as to whether something should be done about it, and if so, whether it should be optional, because it's fundamentally a Vanilla issue - if indeed, it is an issue (and I'll explain my thinking in a moment).

The problem is, as stated, that the Minecraft server tracks total experience and levels (and the current fraction of experience between two levels) separately. When changing a player's total experience via the `/xp` command, Minecraft also ensures that the experience levels are updated. However, this does not happen the other way around, if the _levels_ are updated, the xp is _not_. This doesn't just happen using the `/xp` command, you also observe this behaviour when you repair an item using an anvil and enchant an item using the enchantment table.

[...]

PR #1358, opened 2017-05-17
Automatically recalculate the total XP after updating the experience levels.

JBYoshi:
This will also recalculate the XP whenever the player data is loaded. This should repair any broken calls from the past.

Fixes SpongePowered/SpongeCommon#1322.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
@@ -46,6 +46,7 @@ import net.minecraft.inventory.InventoryEnderChest;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.item.ItemSword;
+import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.network.play.server.SPacketEntityVelocity;
 import net.minecraft.scoreboard.Scoreboard;
 import net.minecraft.scoreboard.Team;
@@ -84,10 +85,10 @@ import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.Redirect;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
-import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
 import org.spongepowered.common.SpongeImpl;
 import org.spongepowered.common.SpongeImplHooks;
 import org.spongepowered.common.data.manipulator.mutable.entity.SpongeHealthData;
+import org.spongepowered.common.data.processor.common.ExperienceHolderUtils;
 import org.spongepowered.common.entity.EntityUtil;
 import org.spongepowered.common.event.damage.DamageEventHandler;
 import org.spongepowered.common.interfaces.ITargetedLocation;
@@ -131,7 +132,6 @@ public abstract class MixinEntityPlayer extends MixinEntityLivingBase implements
     @Shadow public abstract int xpBarCap();
     @Shadow public abstract float getCooledAttackStrength(float adjustTicks);
     @Shadow public abstract float getAIMoveSpeed();
-    @Shadow public abstract void addExperience(int amount);
     @Shadow public abstract void onCriticalHit(net.minecraft.entity.Entity entityHit);
     @Shadow public abstract void onEnchantmentCritical(net.minecraft.entity.Entity entityHit); // onEnchantmentCritical
     @Shadow public abstract void addExhaustion(float p_71020_1_);
@@ -148,10 +148,13 @@ public abstract class MixinEntityPlayer extends MixinEntityLivingBase implements
     @Shadow public abstract Scoreboard getWorldScoreboard();
     @Shadow public abstract String getName();
     @Shadow @Nullable public abstract Team getTeam();
+    @Shadow public abstract void addExperienceLevel(int levels);
+    @Shadow public abstract void addScore(int scoreIn);
 
     private boolean affectsSpawning = true;
     private UUID collidingEntityUuid = null;
     private Vector3d targetedLocation;
+    private boolean dontRecalculateExperience;
 
     @Inject(method = "<init>(Lnet/minecraft/world/World;Lcom/mojang/authlib/GameProfile;)V", at = @At("RETURN"))
     public void construct(World worldIn, GameProfile gameProfileIn, CallbackInfo ci) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
```diff
@@ -221,6 +222,52 @@ public void setTotalExperience(int exp) {
...
+     * @see #onAddExperienceLevels(int, CallbackInfo)
+     */
+    private void recalculateTotalExperience() {
+        if (!dontRecalculateExperience.get()) {
+            this.experienceTotal = this.xpAtLevel(this.experienceLevel) + Math.round(this.experience * this.xpBarCap());
+        }
+    }
+
+    @Inject(method = "addExperienceLevel", at = @At("RETURN"))
+    public void onAddExperienceLevels(int levels, CallbackInfo ci) {
```
stephan-gh: All your injections can and should be private ;)

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
```diff
@@ -155,9 +153,12 @@
     @Shadow public abstract boolean isPlayerSleeping();
     @Shadow public abstract void wakeUpPlayer(boolean immediately, boolean updateWorldFlag, boolean setSpawn);
 
+    @Shadow public abstract void addExperienceLevel(int levels);
+
     private boolean affectsSpawning = true;
     private UUID collidingEntityUuid = null;
     private Vector3d targetedLocation;
+    private final ThreadLocal<Boolean> dontRecalculateExperience = ThreadLocal.withInitial(() -> false);
```
stephan-gh: Is this ever called anywhere except from the main thread?

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
```diff
@@ -221,6 +222,52 @@ public void setTotalExperience(int exp) {
...
 
+    /**
+     * {@link EntityPlayer#addExperienceLevel(int)} and
+     * {@link EntityPlayer#removeExperienceLevel(int)} don't update the total
+     * experience. This recalculates it for plugins to properly make use of it.
+     *
+     * @see #onAddExperienceLevels(int, CallbackInfo)
+     */
+    private void recalculateTotalExperience() {
+        if (!dontRecalculateExperience.get()) {
```
gabizou: since this is an instance variable, regardless whether it's thread local, you should have it qualified with `this`.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
```diff
@@ -221,6 +222,52 @@ public void setTotalExperience(int exp) {
...
+    /**
+     * {@link EntityPlayer#addExperienceLevel(int)} and
+     * {@link EntityPlayer#removeExperienceLevel(int)} don't update the total
+     * experience. This recalculates it for plugins to properly make use of it.
+     *
+     * @see #onAddExperienceLevels(int, CallbackInfo)
+     */
+    private void recalculateTotalExperience() {
+        if (!this.dontRecalculateExperience) {
+            this.experienceTotal = this.xpAtLevel(this.experienceLevel) + Math.round(this.experience * this.xpBarCap());
```
dualspiral: There is an edge case here where you'll potentially go out of sync with the levels, if your experience level has gone down and `Math.round(this.experience * this.xpBarCap()) == this.xpBarCap()` (so you'd actually have the total experience for the _next_ level), you need to either bump this up to the next level, or remove an experience point - I guess the latter is more appropriate here as you're trying to hit a target level.

You also need to recalculate `this.experience` - as you performed a rounding operation, this value will no longer be correct.

dualspiral (changes requested):

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
JBYoshi: Apparently this calculation gets pretty wonky... I'll start by addressing the third point on recalculating `this.experience`.

### Should `this.experience` be recalculated?

Let's say you were on a level that requires 8 XP to move on to the next level, and you had 6 XP towards it. If you change levels using `/xp`, Minecraft preserves the percentage of XP in your XP bar. So if someone pulled you down to a level that required 4 XP to move on, you would have 3 XP on your way to the next level.

On the other hand, if they first pulled you down to a level that needed 6 XP to move on (in a separate command), you would have 75% of 6 XP. Vanilla Minecraft doesn't care about it; when they pull you down to the level needing 4 XP, you're in exactly the same state as above. On the other hand, when tweaking `this.experience` in the recalculation of `totalExperience`, you would either end up with 2 of 4 XP (if you round down; 75% would go to 4/6, or 67%, and then that would go down to 2/4, or 50%) or 3 of 4 XP (if you round to the nearest whole number; 75% would go to 5/6, or 83%, and then that would back to 75%) - though the latter wouldn't necessarily work out either if there were more steps.

In this case, I left `this.experience` alone because I see `totalExperience` as based off `experience` and `experienceLevel`; if the experience changes, `totalExperience` will properly follow it.

### Round to the nearest value or down?
Vanilla Minecraft would say round down. If you had 5 XP on a level that needed 6 XP to continue, you would be 83% of the way to the next level. If you then went down to a level that needed 2 XP to continue, you would still be 83% of the way - except the next experience point you collect would progress you 33% of the way, pushing you slightly over. If you round to the nearest whole number, you would be pushed on to the next level, which is inconsistent with vanilla.

The only problem? Floating point inaccuracies. In my testing when I had rounding down, I gave myself 100 XP and then ran `/xp 0L <player name>` to simply call `recalculateTotalExperience()`. Then the `totalExperience` was recalculated, but was off slightly due to the floating point error: I ended up with 99 XP. I had supposedly been 9 XP through the level, and it needed 21 XP to continue, but the code actually said I was `8.999997` XP through the level. Rounding down then stole an XP point.

Another option is a threshold. That would work out nicely, except that I've found Minecraft's `addExperience()` method exaggerates the floating point inaccuracies to the point where a single XP point can be smaller on the scale than its inaccuracy.

All in all, I have no idea what would work best, considering that we still need Sponge to work with existing save files that used the inaccurate floating point values (otherwise I could completely rip out the current XP calculation system and replace it with something more accurate). Any ideas would be greatly appreciated.

dualspiral: Oh, I agree, the system is badly done, but we should still try to keep things in sync as much as possible.

With regards to `this.experience`, if you leave it alone after performing your experience calculation, you run the risk of desyncing again further down the road. The chances of this getting to be a large issue isn't high, but because you have a little bit extra on the XP bar, once each XP point represents less than this extra you've got, (which can happen if you just continue to add experience by points alone using the standard Minecraft method) you'll start to level up early again. I've not had time to test this yet, but it certainly is theoretically possible, which is why I think we _should_ be updating it once the `totalExperience` is synced up, if only for this reason alone.

Total experience isn't affected by not resetting `this.experience`, but the leveling mechanics are. 

> Vanilla Minecraft would say round down.

Then that's what we should be doing, realistically. Don't get me wrong, I understand your point and reasoning, but for a quick fix in the case of adding no levels, that is, the experience is already OK for the current level, is that we should just not recalculate the experience. It doesn't completely remove the issue that changing levels might be out by an XP point, sure, but it does remove the obviousness if someone doesn't actually change levels.

Maybe it's worth recalculating `this.experience` when completing `addExperience` to reduce the inaccuracy?

JBYoshi: Here you go then!

JBYoshi:
Note for future readers: There was a bunch of discussion in the "outdated diff" comments of dualspiral's review above about the ideas behind this PR; feel free to look there for more information.

dualspiral:
I was more expecting to just inject code at the end of `addExperience` that does the recalculation. Don't get me wrong, I'm not _personally_ against rewriting the code - we just need to keep our eyes open for if this mechanic changes.

If others are alright with it, it looks good to me.

dualspiral (approved):

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
```diff
@@ -221,6 +224,88 @@ public void setTotalExperience(int exp) {
...
+        recalculateTotalExperience();
+    }
+
+    @Inject(method = "removeExperienceLevel", at = @At("RETURN"))
+    private void onRemoveExperienceLevels(int levels, CallbackInfo ci) {
+        recalculateTotalExperience();
+    }
+
+    /**
+     * @author JBYoshi
```
gabizou: Include a date please.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
```diff
@@ -221,6 +224,88 @@ public void setTotalExperience(int exp) {
...
+     * @reason This makes the experience updating more accurate and disables
+     * the totalExperience recalculation above for this method, which would
+     * otherwise have weird intermediate states.
+     */
+    @Overwrite
+    public void addExperience(int amount)
+    {
```
gabizou: No need to keep MCP formatting, shorten the line length by following sponge code style.

gabizou (commented):
Looks good, just some format styling.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
JBYoshi: Fixed.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
JBYoshi: Fixed.

parlough:
@JBYoshi Mind updating this? It would be good to get in.

JBYoshi:
Can this be merged? I have some other stuff I'm working on that depends on this.

PR #1360, opened 2017-05-19
Fix a thread safety problem which may occur in SpongeEventManager

ustc-zzzz:
There is a thread safety problem in `SpongeEventManager`, and it could cause dead locks.
Here is a typical crash log: <http://paste.ubuntu.com/24604406/>

The reason is that, on one hand, when a thread tries to post a event, it locks the `LoadingCache` provided by Caffeine first to get something, and then the `CacheLoader` callback tries to lock the `lock` field in `SpongeEventManager`, on the other hand, another thread which wants to register/unregister events locks the `lock` field first, and then tries to lock the `LoadingCache` in order to make all the caches invalid.

After browsing through the relevant code, I found that the lock before calling the `invalidateAll` method of `LoadingCache` is unnecessary. The operation of the `LoadingCache` is thread safe and the call to the method can be done after the lock is released.

That's all of what this Pull Request does.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
@@ -168,19 +168,19 @@ public class SpongeEventManager implements EventManager {
     }
 
     private void register(List<RegisteredListener<? extends Event>> handlers) {
-        synchronized (this.lock) {
-            boolean changed = false;
+        boolean changed = false;
 
+        synchronized (this.lock) {
             for (RegisteredListener<?> handler : handlers) {
                 if (this.handlersByEvent.put(handler.getEventClass(), handler)) {
                     changed = true;
                     this.checker.registerListenerFor(handler.getEventClass());
                 }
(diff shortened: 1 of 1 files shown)


ustc-zzzz:
I'm not sure if my PR has something wrong, but the dead lock does exist, and it did not appear since I had patched it up and built the project by my self.
Please tell me what I still need to do with my PR.

stephan-gh:
@ustc-zzzz Thank you for your PR! There is nothing you need to do right now - we need to review and verify your PR before we can merge it.

stephan-gh (approved):

ustc-zzzz:
Well, could I ask for a favor? Could this commit be cherry picked and merged into those stable branches such as `stable-5` and `stable-6`? I consider that it is relatively common that a plugin register several events while the server is running with minecraft worlds loaded, and the bug fix is a little bit important. @Minecrell

stephan-gh:
@ustc-zzzz Done.

Issue #848, opened 2016-07-17
SpongeTeleportHelper + slabs


ryantheleach:
SpongeTeleportHelper probably needs to be improved to be able to deal with slabs/rails/carpets/ other ground coverers.

One way that it can be improved is to check 1 block up first, before it starts to check in a radius around the location, as after gravity takes effect they will end up in the correct position.

What would be better would be some real kind of hitbox checking, check that the block isn't on a configurable blacklist (mod support) and that the player will hit the hitbox and be able to stand there.

Zidane:
@dualspiral @ryantheleach 

Forgot to check if this ticket is still an issue with @dualspiral 's changes. Can one of you investigate and see?

Faithcaio:
Signs are also very unsafe it seems.

stephan-gh:
@dualspiral Can you take a look at this issue? Is it still an issue with your refactor in https://github.com/SpongePowered/SpongeCommon/pull/905?

dualspiral:
Yep, will put this on the pile to look at.

dualspiral:
Looking into this:

* carpets, rails and signs are fine (signs are passable blocks, I was able to warp straight into a sign)
* half blocks and stairs are not OK, and the safe teleportation will likely move the player sideways rather than up - with the routine as it is, it _could_ teleport a player one block up, it just doesn't currently.

I just need to tweak the algorithm a bit, which I'm doing as part of a bigger change that allows plugins to determine their own criteria for whether a block is safe/floor is safe/blockstate is passable, should have a PR for this soon.

Faithcaio:
@dualspiral Are you sure you teleported INTO a sign?
I keep getting placed above or next to it using `Player#setLocationSafely`.
`BlockSign` is `Material.WOOD`
Maybe check for IBlockProperties#causesSuffocation + Lava?

dualspiral:
Yep, I was put into the middle of the sign, rather than beside it. Odd, as you said, that probably shouldn't have worked, probably just fell onto it.

I think I'm going to do as you suggest regardless, I've been wanting to move this away from just material based as much as possible to try and cater for mods more. I just hadn't gotten around to tweaking my kernels yet.

PR #1362, opened 2017-05-20
Rework some of the SpongeTeleportHelper

dualspiral:
[Sponge API](https://github.com/SpongePowered/SpongeAPI/pull/1563) | **Sponge Common** | [Original Issue](https://github.com/SpongePowered/SpongeCommon/issues/848)

This PR implements the API proposed in the API PR, including the various TeleportHelperFilters. This PR also tweaks the algorithm so that a block above a half-block slab should be selected as opposed to a block to the side.

One thing I haven't worked out yet - configurable default blacklist for blocks, in order to support mods. That's next to the todo list, but I wanted to get this bit ready now.

See the API PR for more information on the API side.


src/main/java/org/spongepowered/common/registry/type/world/TeleportHelperKernelRegistryModule.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Optional;
+
+public class TeleportHelperKernelRegistryModule implements AdditionalCatalogRegistryModule<TeleportHelperKernel> {
+
+    @RegisterCatalog(TeleportHelperKernels.class)
+    private final Map<String, TeleportHelperKernel> kernelMap = Maps.newHashMap();
+
+    private final Map<String, TeleportHelperKernel> idMap = Maps.newHashMap();
+
+    @Override public Optional<TeleportHelperKernel> getById(String id) {
```
kashike: Annotation should be on it's own line.
```java
    @Override
    public Optional...
```

src/main/java/org/spongepowered/common/world/teleport/DefaultTeleportHelperKernel.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Set;
+
+public class DefaultTeleportHelperKernel implements TeleportHelperKernel {
+
+    // Materials it is NOT safe to put players on top of.
+    private static final Set<Material> NOT_SAFE_FLOOR = ImmutableSet.of(Material.AIR, Material.CACTUS, Material.FIRE, Material.LAVA);
+
+    // Materials that are passable and not harmful to the player body.
+    private static final Set<Material> SAFE_BODY = ImmutableSet.of(
+        Material.AIR, Material.PLANTS, Material.WATER, Material.REDSTONE_LIGHT, Material.CIRCUITS, Material.SNOW, Material.PORTAL, Material.WEB,
```
kashike: Wrong indentation.

src/main/java/org/spongepowered/common/world/teleport/DefaultTeleportHelperKernel.java
```diff
@@ -0,0 +1,68 @@
...
+            Material.VINE
+    );
+
+    @Override public String getId() {
```
kashike: Annotation should be on it's own line.
```java
    @Override
    public String...
```

kashike (changes requested):

src/main/java/org/spongepowered/common/registry/type/world/TeleportHelperKernelRegistryModule.java
```diff
@@ -0,0 +1,79 @@
...
+import org.spongepowered.common.world.teleport.SurfaceOnlyTeleportHelperFilter;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Optional;
+
+public class TeleportHelperKernelRegistryModule implements AdditionalCatalogRegistryModule<TeleportHelperFilter> {
+
+    @RegisterCatalog(TeleportHelperFilters.class)
+    private final Map<String, TeleportHelperFilter> kernelMap = Maps.newHashMap();
```
kashike: Should be renamed to match the new name of the interface.

src/main/java/org/spongepowered/common/registry/type/world/TeleportHelperKernelRegistryModule.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private final Map<String, TeleportHelperFilter> idMap = Maps.newHashMap();
+
+    @Override
+    public Optional<TeleportHelperFilter> getById(String id) {
+        return Optional.ofNullable(this.idMap.get(id));
```
kashike: `id` should be `toLowerCase(Locale.ENGLISH)`'d

kashike (changes requested):

dualspiral:
I've added support for a configurable blacklist of blocks, so that behaviour of the helper can be altered via the global config.

There are a few things to note with how I've done this:

* I've written the block blacklist in the config module to accept both `BlockType` and `BlockState` ids. My reasoning is that you might want to block `minecraft:stained_glass` from being a target no matter what state it's in, and not have to list all its states (colours in this case).
* To prevent potentially expensive lookup operations each time, I cache the `BlockType`s and `BlockState`s, and invalidate the list when the config is reloaded - [it gets invalidated here](https://github.com/SpongePowered/SpongeCommon/pull/1362/files#diff-6bdb73ca8e3f612a09b73db110192b5cR523).
* The config blacklist works with all Sponge (i.e. this PR) provided filters, but a plugin provided filter doesn't necessarily use the block (it's always available as a separate filter, `TeleportHelperFilters.CONFIG`). However, I'm considering making it a mandatory filter completely backed into the helper, because I feel like a plugin probably shouldn't override a user. Thoughts?

ryantheleach:
In general "I feel like a plugin probably shouldn't override a user." 

I feel like this statement should be used as a default, i.e. user choices in configs in sponge should be applied by default, but have a method to avoid them if it directly conflicts with a plugins own configs / intentions.

e.g. a plugin supplying a filter should supplement the default, unless they explicitly clear it.

In this specific case I'm not sure it matters one way or another (except for consistency), considering it's a safe teleporter.

dualspiral:
So, what I've opted to do is add an option to make the blacklist mandatory in the config (which adds the `CONFIG` filter to the set of filters _always_). 

I have also added a method in the API to use the blacklist (mostly as a convenience function) that adds the blacklist (`CONFIG`) filter automatically regardless of whether the blacklist is forced or not, just to give the blacklist filter more visbility (plugins might be more inclined to use a method that specifies the blacklist, rather than a different filter).

This should give plugins the flexibilty of deciding whether to use this blacklist or not, and users the ability to override this if they feel like they need to.

Zidane:
Merged in https://github.com/SpongePowered/SpongeCommon/commit/5a35b18813d49d583c53a9eb2afcf2e30663eb36

PR #1368, opened 2017-05-24
Don't convert back and forth between Explosion objects and their unpacked data.

JBYoshi:
This PR:

* Inlines `newExplosion()` in `MixinWorld.triggerExplosion()`
* Moves the explosion code in `MixinWorldServer` from `triggerExplosion()` and `newExplosion()` to `triggerExplosionInternal()`
* Corrects a few defaults in `SpongeExplosionBuilder` that badly messed up my testing

Fixes SpongePowered#1367.


src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -1602,68 +1590,11 @@ public SpongeBlockSnapshot createSpongeBlockSnapshot(IBlockState state, IBlockSt
...
-
-        // Sponge Start - end processing
-        this.processingExplosion = false;
-        if (CauseTracker.ENABLED) {
-            CauseTracker.getInstance().completePhase(GeneralPhase.State.EXPLOSION);
-        }
-        // Sponge End
+        // Sponge start - move this behavior to another method to allow precreated explosions to be used
+        // All of the original behavior is in triggerExplosion().
+        triggerExplosion((org.spongepowered.api.world.explosion.Explosion) explosion, Cause.of(NamedCause.source(entityIn == null ? this : entityIn)));
```
gabizou: I'd rather use the CauseTracker to define what `Cause` to use for the explosion source. Since the source passed in as `Entity` is usually an entity ticking, or in some cases a `TileEntity` passing null for explosions. You may also want to verify the explosion optimization (not really optimization, it's an optimization for clients on the server due to the unperformant explosion packets and what they entail) still functions as expected.

gabizou (changes requested):
Some cause tracking details are missing in my opinion.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
JBYoshi: Fixed. Although pretty much the entire PR changed - you may need to re-review it.

src/main/java/org/spongepowered/common/mixin/optimization/MixinWorldServer_Explosion.java
```diff
@@ -48,21 +50,23 @@
 @Mixin(value = WorldServer.class, priority = 1111)
 public abstract class MixinWorldServer_Explosion extends MixinWorld {
```
Zidane, line 44: @gabizou @JBYoshi Any reason why this mixin is on its own and not in MixinWorldServer?

JBYoshi: It looks like it's separate because it's a configurable optimization.

gabizou: @Zidane because it's altering the mechanics of multiple explosions occurring at the same time. This goes about removing the point of sending explosion packets (reducing lag on clients), this was meddled with @amaranth a while back but I haven't verified whether it should be enabled by default now or not.

JBYoshi:
Can this be merged?

src/main/java/org/spongepowered/common/event/tracking/phase/plugin/PluginPhase.java
```diff
@@ -30,10 +30,11 @@
 import net.minecraft.world.WorldServer;
 import org.spongepowered.api.item.inventory.ItemStack;
 import org.spongepowered.common.entity.PlayerTracker;
-import org.spongepowered.common.event.tracking.PhaseTracker;
 import org.spongepowered.common.event.tracking.IPhaseState;
 import org.spongepowered.common.event.tracking.PhaseContext;
+import org.spongepowered.common.event.tracking.PhaseTracker;
 import org.spongepowered.common.event.tracking.phase.TrackingPhase;
+import org.spongepowered.common.event.tracking.phase.general.ExplosionContext;
```
liach: Extra import?

src/main/java/org/spongepowered/common/event/tracking/phase/plugin/PluginPhaseState.java
```diff
@@ -29,7 +29,7 @@
 import org.spongepowered.common.event.tracking.phase.TrackingPhase;
 import org.spongepowered.common.event.tracking.phase.TrackingPhases;
 
-public abstract class PluginPhaseState<P extends PluginPhaseContext<P>> implements IPhaseState<P> {
+public abstract class PluginPhaseState<P extends PhaseContext<P>> implements IPhaseState<P> {
```
liach: Why?

JBYoshi: This was to allow `CustomExplosionState` to use the same context class as `ExplosionState` (of `GeneralPhase`); there's pretty much no difference between the two types.

src/main/java/org/spongepowered/common/event/tracking/phase/plugin/PluginPhase.java
JBYoshi: Whoops :P I will revert that.

JBYoshi: Actually, this was intended. It now uses this class instead of the previous `ExplosionContext` class that was in this package.

JBYoshi:
@liach I found a way to make this work without messing with the `PhaseContext` class hierarchy. The callers now create a context and pass the explosion to it.

ryantheleach:
~qa

@gabizou You offered review advice earlier, so I'm looking to see a final review to see if it's ready, or else give to someone else.

@JBYoshi I'll make a test plugin that tests the original issue tonight.

phit:
@ryantheleach status?

ryantheleach:
I know that I said previously that I was going to make a test plugin, but at this stage I have no idea if I did that or not, I've been somewhat MIA over the last few months.

At present anything that's blocked on me should be considered up for fair game until I can find the time to invest back into Sponge.

ImMorpheus:
Updated (after asking @JBYoshi about it).

Note: `PluginPhase.State.CUSTOM_EXPLOSION` was removed by https://github.com/SpongePowered/SpongeCommon/commit/0e16a5d8ba7678b2c0af56dd4a29e040c65596ab#diff-789d0465f5f0c54bb475b2958e6263d6 so I'm using `GeneralPhase.State.EXPLOSION`.

gabizou (approved):
Looks good to me. Just may need to add explosion event `ShouldFire` checks to reduce potential stack frame creation unnecessarily.

ImMorpheus:
@gabizou I'll add it with other performance improvements once I reopen the explosions part of https://github.com/SpongePowered/SpongeCommon/pull/2031

ImMorpheus:
Conflicts fixed and rebased.

A note (In case some plugins complain about this being a breaking change): the default value of `shouldBreakBlocks` (and `shouldDamageEntities`) is an implementation detail, it's not mentioned in the API. If you're creating an explosion with the ExplosionBuilder without explicitly setting those flags, then it's your fault for relying on the default value.

PR #1372, opened 2017-05-26
Fix issue with aether_legacy inventory

Nitix:
Fix SpongePowered/SpongeForge#1448

It will now check the real size of the player's inventory and add missing slots as regular slot.


gabizou:
What are the ramifications of these changes? How are these slots being represented by sponge to plugins? Are the transactions still being processed?

Nitix:
I totally forgot about this PR but closed as https://github.com/SpongePowered/SpongeCommon/commit/b241c2016cd150fe683d5fccf0db159173f95349 fix it

PR #1383, opened 2017-05-31
Allow reregistration in unregisterPluginListeners().

JBYoshi:
I'm working on a unit testing framework for Sponge, and it would be pretty useful for me to be able to prevent one test from receiving events from another test. However, I unregister the listeners using `unregisterPluginListeners()`, and I get an error when trying to reregister them since the event manager still thinks the objects are registered. This PR fixes that.


stephan-gh:
Is there any reason why the call to remove the object isn't in `unregister(Predicate<...> unregister)` directly? `unregisterListeners` does it manually and now you added it to `unregisterPluginListeners`. It could be just in `unregister` directly if both methods do it.

JBYoshi:
@Minecrell Fixed.

kashike (approved):

stephan-gh (approved):

PR #1384, opened 2017-06-01
Replace a hacky (and now broken) client-side test to allow connecting to external servers.

JBYoshi:
This assumption was broken with 5470f0e613bad359c32d50d3da3f04eec47c9a82. It resulted in an `IllegalStateException` from trying to grab the `Server` below whenever someone tries to connect to an external server. This PR fixes that.


kashike:
Thanks!

PR #1387, opened 2017-06-04
Update Database Dependencies

vectrixdevelops:
After encountering some issues with the current database versions, as well as lacking certain things possible in newer versions, I think they would suite a good ol' update.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -63,10 +63,10 @@ dependencies {
     runtime 'org.apache.logging.log4j:log4j-slf4j-impl:2.0-beta9'
 
     // Database connectors
-    compile 'com.zaxxer:HikariCP:2.5.1'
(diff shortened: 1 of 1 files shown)


kashike:
Does this break compatibility at all?

parlough:
From my look over the changelogs and commits, the changes seem to be mostly bugfixes, better Java 8 compatibility, and other small improvements. No major breaking changes. Best to keep these changes to API 7 just in-case though.

vectrixdevelops:
It should not as far as I have tested and looked through the commits and news of each dependency.

stephan-gh:
Thanks for the PR! There were some new updates since you submitted the PR so I used the chance and updated them again in https://github.com/SpongePowered/SpongeCommon/commit/f1a78588a72517d2803648757fd29d1a3afd9747

PR #1389, opened 2017-06-05
Add cause tracker dump to crash reports

Aaron1011:
This PR adds a cause tracker phase + context dump (as seen in 'Invalid Phase' and 'Runaway Phase' errors) to every crash report. This should aid in debugging problems like CauseTracker-induced stack overflows, where information about the specific block is often not available in the stack trace.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java
@@ -31,6 +31,9 @@ import co.aikar.timings.TimingsManager;
 import com.google.common.collect.ImmutableList;
 import net.minecraft.command.ICommandManager;
 import net.minecraft.command.ICommandSender;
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java
```diff
@@ -791,4 +795,10 @@ public Object onCall(Callable<?> callable) throws Exception {
     private Object onRun(FutureTask<?> task, Logger logger) {
         return SpongeImplHooks.onUtilRunTask(task, logger);
     }
+
+    @Inject(method = "addServerInfoToCrashReport", at = @At("RETURN"), cancellable = true)
+    public void onCrashReport(CrashReport report, CallbackInfoReturnable<CrashReport> cir) {
```
stephan-gh, line 800: You can make this injection method private, there is no need to override it in subclasses.

PR #1393, opened 2017-06-11
[1.10.2] Remove Mixin warning in SignCommandSender.

JBYoshi:
The removed method is implemented in [`MixinLocatedSource`](https://github.com/SpongePowered/SpongeCommon/blob/stable-5/src/main/java/org/spongepowered/common/mixin/core/command/MixinLocatedSource.java#L52).

Fixes SpongePowered/SpongeVanilla#331.


kashike (approved):

gabizou:
The reason why I added it in the first place is to eliminate a bunch of new `Location` objects being created when requesting the `World` object. This is staying as is.

gabizou:
If you want, you can implement the change a different way, but this is a very fast short circuit method that eliminates something that should be as fast as a field access.

gabizou (changes requested):
There's a better way than removing the very fast short circuit.

stephan-gh:
In my opinion, the implementation in `MixinLocatedSource` is kind of "meh", checking `instanceof` on the current object is a pretty bad way to write code. `MixinLocatedSource` could have a generic implementation (so basically the one that just returns the main world) that is then replaced by specific Mixins for the entity and tile entity classes. The warning can be avoided by setting a higher priority for these mixins.

JBYoshi:
I realized one other point with this: `Entity` and `TileEntity` both implement `LocatedSource`, so the original method would override simpler, faster methods. I ended up disbanding `MixinLocatedSource` and reimplementing the remaining methods in `MixinCommandExecuteAtSender` and `MixinSignCommandSender` to delegate to their respective fields.

JBYoshi:
Merged in https://github.com/SpongePowered/SpongeCommon/commit/8d5a244903092003982004b1c07d2b915861db25.

PR #1401, opened 2017-06-18
Offline User Inventory + Equipable

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1624) | *SpongeCommon*

Add support for User Inventories
Also some missing parts of equipable.

### DIFF
--- a/src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
+++ b/src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
@@ -29,6 +29,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.mojang.authlib.GameProfile;
+import net.minecraft.inventory.EntityEquipmentSlot;
 import net.minecraft.nbt.CompressedStreamTools;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.nbt.NBTTagList;
@@ -38,6 +39,7 @@ import org.spongepowered.api.data.DataHolder;
 import org.spongepowered.api.data.DataSerializable;
 import org.spongepowered.api.data.Queries;
 import org.spongepowered.api.data.type.HandType;
+import org.spongepowered.api.data.type.HandTypes;
 import org.spongepowered.api.entity.ArmorEquipable;
 import org.spongepowered.api.entity.Tamer;
 import org.spongepowered.api.entity.living.player.Player;
@@ -45,12 +47,15 @@ import org.spongepowered.api.entity.living.player.User;
 import org.spongepowered.api.item.inventory.Carrier;
 import org.spongepowered.api.item.inventory.ItemStack;
 import org.spongepowered.api.item.inventory.equipment.EquipmentType;
+import org.spongepowered.api.item.inventory.equipment.EquipmentTypes;
 import org.spongepowered.api.item.inventory.type.CarriedInventory;
 import org.spongepowered.api.util.RespawnLocation;
 import org.spongepowered.common.SpongeImpl;
 import org.spongepowered.common.data.nbt.CustomDataNbtUtil;
+import org.spongepowered.common.data.type.SpongeEquipmentType;
 import org.spongepowered.common.data.util.DataQueries;
 import org.spongepowered.common.data.util.NbtDataUtil;
+import org.spongepowered.common.item.inventory.util.ItemStackUtil;
 import org.spongepowered.common.world.WorldManager;
 
 import java.io.File;
@@ -62,6 +67,8 @@ import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
+import java.util.function.Consumer;
+import java.util.function.Function;
 
 import javax.annotation.Nullable;
 
@@ -84,6 +91,9 @@ public class SpongeUser implements ArmorEquipable, Tamer, DataSerializable, Carr
 
     private final Map<UUID, RespawnLocation> spawnLocations = Maps.newHashMap();
 
+    private SpongeUserInventory inventory; // lazy load when accessing inventory
+    private NBTTagCompound nbt;
+
     public SpongeUser(GameProfile profile) {
         this.profile = profile;
     }
@@ -94,7 +104,9 @@ public class SpongeUser implements ArmorEquipable, Tamer, DataSerializable, Carr
 
     public void readFromNbt(NBTTagCompound compound) {
         this.reset();
+        this.nbt = compound;
         // See EntityPlayer#readEntityFromNBT
+
         final NBTTagCompound spongeCompound = compound.getCompoundTag(NbtDataUtil.FORGE_DATA).getCompoundTag(NbtDataUtil.SPONGE_DATA);
         CustomDataNbtUtil.readCustomData(spongeCompound, ((DataHolder) this));
         if (!spongeCompound.hasNoTags()) {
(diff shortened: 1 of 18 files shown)


src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
```diff
@@ -95,6 +97,10 @@ private void reset() {
     public void readFromNbt(NBTTagCompound compound) {
         this.reset();
         // See EntityPlayer#readEntityFromNBT
+
+        NBTTagList nbttaglist = compound.getTagList("Inventory", 10);
```
gabizou: s/"Inventory"/`NbtDataUtil.INVENTORY` or something thereof.

src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
```diff
@@ -126,6 +132,8 @@ public void readFromNbt(NBTTagCompound compound) {
 
     public void writeToNbt(NBTTagCompound compound) {
 
+        compound.setTag("Inventory", this.inventory.writeToNBT(new NBTTagList()));
```
gabizou: here as well.

gabizou (changes requested):
string constants for NBT should be in the util class.

src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
kashike: Can give this parameter a nicer name, too.

kashike (changes requested):

src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
```diff
@@ -316,4 +313,82 @@ public void save() {
...
+            for (EntityEquipmentSlot slot : slots) {
+                this.setItemStackToSlot(slot, ItemStackUtil.toNative(item));
+                // TODO check canequip
+                return;
+            }
+        }
+    }
+
+    private net.minecraft.item.ItemStack getItemStackFromSlot(EntityEquipmentSlot slotIn)
+    {
```
gabizou: why the weird formatting?

src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
```diff
@@ -316,4 +313,82 @@ public void save() {
...
+        {
+            return this.inventory.getCurrentItem();
+        }
+        else if (slotIn == EntityEquipmentSlot.OFFHAND)
+        {
+            return this.inventory.offHandInventory.get(0);
+        }
+        else
+        {
+            return slotIn.getSlotType() == EntityEquipmentSlot.Type.ARMOR ? this.inventory.armorInventory.get(slotIn.getIndex()) : net.minecraft.item.ItemStack.EMPTY;
```
gabizou: You could multiline this ternary to avoid issues with figuring out where an error occurs in the future (if there are any that occur in the future)

src/main/java/org/spongepowered/common/entity/player/SpongeUserInventory.java
```diff
@@ -0,0 +1,391 @@
...
+    public final NonNullList<ItemStack> offHandInventory = NonNullList.<ItemStack>withSize(1, ItemStack.EMPTY);
+    public final List<NonNullList<ItemStack>> allInventories;
+    /** The index of the currently held item (0-8). */
+    public int currentItem;
+    /** The player whose inventory this is. */
+    public SpongeUser player;
+    private boolean dirty = false;
+
+    public SpongeUserInventory(SpongeUser playerIn)
+    {
```
gabizou: awkward formatting again?

src/main/java/org/spongepowered/common/entity/player/SpongeUserInventory.java
```diff
@@ -0,0 +1,391 @@
...
+import java.util.List;
+
+public class SpongeUserInventory implements IInventory {
+
+    // sourced from InventoryPlayer
+
+    /** An array of 36 item stacks indicating the main player inventory (including the visible bar). */
+    public final NonNullList<ItemStack> mainInventory = NonNullList.<ItemStack>withSize(36, ItemStack.EMPTY);
+    /** An array of 4 item stacks containing the currently worn armor pieces. */
+    public final NonNullList<ItemStack> armorInventory = NonNullList.<ItemStack>withSize(4, ItemStack.EMPTY);
```
gabizou: Why are these public? Sure it might be more convenient with field accesses, but we generally try to use methods for accessors to have better future updatability and some control.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinArmorEquipable.java
```diff
@@ -35,86 +35,75 @@
...
 
-    public Optional<ItemStack> equipable$getHelmet() {
-        @Nullable final net.minecraft.item.ItemStack itemStack = this.getItemStackFromSlot(EntityEquipmentSlot.HEAD);
-        return Optional.ofNullable(itemStack.isEmpty() ? null : (ItemStack) itemStack.copy());
+    @Override
+    public boolean canEquip(EquipmentType type) {
+        return true;
     }
 
-    public void equipable$setHelmet(ItemStack helmet) {
```
gabizou, line 59: Why are these implementations being removed? Wouldn't it make sense to keep them around for slightly faster implementation?

src/main/java/org/spongepowered/common/registry/type/item/EquipmentTypeRegistryModule.java
```diff
@@ -60,12 +62,17 @@ public void registerDefaults() {
...
         // TODO - evaluate whether the id's are used for any other purpose than the catalog id type to look up
-        final SpongeEquipmentTypeWorn head = new SpongeEquipmentTypeWorn("head", 0);
+        final SpongeWornEquipmentType head = new SpongeWornEquipmentType("head", EntityEquipmentSlot.HEAD);
         this.equipmentTypeMap.put("headwear", head);
         this.equipmentTypeMap.put("head", head);
-        this.equipmentTypeMap.put("chestplate", new SpongeEquipmentTypeWorn("chestplate", 1));
-        this.equipmentTypeMap.put("leggings", new SpongeEquipmentTypeWorn("leggings", 2));
-        this.equipmentTypeMap.put("boots", new SpongeEquipmentTypeWorn("boots", 3));
-        this.equipmentTypeMap.put("worn", new SpongeEquipmentTypeWorn("worn", 0));
+        this.equipmentTypeMap.put("chestplate", new SpongeWornEquipmentType("chestplate", EntityEquipmentSlot.CHEST));
```
gabizou: At this point, why not make these registrations method calls instead of repeating the same field access over and over again with the same strings being used?

testplugins/src/main/java/org/spongepowered/test/OfflineInventoryTest.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+@Plugin(id = "offlineinventorytest", name = "Offline Inventory Test", description = "A plugin to test offline inventories")
+public class OfflineInventoryTest {
+
+    @Inject private Logger logger;
+
+    @Listener
+    public void onDisconnect(ClientConnectionEvent.Disconnect event, @Root Player player) {
+        UUID uuid = player.getUniqueId();
+        Sponge.getScheduler().createTaskBuilder().delayTicks(20).execute(() -> this.run(uuid)).submit(this);
```
gabizou, line 56: Can this be configured just in case we don't want to see this while testing for a bit? (I've had some qualms about the "This Inventory is brought to you by Minecraft" sort of thing popping up when testing Causes branch).

gabizou (changes requested):
Some minor formatting and other minor changes requested for cleaner overall readability.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinArmorEquipable.java
Faithcaio: I default implemented them in the API with equip(type, item)

src/main/java/org/spongepowered/common/entity/player/SpongeUserInventory.java
Faithcaio: I basicly just copied the vanilla InventoryPlayer and adjusted it to work with User.

testplugins/src/main/java/org/spongepowered/test/OfflineInventoryTest.java
Zidane: @gabizou 

That is the issue with the test plugins ;). The better fix would be to figure out how to disable their sourceset without the IDEs bricking themselves.

@Minecrell ^

stephan-gh: I wanted to make the test plugins opt-in originally, but @gabizou wanted them to be always enabled.

Zidane: @Minecrell 

If you can make not including them in gradle import not break IDEs then it would be resolved.

Faithcaio: so should I change something here?

Issue #1402, opened 2017-06-18
Inventory Event Implementation Progress


Faithcaio:
```
<Zidane> Faithcaio, we should focus more on making sure the inventory events are being fired
<Zidane> and the lookup methods are all working
<Zidane> (So we can put Inventory behind us for the most part, lol!)_
```
`? extends TargetInventoryEvent`
 - [x] org.spongepowered.api.event.item.inventory.TargetInventoryEvent
 - [x] org.spongepowered.api.event.entity.ChangeEntityEquipmentEvent
 - [x] org.spongepowered.api.event.entity.ChangeEntityEquipmentEvent.TargetLiving
 - [x] org.spongepowered.api.event.entity.ChangeEntityEquipmentEvent.TargetHumanoid
 - [x] org.spongepowered.api.event.entity.ChangeEntityEquipmentEvent.TargetPlayer
 - [x] org.spongepowered.api.event.item.inventory.ChangeInventoryEvent.Equipment
   equipmentslot-clicks/equip item in hand/equip by dispenser(this last one fires item-pickup)
 - [x] org.spongepowered.api.event.item.inventory.ChangeInventoryEvent
 - [x] org.spongepowered.api.event.item.inventory.ClickInventoryEvent
 - [x] org.spongepowered.api.event.item.inventory.ClickInventoryEvent.Creative
   maybe detect if its not in the creative menu?
 - [x] org.spongepowered.api.event.item.inventory.ClickInventoryEvent.Drag
 - [x] org.spongepowered.api.event.item.inventory.ClickInventoryEvent.Drag.Primary
 - [x] org.spongepowered.api.event.item.inventory.ClickInventoryEvent.Drag.Secondary
 - [x] org.spongepowered.api.event.item.inventory.ClickInventoryEvent.Drop
 - [x] org.spongepowered.api.event.item.inventory.ClickInventoryEvent.Drop.Full
 - [x] org.spongepowered.api.event.item.inventory.ClickInventoryEvent.Drop.Single
    drop Full/Single with Q/ctrl+Q is not fired when inventory is closed (active hotbar item)
 - [x] org.spongepowered.api.event.item.inventory.ClickInventoryEvent.Drop.Outside
   Outside events dont fire when in creative instead only .Creative (actually we can't know better)
 - [x] org.spongepowered.api.event.item.inventory.ClickInventoryEvent.Drop.Primary in Outside
[...]

PR #1407, opened 2017-06-22
fire ClickInventoryEvent.Drop.Single and Full

Faithcaio:
https://github.com/SpongePowered/SpongeCommon/issues/1402

Fires ClickInventoryEvent.Drop.Single and Full when dropping items without a container open.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
@@ -305,7 +305,7 @@ public abstract class MixinEntityPlayer extends MixinEntityLivingBase implements
      */
     @Overwrite
     @Nullable
-    public EntityItem dropItem(@Nullable ItemStack itemStackIn, boolean unused) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
```diff
@@ -322,7 +322,7 @@ public EntityItem dropItem(@Nullable ItemStack itemStackIn, boolean unused) {
     @Nullable
     @Overwrite
     public EntityItem dropItem(@Nullable ItemStack droppedItem, boolean dropAround, boolean traceItem) {
-        if (droppedItem == null || droppedItem.getCount() == 0 || droppedItem.getItem() == null) {
+        if (droppedItem.isEmpty()) {
```
gabizou: This can still be null, can it not?

Faithcaio: Should not be possible. but I'll check

Faithcaio: I'll remove the Nullable

Issue #1406, opened 2017-06-22
Keys.REPRESENTED_PLAYER not working with items


pie-flavor:
`Keys.REPRESENTED_PLAYER` on a skull item does not work.

Tested via the following code, executed on SpongeVanilla 1.12-7.0.0-BETA-288:
```java
Player player = Sponge.getServer().getPlayer("pie_flavor").get();
ItemStack stack = ItemStack.builder()
        .itemType(ItemTypes.SKULL)
        .keyValue(Keys.SKULL_TYPE, SkullTypes.PLAYER)
        .keyValue(Keys.REPRESENTED_PLAYER, player.getProfile())
        .build();
player.getInventory().offer(stack);
```
The item shows up as a bare Steve head, no extra data. Examination of the internals returned this (and nothing else):
```
{id:"minecraft:skull",Count:1b,Damage:3s}
```

Zidane:
@Aaron1011 Can you fix this up since you did skulls?

parlough:
@Zidane Was this not properly solved with #1409?

Zidane:
Oh, completely spaced on this.

PR #1409, opened 2017-06-25
Keys.SKULL_TYPE not working

rexbut:
@Aaron1011 already fixed the problem in commit https://github.com/SpongePowered/SpongeCommon/commit/8970a6a88b44ebba77d5752d8e4f1f0fde98e781 but commit https://github.com/SpongePowered/SpongeCommon/commit/9d594880e16905f61784f37d0bc890da05981caf canceled the fix.

This also solves problem https://github.com/SpongePowered/SpongeCommon/issues/1406

### DIFF
--- a/src/main/java/org/spongepowered/common/data/processor/common/SkullUtils.java
+++ b/src/main/java/org/spongepowered/common/data/processor/common/SkullUtils.java
@@ -24,7 +24,6 @@
  */
 package org.spongepowered.common.data.processor.common;
 
-import com.google.common.collect.Iterables;
 import net.minecraft.init.Items;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
@@ -36,6 +35,7 @@ import org.spongepowered.api.data.type.SkullTypes;
 import org.spongepowered.api.profile.GameProfile;
 import org.spongepowered.common.SpongeImpl;
 import org.spongepowered.common.data.manipulator.mutable.SpongeRepresentedPlayerData;
+import org.spongepowered.common.data.type.SpongeSkullType;
 import org.spongepowered.common.data.util.NbtDataUtil;
 import org.spongepowered.common.interfaces.block.tile.IMixinTileEntitySkull;
 
@@ -57,7 +57,9 @@ public class SkullUtils {
     }
 
     public static SkullType getSkullType(int skullType) {
-        return Iterables.get(SpongeImpl.getRegistry().getAllOf(SkullType.class), skullType);
+        return SpongeImpl.getRegistry().getAllOf(SkullType.class).stream()
+                .filter(type -> type instanceof SpongeSkullType && ((SpongeSkullType) type).getByteId() == skullType)
+                .findAny().orElse(DEFAULT_TYPE);
     }
 
     public static boolean isValidItemStack(Object container) {
--- a/src/main/java/org/spongepowered/common/data/processor/common/SkullUtils.java
+++ b/src/main/java/org/spongepowered/common/data/processor/common/SkullUtils.java
@@ -57,9 +57,12 @@ public class SkullUtils {
     }
 
     public static SkullType getSkullType(int skullType) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/data/processor/common/SkullUtils.java
```diff
@@ -57,7 +57,9 @@ public static boolean supportsObject(Object object) {
     }
 
     public static SkullType getSkullType(int skullType) {
-        return Iterables.get(SpongeImpl.getRegistry().getAllOf(SkullType.class), skullType);
+        return SpongeImpl.getRegistry().getAllOf(SkullType.class).stream()
```
gabizou: I almost want to say that I'd rather a traditional for loop to a stream, since streams are not quite yet optimized. Given the potential requirement of this being considered a "hot" method in some cases, it's best to be fast than to use fancy new API's, at least in the current implementation with streams.

gabizou (changes requested):

ryantheleach:
This doesn't seem to solve #1406 (in my dev environment, may be an error with fetching skulls from mojang though)

Nor did the preexisting code break skulls?

https://gist.github.com/ryantheleach/422e144331286524f3b746a9ec6ad4ef

What testing procedures did you use @rexbut ?

ryantheleach:
**Edit: Ignore this, it looks good. my debugging was only getting called for the defaults for the other skulls.**

> This fixes #1406
> 
> But it didn't seem to have any difference for the other skull types, vs the previous code.
> 
> https://gist.github.com/ryantheleach/422e144331286524f3b746a9ec6ad4ef
> 
> What testing procedures did you use @rexbut ?
> 
> So both the original and the new version, whilst they work, still seem incorrect.
> 
> I would have been expecting the id's to be fixed.

ryantheleach:
CI is failing due to bleeding not being updated in rexbut/SpongeAPI

Code works well against the test plugin I made, but the test plugin might need style tweaks? I kind of rushed it a little near the end.

parlough (approved):

testplugins/src/main/java/org/spongepowered/test/SkullTest.java
```diff
@@ -0,0 +1,108 @@
...
+            .extendedDescription(Text.of("commands:\n", "Give you a skull"))
+            .arguments(playerOrSource(PLAYER), GenericArguments.integer(SKULL))
+            .executor(SkullTest::giveSkull)
+            .build();
+    }
+
+    private static CommandResult giveSkull(CommandSource commandSource, CommandContext commandContext) {
+        Player target = (Player) commandContext.getOne(PLAYER).get();
+        int skulltype = (int) commandContext.getOne(SKULL).get();
+        SkullType[] skulls = {
```
gabizou: Why are you using integers? Why not just use the actual catalog id that we provide, and the command would be able to parse correctly? It would avoid the issue of figuring out invalid integers.

gabizou (changes requested):
Get rid of these magic numbers...

testplugins/src/main/java/org/spongepowered/test/SkullTest.java
ryantheleach: It went through a few iterations, I was also wanting an additional option, (that could have been satisfied by a sub-command ) to spawn a head from an unknown game profile.

The false reasoning (since edits happened) behind using int's is that I was originally going to hook straight into Skull Utils to test, then decided that I'd use the SkullTypes as registered.

The reason for less validation, was if it had of hooked into SkullUtils direct, it would have been extensible for future testing, should any skulls be introduced that wern't in the registry.

But that also ended up void, as SkullUtils didn't get edited to use internal values and not the registry.

I didn't like the potential of having wrong ints in the registry, end up working somehow, because it happened to be wrong twice.

Basically, I wrote the tests first, didn't revisit, I'll change it in 8 hours if rexbut doesn't beat me to it.

ryantheleach:
@gabizou Is this good to go now?

parlough (approved):

PR #1411, opened 2017-06-27
Add TestPlugin suite for in-development testing

gabizou:
**SpongeCommon** | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1606) | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/334)
This adds a unique TestPlugins project to SpongeCommon that would allow for implementation and API to be tested in a plugin-like environment.

What this does:
- Adds a "in-dev" plugin that is exposed in development environments that can be used without manually adding plugins back and forth to the module/project dependencies
- Avoids the test plugin output to be included in shadow jars for production jars (source jars and javadoc jars included)
- Allows for future unit testing integration with pre-defined listeners


testplugins/src/main/java/org/spongepowered/test/TestPlugin.java
```diff
@@ -0,0 +1,50 @@
...
+import org.spongepowered.api.plugin.Plugin;
+
+@Plugin(
+    id = "testplugin",
+    name = "TestPlugin",
+    version = "0.0.1-SNAPSHOT",
+    url = "spongepowered.org",
+    authors = "SpongePowered"
+)
+public class TestPlugin {
```
kashike: Perhaps it would be better to put each test plugin in their own class - `ScaledHealthTest`, in this case. Some can get quite large when dealing with complex functionality.

gabizou: Would need to test with both Forge and Vanilla environments to assure that neither will produce an invalid plugin container error with multiple "plugins" in the same "container" so to speak.

gradle/implementation.gradle
```diff
@@ -12,10 +13,12 @@ ext.apiSuffix = (api.version - '-SNAPSHOT') + '-BETA'
 dependencies {
     compile common
     compile common.sourceSets.java6.output
+//    compile testPlugins
```
kashike: remove

gradle/implementation.gradle
```diff
@@ -12,10 +13,12 @@ ext.apiSuffix = (api.version - '-SNAPSHOT') + '-BETA'
...
 }
 
 sourceSets {
     java6.compileClasspath += main.compileClasspath + main.output
+//    testPlugins.compileClasspath += api.compileClasspath + api.output
```
kashike: remove

kashike (changes requested):

testplugins/gradle.properties
```diff
@@ -0,0 +1 @@
+include ':SpongeCommon'
```
stephan-gh: This doesn't do anything, `gradle.properties` is for key value pairs of properties exposed to the Gradle project. Includes must be in `settings.gradle` but is it really needed?

stephan-gh (changes requested):

testplugins/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Sun Jan 08 11:37:06 CET 2017
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-bin.zip
```
stephan-gh: No need for a separate wrapper for the subproject

stephan-gh (changes requested):

gabizou:
@Minecrell @kashike looks good?

kashike (approved):

PR #1412, opened 2017-06-30
Update command system to 1.12.

JBYoshi:
Fixes SpongePowered#1390.


src/main/java/org/spongepowered/common/command/SpongeProxySource.java
```diff
@@ -35,133 +36,151 @@
...
     }
 
     @Override
-    public String getIdentifier() {
-        return getOriginalSource().getIdentifier();
+    public MessageChannel getMessageChannel() {
+        return this.channel;
+    }
+
+    @Override public void setMessageChannel(MessageChannel channel) {
```
stephan-gh: Add a new line after the `@Override` please :)

src/main/java/org/spongepowered/common/mixin/core/command/MixinFunctionObject.java
```diff
@@ -0,0 +1,43 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+import java.util.Map;
+
+@Mixin(FunctionObject.class)
+public abstract class MixinFunctionObject {
+
+    @Redirect(method = "create", at = @At(value = "INVOKE", target = "Ljava/util/Map;containsKey(Ljava/lang/Object;)Z"))
```
stephan-gh: You need `remap = false` on the `@At` annotation to make sure Mixin doesn't complain when compiling

stephan-gh (changes requested):

src/main/java/org/spongepowered/common/command/SpongeProxySource.java
JBYoshi: Fixed.

src/main/java/org/spongepowered/common/mixin/core/command/MixinFunctionObject.java
JBYoshi: Fixed.

PR #1419, opened 2017-07-07
Use ext.refMap to fix compatibility with later Gradle versions

Grinch:
This does not break compatibility with the wrapper version.


kashike (approved):

kashike:
Thanks!

PR #1429, opened 2017-07-09
Add Test Plugin : Score Text Test

ryantheleach:
@pie-flavor tipped me off that ScoreTexts might not be working as intended, so I created a quick test plugin for use for fixing them.

image

### DIFF
--- a/testplugins/src/main/java/org/spongepowered/test/ScoreTextTest.java
+++ b/testplugins/src/main/java/org/spongepowered/test/ScoreTextTest.java
@@ -1,0 +1,72 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.test;
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.command.CommandException;
+import org.spongepowered.api.command.CommandResult;
+import org.spongepowered.api.command.CommandSource;
+import org.spongepowered.api.command.args.CommandContext;
+import org.spongepowered.api.command.spec.CommandSpec;
+import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.api.event.Listener;
+import org.spongepowered.api.event.game.state.GameStartingServerEvent;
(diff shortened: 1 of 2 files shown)


testplugins/src/main/java/org/spongepowered/test/ScoreTextTest.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+        final Player player = (Player) src;
+        Sponge.getCommandManager().process(src, "scoreboard objectives add test dummy test");
+        Sponge.getCommandManager().process(src, "scoreboard players add @p test 1");
+
+        final Scoreboard sb = Sponge.getServer().getServerScoreboard().get();
+        final Objective test = sb.getObjective("test").get();
+        final Score score = test.getScore(player.getTeamRepresentation()).get();
+
+        src.sendMessage(Text.of(TextColors.GREEN, Text.of("1.")));
```
gabizou: Constants instead of creating additional text objects every time?

testplugins/src/main/java/org/spongepowered/test/ScoreTextTest.java
```diff
@@ -0,0 +1,98 @@
...
+                        .executor(this::showTest)
+                        .description(Text.of("Creates a dummy objective 'test', adds 1 to it, then shows you the ScoreText"))
+                        .build(),
+                "scoretext"
+        );
+    }
+
+    private CommandResult showTest(CommandSource src, CommandContext ctx) throws CommandException {
+        if (!(src instanceof Player)) {
+            throw new CommandException(Text.of("Must be ran by player"));
```
gabizou: Should we be throwing an exception? or returning `CommandResult.emptyResult()`? I have been told differently so many times I don't recall anymore... @kashike @dualspiral ?

gabizou (commented):
Just double check the return result. Constants are not required. but might be nicer with style.

testplugins/src/main/java/org/spongepowered/test/ScoreTextTest.java
ryantheleach: Throwing a CommandException results in an error message being displayed to the user.

Returning CommandResult.emptyResult() is suitable for commands that 'succeeded' with no targets, i.e. No-op.

dualspiral: Note that, for a basic command (not a sub command) both throwing a `CommandException` and returning an empty result will end up returning an empty result, the difference is exactly what @ryantheleach said. There is another important difference if an exception is thrown and this is a child command in a `CommandSpec`, it prompts the system to try to run the parent command instead - returning an empty result would not - and this can be a useful guide for whether to throw an exception or not.

So, just as the command elements would throw an exception if the argument string was not in an expected state, the command source would not in an expected state - I'd argue throwing an exception makes sense.

Zidane:
@ryantheleach Can you get the concerns above taken care of and rebase to one commit?

parlough:
@ryantheleach This one's good to merge? It's always behind a command, so should be fine.

ryantheleach:
@Meronat to the best of my knowledge it's ready to merge. 
  
The command result always being success was intended, as my train of thought is that the failure case for a test command, that's handled by the command and still gives a readable (failed) result, is a success.

ImMorpheus:
@ryantheleach 
Is this still relevant after dualspiral selector PR ?

See https://github.com/SpongePowered/SpongeCommon/blob/dac83ee0bfeb62847def30967d71347885948468/testplugins/src/main/java/org/spongepowered/test/SelectorTest.java

ryantheleach:
@ImMorpheus The issue related to this test plugin has been fixed, but there was a separate issue with Score texts specifically. I'm ambivalent on whether it should be merged or not.

As per my last comment, it's been reviewed and done, and ready to merge. Just to my knowledge, I don't have authority(as opposed to ability) to merge PR's, otherwise I would have done so already.

testplugins/src/main/java/org/spongepowered/test/SelectorTextTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.spongepowered.api.scoreboard.Scoreboard;
+import org.spongepowered.api.scoreboard.objective.Objective;
+import org.spongepowered.api.text.LiteralText;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.format.TextColors;
+import org.spongepowered.api.text.selector.Selector;
+
+import javax.inject.Inject;
+
+@Plugin(id = "selectortexttest", name = "Selector Text test", description = "A plugin to test selector texts")
```
ImMorpheus: version = "0.0.0"

ImMorpheus:
```suggestion
@Plugin(id = "selectortexttest", name = "Selector Text test", description = "A plugin to test selector texts", version = "0.0.0")
```

gabizou: Agreed, just add the version to suppress the warning for compiling.

gabizou (approved):
One compiler warning to avoid by adding a version string. Otherwise good to go.

PR #1430, opened 2017-07-09
Replace references to EntitySkeleton.

JBYoshi:
- In MixinArmorEquipable, I simply replaced EntitySkeleton with AbstractSkeleton.
- In MixinRanger, I saw the comment about mixing in to implementors of an interface. Mixin doesn't yet allow that for performance reasons, but it is possible to mix in to an interface directly, so I did that.
- In MixinEntityProjectileSource, I noticed that MixinEntityLivingBase already implemented both of the methods, and MixinEntityProjectileSource had also been removed from the mixin JSON file, so I simply deleted the class.

Fixes SpongePowered#1428.


src/main/java/org/spongepowered/common/mixin/core/entity/MixinRanger.java
```diff
@@ -24,22 +24,12 @@
...
 
-@Mixin({
-        EntityHuman.class,
-        EntitySkeleton.class,
-        EntitySnowman.class,
-        EntityWitch.class,
-        EntityWither.class
-})
+@Mixin(IRangedAttackMob.class)
 // TODO Remove when we can mix into implementors of an interface
```
liach: No need for this now

JBYoshi: Removed.

gabizou (approved):

liach (approved):

ghost:
Please backport this to 1.11.2 too, thanks.

PR #1435, opened 2017-07-13
Fix InjectionPoint support on constructor/method parameters.

Cybermaxke:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1607) | **Common**

I ported it over from https://github.com/LanternPowered/LanternServer, but still needs some testing in the sponge environment.

The spi `InjectionPoint` works fine when you target `Field`s, but once you target `Method` or `Constructor` there is information about which parameter missing in the spi injection point. The current providers that use `InjectionPoint`s also ignore the fact that the target could be a parameter, and the method annotations are used here instead of the parameter ones.
Adding a convenient custom `InjectionPoint` makes it easier to get direct access to the information about the injected field/parameter.
There is currently no way to get the actual field/method, but this could be added if desired. IMO I don't see the need why any injection behavior would be different on a field or method, so there is no point in exposing it.

@kashike


src/main/java/org/spongepowered/common/inject/SpongeInjectionPoint.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override
+    public Annotation[] getDeclaredAnnotations() {
+        return Arrays.copyOf(this.annotations, this.annotations.length);
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper("InjectionPoint")
```
kashike: `MoreObjects.toStringHelper(this)`

kashike (changes requested):

src/main/java/org/spongepowered/common/inject/InjectionPointProvider.java
```diff
@@ -66,20 +73,35 @@ public boolean matches(Binding<?> binding) {
         }
     }
 
+    @Nullable
     private static InjectionPoint findInjectionPoint(List<DependencyAndSource> dependencyChain) {
```
kashike: It's worth noting that [`ProvisionListener$ProvisionInvocation#getDependencyChain()`](https://github.com/google/guice/blob/master/core/src/com/google/inject/spi/ProvisionListener.java#L66-L102) has been deprecated ([not long after `4.1` was released](https://github.com/google/guice/commit/6f185140d88feafeec65039a213f08a0abeff268)) and scheduled for removal in `4.4`, as is [`DependencyAndSource`](https://github.com/google/guice/blob/master/core/src/com/google/inject/spi/DependencyAndSource.java#L31-L33). 

Would you be able/willing to see about making this use the alternative method (as suggested in the javadocs) now, rather than later, to ensure that what we do is still possible down the road? If not, could you add a comment to this class, please?

Cybermaxke: Yes, I will look into the alternative method.

kashike (approved):

kashike:
Thanks.

PR #1437, opened 2017-07-14
Expand and implement the Instrument/Noteblock API.

Cybermaxke:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1608) | **Common**

This PR implements the Instrument API and the expansions added by the API PR.

### DIFF
--- a/src/main/java/org/spongepowered/common/registry/type/data/InstrumentTypeRegistryModule.java
+++ b/src/main/java/org/spongepowered/common/registry/type/data/InstrumentTypeRegistryModule.java
@@ -1,0 +1,75 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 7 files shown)


parlough:
Would be good to add a test plugin for some of these additions :)

parlough (approved):
Tested these additions and they seemed to work perfectly. 

Looks good to me. Cheers! 🍰

src/main/java/org/spongepowered/common/mixin/core/tileentity/MixinTileEntityNote.java
```diff
@@ -54,4 +72,42 @@ public void supplyVanillaManipulators(List<DataManipulator<?, ?>> manipulators)
...
+        super.writeToNbt(compound);
+        if (this.overriddenInstrument != null) {
+            compound.setString(OVERRIDDEN_INSTRUMENT_KEY, this.overriddenInstrument.getId());
+        }
+    }
+
+    @Redirect(method = "triggerNote(Lnet/minecraft/world/World;Lnet/minecraft/util/math/BlockPos;)V",
+            at = @At(value = "INVOKE", target = "Lnet/minecraft/world/World;addBlockEvent("
+                    + "Lnet/minecraft/util/math/BlockPos;Lnet/minecraft/block/Block;II)V"))
+    public void onTriggerNote(World target, BlockPos blockPos, Block block, int instrument, int note) {
```
stephan-gh: No need to have this public (can be private)

stephan-gh (changes requested):

src/main/java/org/spongepowered/common/mixin/core/tileentity/MixinTileEntityNote.java
Cybermaxke: Fixed

src/main/java/org/spongepowered/common/registry/type/data/InstrumentTypeRegistryModule.java
```diff
@@ -0,0 +1,99 @@
...
+import org.spongepowered.api.registry.util.RegisterCatalog;
+import org.spongepowered.common.data.type.SpongeInstrumentType;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Optional;
+
+public class InstrumentTypeRegistryModule implements AlternateCatalogRegistryModule<InstrumentType> {
```
gabizou: Isn't there a better registry module you can extend to remove all the boilerplate?

testplugins/src/main/java/org/spongepowered/test/InstrumentTestPlugin.java
```diff
@@ -0,0 +1,66 @@
...
+import org.spongepowered.api.plugin.Plugin;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.util.Tristate;
+
+@SuppressWarnings("ConstantConditions")
+@Plugin(id = "instrument-test")
+public class InstrumentTestPlugin {
+
+    @Listener
+    public void onUseItem(InteractBlockEvent.Primary.MainHand event, @First Player player) {
```
gabizou: This should be toggle able in my opinion. If the plugin is disabled or enabled, because by default having. a plugin active where it mentions "you've done x", is rather annoying and sometimes startling.

gabizou (changes requested):

testplugins/src/main/java/org/spongepowered/test/InstrumentTestPlugin.java
Cybermaxke: This should also be done in some of the other test plugins, the `InventoryPluginTest` one is very annoying.

gabizou: Agreed.

Issue #1431, opened 2017-07-10
Inventory#query(Object[]) returns an empty inventory every time


pie-flavor:
Code: 
```java
player.getInventory().query(ItemTypes.STICK, ItemStack.of(ItemTypes.BLAZE_ROD, 5))
```
Inventory contains a stick and a stack of five blaze rods.
The returned `Inventory` is empty.

SpongeVanilla 1.11.2-6.1.0-BETA-11

JBYoshi:
A quick note: I decided to have the method ignore the stack size. Let me know if it should be changed.

pie-flavor:
I think there should be two methods, one for each behavior, just like the current ItemStack query methods.

JBYoshi:
@pie-flavor I have updated the behavior for API versions 6 and 7. I haven't changed API 5 because it uses a stack size of -1 to specify that the stack size should be ignored.

Mumfrey:
No, `queryAny` is bad and I can't believe you added this without discussing it.

Mumfrey:
I mean. This is where inconsistency starts, there are so many ways this could have been done nicely and adding an extra, totally redundant method is a bloody stupid way of going about it and makes the whole thing inconsistent.

`query` - queries the things
`queryAny` - queries *any of the things*?   ... waht?

This is half-baked, inconsistent, and needs to be removed ASAP. Fortunately, the half-bakedness continues to the implementation where you **didn't test your build before you pushed the code**, meaning it's broken.

You need to revert these changes.

Mumfrey:
To be clear, the changes to the strategy are fine and I approve. But I'm absolutely bloody furious that you would change the `Inventory` interface in such a careless manner without so much as a discussion.

JBYoshi:
@Mumfrey Reverted.

Mumfrey:
Thankyou. I see that there's another `queryAny` which was added by @Faithcaio previously which I'm not desperately happy about either but it's been there too long to change. We should probably remove it in API 7.

Propose some better combinatorial approaches and we can discuss them.

Faithcaio:
The queryAny I added was to query ItemStacks of **any** size as it was no longer possible to created -1 sized stacks.

Faithcaio:
just leaving this here:
query(Predicate<ItemStack>...)

pie-flavor:
@Mumfrey That was... an explosion, to say the least. How would _you_ recommend we go about adding support for both querying including stack size, and querying ignoring it?

PR #1440, opened 2017-07-15
Implement rewrite of inventory queries.

JBYoshi:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1610) | **SpongeCommon**

When inventories were first added, they had support for querying for sub-inventories based on various attributes about them. One of those attributes was the type of `ItemStack` they contained. However, sometimes, developers want to use the quantity of the `ItemStack`, and sometimes they want to ignore it but keep the data on it. Originally, the way to specify this was through a quantity of `-1`. Then Minecraft dropped support for negative quantities. Sponge left in its wake a `queryAny()` method that ignored the quantity. In a discussion earlier today, Mumfrey decided this system needed to change, and I decided to be the one to write it. This version exposes a `QueryOperation` that tells the query how to treat each argument. It should also be more flexible than the current system in case Minecraft breaks some other behavior in the future.

Closes https://github.com/SpongePowered/SpongeCommon/issues/1431.


src/main/java/org/spongepowered/common/item/inventory/query/SpongeQueryOperationType.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public SpongeQueryOperationType(String id, Function<T, SpongeQueryOperation> newInstance) {
+        super(id);
+        this.newInstance = newInstance;
+    }
+
+    @Override
+    public QueryOperation of(T arg) {
+        checkNotNull(arg);
+        return newInstance.apply(arg);
```
kashike: `this.`

src/main/java/org/spongepowered/common/registry/CommonModuleRegistry.java
```diff
@@ -41,8 +41,51 @@
 import org.spongepowered.api.data.meta.PatternLayer;
 import org.spongepowered.api.data.persistence.DataFormat;
 import org.spongepowered.api.data.persistence.DataTranslator;
-import org.spongepowered.api.data.type.*;
```
kashike: undo

kashike (changes requested):

src/main/java/org/spongepowered/common/registry/CommonModuleRegistry.java
JBYoshi: I went back through all the imports only to find that IntelliJ undid all of my changes... then I pressed the "undo" button and it immediately went back through and "fixed" the imports for me... :rage:

src/main/java/org/spongepowered/common/item/inventory/query/SpongeQueryOperationType.java
JBYoshi: Fixed.

Zidane:
@JBYoshi Is this ready to go?

@Faithcaio @Mumfrey How is this looking?

PR #1444, opened 2017-07-15
Fix saturation property value to match actual saturation gained - Fixes #1418

### LINKED ISSUE #1418: SaturationProperty does not match actual saturation replenished
https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/data/property/store/item/SaturationPropertyStore.java

So the SaturationProperty is all about amount of saturation replenished, but Minecraft seems to use a different value internally when it registers the food items, then when it adds the stats to the player it uses this calculation:

```java
this.foodSaturationLevel = Math.min(this.foodSaturationLevel + (float)foodLevelIn * foodSaturationModifier * 2.0F, (float)this.foodLevel);
```

For what reason this internal value is used, I have no idea, but I feel it would be best to modify the SaturationPropertyStore to add this calculation within there to display what that actual amount of saturation gained will be instead of some internal number that is changed later anyway. These new numbers would then match the wiki values. 

Maybe I am missing a reason this internal number is used, so if so please point that out to me.

Once there's some agreement I can make this PR.

parlough:
I went over this in more detail in https://github.com/SpongePowered/SpongeCommon/issues/1418

But essentially the `SaturationProperty` value is the internal one that Mojang uses for what seems like no apparent reason as they just multiply it by the food value and by 2 when applying it to the player. To allow this property to accurately display the saturation the player will gain from the food, I added this calculation to the saturation property store.

To show this with some actual numbers and food, with the following code: 

```java
@Listener
public void onRightClick(InteractItemEvent event, @First Player player) {
    final Optional<SaturationProperty> saturationProperty = event.getItemStack().getProperty(SaturationProperty.class);
    if (saturationProperty.isPresent()) {
        player.sendMessage(Text.of("Saturation value for " + event.getItemStack().getType().getId() + ": " + saturationProperty.get().getValue()));
    } else {
        player.sendMessage(Text.of("No saturation property."));
    }
}
```

Clicking with the following items before this change returned this values: 

beforesaturation

Now after this change returns these values:

aftersaturation

Which match the saturation gained and what the Minecraft wiki states:

Raw rabbit:

raw_rabbit

Cooked chicken:

cooked_chicken

Raw porkchop:

image

_As a whole, this is still a proposal, but I do believe this change should be made unless I am missing something. This way the actual amount of saturation of a food can be accurately retrieved and other SpongeAPI implementations don't have to use Mojang's strange/unnecessary math in the first place._


parlough:
I didn't really I had this in my bleeding branch. I will fix it :L

PR #1461, opened 2017-07-23
Add configuration option to force the adoption of the Sponge permission handler

kashike:
[**SpongeCommon**](https://github.com/SpongePowered/SpongeCommon/pull/1461) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1660)

### DIFF
--- a/src/main/java/org/spongepowered/common/config/category/PermissionCategory.java
+++ b/src/main/java/org/spongepowered/common/config/category/PermissionCategory.java
@@ -1,0 +1,39 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/config/category/CommandsCategory.java
```diff
@@ -41,6 +41,9 @@
             + "changes on the all worlds.")
     private Map<String, Boolean> multiWorldCommandPatches = new HashMap<>();
 
+    @Setting(value = "force-permission-handler", comment = "Forces the Sponge permission handler to be used rather than another mod's permission handler")
+    private boolean forcePermissionHandler = true;
```
dualspiral: Why is this in the commands category? Permissions have nothing to do with commands.

ryantheleach:
This is in SC solely for the config right? May want to have it explicit in the config file that this is a MC-Forge only setting.

gabizou:
Couldn't this be mixed into by SpongeForge? It wouldn't make sense to have this in SpongeVanilla server configurations. Since the class load by configurate would be super early, we'd still be able to mix in the field in pre-init phases (I've dabbled with mixing into configurate itself before sponge's configurations loaded before).

ImMorpheus (approved):

PR #1465, opened 2017-07-27
Bump Mixin version to 0.7

caseif:
This PR bumps the `stable-5` Mixin version to 0.7 and resolves incompatibilities which arise as a result.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/data/MixinDataHolder.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/data/MixinDataHolder.java
@@ -413,9 +413,4 @@ public abstract class MixinDataHolder implements DataHolder {
         return Collections.emptyList();
     }
 
-    @Override
-    public DataHolder copy() {
-        return this;
-    }
-
 }
--- a/build.gradle
+++ b/build.gradle
@@ -50,7 +50,7 @@ ext.apiVersion = version
 
 dependencies {
     compile api
-    compile('org.spongepowered:mixin:0.6.15-SNAPSHOT') {
+    compile('org.spongepowered:mixin:0.7-SNAPSHOT') {
         exclude module: 'launchwrapper'
     }
 
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinSpongeUser.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinSpongeUser.java
@@ -25,6 +25,7 @@
 package org.spongepowered.common.mixin.core.entity.player;
 
 import com.google.common.base.Objects;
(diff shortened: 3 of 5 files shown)


Zidane:
I'm not going to review this until @Mumfrey lets me know that 0.7 is fine for general use (I'm aware it is released).

Mumfrey:
> I'm not going to review this until @Mumfrey lets me know that 0.7 is fine for general use (I'm aware it is released).

Yeah I'd actually recommend going to 0.7 for general stability and feature improvements.

src/main/java/org/spongepowered/common/mixin/core/data/MixinDataHolder.java
```diff
@@ -413,9 +413,4 @@ public DataContainer toContainer() {
         return Collections.emptyList();
     }
 
-    @Override
-    public DataHolder copy() {
```
gabizou, line 417: I believed this to be fixed , why is it being removed now?

stephan-gh:
@caseif I have backported the update to Mixin 0.6.15 for now to fix https://github.com/SpongePowered/SpongeVanilla/issues/338 which has caused conflicts in your PR. Would you mind fixing them?

I personally think we should do the update to Mixin 0.7 in bleeding first before we backport it to the stable branches eventually.

caseif:
Should be resolved now.

src/main/java/org/spongepowered/common/mixin/core/data/MixinDataHolder.java
caseif: Mixin 0.7 claims that it clashes with lower-level implementations. The override has been migrated to classes further down in the hierarchy where necessary `MixinSpongeUser` and `MixinTileEntity`.

Mumfrey: What are you talking about "down the hierarchy", these bridge methods conflict **in the same class**. The *correct* fix was already applied on `bleeding`, by me. Why are you using a different approach?

ryantheleach: I don't see your changes in a recent commit @Mumfrey I'd appreciate if someone could narrow it down for me.

caseif: Sorry, you're right. I was referring to moving the implementation from `MixinDataHolder` to more specific mixins, but didn't realize there was a better solution. Will fix later today.

Zidane:
Thanks @caseif for the PR but Minecrell went ahead and took care of this among some other house cleaning.

PR #1471, opened 2017-08-01
Fix misplaced phase complete in chunk generation (SpongeForge#1669)

Xfel:
As seen in [SpongeForge#1669,](https://github.com/SpongePowered/SpongeForge/issues/1669) the cause tracking system stumbled into some issues during the generation of ocean monuments. The apparent reason for this is a misplaced mixin injection. I refined the injection point to only target the desired instruction.


kashike (approved):

gabizou:
I'd rather use a Slice like [here](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java#L998) so we can see what is being done.

Mumfrey:
Don't use `ordinal`, you should use a `slice` for this.

PR #1305, opened 2017-04-14
Fix /help 2 for paging, Fix tab completion, add Tips.

ryantheleach:
Requires: https://github.com/SpongePowered/SpongeAPI/pull/1539

Alternative to: https://github.com/SpongePowered/SpongeCommon/pull/1303

Fixes: https://github.com/SpongePowered/SpongeCommon/issues/1242

/page is still preferable to /help as /help reconstructs the entire pagination each time. 

/help 2 however will still continue to work even after the pagination expires.

I've also added the default minecraft help tip to the text, and changed the header to more closely match the default minecraft header.

image

image

image

Only current downside, is that the error handling slightly pollutes the usage text for /help.

image

Also now with tab completion.

image

PR #1475, opened 2017-08-05
Allow sending a specific page of a pagination list and add a test plugin

parlough:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1622) | SpongeCommon

Accomplishes part of https://github.com/SpongePowered/SpongeCommon/pull/1305

Also cleans up some messages and code and makes it so a pagination list defaults to an empty list if no contents are specified.

### DIFF
--- a/src/main/java/org/spongepowered/common/service/pagination/IterablePagination.java
+++ b/src/main/java/org/spongepowered/common/service/pagination/IterablePagination.java
@@ -59,11 +59,15 @@ class IterablePagination extends ActivePagination {
     @Override
     protected Iterable<Text> getLines(int page) throws CommandException {
         if (!this.countIterator.hasNext()) {
-            throw new CommandException(t("Already at end of iterator"));
+            throw new CommandException(t("You're already at the end of the pagination list iterator."));
+        }
+
+        if (page < 1) {
+            throw new CommandException(t("Page %s does not exist!", page));
         }
 
         if (page <= this.lastPage) {
-            throw new CommandException(t("Cannot go backward in an IterablePagination"));
+            throw new CommandException(t("You cannot go to previous pages in an iterable pagination."));
         } else if (page > this.lastPage + 1) {
             getLines(page - 1);
         }
@@ -119,6 +123,6 @@ class IterablePagination extends ActivePagination {
 
     @Override
     public void previousPage() throws CommandException {
-        throw new CommandException(t("Cannot go backwards in a streaming pagination"));
+        throw new CommandException(t("You cannot go to previous pages in an iterable pagination."));
     }
 }
--- a/src/main/java/org/spongepowered/common/service/pagination/ListPagination.java
+++ b/src/main/java/org/spongepowered/common/service/pagination/ListPagination.java
@@ -80,12 +80,13 @@ class ListPagination extends ActivePagination {
 
     @Override
     protected Iterable<Text> getLines(int page) throws CommandException {
-        if (this.pages.size() == 0) {
+        final int size = this.pages.size();
+        if (size == 0) {
             return ImmutableList.of();
         } else if (page < 1) {
             throw new CommandException(t("Page %s does not exist!", page));
-        } else if (page > this.pages.size()) {
-            throw new CommandException(t("Page %s is too high", page));
+        } else if (page > size) {
+            throw new CommandException(t("Page %s is greater than the max of %s!", page, size));
         }
         return this.pages.get(page - 1);
     }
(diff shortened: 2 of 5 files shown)


JBYoshi (approved):

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationList.java
```diff
@@ -100,9 +101,9 @@ public int getLinesPerPage() {
     }
 
     @Override
-    public void sendTo(final MessageReceiver receiver) {
-        checkNotNull(this.contents, "contents");
-        checkNotNull(receiver, "source");
+    public void sendTo(final MessageReceiver receiver, int page) {
+        checkNotNull(this.contents, "The contents of the pagination list cannot be null!");
```
ST-DDT: This is a final field. This should be validated in the constructor.

ST-DDT (changes requested):

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationList.java
parlough: It's validated in the builder build method, so I just removed this check.

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationBuilder.java
```diff
@@ -105,7 +106,9 @@ public SpongePaginationBuilder(SpongePaginationService service) {
 
     @Override
     public PaginationList build() {
-        checkNotNull(this.contents, "contents");
+        if (this.contents == null) {
+            this.contents = ImmutableList.of();
+        }
```
ST-DDT: IMO its more likely to be an error to not specify the contents. So this should throw an ISException. But it does not really matter.

ST-DDT (approved):

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationBuilder.java
parlough: I have made this change

src/main/java/org/spongepowered/common/service/pagination/IterablePagination.java
```diff
@@ -63,7 +63,7 @@ public IterablePagination(Supplier<Optional<MessageReceiver>> src, PaginationCal
         }
 
         if (page <= this.lastPage) {
-            throw new CommandException(t("Cannot go backward in an IterablePagination"));
+            throw new CommandException(t("You cannot go backwards in an IterablePagination."));
```
kashike: Perhaps phrase this exception message a different way? This sounds weird.

src/main/java/org/spongepowered/common/service/pagination/ListPagination.java
```diff
@@ -85,7 +85,7 @@ public ListPagination(Supplier<Optional<MessageReceiver>> src, PaginationCalcula
         } else if (page < 1) {
             throw new CommandException(t("Page %s does not exist!", page));
         } else if (page > this.pages.size()) {
-            throw new CommandException(t("Page %s is too high", page));
+            throw new CommandException(t("Page %s is too high!", page));
```
kashike: Include the maximum page?
`"Page %s is greater than %s."`?

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationBuilder.java
```diff
@@ -91,7 +92,7 @@ public SpongePaginationBuilder(SpongePaginationService service) {
 
     @Override
     public PaginationList.Builder padding(Text padding) {
-        checkNotNull(padding, "padding");
+        checkNotNull(padding, "The padding cannot be null!");
```
kashike, line 96: unnecessary

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationBuilder.java
```diff
@@ -105,7 +106,9 @@ public SpongePaginationBuilder(SpongePaginationService service) {
 
     @Override
     public PaginationList build() {
-        checkNotNull(this.contents, "contents");
+        if (this.contents == null) {
+            throw new IllegalStateException("The contents of the pagination list cannot be null!");
```
kashike: use `checkState`

kashike (changes requested):

src/main/java/org/spongepowered/common/service/pagination/IterablePagination.java
parlough: `You cannot go to a previous page in an IterablePagination.`?

kashike: Sure.

PR #1479, opened 2017-08-08
Add an optimized version of Entity.getNearbyEntities(double).

JBYoshi:
The `default`-implemented version in the API checks every entity in the world. This version only checks entities in chunks that could be within the specified distance.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
@@ -833,6 +833,13 @@ public abstract class MixinEntity implements IMixinEntity {
         return entity != null && entity.addPassenger(this);
     }
 
+    @Override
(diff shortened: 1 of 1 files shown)


NeumimTo:
It would be actually helpful to also add a method World#getNearbyEntities(Position, radius), a method, where we wont need to use an entity as the centre. 
This could be extended maybe to World#getNearbyEntities(Position, radius, loadchunks) ...

Zidane:
@NeumimTo 

Open issue on API then. Not the focus of this PR.

Zidane:
@JBYoshi 

The default implemented version in the API, frankly, sucks and any implementation (SpongeCommon, Lantern) will surely override it. Can you do an API PR and further this PR to correct that?

@Meronat You'll need to update Lantern for that.

JBYoshi:
Superseded by SpongePowered/SpongeAPI#1630.

PR #1481, opened 2017-08-13
Use pagination for '/sponge plugins' command

### LINKED ISSUE #1239: Sponge Plugins list String too big

Grinch:
**SpongeCommon** | **[SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1719)**

Fixes #1239, Supersedes #1250.

Base work done by @zachoooo.

**Screenshots**

### DIFF
--- a/src/main/java/org/spongepowered/common/SpongeImplHooks.java
+++ b/src/main/java/org/spongepowered/common/SpongeImplHooks.java
@@ -27,6 +27,7 @@ package org.spongepowered.common;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Streams;
 import net.minecraft.block.Block;
 import net.minecraft.block.ITileEntityProvider;
@@ -59,11 +60,19 @@ import net.minecraft.world.WorldProvider;
 import net.minecraft.world.WorldServer;
 import net.minecraft.world.chunk.Chunk;
 import net.minecraft.world.storage.MapStorage;
+import org.spongepowered.api.command.args.ChildCommandElementExecutor;
+import org.spongepowered.api.command.spec.CommandSpec;
 import org.spongepowered.api.item.inventory.ItemStack;
 import org.spongepowered.api.item.inventory.crafting.CraftingGridInventory;
 import org.spongepowered.api.item.recipe.crafting.CraftingRecipe;
 import org.apache.logging.log4j.Logger;
 import org.spongepowered.api.event.cause.NamedCause;
+import org.spongepowered.api.plugin.PluginContainer;
+import org.spongepowered.api.service.pagination.PaginationList;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.action.TextActions;
+import org.spongepowered.api.text.format.TextColors;
+import org.spongepowered.common.command.SpongeCommand;
 import org.spongepowered.common.event.tracking.CauseTracker;
 import org.spongepowered.common.event.tracking.ItemDropData;
 import org.spongepowered.common.item.inventory.util.InventoryUtil;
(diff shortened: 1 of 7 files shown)


parlough:
I don't really see a need for the numbers and the light blue doesn't seem to fit that well. Maybe try light green instead?

Grinch:
@Meronat Is [this](http://i.imgur.com/KTWJtBf.png) a bit better?

vectrixdevelops:
@Grinch if I may make a suggestion, maybe when you hover over a plugin, you could also add a line for that plugins dependencies and the ones that aren't optional and are not present could show up red to assist the user in understanding what is missing. Not sure. 😄

Grinch:
@connorhartley Pretty sure when a plugin loads, and it is missing a non-optional dependency that it will throw an exception and not load further.

vectrixdevelops:
@Grinch true, but on the current list the plugin shows up as red to say it couldn't load. Not sure if this is still the effect you have on this list. If not, I wouldn't worry about my suggestion. :P

Grinch:
@connorhartley This does not take place in the current code of SpongeCommon: https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/command/SpongeCommand.java#L500

**Edit**: Make no mistake, I like your idea. I'm just not sure how one even checks if a plugin loaded successfully since it wouldn't be registered I imagine as well as the tooltips in Minecraft are finicky as it is so adding a list of dependencies could result in some plugins having really bad tooltips.

vectrixdevelops:
@Grinch Yeah, I am not sure why I am thinking it does that when there is absolutely no way that could happen. Oh well.

**Edit:**  I have bamboozled myself.

src/main/java/org/spongepowered/common/SpongeImplHooks.java
```diff
@@ -334,4 +345,31 @@ public static void onCraftingRecipeRegister(CraftingRecipe recipe) {
...
+
+    public static Text getAdditionalCommandDescriptions() {
+        return Text.EMPTY;
+    }
+
+    public static void registerAdditionalCommands(ChildCommandElementExecutor flagChildren, ChildCommandElementExecutor nonFlagChildren) {
+        // Overwritten in SpongeForge
+    }
+
+    public static List<PluginContainer> getSortedPluginList() {
```
kashike: Why is this here? It's only used in the command.

src/main/java/org/spongepowered/common/command/SpongeCommand.java
```diff
@@ -472,64 +481,88 @@ private static CommandSpec getPluginsCommand() {
...
-                        }
-                    } else {
-                        Collection<PluginContainer> plugins = SpongeImpl.getGame().getPluginManager().getPlugins();
-                        Text.Builder build = Text.builder(String.format("Plugins (%d): ", plugins.size()));
-                        boolean first = true;
-                        for (PluginContainer next : plugins) {
-                            if (!first) {
-                                build.append(SEPARATOR_TEXT);
+                            final PaginationList.Builder builder = PaginationList.builder();
+                            builder.title(Text.of(TextColors.YELLOW, "Plugins", TextColors.WHITE, " (", sortedContainers.size(), ")"))
```
kashike: Perhaps `GOLD` instead?

src/main/java/org/spongepowered/common/SpongeImplHooks.java
```diff
@@ -334,4 +345,31 @@ public static void onCraftingRecipeRegister(CraftingRecipe recipe) {
...
+        final List<String> pluginStatics = Lists.newArrayList("minecraft", "mcp", "spongeapi", "sponge");
```
kashike: Can be a constant.

src/main/java/org/spongepowered/common/SpongeImplHooks.java
Grinch: It is overridden in SpongeForge to change the static entries at the beginning of the list.

src/main/java/org/spongepowered/common/command/SpongeCommand.java
Grinch: Zidane wanted whatever is generally used for Sponge elsewhere.

src/main/java/org/spongepowered/common/SpongeImplHooks.java
Grinch: To clarify actually. It is overwritten by SpongeForge to remove any mods from the list.

PR #1483, opened 2017-08-14
Method Tracking System

Cybermaxke:
**Common** | [Forge](https://github.com/SpongePowered/SpongeForge/pull/1723)

This PR adds a `ClassTransformer` that allows the methods of target classes (mainly interfaces) to be redirected so actions can be executed before and after the method, and even cancel the method or modify it's parameters.

The main reason for this system was to track calls to `IItemHandler`, `IFluidTank`, `IFluidHandler` and `IFluidBlock`. These are all forge interfaces and will be tracked in `SpongeForge`.

The way the tracker class system is designed is that no classes will be loaded that could trigger class loading, ASM will be used to read all the `TrackerMethod`s from registered tracker classes. This is already a preparation for the separated core/main mod refactor.

A example of the usage can be found in the `SpongeForge` PR.

### DIFF
--- a/src/main/java/org/spongepowered/common/launch/transformer/tracker/TrackerClassTransformer.java
+++ b/src/main/java/org/spongepowered/common/launch/transformer/tracker/TrackerClassTransformer.java
@@ -1,0 +1,191 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.launch.transformer.tracker;
+
+import static org.spongepowered.asm.lib.Opcodes.ACC_PRIVATE;
+import static org.spongepowered.asm.lib.Opcodes.ACC_STATIC;
+import static org.spongepowered.asm.lib.Opcodes.ALOAD;
+import static org.spongepowered.asm.lib.Opcodes.ASM5;
+import static org.spongepowered.asm.lib.Opcodes.CHECKCAST;
+import static org.spongepowered.asm.lib.Opcodes.F_SAME;
+import static org.spongepowered.asm.lib.Opcodes.IFEQ;
+import static org.spongepowered.asm.lib.Opcodes.ILOAD;
+import static org.spongepowered.asm.lib.Opcodes.INSTANCEOF;
+import static org.spongepowered.asm.lib.Opcodes.INVOKEINTERFACE;
+import static org.spongepowered.asm.lib.Opcodes.INVOKESTATIC;
+import static org.spongepowered.asm.lib.Opcodes.INVOKEVIRTUAL;
+import static org.spongepowered.asm.lib.Opcodes.IRETURN;
+
+import net.minecraft.launchwrapper.IClassTransformer;
+import org.spongepowered.asm.lib.ClassReader;
+import org.spongepowered.asm.lib.ClassVisitor;
+import org.spongepowered.asm.lib.ClassWriter;
+import org.spongepowered.asm.lib.Label;
+import org.spongepowered.asm.lib.MethodVisitor;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+public class TrackerClassTransformer implements IClassTransformer {
+
+    @Override
+    public byte[] transform(String name, String transformedName, byte[] basicClass) {
+        if (basicClass == null || TrackerRegistry.trackerClasses.contains(name)) {
+            return basicClass;
+        }
+
+        final ClassReader classReader = new ClassReader(basicClass);
+        final ClassWriter classWriter = new ClassWriter(classReader, 0);
+
+        final TrackerClassVisitor classVisitor = new TrackerClassVisitor(classWriter);
+        classReader.accept(classVisitor, 0);
+
+        return classWriter.toByteArray();
+    }
+
+    private static class TrackerClassVisitor extends ClassVisitor {
+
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/common/launch/transformer/tracker/TrackerClassTransformer.java
```diff
@@ -0,0 +1,210 @@
...
+                final MethodInsnNode newMethodInsnNode = new MethodInsnNode(INVOKESTATIC,
+                        classNode.name, methodName, desc, false);
+                it.set(newMethodInsnNode);
+            }
+            change = true;
+        }
+        return change;
+    }
+
+    private static int getReturnOpcode(Type type) {
```
Mumfrey: Why would you not just use `Type.getOpcode(Opcodes.IRETURN)` ?

src/main/java/org/spongepowered/common/launch/transformer/tracker/TrackerClassTransformer.java
```diff
@@ -0,0 +1,210 @@
...
+            return DRETURN;
+        } else if (type == Type.FLOAT_TYPE) {
+            return FRETURN;
+        } else if (type == Type.FLOAT_TYPE) {
+            return LRETURN;
+        }
+        return ARETURN;
+    }
+
+    private static int getLoadOpcode(Type type) {
```
Mumfrey: Why would you not just use `Type.getOpcode(Opcodes.ILOAD)` ?

src/main/java/org/spongepowered/common/launch/transformer/tracker/TrackerClassTransformer.java
```diff
@@ -0,0 +1,210 @@
...
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.ListIterator;
+import java.util.Map;
+import java.util.Set;
+
+public class TrackerClassTransformer implements IClassTransformer {
+
+    @Override
+    public byte[] transform(String name, String transformedName, byte[] basicClass) {
```
Mumfrey, line 55: Missing null check, breaks transformer contract.

src/main/java/org/spongepowered/common/launch/transformer/tracker/TrackerClassTransformer.java
```diff
@@ -0,0 +1,210 @@
...
+
+        final ClassReader classReader = new ClassReader(basicClass);
+
+        final ClassNode classNode = new ClassNode(ASM5);
+        classReader.accept(classNode, ClassReader.EXPAND_FRAMES);
+
+        final Set<String> addedMethods = new HashSet<>();
+        boolean change = false;
+        for (MethodNode methodNode : new ArrayList<>(classNode.methods)) {
+            if (processMethodNode(classNode, methodNode, addedMethods)) {
```
Mumfrey: This reeks of needing to be visitor pattern not tree pattern. Why are you using tree for this?

src/main/java/org/spongepowered/common/launch/transformer/tracker/TrackerClassTransformer.java
```diff
@@ -0,0 +1,210 @@
...
+                change = true;
+            }
+        }
+
+        if (!change) {
+            return basicClass;
+        }
+
+        final ClassWriter classWriter = new ClassWriter(ASM5);
```
Mumfrey: Not using the correct constructor for `ClassWriter`, but really this whole thing should be visitor not tree.

src/main/java/org/spongepowered/common/launch/transformer/tracker/TrackerClassTransformer.java
```diff
@@ -0,0 +1,210 @@
...
+        classNode.accept(classWriter);
+
+        return classWriter.toByteArray();
+    }
+
+    private static boolean processMethodNode(ClassNode classNode, MethodNode methodNode, Set<String> addedMethods) {
```
Mumfrey: visitor?

src/main/java/org/spongepowered/common/launch/transformer/tracker/TrackerRegistry.java
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * Initializes the method tracking.
+     */
+    public static void initialize() {
+        if (INITIALIZED) {
+            return;
+        }
+        INITIALIZED = true;
+        final String pack = "org.spongepowered.common.launch.transformer.tracker.";
```
Mumfrey: inline constant

Mumfrey:
I'd like to take the opportunity to point out that the reason Mixin doesn't support this, is that this kind of global class processing is always going to be expensive. I feel like using tree in this instance is just exacerbating the problem and that if you're going to do this you should use visitor instead to reduce the overhead of this processing. Looking at your code I think that it will actually make it easier to read as well since you're basically *emulating* visitor pattern but with the added overhead of it being tree.

You're also duplicating some helper method functionality that already exists in `Type` with some unnecessary helper methods of your own, I have annoted these in the above review.

src/main/java/org/spongepowered/common/launch/transformer/tracker/TrackerClassTransformer.java
Cybermaxke: I didn't know you could do that, I will change it.

src/main/java/org/spongepowered/common/launch/transformer/tracker/TrackerClassTransformer.java
Cybermaxke: I started with a tree, but ended up not needing it. I will change it to a ClassVisitor.

Mumfrey:
Also note that when you have a `ClassReader`/`ClassWriter` pair you should always pass the reader into the writer's constructor when using "mostly add" operations since this enables the built-in optimisations in ASM for this scenario.

src/main/java/org/spongepowered/common/launch/transformer/tracker/TrackerClassTransformer.java
Cybermaxke: @Mumfrey When could one of these parameters be `null`? It doesn't seem to be handled in `SpongeSuperclassTransformer` as well:
https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/launch/transformer/SpongeSuperclassTransformer.java#L43

stephan-gh: `basicClass` is null when the class could be found on the classpath. (One of the transformers may want to generate that class at runtime)

Mumfrey: > When could one of these parameters be null?

`basicClass` can be `null` when the class is non-existent or is expected to be generated at runtime, aka *runtime-synthetic* class. If your transformer doesn't synthesize the class then it should return `null` per the transformer contract, throwing a `NullPointerException` when trying to read the class bytecode which doesn't exist is not an acceptable behaviour for a transformer because it leads to unexpected `ClassNotFoundError`s which are actually because the transformer which *should* generate the class never gets a chance to. For example Mixin generates different classes at runtime to support target-specialised inner classes, and `Args` subclasses for `@ModifyArgs`.

> It doesn't seem to be handled in SpongeSuperclassTransformer as well:

Then that's a mistake and should be rectified, and the person who wrote it duly chastised.

*EDIT: @Minecrell beat me to it, but the above is relevant all the same*

Mumfrey: It looks like `SpongeSuperclassTransformer` actually gets away with it because it only does any processing for pre-registered classes. So it's not gonna hit the sitation where it tries to transform a runtime-synthetic class anyway unless something goes horribly wrong. It should probably still null-check anyway, just as good practice, but it's not an unforgivable situation because the design of the transformer is such that it mitigates the issue.

bloodmc:
Please let me know when this is ready for pull as GriefPrevention could really use this system to track fluids and item transportation with mods.

Cybermaxke:
@bloodmc It is finished as far I know, unless I missed something.

bloodmc:
@Mumfrey does this PR look good to pull?

Cybermaxke:
Also a small note that only one tracker can be registered for every method. But you can mixin into the tracker method classes.

PR #960, opened 2016-09-30
Implement Scoreboard Tags API

AlphaModder:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1374) | **Common** | Forge

This implements the scoreboard tags API from [SpongeAPI/#1374](https://github.com/SpongePowered/SpongeAPI/pull/1374).

This is my first time creating and implementing data, and also my first time working with submodules, so it's possible that I've done something horribly wrong.

PR #1490, opened 2017-08-16
Implements the API for scoreboard tags

parlough:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1633) | **SpongeCommon**

Implements the data API interfaces for the TagsData often used by scoreboard related commands.

Supersedes https://github.com/SpongePowered/SpongeCommon/pull/960

### DIFF
--- a/src/main/java/org/spongepowered/common/data/manipulator/mutable/entity/SpongeTagsData.java
+++ b/src/main/java/org/spongepowered/common/data/manipulator/mutable/entity/SpongeTagsData.java
@@ -1,0 +1,53 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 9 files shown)


src/main/java/org/spongepowered/common/data/processor/data/entity/TagsDataProcessor.java
```diff
@@ -0,0 +1,93 @@
...
+        super(Entity.class, Keys.TAGS);
+    }
+
+    @Override
+    protected boolean set(Entity dataHolder, Set<String> value) {
+        for (String oldTag : dataHolder.getTags()) {
+            dataHolder.removeTag(oldTag);
+        }
+
+        // Minecraft does not allow there to be more than 1024 tags.
```
parlough: Another option is to add all of them one by one, keeping track of which ones were added successfully, then removing them all again if one of them failed to be added.

parlough: I'm actually going to switch to that option in-case some entity in a mod overrides the `addTag()` method.

src/main/java/org/spongepowered/common/data/manipulator/mutable/entity/SpongeTagsData.java
```diff
@@ -0,0 +1,53 @@
...
+import org.spongepowered.common.data.manipulator.immutable.entity.ImmutableSpongeTagsData;
+import org.spongepowered.common.data.manipulator.mutable.common.AbstractSingleSetData;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class SpongeTagsData extends AbstractSingleSetData<String, TagsData, ImmutableTagsData> implements TagsData {
+
+    public SpongeTagsData() {
+        this(new HashSet<>());
```
kashike: `Collections.emptySet()`

kashike (changes requested):

parlough:
Will update this soon so we can get it into API 7.1

PR #1491, opened 2017-08-17
Move the implementation of Titles to SpongeCommon

parlough:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1634) | **SpongeCommon**

This moves the implementation of `Title` to Sponge to follow what has been a continued goal of the API. 

I believe the implementations are functionally the same, but if you see any difference point it out. Or point out if you see any changes which would be good to make now.

If anyone knows a plugin which utilizes titles extensively, it would be nice to know so I can test to compare this PR versus bleeding better. Thanks!


gabizou:
Was handled when https://github.com/SpongePowered/SpongeAPI/pull/1744 and it's respective PR was merged.

PR #402, opened 2016-01-05
Implementation of SizeData

hsyyid:
Implementation of `SizeData`.
- [x] Registration of field getters and setters
- [x] Accurately used the appropriate `AbstractData` implementation

Implementation of `ImmutableSizeData`
- [x] Accurately extended the appropriate `AbstractImmutableData` implementation
- [x] Accurately instantiating final instance fields with an `ImmutableValue` counter part for any value getters
- [x] If necessary, creating a new `ImmutableValue` for a value that should not be cached, or, using the existing caching utils.

Implementation of the `VALUE_PROCESSOR`(s):
- [x] Appropriately extended `AbstractSpongeValueProcessor`
- [x] Individual processors for each source type possible (one for `ItemStack`/`TileEntity`/`Entity` as necessary).

Registration:
- [x] Registered the `Key` correctly by `FIELD_NAME` in the `KeyRegistry`
- [x] Registered the `DataProcessor`s and `DataManipulatorBuilder` in `SpongeSerializationRegistry`
- [x] Registered the `ValueProcessor`s in the `SpongeSerializationRegistry`

``` java
package io.github.hsyyid.testplugin;

import org.spongepowered.api.Sponge;
import org.spongepowered.api.block.BlockTypes;
import org.spongepowered.api.command.CommandException;
import org.spongepowered.api.command.CommandResult;
import org.spongepowered.api.command.CommandSource;
import org.spongepowered.api.command.args.CommandContext;
import org.spongepowered.api.command.args.GenericArguments;
import org.spongepowered.api.command.spec.CommandExecutor;
import org.spongepowered.api.command.spec.CommandSpec;
import org.spongepowered.api.data.key.Keys;
import org.spongepowered.api.entity.Entity;
import org.spongepowered.api.entity.EntityType;
import org.spongepowered.api.entity.living.player.Player;
import org.spongepowered.api.event.Listener;
import org.spongepowered.api.event.cause.Cause;
import org.spongepowered.api.event.game.state.GameInitializationEvent;
import org.spongepowered.api.plugin.Plugin;
import org.spongepowered.api.text.Text;
import org.spongepowered.api.text.format.TextColors;
import org.spongepowered.api.util.blockray.BlockRay;
import org.spongepowered.api.util.blockray.BlockRayHit;
import org.spongepowered.api.world.Location;
import org.spongepowered.api.world.World;
import org.spongepowered.api.world.extent.Extent;

import java.util.Optional;

[...]

PR #1492, opened 2017-08-18
Implement SizeData

parlough:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1636) | **SpongeCommon**

Supersedes https://github.com/SpongePowered/SpongeCommon/pull/402

Implements size data based on Hassan's changes, but with some changes to the API and implementation, including the inclusion of a modified version of his test plugin. 

With this PR you can modify the base_size(width) and height, but developers have to keep in mind that this only changes the bounding box, it isn't super reliable, and Sponge(maybe Minecraft idk) limits the max size of a bounding box to 1000.

I am open to changes everywhere, from documentation to functionality, so please give this a good review.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/manipulator/immutable/entity/ImmutableSpongeSizeData.java
+++ b/src/main/java/org/spongepowered/common/data/manipulator/immutable/entity/ImmutableSpongeSizeData.java
@@ -1,0 +1,124 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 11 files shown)


parlough:
I can go ahead and fix these merge conflicts sometime today or tomorrow. If anyone would like to see any other changes please bring them up.

src/main/java/org/spongepowered/common/data/manipulator/immutable/entity/ImmutableSpongeSizeData.java
```diff
@@ -0,0 +1,124 @@
...
+    private final float scale;
+
+    private final ImmutableBoundedValue<Float> baseValue;
+    private final ImmutableBoundedValue<Float> heightValue;
+    private final ImmutableBoundedValue<Float> scaleValue;
+
+    public ImmutableSpongeSizeData(float base, float height, float scale) {
+        super(ImmutableSizeData.class);
+
+        checkArgument(scale > 0, "The base size must be greater than 0.");
```
kashike: Should be `base > 0`.

kashike (changes requested):

gabizou:

PR #1495, opened 2017-08-18
Implement parrot data and variants

parlough:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1638) | **SpongeCommon**

This pull request implements parrot data and parrot variants, as well as adding a cute little test plugin.

For now I default to red in places as that is the parrot in vanilla with the 0 meta, so it made sense, but I'm open to changing this.

Cheers 🐦 

(Here's me using the test plugin to make a ton of red parrots)

### DIFF
--- a/src/main/java/org/spongepowered/common/data/manipulator/mutable/entity/SpongeParrotData.java
+++ b/src/main/java/org/spongepowered/common/data/manipulator/mutable/entity/SpongeParrotData.java
@@ -1,0 +1,48 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data.manipulator.mutable.entity;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import org.spongepowered.api.data.key.Keys;
+import org.spongepowered.api.data.manipulator.immutable.entity.ImmutableParrotData;
+import org.spongepowered.api.data.manipulator.mutable.entity.ParrotData;
+import org.spongepowered.api.data.type.ParrotVariant;
+import org.spongepowered.api.data.type.ParrotVariants;
+import org.spongepowered.common.data.manipulator.immutable.entity.ImmutableSpongeParrotData;
+import org.spongepowered.common.data.manipulator.mutable.common.AbstractSingleCatalogData;
+
+public class SpongeParrotData extends AbstractSingleCatalogData<ParrotVariant, ParrotData, ImmutableParrotData> implements ParrotData {
+
+    public SpongeParrotData() {
+        this(ParrotVariants.RED);
+    }
+
+    public SpongeParrotData(ParrotVariant variant) {
+        super(ParrotData.class, checkNotNull(variant, "The parrot variant cannot be null!"),
+                Keys.PARROT_VARIANT, ImmutableSpongeParrotData.class);
+    }
+
+}
--- a/src/main/java/org/spongepowered/common/data/processor/data/entity/ParrotDataProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/data/entity/ParrotDataProcessor.java
@@ -1,0 +1,84 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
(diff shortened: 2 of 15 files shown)


src/main/java/org/spongepowered/common/data/processor/data/entity/ParrotDataProcessor.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Override
+    protected Optional<ParrotVariant> getVal(EntityParrot dataHolder) {
+        return Optional.of(SpongeEntityConstants.PARROT_VARIANT_IDMAP.get(dataHolder.getVariant()));
+    }
+
+    @Override
+    protected ImmutableValue<ParrotVariant> constructImmutableValue(ParrotVariant value) {
+        return ImmutableSpongeValue.cachedOf(Keys.PARROT_VARIANT, ParrotVariants.RED, value);
```
kashike: `this.key`

src/main/java/org/spongepowered/common/data/processor/data/entity/ParrotDataProcessor.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Override
+    protected Value<ParrotVariant> constructValue(ParrotVariant actualValue) {
+        return new SpongeValue<>(Keys.PARROT_VARIANT, ParrotVariants.RED, actualValue);
```
kashike: `this.key`

src/main/java/org/spongepowered/common/data/manipulator/mutable/entity/SpongeParrotData.java
```diff
@@ -0,0 +1,48 @@
...
+import org.spongepowered.common.data.manipulator.mutable.common.AbstractSingleCatalogData;
+
+public class SpongeParrotData extends AbstractSingleCatalogData<ParrotVariant, ParrotData, ImmutableParrotData> implements ParrotData {
+
+    public SpongeParrotData(ParrotVariant variant) {
+        super(ParrotData.class, checkNotNull(variant, "The parrot variant cannot be null!"),
+                Keys.PARROT_VARIANT, ImmutableSpongeParrotData.class);
+    }
+
+    public SpongeParrotData() {
```
kashike: move this one before the other constructor

kashike (changes requested):

testplugins/src/main/java/org/spongepowered/test/ParrotDataTest.java
```diff
@@ -0,0 +1,53 @@
+package org.spongepowered.test;
```
kashike, line 25: l

parlough: generic-drivers-license

kashike: expired

parlough: Should be good now. I soaked up the bad smell with a Sponge license.

src/main/java/org/spongepowered/common/data/processor/data/entity/ParrotDataProcessor.java
```diff
@@ -0,0 +1,87 @@
...
+public class ParrotDataProcessor extends
+        AbstractEntitySingleDataProcessor<EntityParrot, ParrotVariant, Value<ParrotVariant>, ParrotData, ImmutableParrotData> {
+
+    public ParrotDataProcessor() {
+        super(EntityParrot.class, Keys.PARROT_VARIANT);
+    }
+
+    @Override
+    protected boolean set(EntityParrot dataHolder, ParrotVariant value) {
+        if (value instanceof SpongeParrotVariant) {
```
gabizou: If you're anticipating non-sponge implemented variants, you should throw an exception for a bad data transaction.

gabizou (changes requested):
One thing, but otherwise, looks good.

src/main/java/org/spongepowered/common/data/processor/data/entity/ParrotDataProcessor.java
Meronat: I was just following what is present in the Rabbit and Ocelot data processors. Would you like me to make this change there as well?

src/main/java/org/spongepowered/common/registry/type/entity/ParrotVariantRegistryModule.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Optional;
+
+public class ParrotVariantRegistryModule implements CatalogRegistryModule<ParrotVariant> {
+
+    @RegisterCatalog(ParrotVariants.class)
+    private final Map<String, ParrotVariant> parrotVariantMap = new HashMap<>();
+
+    @Override
+    public Optional<ParrotVariant> getById(String id) {
+        return Optional.ofNullable(this.parrotVariantMap.get(checkNotNull(id).toLowerCase(Locale.ENGLISH)));
```
Zidane, line 49: I've seen in other modules we check if it doesn't have a domain, like minecraft: or almura:, and if not we prepend minecraft: before checking the map...is that needed here? Did the tests pass? 

@gabizou you wrote the modules tests, is that something that needs to be accounted for and should we just do it always?

src/main/java/org/spongepowered/common/registry/type/entity/ParrotVariantRegistryModule.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public Collection<ParrotVariant> getAll() {
+        return ImmutableList.copyOf(this.parrotVariantMap.values());
+    }
+
+    @Override
+    public void registerDefaults() {
+        this.parrotVariantMap.putAll(SpongeEntityConstants.PARROT_VARIANTS);
```
Zidane, line 59: @gabizou Not sure its worth it to have a constants class for the Entity stuff but not do this for many other processors. Either we should standardize this or drop this concept and have the mappings in each module. I almost prefer to keep it in the module as it is all kept together under a common struct (parrot stuff with parrot).

src/main/java/org/spongepowered/common/registry/type/entity/ParrotVariantRegistryModule.java
parlough: I was just following how it was done previously, but I did feel it was weird that prepending minecraft: did not work. I'd be willing to make this change so both work for this and other similar types.

src/main/java/org/spongepowered/common/registry/type/entity/ParrotVariantRegistryModule.java
parlough: I can go ahead and move them all to their own in this PR if desired or in a new one after this PR.

src/main/java/org/spongepowered/common/registry/type/entity/ParrotVariantRegistryModule.java
ST-DDT: Sooner or later we should address this globally. But for the start it is probably enough. If you are unsure use the test provided in https://github.com/SpongePowered/SpongeCommon/pull/1283 and verify that there are no issues with your classes. That PR is not merged yet so you have to do it yourself (Maybe adjust them to only test your catalog types since there are some broken ones out there)

src/main/java/org/spongepowered/common/registry/type/entity/ParrotVariantRegistryModule.java
gabizou: We already have [DataConstants](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/data/util/DataConstants.java#L56) where you should be able to chuck default ones in the Catalogued section. But I'm in agreement that the defaults should be in the same module, SpongeEntityCosntants is going away very soon.

kashike (approved):

parlough:
Should be good to go if people want to review. Entity constants/catalog changes can be PR'd after, especially after some related catalog changes get in.

parlough:
I have expanded this PR to implement more of the parrot methods as well as the tame entity event and sitting data as well. Parrots should be good to go :) just need to do some final testing with the new additions

parlough:
I forgot my license at home officer, please let me off the hook this time.

parlough:
Okay I've tested these changes and they seem to work well to me.

PR #1500, opened 2017-08-19
Fix unintentionally class loading caused by Tracker redirect methods.

Cybermaxke:


JBYoshi:
Did anyone else notice this was #1500?

PR #1519, opened 2017-09-08
Bugfixes and improvements for selectors

rexbut:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1645) | **SpongeCommon**

I am sorry to have recreated a PR but I have no news from @kenzierocks : https://github.com/SpongePowered/SpongeCommon/pull/793

* Adds new selector 1.12

* Improvements and fixes for selectors based on 1.9 and my own review of the code.

* Selectors now use a List, the resolver returns the elements ordered properly as well as the right number of elements. Removed the force argument as I don't remember what it was really useful for.

### DIFF
--- a/src/main/java/org/spongepowered/common/registry/type/text/ArgumentRegistryModule.java
+++ b/src/main/java/org/spongepowered/common/registry/type/text/ArgumentRegistryModule.java
@@ -64,7 +64,7 @@ public final class ArgumentRegistryModule implements RegistryModule {
         this.argumentTypeMap.put("radius", radius);
 
         // GAME_MODE
-        this.argumentTypeMap.put("game_mode", factory.createArgumentType("m", GameMode.class));
+        this.argumentTypeMap.put("game_mode", factory.createInvertibleArgumentType("m", GameMode.class));
 
         // COUNT
         this.argumentTypeMap.put("count", factory.createArgumentType("c", Integer.class));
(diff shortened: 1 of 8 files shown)


ryantheleach:
@rexbut As this is targeting API 8 due to API changes being required, it's predicted that this will fall into 1.13 rather then 1.12

This means that this PR will likely need to be updated for the 1.13 codebase when it lands. Are you willing to update it when that happens?

rexbut:
@ryantheleach Yes I will upgrade my modification on 1.13

dualspiral:
It's a shame this has stalled (sorry), but it would be good to see as much of this as possible on API 7 as selectors could do with some love, especially as 1.12/API 7 will be an LTS version. I was looking to fix a couple of issues with selectors (specifically `@p` selecting more than one player and not necessarily the nearest player!), but it would be far better to look at getting this in.

I can see that there are breaks in the API to fix some of these issues (the big one being that sets->lists), but I'm fairly sure we can do all this without an API break so it's sutable for 1.12.

Specifically, the idea of using lists is for sorting in the implementation. We can use an `ImmutableSet` for now [as the order is preserved](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/ImmutableSet.html#copyOf(java.util.Collection)). We could also use an `ImmutableSortedSet` with comparitor. Similarly, [we don't need to change this](https://github.com/SpongePowered/SpongeAPI/pull/1645/files#diff-5a900ae119b2ea082d6e3d34ade1f9b5R66), `ArgumentType.Invertible` implements `ArgumentType`, though it would be nice to change that in API 8. for clarity. Same with the lists. Making these changes (with added documentation on the fact that the set will be ordered) would mean that the API would be compatible with API 7 and we can proceed with this.

Does this proposal seem reasonable to people? Does anyone in @SpongePowered/developers have any input? Finally, @rexbut, would you be willing to continue this, or would you rather not? Regardless, thanks for your contribution to this!

dualspiral:
Superseded by https://github.com/SpongePowered/SpongeCommon/pull/2033

PR #1528, opened 2017-09-15
Update implementation for ContextValue

gabizou:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1648) | **SpongeCommon**

This updates the implementation, borrowing the implementation of previously `NamedCauseFilter`.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/filter/delegate/NamedCauseFilterSourceDelegate.java
+++ b/src/main/java/org/spongepowered/common/event/filter/delegate/NamedCauseFilterSourceDelegate.java
@@ -1,0 +1,113 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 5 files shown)


src/test/java/org/spongepowered/common/event/listener/ContextValueCauseListener.java
```diff
@@ -0,0 +1,32 @@
+package org.spongepowered.common.event.listener;
```
kashike, line 1: license header

kashike (changes requested):

src/main/java/org/spongepowered/common/event/filter/delegate/NamedCauseFilterSourceDelegate.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+
+public class NamedCauseFilterSourceDelegate extends CauseFilterSourceDelegate {
+
+    private final ContextValue anno;
+
+    public NamedCauseFilterSourceDelegate(ContextValue anno) {
+        this.anno = anno;
```
Deamon5550, line 56: Assert here that the id in the annotation is a valid key and throw an exception if not.

src/main/java/org/spongepowered/common/event/filter/delegate/NamedCauseFilterSourceDelegate.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @Override
+    protected void insertCauseCall(MethodVisitor mv, Parameter param, Class<?> targetType) {
+        mv.visitMethodInsn(INVOKEVIRTUAL, Type.getInternalName(Cause.class), "getContext", "()" + Type.getDescriptor(EventContext.class), false);
```
Deamon5550, line 61: Throw an exception here if the parameter type does not match the type of the key.

src/main/java/org/spongepowered/common/event/filter/delegate/NamedCauseFilterSourceDelegate.java
Deamon5550: This filter delegate should be creating a field in the filter holding the `EventContextKey` which is fetched from the registry by id when the filter is initialized (look at the subtype filter for how creating a field and putting stuff into the initializer works)

src/test/java/org/spongepowered/common/event/listener/ContextValueCauseListener.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    @Listener
+    public void namedCauseListenerInclude(EventFilterTest.SubEvent event, @ContextValue(value = EventContextKeys.Named.OWNER_NAME, typeFilter=User.class) Object state) {
+        this.namedCauseListenerCalledInc = true;
+    }
+
+    @Listener
+    public void namedCauseListenerExclude(EventFilterTest.SubEvent event, @ContextValue(value = EventContextKeys.Named.OWNER_NAME, typeFilter=User.class, inverse = true) Object state) {
+        this.namedCauseListenerCalledEx = true;
```
Deamon5550, line 28: this needs to be inverted since you are testing for false

Deamon5550 (changes requested):

ryantheleach:
Closed by @Zidane , See API PR.

Deamon5550:
@ryantheleach this PR is still needed as the `ContextValue` filter is still not implemented. The api pr had adjacent changes which is what @Zidane was rejecting.

PR #1536, opened 2017-09-23
Fix HarvestingProperty for pickaxes

DawnTheFawn1:
This fixes obtaining the correct set of blocks in HarvestingProperty for pickaxes. I made a plugin that can easily demonstrate this change. When a player begins breaking a block, the plugin will automatically switch the item in their hand with the correct item for breaking that block. Currently, if a player has a diamond pickaxe in their inventory and tries to break obsidian it won't switch but with this change it does. [Plugin source](https://github.com/RysingDragon/ToolSwitch/blob/master/src/main/java/com/rysingdragon/toolswitch/ToolSwitch.java)

### DIFF
--- a/src/main/java/org/spongepowered/common/data/property/store/item/HarvestingPropertyStore.java
+++ b/src/main/java/org/spongepowered/common/data/property/store/item/HarvestingPropertyStore.java
@@ -27,6 +27,7 @@ package org.spongepowered.common.data.property.store.item;
 import com.google.common.collect.ImmutableSet;  
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.item.Item;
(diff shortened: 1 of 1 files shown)


kashike (approved):

kashike:
Thanks, @RysingDragon.

gabizou:
I would have almost preferred that the test plugin would have been added to testplugins for implementation. Since it's a toggle, it would've been a good test to add (and a good example of what they should be).

DawnTheFawn1:
@gabizou I understand, I originally created the plugin as I thought it was a nice little thing the community would appreciate. I've already uploaded it on Ore, I'm not sure if you have a policy against this or something, but I wouldn't mind if you used the source and still wanted to add it to the testplugins.

PR #1548, opened 2017-09-30
Fix use_item context in specific InteractItem events.

Zidane:
FishingEvent, in certain scenarios, will not have the use_item context
in it's cause. I elected to always provide that context if we're dealing
with a TryUseItem packet.

@bloodmc Need your input here

### DIFF
--- a/src/main/java/org/spongepowered/common/network/PacketUtil.java
+++ b/src/main/java/org/spongepowered/common/network/PacketUtil.java
@@ -229,6 +229,12 @@ public class PacketUtil {
             }
         } else if (packetIn instanceof CPacketPlayerTryUseItem) {
             CPacketPlayerTryUseItem packet = (CPacketPlayerTryUseItem) packetIn;
+
+            // Put the item being used into the context to handle cases where items are firing events AFTER InteractItemEvent is called for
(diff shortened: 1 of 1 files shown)


Deamon5550:
Can you throw a javadoc on that method to document its behaviour as being not cause-stack neutral.

src/main/java/org/spongepowered/common/network/PacketUtil.java
```diff
@@ -166,6 +166,13 @@ private static boolean creativeCheck(Packet<?> packetIn, EntityPlayerMP playerMP
         return packetIn instanceof CPacketCreativeInventoryAction;
     }
 
+    /**
+     * Handles firing of events that are triggered before server logic. Be aware that this method makes changes to the current context
+     * of the current frame of the {@link CauseStackManager.StackFrame} (namely to set the used_item).
```
gabizou, line 171: Perhaps we should have a nice documented annotation of our own just to self document methods that are not stack neutral? It would help in self documentation in some cases (of course, something that we should be able to instantly see and understand that something is stack altering to the trained eye versus the untrained eye).

ImMorpheus:
Superseded by https://github.com/SpongePowered/SpongeCommon/pull/2204

PR #1549, opened 2017-09-30
Fix cause for natural entity spawns.

Zidane:
Before no cause was present so the game object would report. Now the
World object is pushed as a cause before processing any phases for
entity spawns.

Also, for triggering spawns in GeneralGenerationPhase we weren't passing
the world spawner as the context of the spawn.

@gabizou please review this.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/CauseTracker.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/CauseTracker.java
@@ -623,6 +623,7 @@ public final class CauseTracker {
             // capture all entities until the phase is marked for completion.
             if (!isForced) {
                 try {
+                    Sponge.getCauseStackManager().pushCause(world);
                     return ((IPhaseState) phaseState).spawnEntityOrCapture(context, entity, chunkX, chunkZ);
                 } catch (Exception | NoClassDefFoundError e) {
                     // Just in case something really happened, we should print a nice exception for people to
--- a/src/main/java/org/spongepowered/common/event/tracking/CauseTracker.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/CauseTracker.java
@@ -623,7 +623,6 @@ public final class CauseTracker {
             // capture all entities until the phase is marked for completion.
             if (!isForced) {
                 try {
-                    Sponge.getCauseStackManager().pushCause(world);
                     return ((IPhaseState) phaseState).spawnEntityOrCapture(context, entity, chunkX, chunkZ);
                 } catch (Exception | NoClassDefFoundError e) {
                     // Just in case something really happened, we should print a nice exception for people to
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/event/tracking/CauseTracker.java
```diff
@@ -623,6 +623,7 @@ public boolean spawnEntity(World world, Entity entity) {
             // capture all entities until the phase is marked for completion.
             if (!isForced) {
                 try {
+                    Sponge.getCauseStackManager().pushCause(world);
```
gabizou: Why are we pushing a world here? It's not always the world's cause to be spawning an entity, it should be up to the phase state to push the world cause.

src/main/java/org/spongepowered/common/event/tracking/phase/generation/GeneralGenerationPhaseState.java
```diff
@@ -134,16 +134,20 @@ public final void unwind(G context) {
...
         final ArrayList<Entity> entities = new ArrayList<>(1);
         entities.add(entity);
-        final SpawnEntityEvent event = SpongeEventFactory.createSpawnEntityEventSpawner(Sponge.getCauseStackManager().getCurrentCause(),
-            entities);
-        SpongeImpl.postEvent(event);
-        if (!event.isCancelled() && event.getEntities().size() > 0) {
-            for (Entity item : event.getEntities()) {
-                ((IMixinWorldServer) item.getWorld()).forceSpawnEntity(item);
+        try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
+            frame.addContext(EventContextKeys.SPAWN_TYPE, SpawnTypes.WORLD_SPAWNER);
```
gabizou: Should be adding the world to the cause here, in the frame.

gabizou: Or, you could be pushing the world to the cause stack the moment the phase is switched (in the context of when world generation phase states are switched on).

Zidane: @gabizou 

Thanks for the feedback. Based on your comments, I'd need to push the world into the stack...here?

https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldEntitySpawner.java#L344

gabizou: You could, but if you’re pushing it outside of a frame, like that injection point, you need to pop that cause off at the return point. If you do it here where you are throwing an event, you can ignore having to pop the cause off the stack since the frame closing will do that for you.

ryantheleach:
Should spawning be happening under this General phase? Or should spawning have it's own phase?

gabizou:
> Should spawning be happening under this General phase? Or should spawning have it's own phase?

This is the generalized spawning. When you want to spawn something in implementation where the spawns themselves are a byproduct, and not a step in a multi step process (like ender dragons or dragon respawns), you don't need to add another phase. In this case, this is the super-generic-blind-firing-don't-care-what-happens phase that will just spawn things normally as they should be expected in all circumstances.

gabizou (approved):

JBYoshi:
Can this be merged? It's been marked "ready to merge" for over a month now.

Issue #1286, opened 2017-04-06
Can't retrieve a ParticleType with just the name, need "minecraft:"


Xenmai:
I've noticed you can't retrieve a ParticleType with Sponge.getRegistry().getType(ParticleType.class, "heart"), you need to use "minecraft:heart". I was not expecting this because, when retrieving blocks, you can use just "stone" without the need of "minecraft:". Seems like a small issue. Thanks beforehand for your time.

phit:
https://github.com/SpongePowered/SpongeCommon/pull/1554

ImMorpheus:
PR merged. 

`Sponge.getRegistry().getType(ParticleType.class, "heart")` is still returning Optional#empty()

ImMorpheus:
PR merged

PR #1552, opened 2017-10-02
Improve registry namespace usage

kashike:
brace yourself


Pokechu22:
From my understanding, this fixes #1286.

kashike:
It will, yes.

Pokechu22:
One other thing - should the namespace also be lower-cased?  Right now you're lowercasing the value but not the namespace; the vanilla `ResourceLocation` lowercases both.  I don't _think_ that that code would be used externally (so it wouldn't really matter too much) but it's worth considering.

kashike:
I lowercase the entire string if it includes the namespace: https://github.com/SpongePowered/SpongeCommon/pull/1552/files#diff-fe71dfd1d29d9150251c9f6ca2af9064R41

Pokechu22:
Right, but what about passing an uppercase `namespace` parameter to `namespacedId`?  I doubt it would happen in practice (I'd assume that the parameter wouldn't ever be changable from user code), but it's a case to consider.

src/main/java/org/spongepowered/common/registry/SpongeGameRegistry.java
```diff
@@ -140,7 +140,7 @@
 
     private RegistrationPhase phase = RegistrationPhase.PRE_REGISTRY; // Needed for module phase registrations
 
-    protected final Map<Class<? extends CatalogType>, CatalogRegistryModule<?>> catalogRegistryMap = new IdentityHashMap<>();
+    public final Map<Class<? extends CatalogType>, CatalogRegistryModule<?>> catalogRegistryMap = new IdentityHashMap<>();
```
kashike, line 143: only a testing change, will not be included in merge

src/main/java/org/spongepowered/common/registry/type/AbstractAlternateCatalogRegistryModule.java
```diff
@@ -0,0 +1,55 @@
...
+    public final Map<String, C> map = new HashMap<>();
+
+    @Override
+    public final Optional<C> getById(final String id) {
+        return Optional.ofNullable(this.map.get(Namespaces.namespacedId(id)));
+    }
+
+    @Override
+    public final Collection<C> getAll() {
+        return Collections.unmodifiableCollection(this.map.values());
```
ST-DDT: Please use a set to avoid returning duplicates.

src/main/java/org/spongepowered/common/registry/type/text/TextSerializerRegistryModule.java
```diff
@@ -65,20 +47,11 @@ public void registerDefaults() {
...
-    }
-
-    @Override
-    public Map<String, TextSerializer> provideCatalogMap() {
-        final HashMap<String, TextSerializer> map = new HashMap<>();
-        for (Map.Entry<String, TextSerializer> entry : this.textSerializerMappings.entrySet()) {
-            map.put(entry.getKey().replace("minecraft:", "").replace("sponge:", ""), entry.getValue());
-        }
-        return map;
+        this.map.put(serializer.getId(), serializer);
```
ST-DDT, line 51: This should check whether the id is registered already and lowercase the id to prevent it from becoming inaccessible.

src/main/java/org/spongepowered/common/registry/type/text/TextStyleRegistryModule.java
```diff
@@ -57,7 +56,7 @@
         if (id.equals("NONE")) {
```
ST-DDT: Potential NPE. Flip this statement.

ST-DDT (changes requested):
Maybe it would be a good idea to also create abstract parent classes for the other registry type modules.

src/main/java/org/spongepowered/common/registry/type/text/TextSerializerRegistryModule.java
ST-DDT: Also missing null check

src/main/java/org/spongepowered/common/registry/type/tileentity/StructureModeRegistryModule.java
```diff
@@ -24,51 +24,30 @@
...
-
-    @Override
-    public Optional<StructureMode> getById(String id) {
-        return Optional.ofNullable(this.structureModes.get(checkNotNull(id, "id").toLowerCase(Locale.ENGLISH)));
-    }
-
-    @Override
-    public Collection<StructureMode> getAll() {
-        return ImmutableSet.copyOf(this.structureModes.values());
+        this.map.put("corner", (StructureMode) (Object) TileEntityStructure.Mode.CORNER);
```
ST-DDT, line 39: This should also be prefixed with `minecraft:` (although it will be added by default) to match the other registry classes.
(Or it should be removed from all)

ST-DDT (changes requested):

kashike:
Moving to #1554

PR #1554, opened 2017-10-02
Catalogue key

kashike:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1655) | [**SpongeCommon**](https://github.com/SpongePowered/SpongeCommon/pull/1554)


ST-DDT:
It might be useful to merge https://github.com/SpongePowered/SpongeCommon/pull/1283 before this to ensure all methods are properly implemented everywhere.

src/main/java/org/spongepowered/common/item/inventory/lens/impl/comp/Inventory2DLensImpl.java
```diff
@@ -92,7 +92,7 @@ protected void init(SlotProvider<IInventory, ItemStack> slots) {
      * that the target inventory is {@link #stride} slots wide.
      *
      * @param slots the provider of the slots
-     * @param spanning Set to true to create spanning slots, false to create
```
Zidane: Tsk tsk.

src/main/java/org/spongepowered/common/registry/AbstractCatalogRegistryModule.java
```diff
@@ -0,0 +1,53 @@
...
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.registry.AlternateCatalogRegistryModule;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Optional;
+
+public abstract class AbstractCatalogRegistryModule<C extends CatalogType> implements AlternateCatalogRegistryModule<C> {
+
+    protected final RegistryMap<C> map = RegistryMap.create();
```
Zidane: Create method seems like extra fluff, should just do full line initialization.

src/main/java/org/spongepowered/common/item/inventory/lens/impl/comp/Inventory2DLensImpl.java
kashike: :eyes:

src/main/java/org/spongepowered/common/registry/RegistryMapImpl.java
```diff
@@ -0,0 +1,71 @@
...
+
+import org.spongepowered.api.CatalogKey;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+final class RegistryMapImpl<V> extends HashMap<CatalogKey, V> implements RegistryMap<V> {
```
Deamon5550: Why do we have an interface with exactly one implementation which is essentially the trivial implementation. For that matter why do we have `RegistryMap` in the first place, it seems to just be a `Map<CatalogKey, V>` with the only additional method being `forCatalogRegistration` which just converts it to a map keyed on strings. Which could be concerted to a utility method, or even better since it is used in exactly one place just inlined into `provideCatalogMap`

Deamon5550 (changes requested):

src/main/java/org/spongepowered/common/registry/RegistryMapImpl.java
kashike: The `RegistryMap` interface and `RegistryMapImpl` came from my old PR. In any case, it handles things like ensuring `values` is unmodifiable, `getOptional` wrapper, and `forCatalogRegistration` currently. It results in a lot less duplicated code throughout all the modules.
It can be collapsed into one class now that it isn't a `String` key.

src/main/java/org/spongepowered/common/registry/RegistryMap.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * A special map for use in registry modules.
+ *
+ * @param <V> The value type
+ */
+public final class RegistryMap<V> extends HashMap<CatalogKey, V> {
```
liach: How about extending a `ForwardingMap` instead?

PR #1557, opened 2017-10-04
view distance

kashike:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1659) | [**SpongeCommon**](https://github.com/SpongePowered/SpongeCommon/pull/1557)

### DIFF
--- a/src/main/java/org/spongepowered/common/config/category/WorldCategory.java
+++ b/src/main/java/org/spongepowered/common/config/category/WorldCategory.java
@@ -103,6 +103,15 @@ public class WorldCategory extends ConfigCategory {
 
     @Setting(value = "weather-ice-and-snow", comment = "Enable to allow the natural formation of ice and snow in supported biomes.")
     private boolean weatherIceAndSnow = true;
+
+    public static final int USE_SERVER_VIEW_DISTANCE = -1;
+    @Setting(
+            value = "view-distance",
+            comment = "The view distance."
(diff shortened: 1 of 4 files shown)


Deamon5550 (approved):

src/main/java/org/spongepowered/common/config/category/WorldCategory.java
```diff
@@ -103,6 +103,15 @@
 
     @Setting(value = "weather-ice-and-snow", comment = "Enable to allow the natural formation of ice and snow in supported biomes.")
     private boolean weatherIceAndSnow = true;
+
+    public static final int USE_SERVER_VIEW_DISTANCE = -1;
```
JBYoshi, line 107: Should this be exposed in the API?

VapidLinus: Would a `clearCustomViewDistance()` method make sense instead?

src/main/java/org/spongepowered/common/config/category/WorldCategory.java
```diff
@@ -216,4 +225,12 @@ public boolean getWeatherThunder() {
...
     }
+
+    public int getViewDistance() {
+        return this.viewDistance;
+    }
+
+    public void setViewDistance(final int viewDistance) {
+        this.viewDistance = viewDistance;
+    }
 }
```
parlough: New line while you're at it?

parlough (approved):

PR #1565, opened 2017-10-12
Optimize and improve experience calculations.

JBYoshi:
This PR has two primary changes. First, it replaces the float-based "experience since level" calculations with integer-based ones [to avoid rounding errors](https://github.com/SpongePowered/SpongeCommon/pull/1358). Second, it replaces the loops used to calculate experience in a given level with `Math.pow()` and `Math.sqrt()` for performance and ease of updating.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/manipulator/mutable/entity/SpongeExperienceHolderData.java
+++ b/src/main/java/org/spongepowered/common/data/manipulator/mutable/entity/SpongeExperienceHolderData.java
@@ -24,6 +24,8 @@
  */
 package org.spongepowered.common.data.manipulator.mutable.entity;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
 import org.spongepowered.api.data.DataContainer;
 import org.spongepowered.api.data.key.Keys;
 import org.spongepowered.api.data.manipulator.immutable.entity.ImmutableExperienceHolderData;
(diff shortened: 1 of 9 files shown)


me4502:
Have you tested performance of this? Loops can be potentially vectorised, and sqrt is generally slow. May be worth setting up a micro benchmark to compare so the performance claim holds up.

JBYoshi:
Test code:
```java
DataManipulatorBuilder<ExperienceHolderData, ImmutableExperienceHolderData> builder =
        Sponge.getDataManager().getManipulatorBuilder(ExperienceHolderData.class).get();
long startTime = System.nanoTime();
for (int i = 0; i < 100000; i++) {
    builder.create().set(Keys.TOTAL_EXPERIENCE, 9001);
}
long endTime = System.nanoTime();
p.sendMessage(Text.of(String.format("%.6f milliseconds", (endTime - startTime) / 1000000.0)));
```

Original:
196.7 ms, 156.6 ms, 116.2 ms, 130.9 ms, 121.7 ms, 117.9 ms, 131.9 ms, 116.6 ms, 121.8 ms

Modified:
139.5 ms, 98.2 ms, 119.7 ms, 88.2 ms, 78.2 ms, 81.9 ms, 74.4 ms, 81.2 ms, 46.3 ms

(Note that I ran both of these in the debugger. However, I did nothing else in it before running the tests, and I did not change any configurations.)

PR #1573, opened 2017-10-27
Add exhaustion to damage source

parlough:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1674) | **SpongeCommon**

Adds the ability to set and retrieve the exhaustion of a damage source. Also adds a test plugin and fixes some small comments. I'm open to changes of the defaults for exhaustion and where I set the exhaustion when building the damage sources.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/damage/DamageEventHandler.java
+++ b/src/main/java/org/spongepowered/common/event/damage/DamageEventHandler.java
@@ -105,7 +105,7 @@ public class DamageEventHandler {
 
     public static Optional<List<DamageFunction>> createArmorModifiers(EntityLivingBase entityLivingBase,
             DamageSource damageSource, double damage) {
-        if (!damageSource.isDamageAbsolute()) {
+        if (!damageSource.isUnblockable()) {
             damage *= 25;
             net.minecraft.item.ItemStack[] inventory = Iterables.toArray(entityLivingBase.getArmorInventoryList(), net.minecraft.item.ItemStack.class);
             List<DamageFunction> modifiers = new ArrayList<>();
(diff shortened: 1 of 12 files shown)


testplugins/src/main/java/org/spongepowered/test/DamageSourceTest.java
```diff
@@ -0,0 +1,109 @@
...
+                            return CommandResult.success();
+                        })
+                        .arguments(GenericArguments.doubleNum(Text.of("damage")))
+                        .build(),
+                "dsdamage");
+    }
+
+    @Listener(order = Order.POST)
+    public void onPlayerDamage(DamageEntityEvent event, @Getter("getTargetEntity") Player player) {
+        event.getCause().first(DamageSource.class).ifPresent(d -> {
```
gabizou: Could do with another filter instead of a nested lambda.

testplugins/src/main/java/org/spongepowered/test/DamageSourceTest.java
```diff
@@ -0,0 +1,109 @@
...
+public final class DamageSourceTest {
+
+    private final Set<UUID> activated = new HashSet<>();
+
+    @Listener
+    public void onInit(GameInitializationEvent event) {
+        Sponge.getCommandManager().register(this,
+                CommandSpec.builder()
+                        .executor((src, args) -> {
+                            if (src instanceof Player) {
```
gabizou, line 59: If non player command sources aren’t valid, should return a specific type of return, right @dualspiral?

gabizou (commented):
Overall looks clean. Just a brief review from the phone. Will dive deeper in a bit.

testplugins/src/main/java/org/spongepowered/test/DamageSourceTest.java
parlough: Correct. I just put this test together quickly I guess. Will clean it up and address both issues.

parlough: Issue is addressed.

testplugins/src/main/java/org/spongepowered/test/DamageSourceTest.java
parlough: Addressed.

testplugins/src/main/java/org/spongepowered/test/DamageSourceTest.java
dualspiral: Yeah, an exception should be thrown, as it is now.

parlough:
So we've been calculating damage wrong in some cases and there there was a problem where we calculated enchantment damage modifiers even if the damage was absolute, so I added a check there. We were also not calculating armor damage modifiers if the damage was absolute when we should have been, so I removed that check and replaced it with the proper check to see if the damage source is bypassing armor. 

Other concerns have been addressed as well, most were in the API.

src/main/java/org/spongepowered/common/event/damage/DamageEventHandler.java
```diff
@@ -227,78 +227,80 @@ public static EquipmentType resolveEquipment(int slot) {
...
-        boolean first = true;
-        int totalModifier = 0;
-        for (net.minecraft.item.ItemStack itemStack : inventory) {
-            if (itemStack.isEmpty()) {
-                continue;
-            }
-            NBTTagList enchantmentList = itemStack.getEnchantmentTagList();
-            if (enchantmentList == null) {
-                continue;
+        if (!damageSource.isDamageAbsolute()) {
```
parlough, line 230: No idea why this sections looks like it has huge differences in different places. This was the only change I made, obviously indenting everything inside.

Meronat:
Will update this once I get a chance soon!

parlough:
I will update this soon(hopefully tonight or tomorrow), most likely splitting out the fixes to our damage handling. Cheers.

Dockter:
Hope to see this in API 7

parlough:
@Dockter I'll try to get it done tonight then!

PR #1575, opened 2017-10-28
Implement cooldown tracker API

parlough:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1675) | **SpongeCommon**

Implements the cooldown tracker API and adds a test plugin. Check out the API PR for a little more information.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/util/MixinCooldownTracker.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/util/MixinCooldownTracker.java
@@ -1,0 +1,104 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.mixin.core.util;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import net.minecraft.item.Item;
+import net.minecraft.util.CooldownTracker;
+import org.spongepowered.api.item.ItemType;
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+
+import java.util.Map;
+import java.util.OptionalDouble;
+import java.util.OptionalInt;
+
+@Mixin(CooldownTracker.class)
+public abstract class MixinCooldownTracker implements org.spongepowered.api.entity.living.player.CooldownTracker {
+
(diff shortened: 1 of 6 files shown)


src/main/java/org/spongepowered/common/mixin/core/util/MixinCooldownTracker.java
```diff
@@ -0,0 +1,89 @@
...
+import org.spongepowered.api.item.ItemType;
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+
+import java.util.Map;
+import java.util.OptionalDouble;
+import java.util.OptionalInt;
+
+@Mixin(value = CooldownTracker.class)
```
ST-DDT: You can omit the `value =` part because it is implicit.

src/main/resources/META-INF/common_at.cfg
```diff
@@ -247,6 +247,9 @@ public net.minecraft.tileentity.TileEntityHopper func_145896_c(I)V # setTransfer
 public net.minecraft.tileentity.TileEntitySkull field_145910_i # skullRotation
 
 protected-f net.minecraft.util.EntityDamageSourceIndirect field_76387_p # indirectEntity
+public net.minecraft.util.CooldownTracker$Cooldown
+public net.minecraft.util.CooldownTracker$Cooldown field_185137_a # Lnet/minecraft/util/CooldownTracker$Cooldown;createTicks:I
+public net.minecraft.util.CooldownTracker$Cooldown field_185138_b # Lnet/minecraft/util/CooldownTracker$Cooldown;expireTicks:I
```
ST-DDT: Afaict you dont use these and thus they can be removed.

parlough: I added these to access the inner `Cooldown` class and its field `expireTicks`as they weren't public. Is there a better way to do this? Don't really know much about this piece. 

I can remove the createTicks one though as I don't use it anymore.

src/main/java/org/spongepowered/common/mixin/core/util/MixinCooldownTracker.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Override
+    public boolean hasCooldown(final ItemType type) {
+        checkNotNull(type, "Item type cannot be null!");
+        return hasCooldown((Item) type);
+    }
+
+    @Override
+    public OptionalInt getCooldown(final ItemType type) {
+        checkNotNull(type, "Item type cannot be null!");
```
gabizou, line 58: I'd even go as far as sanity checking the `ItemType` is an `instanceof` `Item`. There are some corner cases where an `ItemType` in `ItemTypes` may not have been properly mapped. At that point, we can safely and properly return the proper description instead of getting a `ClassCastException`.

src/main/java/org/spongepowered/common/mixin/core/util/MixinCooldownTracker.java
```diff
@@ -0,0 +1,94 @@
...
+            if (remainingCooldown > 0) {
+                return OptionalInt.of(remainingCooldown);
+            }
+        }
+        return OptionalInt.empty();
+    }
+
+    @Override
+    public void setCooldown(final ItemType type, final int ticks) {
+        checkNotNull(type, "Item type cannot be null!");
```
gabizou, line 74: Same here.

src/main/java/org/spongepowered/common/mixin/core/util/MixinCooldownTracker.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    public void resetCooldown(final ItemType type) {
+        setCooldown(type, 0);
+    }
+
+    @Override
+    public OptionalDouble getFractionRemaining(final ItemType type) {
+        checkNotNull(type, "Item type cannot be null!");
+
+        final float cooldown = getCooldown((Item) type, 0);
```
gabizou: And here.

testplugins/src/main/java/org/spongepowered/test/CooldownTrackerTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+@Plugin(id = "cooldowntracker", name = "Cooldown Tracker", description = "A plugin to test the cooldown tracker.")
+public final class CooldownTrackerTest {
+
+    @Listener
+    public void onInit(GameInitializationEvent event) {
+        Sponge.getCommandManager().register(this,
+                CommandSpec.builder()
+                        .executor((src, args) -> {
+                            if (src instanceof Player) {
```
gabizou: Inverse the if statement so the main block of code has less spacing as necessary.

testplugins/src/main/java/org/spongepowered/test/CooldownTrackerTest.java
```diff
@@ -0,0 +1,92 @@
...
+                            }
+                            throw new CommandException(Text.of(TextColors.RED, "You must be a player to execute this command!"));
+                        })
+                        .build(),
+                "cooldowntest");
+
+        Sponge.getCommandManager().register(this,
+                CommandSpec.builder()
+                        .executor((src, args) -> {
+                            if (src instanceof Player) {
```
gabizou: And here.

gabizou (changes requested):
Minor sanity checks and style changes, otherwise good.

parlough:
@gabizou I believe I have addressed your concerns. I wasn't sure the best way to go about it, so how does this look?

PR #1579, opened 2017-10-29
Allow Sponge to use Java 9

darktohka:
After Java 9, version strings do not match the 1.x.x scheme anymore.

For example: Java 8 - 1.8.0_144
Java 9 - 9
Java 9.0.1 - 9.0.1

Because of this change, Sponge thinks that Java 9 is an older version than Java 8, and terminates the server on startup. This pull request proposes a solution to this problem.


stephan-gh:
See https://github.com/SpongePowered/SpongeVanilla/issues/355#issuecomment-335770593. Removing the version check won't make Sponge magically work on Java 9. If anything, you should change the version checker to say that Java 9 is not (yet) compatible with Sponge.

PR #1596, opened 2017-11-14
SF-1836: Fallback optimzation to vanilla if null

ryantheleach:
Fixes: https://github.com/SpongePowered/SpongeForge/issues/1836

If marking the entity would fail with a NPE due to mods passing in null to the 'unused' variable in vanilla, do the calculations that would normally be bypassed.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
@@ -239,7 +239,7 @@ public abstract class MixinWorld implements World, IMixinWorld {
     @Shadow public abstract void onEntityRemoved(net.minecraft.entity.Entity entityIn);
     @Shadow public abstract void updateEntity(net.minecraft.entity.Entity ent);
     @Shadow public abstract boolean isBlockLoaded(BlockPos pos);
-    @Shadow public abstract void markChunkDirty(BlockPos pos, net.minecraft.tileentity.TileEntity unusedTileEntity);
+    @Shadow public void markChunkDirty(BlockPos pos, net.minecraft.tileentity.TileEntity unusedTileEntity){};
    // @Shadow public abstract List<Entity> getEntitiesInAABBexcluding(@Nullable net.minecraft.entity.Entity entityIn, AxisAlignedBB boundingBox, @Nullable Predicate <? super net.minecraft.entity.Entity > predicate);
     @Shadow public abstract boolean addWeatherEffect(net.minecraft.entity.Entity entityIn);
     @Shadow public abstract Biome getBiome(BlockPos pos);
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -1929,6 +1929,10 @@ protected boolean isChunkLoaded(int x, int z, boolean allowEmpty) {
     @Override
     public void markChunkDirty(BlockPos pos, net.minecraft.tileentity.TileEntity unusedTileEntity)
     {
+        if(unusedTileEntity == null){
```
kashike: formatting

kashike (changes requested):

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
```diff
@@ -239,7 +239,7 @@
     @Shadow public abstract void onEntityRemoved(net.minecraft.entity.Entity entityIn);
     @Shadow public abstract void updateEntity(net.minecraft.entity.Entity ent);
     @Shadow public abstract boolean isBlockLoaded(BlockPos pos);
-    @Shadow public abstract void markChunkDirty(BlockPos pos, net.minecraft.tileentity.TileEntity unusedTileEntity);
+    @Shadow public void markChunkDirty(BlockPos pos, net.minecraft.tileentity.TileEntity unusedTileEntity){};
```
dualspiral, line 242: This change isn't necessary - unless the method is private, shadow methods should be marked abstract (even if the parent isn't)

ryantheleach: Maybe it's because I didn't have a body then, thanks.

ryantheleach: @dualspiral What do I need to do to fix it then? As far as I'm aware it is necessary, as MixinWorldServer needs to be able to call the super method if I want the code to compile. Since it's shadowed it's just a placeholder anyway right?

simon816: Yeah @ryantheleach is correct, the call to `super` requires the parent method to not be abstract.

ryantheleach: Validated it in debugging by making hoppers pass items back and forth, then used evaluate expression to set the local to null.

It correctly ran the vanilla method.

dualspiral: Ah, my mistake!

gabizou (approved):

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java
liach: I suggest throwing an `AbstractMethodError` here (to be consistent with shadowed static methods)

ryantheleach: @liach Is there a reason for that? it takes up more room, so was wondering if there was ever a usecase where it could be called. Also why AbstractMethodError?

gabizou: There isn't a use case. Shadowed methods that aren't existing will throw an exception at runtime for mixin application anyways.

PR #1601, opened 2017-11-18
Restore Forge Compat for PlayerSetSpawnEvent

phit:
if playerIn never slept, don't give newPlayer a null bedposition, it will achieve nothing and just fire an event mods don't expect

as found by https://github.com/SpongePowered/SpongeForge/issues/1860


ryantheleach:
The original calls `BlockPos blockpos1 = EntityPlayer.getBedSpawnLocation(this.mcServer.getWorld(playerIn.dimension), blockpos, flag);
`

Before setting the spawnpoint. Does this have a dramatic change of behavior?

Given the date, I suggest someone needs to fully review the Overwrite, with phits merged as a quick fix.

PR #1603, opened 2017-11-19
Use active config, instead of always world config

phit:
Use active config like everywhere else where it isn't currently used.
Fixes SpongePowered/SpongeForge#1201


kashike (approved):

kashike:
Thanks, @phit.

PR #1605, opened 2017-11-19
[WIP] Refactor and implement ChangeDataHolderEvent.ValueChange

Aaron1011:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1695) | **SpongeCommon** | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1866)

See the SpongeAPI PR for more details.

This is the basic pattern for firing a `ChangeDataHolderEvent.ValueChange` event for a given key:

1. Redirect all locations in Vanilla/Forge where the value is set. Ideally, there will be a public setter method for you to inject into.
2. Instead of setting the underlying field (or `DataParameter, etc`), call `offerWithEvent` on the appropriate `DataHolder`, putting any relevant objects into the `Cause`
3. In the `ValueProcessor` for the key in question, **bypass** the setter you redirected before, if applicable. This ensures that setting the value does not end up recursing infinitely between your redirected setter and the value processor.

Basically, the idea is to ensure that all changes done by mods/vanilla/forge end up firing an event, while API changes (e.g. anything going through the value processor) does **not**. The decision as to whether or not to fire an event for a plugin change is handled on the API side, and is not the implementation's concern.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/AbstractArchetype.java
+++ b/src/main/java/org/spongepowered/common/data/AbstractArchetype.java
@@ -199,10 +199,22 @@ public abstract class AbstractArchetype<C extends CatalogType, S extends Locatab
                 .flatMap(processor -> processor.readValue(this.data));
     }
 
+    @SuppressWarnings("unchecked")
+    @Override
+    public <R> Optional<R> getDefault(Key<? extends BaseValue<R>> key) {
+        return DataUtil.getNbtProcessor(this.getDataType(), key)
+                .flatMap(processor -> processor.readFrom(this.data).map(v -> v.getDefault()));
+    }
+
     @Override
     public <R, V extends BaseValue<R>> Optional<V> getValue(Key<V> key) {
         return DataUtil.getNbtProcessor(this.getDataType(), key)
                 .flatMap(processor -> processor.readFrom(this.data));
+    }
+
+    @Override
+    public <R, V extends BaseValue<R>> Optional<V> getDefaultValue(Key<V> key) {
+        throw new UnsupportedOperationException(); // TODO - can we construct a new value here?
     }
 
     @Override
--- a/src/main/java/org/spongepowered/common/data/AbstractArchetype.java
+++ b/src/main/java/org/spongepowered/common/data/AbstractArchetype.java
@@ -42,6 +42,8 @@ import org.spongepowered.api.data.persistence.InvalidDataException;
 import org.spongepowered.api.data.value.BaseValue;
 import org.spongepowered.api.data.value.immutable.ImmutableValue;
 import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.api.event.cause.Cause;
+import org.spongepowered.api.event.data.ChangeDataHolderEvent;
 import org.spongepowered.common.SpongeImpl;
 import org.spongepowered.common.data.nbt.CustomDataNbtUtil;
 import org.spongepowered.common.data.nbt.NbtDataType;
@@ -120,6 +122,11 @@ public abstract class AbstractArchetype<C extends CatalogType, S extends Locatab
         return DataUtil.getNbtProcessor(this.getDataType(), key)
                 .map(processor -> processor.offer(this.data, value))
                 .orElseGet(DataTransactionResult::failNoData);
+    }
+
+    @Override
+    public <E> Optional<ChangeDataHolderEvent.ValueChange> offerWithEvent(Key<? extends BaseValue<E>> key, E value, Cause cause) {
+        throw new UnsupportedOperationException();
     }
 
     @SuppressWarnings({ "unchecked", "rawtypes" })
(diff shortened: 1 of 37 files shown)


src/main/java/org/spongepowered/common/data/AbstractArchetype.java
```diff
@@ -122,6 +124,11 @@ public boolean supports(Class<? extends DataManipulator<?, ?>> holderClass) {
                 .orElseGet(DataTransactionResult::failNoData);
     }
 
+    @Override
+    public <E> Optional<ChangeDataHolderEvent.ValueChange> offerWithEvent(Key<? extends BaseValue<E>> key, E value, Cause cause) {
+        throw new UnsupportedOperationException();
```
gabizou, line 129: Message maybe?

src/main/java/org/spongepowered/common/data/AbstractArchetype.java
```diff
@@ -192,12 +199,24 @@ public DataTransactionResult copyFrom(DataHolder that, MergeFunction function) {
                 .flatMap(processor -> processor.readValue(this.data));
     }
 
+    @SuppressWarnings("unchecked")
+    @Override
+    public <R> Optional<R> getDefault(Key<? extends BaseValue<R>> key) {
+        return DataUtil.getNbtProcessor(this.getDataType(), key)
+                .flatMap(processor -> processor.readFrom(this.data).map(v -> v.getDefault()));
```
gabizou, line 206: multi-line the nested mapping. In the event there's an exception, we want to be able to pinpoint which lambda is having an issue.

src/main/java/org/spongepowered/common/data/DataProcessor.java
```diff
@@ -97,6 +98,8 @@
      */
     DataTransactionResult set(DataHolder dataHolder, M manipulator, MergeFunction function);
 
+    //void setWithEvent(DataHolder container, M manipulator, Cause cause);
```
gabizou, line 101: Intended to be used later maybe? comments welcome.

src/main/java/org/spongepowered/common/data/processor/common/AbstractBlockOnlyDataProcessor.java
```diff
@@ -25,13 +25,16 @@
...
+import org.spongepowered.api.data.DataHolder;
 import org.spongepowered.api.data.DataTransactionResult;
 import org.spongepowered.api.data.key.Key;
 import org.spongepowered.api.data.manipulator.DataManipulator;
 import org.spongepowered.api.data.manipulator.ImmutableDataManipulator;
 import org.spongepowered.api.data.value.BaseValue;
 import org.spongepowered.api.data.value.ValueContainer;
 import org.spongepowered.api.data.value.immutable.ImmutableValue;
+import org.spongepowered.api.event.cause.Cause;
+import org.spongepowered.api.event.data.ChangeDataHolderEvent;
```
gabizou, line 37: imports?

src/main/java/org/spongepowered/common/mixin/core/util/MixinFoodStats.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Shadow private int foodLevel;
+    private Player player; // Null when on client
+
+    @Override
+    public void setPlayer(EntityPlayer player) {
+        this.player = (Player) player;
+    }
+
+    @Inject(method = "addStats(Lnet/minecraft/item/ItemFood;Lnet/minecraft/item/ItemStack;)V", at = @At(value = "HEAD"))
```
gabizou, line 55: This should be a redirect if you're modifying the cause stack, this way you can wrap it in a try with autocloseable frames.

testplugins/src/main/java/org/spongepowered/test/FoodChangeTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.spongepowered.api.event.filter.cause.First;
+import org.spongepowered.api.event.filter.data.GetKey;
+import org.spongepowered.api.plugin.Plugin;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.channel.MessageChannel;
+
+import java.util.Optional;
+
+@Plugin(id = "food-change-test", authors = "Aaron1011")
+public class FoodChangeTest {
```
gabizou, line 23: If this is going to constantly print to the player, should make it a toggle to join in (reduces the amount of commenting when debugging implementation).

testplugins/src/main/java/org/spongepowered/test/InjectionTest.java
```diff
@@ -25,6 +25,8 @@
 package org.spongepowered.test;
 
 import org.slf4j.Logger;
+import org.spongepowered.api.data.key.Keys;
```
gabizou, line 28: imports.

gabizou (changes requested):
Small changes, otherwise looks good. Aside from API discussions we've had about removing GetKey annotation.

parlough:
Going to close this for now due to https://github.com/SpongePowered/SpongeAPI/pull/1728 and https://github.com/SpongePowered/SpongeCommon/pull/1670 which provide a method of implementing and listening to the event. 

If you'd like to make other changes, open a new PR with those in relation to what is now in the API and implementation.

PR #1606, opened 2017-11-19
getConnectedChest

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1697) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1606)

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/tileentity/MixinTileEntityChest.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/tileentity/MixinTileEntityChest.java
@@ -50,8 +50,10 @@ import org.spongepowered.common.item.inventory.lens.impl.collections.SlotCollect
 import org.spongepowered.common.item.inventory.lens.impl.comp.GridInventoryLensImpl;
 import org.spongepowered.common.item.inventory.util.InventoryUtil;
 
+import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 
 @NonnullByDefault
 @Mixin(TileEntityChest.class)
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/core/tileentity/MixinTileEntityChest.java
```diff
@@ -170,5 +172,13 @@ public void supplyVanillaManipulators(List<DataManipulator<?, ?>> manipulators)
         return InventoryUtil.getDoubleChestInventory(((TileEntityChest)(Object) this));
     }
 
+    @Override
+    public Optional<Chest> getConnectedChest() {
+        TileEntityChest chest = ObjectUtils.firstNonNull(this.adjacentChestXNeg, this.adjacentChestXPos, this.adjacentChestZNeg, this.adjacentChestZPos);
+        if (chest == null) {
+            return Optional.empty();
+        }
+        return Optional.of(((Chest) chest));
```
parlough: `return Optional.ofNullable((Chest) chest);`

ryantheleach:
In previous versions of Minecraft it was possible to create bugged 3 wide chests.
That + tests in how well it's going to play with mods is what's holding me back from 👍

parlough:
This is simply convenience for chests and I'm assuming mods which do use this class will use the adjacent fields to set if they are connected anywhere, if not there's not much we can do about it with this or connected direction data. We could include a set of all connected chests from each direction, if you're worried about there being ones connected in other directions, and then specify it would be empty if there were none.

Faithcaio:
As as always modded is difficult.
Is there something in Forge that marks blocks as a multiblock structure?
Also see https://github.com/SpongePowered/SpongeAPI/pull/1494.

I'll check the bugged chest thing later today...

ryantheleach:
LGTM

PR #1608, opened 2017-11-22
Implement ConstructPortalEvent

### LINKED ISSUE #1559: Implementation Request: ConstructPortalEvent
This event doesn't seem to be implemented as it doesn't fire when I build a portal. Could you please implement it? Thanks a lot.

Edit: It doesn't fire when travelling through a portal creates another exit portal either.

JBYoshi:
Supports:
- [x] Lighting nether portals
- [x] Nether portals created after teleportation
- [x] Inserting ender eyes into End portals
- [ ] Spawning the End exit portal

Fixes #1559.

### DIFF
--- a/src/main/java/org/spongepowered/common/entity/EntityUtil.java
+++ b/src/main/java/org/spongepowered/common/entity/EntityUtil.java
@@ -30,6 +30,7 @@ import com.flowpowered.math.vector.Vector3d;
 import com.flowpowered.math.vector.Vector3i;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
+import com.google.common.collect.Lists;
 import net.minecraft.advancements.CriteriaTriggers;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.EntityList;
@@ -63,6 +64,7 @@ import net.minecraft.world.WorldProviderEnd;
 import net.minecraft.world.WorldProviderSurface;
 import net.minecraft.world.WorldServer;
 import org.spongepowered.api.Sponge;
+import org.spongepowered.api.block.BlockSnapshot;
 import org.spongepowered.api.entity.EntityType;
 import org.spongepowered.api.entity.Transform;
 import org.spongepowered.api.entity.explosive.FusedExplosive;
@@ -74,7 +76,9 @@ import org.spongepowered.api.event.cause.EventContextKeys;
 import org.spongepowered.api.event.cause.entity.teleport.TeleportTypes;
 import org.spongepowered.api.event.entity.MoveEntityEvent;
 import org.spongepowered.api.event.entity.SpawnEntityEvent;
+import org.spongepowered.api.event.world.ConstructPortalEvent;
 import org.spongepowered.api.item.inventory.ItemStackSnapshot;
+import org.spongepowered.api.world.BlockChangeFlags;
 import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.PortalAgent;
 import org.spongepowered.api.world.World;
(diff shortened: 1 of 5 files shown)


testplugins/src/main/java/org/spongepowered/test/PortalTest.java
```diff
@@ -0,0 +1,68 @@
+package org.spongepowered.test;
```
kashike, line 25: license

kashike (changes requested):

testplugins/src/main/java/org/spongepowered/test/PortalTest.java
JBYoshi: Fixed.

ryantheleach:
Consider adding support not only for the end exit portal, but end gateways.

Zidane:
@JBYoshi 

Status update?

JBYoshi:
@Zidane I found out that implementing this is going to be somewhat more difficult than I thought. My current approach for detecting portal constructions during teleports (by assuming all block placements during a teleport are part of a new portal) is unreliable. On the one hand, it doesn't work for detecting portals created for the first time. On the other hand, it improperly detects the [obsidian platform](https://minecraft.gamepedia.com/Generated_structures#Obsidian_platform) created when entering the End as a portal. I'll probably try to detect whether a block is a portal, but I don't know of a simple way to do that.

ryantheleach:
@JBYoshi I'd consider that platform as part of the portal personally, it's just the receiving end, Just like nether portals have a base when they are forced into air, the one way only exit of the end portal has a set platform in the end at a specific place in the end. As long as it's documented it's fine IMO, unless you can think of some good reasons why we would want to not include blocks created when a portal spawns in the events?

The main usecase I see, is people wanting to prevent all of the blocks being placed / created, so this is fine, maybe some docs or renaming of the method is required to make it more clear.

testplugins/src/main/java/org/spongepowered/test/PortalTest.java
```diff
@@ -0,0 +1,101 @@
...
+    private static final Random RANDOM = new Random();
+    /**
+     * There is no significance in this UUID. I just created it so I didn't
+     * have to go through the effort of setting up custom data.
+     */
+    private static final UUID CREATOR_UUID = UUID.fromString("7fac20d9-cbd4-48c7-88dc-61a33e22cd74");
+
+    @Listener
+    public void onCreatePortal(ConstructPortalEvent e) {
+        if (RANDOM.nextInt(5) == 0) {
```
gabizou: Make this a toggle as well, I don't want to randomly be testing portals and wondering what this new mechanic where they suddenly explode :P

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemEnderEye.java
```diff
@@ -47,4 +58,26 @@ private boolean onspawnEntity(World world, Entity enderEye, World worldIn, Entit
...
+            "Lnet/minecraft/block/state/pattern/BlockPattern;match(Lnet/minecraft/world/World;Lnet/minecraft/util/math/BlockPos;)Lnet/minecraft/block/state/pattern/BlockPattern$PatternHelper;"))
+    @Nullable
+    private BlockPattern.PatternHelper onConstructPortal(BlockPattern pattern, World worldIn, BlockPos pos) {
+        BlockPattern.PatternHelper match = pattern.match(worldIn, pos);
+        if (match == null) {
+            return null;
+        }
+        Vector3d center = VecHelper.toVector3d(match.getFrontTopLeft())
+                .sub(pattern.getThumbLength() / 2.0, pattern.getFingerLength() / 2.0, pattern.getPalmLength() / 2.0);
+        ConstructPortalEvent event = SpongeEventFactory.createConstructPortalEvent(Sponge.getCauseStackManager().getCurrentCause(),
```
gabizou: Would we be able to add the blocks being changed for the portal?

src/main/java/org/spongepowered/common/entity/EntityUtil.java
```diff
@@ -346,6 +361,17 @@ public static EntityPlayerMP transferPlayerToWorld(final EntityPlayerMP player,
                 // TODO determine what it truly is for API 8
                 frame.addContext(EventContextKeys.TELEPORT_TYPE, TeleportTypes.PORTAL);
 
+                if (!context.getCapturedBlockChanges().isEmpty()) {
+                    ConstructPortalEvent constructPortalEvent = SpongeEventFactory.createConstructPortalEvent(Sponge.getCauseStackManager()
```
gabizou: Can shortcut with `frame.getCurrentcause()` since we are already in a frame.

Zidane:
@JBYoshi Either update this or close it.

JBYoshi:
I'm going to go ahead and close this. This code seems to have broken since I last looked at it, and reimplementing it seems to be much more trouble than it's worth. I'm thinking of just getting rid of `ConstructPortalEvent` entirely. `ChangeBlockEvent` should suffice for most cases, and it provides much more information than `ConstructPortalEvent` could.

PR #1610, opened 2017-11-23
Fix players being sent back through a nether portal they just used.

JBYoshi:
While playing around with #1608, I stumbled across this bug in `bleeding`. Immediately after entering a nether portal, the realtime mixin would send the player right back due to the slight burst of lag from loading the nether messing with a trigger in the cooldown code. This PR "snags" the cooldown process for a tick, forcing the collision test to run before re-teleporting the player.

Old process:
- `Entity` sets `timeUntilPortal` to 10 following the original teleportation
- `MixinEntityPlayerMP` updates `timeUntilPortal` to 9
- `BlockPortal` sees `timeUntilPortal` is non-zero and resets it to 10
- `MixinEntityPlayerMP` updates `timeUntilPortal` by (say) 20 ticks, due to a 1-second burst of lag, dropping it to 0
- `BlockPortal` sees `timeUntilPortal` is 0 and initiates a teleport

New process:
- `Entity` sets `timeUntilPortal` to 10 following the original teleportation
- `MixinEntityPlayerMP` updates `timeUntilPortal` to 9
- `BlockPortal` sees `timeUntilPortal` is non-zero and resets it to 10
- `MixinEntityPlayerMP` updates `timeUntilPortal` by (say) 20 ticks, due to a 1-second burst of lag, but "snags" it at 1 for a tick (if it stayed at 1, it would then change to 0 in the following tick)
- `BlockPortal` sees `timeUntilPortal` is non-zero and resets it to 10


kashike (approved):

ryantheleach:
Is this better then just removing the realtime interaction mixin in regards to portal handling?

Can you also describe the vanilla behavior? I'm having difficulty understanding the logic difference, especially the reasoning behind BlockPortal resetting the time to 10.

PR #1611, opened 2017-11-23
Clean up the dye color processors and add support for beds.

JBYoshi:
- Replace the multi-target `DataProcessor` with individually-targeted dual processors.
- Implement `DyeableData` for beds. See SpongePowered/SpongeAPI#1700.


src/main/java/org/spongepowered/common/data/DataRegistrar.java
```diff
@@ -762,9 +770,7 @@ public static void setupSerialization() {
         DataUtil.registerValueProcessor(Keys.VANISH, new VanishValueProcessor());
         DataUtil.registerValueProcessor(Keys.VANISH_IGNORES_COLLISION, new VanishCollisionValueProcessor());
         DataUtil.registerValueProcessor(Keys.VANISH_PREVENTS_TARGETING, new VanishTargetValueProcessor());
-        DataUtil.registerValueProcessor(Keys.DYE_COLOR, new WolfDyeColorValueProcessor());
-        DataUtil.registerValueProcessor(Keys.DYE_COLOR, new SheepDyeColorValueProcessor());
```
gabizou, line 770: Where do these go now?

gabizou (changes requested):
I think you've removed some processor registrations you didn't intend to remove.

src/main/java/org/spongepowered/common/data/DataRegistrar.java
ImMorpheus: Aren't they registered in https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/data/util/DataUtil.java#L298 ?

gabizou: Ah yeah, subtle name change from value to data processor. good catch!

gabizou (approved):

src/main/java/org/spongepowered/common/data/DataRegistrar.java
```diff
@@ -762,9 +770,7 @@ public static void setupSerialization() {
...
-        DataUtil.registerValueProcessor(Keys.DYE_COLOR, new ItemDyeColorValueProcessor());
+        DataUtil.registerValueProcessor(Keys.DYE_COLOR, new WolfDyeColorDataProcessor());
```
ImMorpheus: Is this registered twice ?

src/main/java/org/spongepowered/common/data/processor/data/item/ItemDyeColorDataProcessor.java
```diff
@@ -22,33 +22,50 @@
...
+        final Block block = Block.getBlockFromItem(item);
+
+        return block.equals(Blocks.WOOL)
+                || block.equals(Blocks.STAINED_GLASS)
+                || block.equals(Blocks.CARPET)
+                || block.equals(Blocks.STAINED_GLASS_PANE)
+                || block.equals(Blocks.STAINED_HARDENED_CLAY)
+                || block.equals(Blocks.CONCRETE)
+                || block.equals(Blocks.CONCRETE_POWDER)
+                || block.equals(Blocks.BED);
```
liach: Are bed blocks and bed items paired? I suggest using `Items.BED` in checks.

JBYoshi: You are correct. Let me fix that.

src/main/java/org/spongepowered/common/data/DataRegistrar.java
JBYoshi: It is. 😐

JBYoshi: Fixed.

JBYoshi:
If there are no other comments, I'll merge this tonight.

Issue #651, opened 2016-04-15
Crash when too close to world limit border


ghost:
In the 'org.spongepowered.common.world.storage.SpongeChunkLayout' file, there is a maximum of coordinates in the world. In the vanilla server, when you go further than this limit, there is void, but 
it works. In SpongeCommon it throws an exception: 

Command to teleport: `/tp @p 29999999 100 29999999`.

```
[10:56:49 ERROR]: Encountered an unexpected exception
net.minecraft.util.ReportedException: Exception generating new chunk
    at net.minecraft.server.MinecraftServer.func_71190_q(SourceFile:177) ~[MinecraftServer.class:?]
    at net.minecraft.server.dedicated.DedicatedServer.func_71190_q(SourceFile:299) ~[ko.class:?]
    at net.minecraft.server.MinecraftServer.func_71217_p(SourceFile:535) ~[MinecraftServer.class:?]
    at net.minecraft.server.MinecraftServer.run(SourceFile:451) [MinecraftServer.class:?]
    at java.lang.Thread.run(Thread.java:745) [?:1.8.0_77]
Caused by: java.lang.IllegalArgumentException: Chunk coordinates are not valid1875001, -6
    at com.google.common.base.Preconditions.checkArgument(Preconditions.java:125) ~[minecraft_server.1.8.9.jar:?]
    at org.spongepowered.common.util.gen.ChunkPrimerBuffer.getBlockStart(ChunkPrimerBuffer.java:63) ~[ChunkPrimerBuffer.class:1.8.9-4.1.0-BETA-293]
    at org.spongepowered.common.util.gen.ChunkPrimerBuffer.<init>(ChunkPrimerBuffer.java:57) ~[ChunkPrimerBuffer.class:1.8.9-4.1.0-BETA-293]
    at org.spongepowered.common.world.gen.SpongeChunkProvider.func_73154_d(SpongeChunkProvider.java:206) ~[SpongeChunkProvider.class:1.8.9-4.1.0-BETA-293]
    at net.minecraft.world.gen.ChunkProviderServer.func_73158_c(SourceFile:87) ~[ld.class:?]
    at net.minecraft.world.gen.ChunkProviderServer.func_73154_d(SourceFile:116) ~[ld.class:?]
    at net.minecraft.world.World.func_72964_e(SourceFile:277) ~[adm.class:?]
    at net.minecraft.world.WorldServer.func_147456_g(SourceFile:327) ~[le.class:?]
[...]

PR #1612, opened 2017-11-23
Check the world border when teleporting entities.

### LINKED ISSUE #168: Add world limit protection measures
So, currently you can teleport beyond the limits of Minecraft if using the API. I don't believe this should be possible, and I think we probably shouldn't be crashing like this if someone manages to go beyond the generate-able world:

``` java
net.minecraft.util.ReportedException: Exception generating new chunk
        at net.minecraft.server.MinecraftServer.func_71190_q(MinecraftServer.java:699) ~[MinecraftServer.class:?]
        at net.minecraft.server.dedicated.DedicatedServer.func_71190_q(DedicatedServer.java:364) ~[po.class:?]
        at net.minecraft.server.MinecraftServer.func_71217_p(MinecraftServer.java:598) ~[MinecraftServer.class:?]
        at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:478) [MinecraftServer.class:?]
        at java.lang.Thread.run(Thread.java:745) [?:1.8.0_51]
Caused by: java.lang.IllegalStateException: Optional.get() cannot be called on an absent value
        at com.google.common.base.Absent.get(Absent.java:47) ~[minecraft_server.1.8.jar:?]
        at net.minecraft.world.chunk.Chunk.onConstructed(Chunk.java:88) ~[bfh.class:?]
        at net.minecraft.world.chunk.Chunk.<init>(Chunk.java:93) ~[bfh.class:?]
        at net.minecraft.world.chunk.Chunk.<init>(Chunk.java:97) ~[bfh.class:?]
        at net.minecraft.world.gen.ChunkProviderGenerate.func_73154_d(ChunkProviderGenerate.java:256) ~[bgv.class:?]
        at net.minecraft.world.gen.ChunkProviderServer.originalLoadChunk(ChunkProviderServer.java:162) ~[qs.class:?]
        at net.minecraft.world.gen.ChunkProviderServer.loadChunk(ChunkProviderServer.java:122) ~[qs.class:?]
        at net.minecraft.world.gen.ChunkProviderServer.func_73158_c(ChunkProviderServer.java:92) ~[qs.class:?]
        at net.minecraft.world.gen.ChunkProviderServer.func_73154_d(ChunkProviderServer.java:189) ~[qs.class:?]
        at net.minecraft.world.World.func_72964_e(World.java:304) ~[aqu.class:?]
        at net.minecraft.world.WorldServer.func_147456_g(WorldServer.java:380) ~[qt.class:?]
        at net.minecraft.world.WorldServer.func_72835_b(WorldServer.java:221) ~[qt.class:?]
        at net.minecraft.server.MinecraftServer.func_71190_q(MinecraftServer.java:693) ~[MinecraftServer.class:?]
        ... 4 more
```

JBYoshi:
Fixes #168.
Fixes SpongePowered/SpongeVanilla#357.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
@@ -467,6 +467,7 @@ public abstract class MixinEntity implements IMixinEntity {
     @Override
     public boolean setLocation(Location<World> location) {
         checkNotNull(location, "The location was null!");
+        checkWorldBorder(location.getExtent(), location.getPosition());
         if (isRemoved()) {
             return false;
         }
@@ -555,6 +556,7 @@ public abstract class MixinEntity implements IMixinEntity {
     // to avoid firing a DisplaceEntityEvent.Teleport
     @Override
     public void setLocationAndAngles(Location<World> location) {
+        checkWorldBorder(location.getExtent(), location.getPosition());
         if (((Entity) this) instanceof EntityPlayerMP) {
             ((EntityPlayerMP) (Object) this).connection.setPlayerLocation(location.getX(), location.getY(), location.getZ(), this.rotationYaw, this.rotationPitch);
         } else {
@@ -567,6 +569,7 @@ public abstract class MixinEntity implements IMixinEntity {
 
     @Override
     public void setLocationAndAngles(Transform<World> transform) {
+        checkWorldBorder(transform.getExtent(), transform.getPosition());
         Vector3d position = transform.getPosition();
         EntityPlayerMP player = null;
         if (((Entity) this) instanceof EntityPlayerMP) {
@@ -647,6 +650,14 @@ public abstract class MixinEntity implements IMixinEntity {
             }
         }
         return relocated;
+    }
+
+    private void checkWorldBorder(World world, Vector3d location) {
+        AxisAlignedBB aabb = this.getEntityBoundingBox().offset(location.getX() - this.posX, location.getY() - this.posY,
+                location.getZ() - this.posZ);
+        if (!((net.minecraft.world.World) world).worldBorder.contains(aabb)) {
+            throw new IllegalArgumentException("Location " + location + " is outside the world border");
+        }
     }
 
     @Inject(method = "onUpdate", at = @At("RETURN"))


Zidane:
imo this should never throw an exception because I can tell you that almost no plugin dev is going to check the world border before doing the teleport. It should return false on the location/etc setters.

ImMorpheus:
I think this is part of a larger issue.

I could teleport to 29998983 ~ 29998983 (valid), and start flying toward the worldborder. (#651 would happen)

ryantheleach:
@ImMorpheus Are you sure that happens on the latest version ( I would appreciate you testing #651) Because I've made several tests near the default world border (within the last 6 months or so) (not at those exact co-ordinates) and did not run into problems.

But maybe it's directional, or you need to expand or remove the default world border?

Regardless, I'd consider that a related but separate issue.

ImMorpheus:
Yes, I tested that on 1.12 (bleeding branch).

JoeSGT:
Will this prevent Entity from bypassing the world border thing's like ender pearl?

JoeSGT:
can this be added to API 7 ?

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -649,6 +652,14 @@ public boolean setLocationAndRotation(Location<World> location, Vector3d rotatio
         return relocated;
     }
 
+    private void checkWorldBorder(World world, Vector3d location) {
+        AxisAlignedBB aabb = this.getEntityBoundingBox().offset(location.getX() - this.posX, location.getY() - this.posY,
+                location.getZ() - this.posZ);
+        if (!((net.minecraft.world.World) world).worldBorder.contains(aabb)) {
```
Aaron1011, line 658: I really don't think this is what we want to do. It's perfectly fine for players and entities to be outside the WorldBorder - in fact, the [/wordborder command allows you to cause players to take damage when outside the border](https://minecraft.gamepedia.com/World_border).

Instead of checking `World.worldBorder`, you should check `ChunkLayout#toWorld`, to see if the coordinates actually correspond to a legal chunk (whether or not the x and z coordinates lie from `-30000000` to `30000000`)

Aaron1011 (changes requested):

gabizou:
Superseded by #1958

PR #1602, opened 2017-11-18
split up pickup event

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1693) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1602) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1863)

PR #1615, opened 2017-11-25
livingpickupevent

Faithcaio:
To expand on
https://github.com/SpongePowered/SpongeCommon/pull/1602
This will add Pickup Events for other Entities:
 - [x] EntityLiving
 - [x] Villager

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/passive/MixinEntityVillager.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/passive/MixinEntityVillager.java
@@ -26,10 +26,15 @@ package org.spongepowered.common.mixin.core.entity.passive;
 
 import static com.google.common.base.Preconditions.checkState;
 
+import net.minecraft.entity.EntityLiving;
+import net.minecraft.entity.item.EntityItem;
 import net.minecraft.entity.passive.EntityVillager;
 import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.inventory.EntityEquipmentSlot;
 import net.minecraft.inventory.IInventory;
 import net.minecraft.inventory.InventoryBasic;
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/mixin/core/entity/passive/MixinEntityVillager.java
```diff
@@ -237,4 +248,25 @@ public void supplyVanillaManipulators(List<DataManipulator<?, ?>> manipulators)
     public Optional<Villager> getCarrier() {
         return Optional.of(this);
     }
+
+    @Redirect(method = "updateEquipmentIfNeeded",  at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/passive/EntityVillager;canVillagerPickupItem(Lnet/minecraft/item/Item;)Z"))
+    public boolean onCanVillagerPickUpItem(EntityVillager villager, Item item, EntityItem itemEntity) {
+        boolean result = this.canVillagerPickupItem(item);
+        if (!SpongeCommonEventFactory.callChangeInventoryPickupPreEvent(((EntityLiving)(Object) this), itemEntity)) {
```
JBYoshi, line 255: This should only be fired if `canVillagerPickupItem()` returns `true`.

JBYoshi (changes requested):

Zidane:
@Faithcaio 

Update this for @JBYoshi 's feedback.

src/main/java/org/spongepowered/common/mixin/core/entity/passive/MixinEntityVillager.java
gabizou: Likewise, please javadoc what this method was originally doing if possible for future maintenance (as may be noticed in #1882)

gabizou:

Issue #436, opened 2016-01-16
BanService should provide the CommandSource of a pardon


viveleroi:
When a ban is added, the `addBan` method is called which allows me to access the command source. Yet `removeBan` doesn't appear to be called when pardoning - the `pardon` methods are used directly. This prevents me from knowing the command source.

I would assume that the `removeBan` method should be the "gateway" but am not entirely familiar with the internals/mixin needs.

Aaron1011:
@viveleroi: Sorry for the delay in getting to this. Aren't you able to look up the ban object internally, based off the gameprofile or address passed in?

viveleroi:
I'm not clear on what you're advising. How would the `GameProfile` or address tell me who's performing the pardon?

Aaron1011:
@viveleroi: Ah, sorry - I thought you were asking two different questions here. I'll add a way to do that.

Zidane:
@Aaron1011 

Status update.

ryantheleach:
Due to the age and cause refactor, I am looking for people to test this on 1.12

ryantheleach:
@viveleroi @Aaron1011 If you are still interested in this / if it is still relevant, I have had @RedNesto volunteer to look into this issue, however, I was unable to give them tips on exactly what was being asked.

I've looked at the BanService and it hasn't changed since this ticket was created.

Am I correct in thinking that this issue is about trace-ability of ban causes, and finding out who was responsible for banning a player?

If so, I see 2 options, that the ban that is removed is exposed on pardon, which is an API change, or

getting the cause from the CauseStackManager?

If this was prior to any cause tracking being exposed, I can understand that the problem is (potentially) fixed if the correct causes can be made available through the CSM?

viveleroi:
Close, I was looking for the person responsible for *pardoning* a player. Causes were overhauled multiples times since this was filed, so if that data is now available through the CSM (that sounds like the recommended way to obtain it), then I think this can be closed.

ImMorpheus:
AFAIK PardonUserEvent is not implemented. 
How would you be able to obtain the command source ? 

EDIT:

@ryantheleach viveleroi's initial point is valid and not fixed by the CSM. The method we should implement PardonUserEvent into (https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/service/ban/SpongeBanService.java#L135 ) is never called.
[...]

PR #1616, opened 2017-11-25
Implement PardonUserEvent/PardonIpEvent

ImMorpheus:
From #436 

> The method we should implement PardonUserEvent into (https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/service/ban/SpongeBanService.java#L135 ) is never called.
> As he noted, when a ban is removed, pardon() is called directly instead of removeBan().
> here: https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/service/ban/SpongeIPBanList.java#L93
> and here: https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/service/ban/SpongeUserListBans.java#L84

### DIFF
--- a/src/main/java/org/spongepowered/common/service/ban/SpongeBanService.java
+++ b/src/main/java/org/spongepowered/common/service/ban/SpongeBanService.java
@@ -114,34 +114,35 @@ public class SpongeBanService implements BanService {
 
     @Override
     public boolean pardon(GameProfile profile) {
+        Optional<Ban.Profile> ban = getBanFor(profile);
         this.getUserBanList().removeExpired();
-        boolean hadBan = this.isBanned(profile);
-        UserListUtils.removeEntry(this.getUserBanList(), profile);
-        return hadBan;
+        return ban.isPresent() && removeBan(ban.get());
     }
 
     @Override
     public boolean pardon(InetAddress address) {
-        UserListIPBans banList = this.getIPBanList();
-
-        banList.removeExpired();
-        InetSocketAddress inetSocketAddress = new InetSocketAddress(address, 0);
-        boolean hadBan = this.isBanned(address);
-        UserListUtils.removeEntry(banList, banList.addressToString(inetSocketAddress));
-        return hadBan;
+        Optional<Ban.Ip> ban = getBanFor(address);
+        this.getIPBanList().removeExpired();
+        return ban.isPresent() && removeBan(ban.get());
     }
 
     @Override
     public boolean removeBan(Ban ban) {
+        if (!hasBan(ban)) {
+            return false;
+        }
         if (ban.getType().equals(BanTypes.PROFILE)) {
             User user = Sponge.getServiceManager().provideUnchecked(UserStorageService.class).getOrCreate(((Ban.Profile) ban).getProfile());
             Sponge.getEventManager().post(SpongeEventFactory.createPardonUserEvent(Sponge.getCauseStackManager().getCurrentCause(), (Ban.Profile) ban, user));
 
-            return this.pardon(((Ban.Profile) ban).getProfile());
+            UserListUtils.removeEntry(this.getUserBanList(), ((Ban.Profile) ban).getProfile());
+            return true;
         } else if (ban.getType().equals(BanTypes.IP)) {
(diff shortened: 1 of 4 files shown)


JBYoshi:
Could you also change `pardon()` to call `removeBan()` instead of the other way around so both methods fire the event?

ImMorpheus:
I'm not sure I understand what you mean, `pardon()` is only called inside `removeBan()` to perform the internal logic (`UserListUtils.removeEntry()`) to remove a ban. IMO it shouldn't be called outside that method. Finally, the event is fired in `removeBan()` to remain consistent with "BanUserEvent" and "BanIpEvent", fired inside `addBan()`.

JBYoshi:
It’s publicly exposed in the API.

src/main/java/org/spongepowered/common/service/ban/SpongeBanService.java
```diff
@@ -114,29 +114,35 @@ public boolean isBanned(InetAddress address) {
...
+            return true;
         } else if (ban.getType().equals(BanTypes.IP)) {
-            return this.pardon(((Ban.Ip) ban).getAddress());
+            Sponge.getEventManager().post(SpongeEventFactory.createPardonIpEvent(Sponge.getCauseStackManager().getCurrentCause(), (Ban.Ip) ban));
+
+            InetSocketAddress inetSocketAddress = new InetSocketAddress(((Ban.Ip) ban).getAddress(), 0);
+            UserListUtils.removeEntry(this.getIPBanList(), this.getIPBanList().addressToString(inetSocketAddress));
+            return true;
         }
         throw new IllegalArgumentException(String.format("Ban %s had unrecognized BanType %s!", ban, ban.getType()));
```
ImMorpheus, line 147: Note: This is now "useless": if an unrecognized BanType is passed to `removeBan`, the exception is thrown during `hasBan(ban)`.

src/main/java/org/spongepowered/common/service/ban/SpongeIPBanList.java
```diff
@@ -90,7 +91,8 @@ public void removeEntry(String entry) {
         }
 
         try {
-            getService().pardon(InetAddress.getByName(entry));
+            Optional<Ban.Ip> ban = getService().getBanFor(InetAddress.getByName(entry));
+            ban.ifPresent(ip -> getService().removeBan(ip));
```
JBYoshi: This change shouldn't be necessary anymore.

src/main/java/org/spongepowered/common/service/ban/SpongeUserListBans.java
```diff
@@ -81,7 +82,8 @@ public boolean isEmpty() {
 
     @Override
     public void removeEntry(com.mojang.authlib.GameProfile entry) {
-        getService().pardon((GameProfile) entry);
+        Optional<Ban.Profile> ban = getService().getBanFor((GameProfile) entry);
+        ban.ifPresent(profile -> getService().removeBan(profile));
```
JBYoshi: Same with this.

JBYoshi (approved):
Glitched the reviewer 😐 - let me fix that.

src/main/java/org/spongepowered/common/service/ban/SpongeIPBanList.java
JBYoshi: This change shouldn't be necessary anymore.

src/main/java/org/spongepowered/common/service/ban/SpongeUserListBans.java
JBYoshi: Same here.

JBYoshi (approved):
Looks good!

ImMorpheus:
Fixed merge conflicts.

parlough:
@ImMorpheus Would you mind changing your branch target to `stable-7` and adding a simple test plugin which listens to the events? Once those changes happen I'll look at getting this in!

ImMorpheus:
Done.

I didn't know what to do in the listener, so I sent a message to the payer who caused the event (this is what the related issue was about).
Since the events require user interaction to be fired (/pardon) I think there's no need for a "toggle" option (like TargetAIEventTest).

parlough:
@ImMorpheus Thanks for the change! You forgot the license though.

parlough:
Can't see me

PR #1617, opened 2017-11-26
Remove stability warning from Readme.

ryantheleach:
It came up on discord that this warning could be misleading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,6 @@
 SpongeCommon [![Build Status](https://travis-ci.org/SpongePowered/SpongeCommon.svg?branch=master)](https://travis-ci.org/SpongePowered/SpongeCommon)
 =============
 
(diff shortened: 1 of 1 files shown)


dualspiral:
I’m not sure such a warning should be removed from bleeding - by very definition it _isn’t_ stable and _is_ under heavy development. 

I’d favour such a message to be reworded to specify that the _bleeding_ branch is unstable, but I’m not sure it should be removed outright.

ImMorpheus:
IMO we should remove the warning from the stable-x branches and make stable-7 the default branch.

This would ensure contributors target the stable-7 branch when opening PRs and remove any misleading warning.

parlough:
Superseded by https://github.com/SpongePowered/SpongeCommon/pull/1831

PR #1618, opened 2017-11-26
Cancel Mesa Terrain only if constructed from SpongeForge 

ryantheleach:
Fixes SpongePowered/SpongeForge#1834

This restores vanilla behavior if BiomeMesa is constructed without MixinBiome::buildPopulators being called, which to my knowledge, only gets called under Sponge world generators.

If this is bypassed e.g. mods not knowing about us extracting the Mesa biome into a decorator/populator then it defaults to vanilla's.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/biome/MixinBiomeMesa.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/biome/MixinBiomeMesa.java
@@ -52,9 +52,11 @@ public abstract class MixinBiomeMesa extends MixinBiome {
 
     @Shadow @Final private boolean brycePillars;
     @Shadow @Final private boolean hasForest;
+    private boolean vanillaCompat = true;
 
     @Override
     public void buildPopulators(World world, SpongeBiomeGenerationSettings gensettings) {
+        this.vanillaCompat = false;
         gensettings.getGenerationPopulators().add(new MesaBiomeGenerationPopulator(this.brycePillars, this.hasForest));
         super.buildPopulators(world, gensettings);
         String s = world.getWorldInfo().getGeneratorOptions();
@@ -95,7 +97,9 @@ public abstract class MixinBiomeMesa extends MixinBiome {
      */
     @Inject(method = "genTerrainBlocks(Lnet/minecraft/world/World;Ljava/util/Random;Lnet/minecraft/world/chunk/ChunkPrimer;IID)V", at = @At("HEAD") , cancellable = true)
     public void genTerrainBlocks(World world, Random rand, ChunkPrimer chunk, int x, int z, double stoneNoise, CallbackInfo ci) {
-        ci.cancel();
+        if (this.vanillaCompat) {
+            ci.cancel();
+        }
     }
 
 }


src/main/java/org/spongepowered/common/mixin/core/world/biome/MixinBiomeMesa.java
```diff
@@ -95,7 +97,9 @@ public void buildPopulators(World world, SpongeBiomeGenerationSettings gensettin
      */
     @Inject(method = "genTerrainBlocks(Lnet/minecraft/world/World;Ljava/util/Random;Lnet/minecraft/world/chunk/ChunkPrimer;IID)V", at = @At("HEAD") , cancellable = true)
     public void genTerrainBlocks(World world, Random rand, ChunkPrimer chunk, int x, int z, double stoneNoise, CallbackInfo ci) {
-        ci.cancel();
+        if(this.vanillaCompat){
```
ImMorpheus: Missing space

ryantheleach:
@Fla5hFyre Test Build: 
[TestBuild.jar.zip](https://github.com/SpongePowered/SpongeCommon/files/1504043/TestBuild.jar.zip)

src/main/java/org/spongepowered/common/mixin/core/world/biome/MixinBiomeMesa.java
```diff
@@ -52,9 +52,11 @@
 
     @Shadow @Final private boolean brycePillars;
     @Shadow @Final private boolean hasForest;
+    private boolean vanillaCompat = true;
 
     @Override
     public void buildPopulators(World world, SpongeBiomeGenerationSettings gensettings) {
+        this.vanillaCompat = false;
```
Deamon5550, line 59: This doesn't work with multi-world, as there is only one biome instance for the entire server, so long as any world is using the normal generation all worlds will have the decoration cancelled. A flag set from the generator for whether or not the current generation is calling a modded override would work.

Deamon5550 (changes requested):

src/main/java/org/spongepowered/common/mixin/core/world/biome/MixinBiomeMesa.java
```diff
@@ -95,7 +97,9 @@ public void buildPopulators(World world, SpongeBiomeGenerationSettings gensettin
      */
     @Inject(method = "genTerrainBlocks(Lnet/minecraft/world/World;Ljava/util/Random;Lnet/minecraft/world/chunk/ChunkPrimer;IID)V", at = @At("HEAD") , cancellable = true)
     public void genTerrainBlocks(World world, Random rand, ChunkPrimer chunk, int x, int z, double stoneNoise, CallbackInfo ci) {
-        ci.cancel();
+        if (this.vanillaCompat) {
```
ryantheleach, line 100: @Deamon5550 I've spent a couple of hours revisiting this, Any tips on how to get the generator from inside this injection?

gabizou: @what you could do is take advantage of the `PhaseTracker` and have a boolean flag for whether this biome is usable or not (since world generation is still tracked through a phase state).

dualspiral:
I think, based on the conversation we've had, you might need to add a field to the settings to indicate it's modded.

The following untested code might do what you need it to do - change to requirements:

```java
        if (world.isRemote) {
            // Fix as needed
            if ((SpongeBiomeGenerationSettings) 
                    ((SpongeChunkGenerator) ((WorldServer) world).getChunkProvider().chunkGenerator).getBiomeSettings(this).getModded()) {
                ci.cancel();
            }
        }
```

(disclaimer, this might not work as expected, just through quick messing about)

phit:
https://github.com/SpongePowered/SpongeCommon/commit/ad4859eeff55a9f883a93cb43c3b6657a4f40e4a

Issue #1314, opened 2017-04-20
Block's won't update after explosion


XakepSDK:
1. Place gravel
2. Place TNT
3. Explode tnt
4. Gravel won't fall.
Tested on:
spongevanilla-1.11.2-6.0.0-BETA-250

gabizou:
Could you show a picture of where the gravel/sand is placed in relation to TNT? Otherwise, the gravel/sand are supposed to simply be removed, not "fall".

XakepSDK:
I can't add screenshot now. But you can try it yourself. Place TNT and
place around TNT gravel. Make 3x3x12 gravel structure with TNT in bottom
middle. Activate TNT and wait.
Sorry for my english.

XakepSDK:
https://puu.sh/vrjNS/bc4efd5fff.png
https://puu.sh/vrjOE/d5c99d6dbe.png
This seems not explosions issue, this happens after world generation. Water won't flow, sand won't fall, until i touch it.

gabizou:
> This seems not explosions issue, this happens after world generation. Water won't flow, sand won't fall, until i touch it.

You were reporting gravel not falling, is that still an issue with explosions? Please avoid getting off track.

XakepSDK:
See screenshot. There is issue with block updating. At first, i saw this issue only after TNT explosion, but after 5 mins playing, i found same issues with world generated blocks. Sand won't fall, until i place/break block on this sand, water won't flow.

Deamon5550:
That is how it works in vanilla, all physics updates are disabled during world generation.

XakepSDK:
Oh, this is all my bad english. Sorry, guys. I make video with
demonstration, 14:00 gmt+6

XakepSDK:
No video, sorry.
Gravel:
1. https://puu.sh/vskBd/39bc5ccec0.png
2. https://puu.sh/vskC0/cbf3696aef.png
3. https://puu.sh/vskCp/7b3ad3902a.png
4. https://puu.sh/vskD7/7dffb4be10.png

Sand:
1. https://puu.sh/vskE6/2cb15c926b.png
2. https://puu.sh/vskEx/4ae32f3a00.png
3. https://puu.sh/vskEU/27e865719f.png
4. https://puu.sh/vskFn/a5019c2800.png
5. https://puu.sh/vskFP/7935cc4ba8.png
6. https://puu.sh/vskFP/7935cc4ba8.png
[...]

PR #1619, opened 2017-11-26
Fix explosions not notifying neighbours blocks

RedNesto:
Fixes SpongePowered/SpongeCommon#1314 Fixes SpongePowered/SpongeVanilla#348

### DIFF
--- a/src/main/java/org/spongepowered/common/SpongeImplHooks.java
+++ b/src/main/java/org/spongepowered/common/SpongeImplHooks.java
@@ -24,19 +24,13 @@
  */
 package org.spongepowered.common;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Streams;
 import net.minecraft.block.Block;
 import net.minecraft.block.ITileEntityProvider;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.crash.CrashReport;
-import net.minecraft.entity.Entity;
-import net.minecraft.entity.EntityList;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/SpongeImplHooks.java
```diff
@@ -288,7 +288,7 @@ public static void onTileEntityError(TileEntity tileEntity, CrashReport crashRep
     }
 
     public static void blockExploded(Block block, World world, BlockPos blockpos, Explosion explosion) {
-        world.setBlockState(blockpos, Blocks.AIR.getDefaultState(), 2);
```
phit, line 291: ~~Vanilla source uses 
`world.setBlockState(blockpos, Blocks.AIR.getDefaultState(), 3);`
flag 3 being notify client and server, any reason you are using `setBlockToAir` instead?~~

nvm, `setBlockToAir` is a helper method that does the exact same thing

RedNesto: `setBlockToAir` does the same thing, just a preference in fact.

phit (approved):
Checked [Vanilla](https://github.com/MinecraftForge/MinecraftForge/blob/16bfd8cef1d12ee9ca0de1122addaf9916767ae9/patches/minecraft/net/minecraft/world/Explosion.java.patch#L40) and [Forge](https://github.com/MinecraftForge/MinecraftForge/blob/5cddcc02806e5623945b1783b79c970bb25a9a26/patches/minecraft/net/minecraft/block/Block.java.patch#L808-L812) source, both use flag 3 as well.

PR #1622, opened 2017-11-27
Change entity spawning return types, breaking change - Fixes #1365

### LINKED ISSUE #1365: EntityUniverse#spawnEntities returns whether event was cancelled, not whether entities were spawned
As seen [here](https://github.com/SpongePowered/SpongeCommon/blob/3634fcb/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java#L1093), the method is returning an improper value per [its JavaDoc](https://jd.spongepowered.org/6.0.0/org/spongepowered/api/world/extent/EntityUniverse.html#spawnEntities-java.lang.Iterable-org.spongepowered.api.event.cause.Cause-). Unsure of whether this is an OCD fix since it constitutes a fairly significant bug.

parlough:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1705) | **SpongeCommon**

Refer to API PR for more discussion

Closes https://github.com/SpongePowered/SpongeCommon/issues/1365


testplugins/src/main/java/org/spongepowered/test/EntitySpawnTest.java
```diff
@@ -1,3 +1,27 @@
...
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
```
ryantheleach, line 24: https://github.com/SpongePowered/SpongeAPI/blob/bleeding/scripts/pre-commit

liach: I'd swear that script takes one minute to run

PR #1624, opened 2017-11-28
Fix Sponge not respecting wrapped chunk generator structure lookups / generation

Dockter:
Resolves issues with TerrainControl specifically when looking up structures and properly generating them.

Signed-off-by: Dockter <dockter@almuramc.com>

### DIFF
--- a/src/main/java/org/spongepowered/common/world/gen/SpongeChunkGenerator.java
+++ b/src/main/java/org/spongepowered/common/world/gen/SpongeChunkGenerator.java
@@ -449,18 +449,18 @@ public class SpongeChunkGenerator implements WorldGenerator, IChunkGenerator {
         } else if ("EndCity".equals(structureName)) {
             target = MapGenEndCity.class;
         }
-        if (target == null) {
-            return null;
-        }
-        for (GenerationPopulator gen : this.genpop) {
-            if (target.isInstance(gen)) {
-                return ((MapGenStructure) gen).getNearestStructurePos(worldIn, position, p_180513_4_);
+        if (target != null) {
+            for (GenerationPopulator gen : this.genpop) {
+                if (target.isInstance(gen)) {
+                    return ((MapGenStructure) gen).getNearestStructurePos(worldIn, position, p_180513_4_);
+                }
             }
         }
+
         if (this.baseGenerator instanceof SpongeGenerationPopulator) {
-            return ((SpongeGenerationPopulator) this.baseGenerator).getHandle(this.world).getNearestStructurePos(worldIn, structureName, position,
-                    p_180513_4_);
+            return ((SpongeGenerationPopulator) this.baseGenerator).getHandle(this.world).getNearestStructurePos(worldIn, structureName, position, p_180513_4_);
         }
+
         return null;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/world/gen/SpongeChunkGenerator.java
```diff
@@ -431,6 +431,10 @@ public boolean generateStructures(Chunk chunk, int chunkX, int chunkZ) {
     @Nullable
     @Override
     public BlockPos getNearestStructurePos(World worldIn, String structureName, BlockPos position, boolean p_180513_4_) {
+        if (this.baseGenerator instanceof SpongeGenerationPopulator) {
```
Deamon5550: Not sure I see the purpose of changing the ordering here. It prevents sponge plugins from modifying any behaviour of modded worlds since we're checking for the modded overrime. And this doesn't actually change any behaviour in the absence of a world generation plugin since if we are wrapping a modded world the generation populator list will be empty.

Deamon5550 (changes requested):
The `recreateStructures` change is a good catch, but I have a question about the other change.

src/main/java/org/spongepowered/common/world/gen/SpongeChunkGenerator.java
Dockter: It happened because it was getting here:

https://github.com/AlmuraDev/SpongeCommon/blob/8f3915941467525869937ba1f14ce6d78df8b01a/src/main/java/org/spongepowered/common/world/gen/SpongeChunkGenerator.java#L456

Deamon5550: Ahh yes for custom structures, good point. I think the better solution then is to invert that null check and wrap the for loop after it. Then if either the string is for a custom structure, or if no genpop in the list matches that structure it will fall into the modded override.

Dockter:
Fixed per your request.

src/main/java/org/spongepowered/common/world/gen/SpongeChunkGenerator.java
```diff
@@ -449,23 +449,27 @@ public BlockPos getNearestStructurePos(World worldIn, String structureName, Bloc
...
             return ((SpongeGenerationPopulator) this.baseGenerator).getHandle(this.world).getNearestStructurePos(worldIn, structureName, position,
                     p_180513_4_);
         }
+
         return null;
     }
 
     @Override
     public boolean isInsideStructure(World worldIn, String structureName, BlockPos position) {
+        if (this.baseGenerator instanceof SpongeGenerationPopulator) {
```
Deamon5550: Fix this one too as per `GetNearestStructurePos`

src/main/java/org/spongepowered/common/world/gen/SpongeChunkGenerator.java
```diff
@@ -504,6 +505,10 @@ public void recreateStructures(Chunk chunkIn, int x, int z) {
             ((IChunkGenerator) this.baseGenerator).recreateStructures(chunkIn, x, z);
             return;
```
Deamon5550: I think this return might be an error, preventing plugin structures from working properly.

Deamon5550 (changes requested):
Two more small things, other than that looks good thanks.

src/main/java/org/spongepowered/common/world/gen/SpongeChunkGenerator.java
Dockter: That return was there prior to my changes.  I am not sure I the setup to see what would happen if I removed it.

How shall I proceed?

src/main/java/org/spongepowered/common/world/gen/SpongeChunkGenerator.java
Dockter: Updated the PR to match.  Also removed the below return though I don't have a plugin that is doing structure generation where I can test that aspect of this PR.

src/main/java/org/spongepowered/common/world/gen/SpongeChunkGenerator.java
Deamon5550: Yes I know, but the structures covered by the (effectively) three cases will all be exclusive.

Deamon5550 (approved):

PR #1625, opened 2017-11-28
Update SpongeCommon Color Usage

SimonFlash:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1699) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1625)

This PR updates SpongeCommon for the changes made in SpongeAPI, notably the deprecation of `Color#ofRbg(int)` and the addition of `TextColor#getBackgroundColor()`. Additionally, this also removes the `ColorUtil` class and migrates any methods still used to `ColoredDataProcessor` or `NBTDataUtil`.

Edit: I have no idea what I did the first time around, but thanks to Faithcaio and phit things are no longer a mess. In the meantime, I have been tasked with reading a git book - adiou!


SimonFlash:
The trend throughout most of Common is to have a separate `...Builder` class, but this seems to apply to classes such as `ItemStack` that aren't implemented in Sponge. If we would prefer have a `SpongeColor` and `SpongeColorBuilder` it's an easy change to make.

Also, I'm not sure how to setup the `ColorTest` to work in Common because the registry would need to be initialized to get the color builder. If anyone could provide direction on this that'd be great.

dualspiral:
> Also, I'm not sure how to setup the ColorTest to work in Common because the registry would need to be initialized to get the color builder. If anyone could provide direction on this that'd be great.

Annotate your test class with `@RunWith(LaunchWrapperTestRunner.class)`. That'll run by initialising a dummy server first, letting you test using the registry.

gabizou (changes requested):
If you're going to want this tested, you at the very least want to have a test plugin that will be able to provide some colored items, including dyed leather armor pieces, etc. to validate that the color changes are working (and of course, translation of colors between them).

SimonFlash:
Update for the `TextColor` change in API, a fix for `ColorTest` (junit) to work with the registry, and the addition of a `ColorTest` plugin for proof-of-concept (and rainbows). I've tested this with SpongeForge compiled from a couple hours ago without any issues. Colors are being retrieved properly from items and/or dye colors and can be offered back to the item with `Keys.COLOR` successfully.

Remaining potential projects in this PR:

* Have `#mix` use the Minecraft formula for mixing dyes
* Remove `#mix` completely (uses are few and far between; not used in Sponge. Is breaking)
* Implementation changes from projects in [API](https://github.com/SpongePowered/SpongeAPI/pull/1699#issuecomment-348744940)

phit:
make the testplugin something that's not enabled on every click, preferably something like set color for item in hand when running a command and get color from item in hand

ryantheleach:
> make the testplugin something that's not enabled on every click, 

This is good advice, but considering the growing number of test plugins I'm considering making a basic framework for dealing with this after I've cleared the rest of my backlog.

testplugins/src/main/java/org/spongepowered/test/ColorTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.spongepowered.api.plugin.Plugin;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.util.Color;
+
+@Plugin(id = "color_test", name = "Color Test", description = "Right click to get an items color. Left click to set the color to orange.")
+public class ColorTest {
+
+    @Listener
+    public void onPrimaryClick(InteractItemEvent.Primary event, @First Player player) {
+        player.getItemInHand(HandTypes.MAIN_HAND).ifPresent(item -> {
```
gabizou: If this is a test, please make it toggle able either by command or configuration option. By no means should this always emit some message randomly causing people to freak out about debug messages when working on implementation ;).

testplugins/src/main/java/org/spongepowered/test/ColorTest.java
```diff
@@ -0,0 +1,71 @@
...
+            } else if (item.supports(Keys.DYE_COLOR)) {
+                DyeColor dyeColor = item.get(Keys.DYE_COLOR).get();
+                player.sendMessage(Text.of(id, "'s dye color is ", dyeColor.getId(), " (", String.format("0x%06X", dyeColor.getColor().getRgb()), "),", dyeColor.getColor().toVector3i()));
+            } else {
+                player.sendMessage(Text.of(id, " does not support color or dye color."));
+            }
+        });
+    }
+
+    private static double radians;
```
gabizou: For the sake of style, this should be on the top of the class.

testplugins/src/main/java/org/spongepowered/test/ColorTest.java
```diff
@@ -0,0 +1,71 @@
...
+    private static final double increment = Math.PI / 7, shift = 2 * Math.PI / 3;
```
gabizou: Should be formatted as a constant (
```java
private static final double INCREMENT = Math.PI / 7;
private static final double SHIFT = 2 * Math.PI / 3;
```

testplugins/src/main/java/org/spongepowered/test/ColorTest.java
```diff
@@ -0,0 +1,71 @@
...
+    private static final Vector3d shifts = Vector3d.from(2 * shift, shift, 0), offset = Vector3d.from(127.5);
```
gabizou: And here, don't tail on fields like this. It's bad legibility for others.

src/main/java/org/spongepowered/common/data/processor/data/item/ColoredDataProcessor.java
```diff
@@ -38,14 +40,19 @@
...
 import java.util.Optional;
 
 public class ColoredDataProcessor extends AbstractItemSingleDataProcessor<Color, Value<Color>, ColoredData, ImmutableColoredData> {
 
     public ColoredDataProcessor() {
-        super(ColorUtil::hasColor, Keys.COLOR);
+        super(ColoredDataProcessor::supportsColor, Keys.COLOR);
+    }
+
+    private static boolean supportsColor(ItemStack stack) {
```
gabizou, line 52: Any chance this can be overridden in Forge for compatibility with other mods?

src/main/java/org/spongepowered/common/registry/type/text/TextColorRegistryModule.java
```diff
@@ -60,20 +60,20 @@
...
-        addTextColor(TextFormatting.AQUA, Color.ofRgb(0x55FFFF));
-        addTextColor(TextFormatting.RED, Color.ofRgb(0xFF5555));
-        addTextColor(TextFormatting.LIGHT_PURPLE, Color.ofRgb(0xFF55FF));
-        addTextColor(TextFormatting.YELLOW, Color.ofRgb(0xFFFF55));
+        addTextColor(TextFormatting.DARK_BLUE, Color.of(0x0000AA));
+        addTextColor(TextFormatting.DARK_GREEN, Color.of(0x00AA00));
+        addTextColor(TextFormatting.DARK_AQUA, Color.of(0x00AAAA));
+        addTextColor(TextFormatting.DARK_RED, Color.of(0xAA0000));
+        addTextColor(TextFormatting.DARK_PURPLE, Color.of(0xAA00AA));
+        addTextColor(TextFormatting.GOLD, Color.of(0xFFAA00));
```
gabizou: The more I look at this, the more I prefer the `ofRgb` instead of just flag `of`, since you can define how it's formatted.

src/main/java/org/spongepowered/common/util/SpongeColor.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        @Override
+        public Builder from(Color value) {
+            return rgb(red, green, blue);
+        }
+
+        @Override
+        public Builder reset() {
+            return rgb(0, 0, 0);
```
gabizou: A little wasteful when all you need to do is just set the fields and return this...

src/main/java/org/spongepowered/common/util/SpongeColor.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        @Override
+        public Color build() {
+            return new SpongeColor(red, green, blue);
+        }
+
+    }
+
+    public static final class DataBuilder extends AbstractDataBuilder<Color> {
```
gabizou: To be quite honest, this should be merged with the Builder, and not as a separate class.

gabizou (changes requested):

src/main/java/org/spongepowered/common/data/processor/data/item/ColoredDataProcessor.java
SimonFlash: Unsure, as my experience with both Forge and the Data API is lacking. This method itself was just copied from `ColorUtil` so I really don't know the specifics of it and what would need to be done to support this.

src/main/java/org/spongepowered/common/registry/type/text/TextColorRegistryModule.java
SimonFlash: I'm fine with that; though in my mind `of` is more consistent with the naming of other static methods used throughout Sponge. If we do revert this change, then I would like to be consistent and use `ofVector3i`, `ofJavaColor`, and so on.

src/main/java/org/spongepowered/common/util/SpongeColor.java
SimonFlash: That's true, just seemed simpler to me to do it like that. It might also be advantageous to us to use `-1` as the default value, that way we know a color has been set.

src/main/java/org/spongepowered/common/util/SpongeColor.java
SimonFlash: Can this be done by having `Builder` extend `AbstractDataBuilder<Color>` and implement `Color.Builder`, or is there more to it? Is there any example of this combination used elsewhere in Common that I can reference?

testplugins/src/main/java/org/spongepowered/test/ColorTest.java
SimonFlash: Tell that to whoever set up the inventory tests, I still get "You clicked on slot 7" messages every time I go to pull out an item xD. I'll switch those over to commands that can be used.

I'm not going to respond to the remaining reviews individually, but I'll address them in my edits tonight. I think I had a bit to much fun with rainbows (though they are fun) and let some of the good practices slip away from me. @ryantheleach, could you supply some info on what you'd like to see as a standard for test plugins so I can do that now and not have to rebuild it later?

ryantheleach: You won't have to rebuild it later, just make it somewhat toggleable to pass review, this is something larger that I havn't fully fleshed out yet.

src/main/java/org/spongepowered/common/registry/type/text/TextColorRegistryModule.java
Deamon5550: `ofVector3i` should also be `ofRgb` as its simply an overload to take a vector.

SimonFlash: Fair enough, but I don't see the reasoning to have `ofRgb` over `of` unless `Color` is designed to support multiple color formats, which is unneeded as well as more complex. We represent colors using RGB, so `ofRgb` doesn't provide any additionally information and, as mentioned, is less consistent with the rest of Sponge.

This change in particular, though, is more specific to the API over Common.

Deamon5550: Its ensuring that we don't back ourselves into a confusion corner later if say we wanted to add an `ofHSV` (not a bad idea as it is a much easier color space to work with for many use cases).

SimonFlash: We currently store colors in an RGB format, meaning that we would either need to have a storage option for HSV (preferable) or handle the conversion, which is messy and limits our ability to have getters for HSV values. In a graphical environment, you would have an `RgbColor` class and a `HsvColor` class and use inheritance to handle rendering them. I don't believe this step is at all necessary or within the scope of Sponge. Colors in Minecraft are handled with RGB values, and I strongly recommend to keep this trend as there is not sufficient reason to go to these lengths to support HSV or any other color format.

@gabizou, thoughts on this and the resolution of the `of`/`ofRgb` change?

dualspiral: I think you're overcomplicating and thus overthinking what is being asked here. We're not asking you to implement any of this, we're asking for a function name change to try to futureproof the API. I, for one, am in favour of `ofRgb` instead of `of`.

The process of devloping an API is not to think about what is now, but also about what could be in the future. When I suggested we should give this the whole API/impl split treatment, I was after the current functionality being split to reduce the API footprint to make it easier to add future API methods later and give us flexibility in the implementation. `ofHSV` is a potential extension _that can be done later_. Futher, the whole point of splitting this into API/impl is that we can do that sort of subclassing if we need to without impacting the API. By naming it `ofRgb` now, we are ensuring that we don't need to break backwards compatibility because we have a method that is already describing what is going on. If we leave it as `of`, and then want to add `ofHSV`, we then end up with inconsistent naming which would likely lead to the deprecation of `of` and adding `ofRgb` anyway. `ofRgb` is also more descriptive to just look at.

No-one is saying that we should stop storing values as RGB, and these are potentially valid and interesting questions, but these are implementation details that can be picked up later. We want to design the API such that any implementation detail, be it storage detail or otherwise, can fit the API nicely without needing any more breakages.

src/main/java/org/spongepowered/common/util/SpongeColor.java
dualspiral: Yes, you can do that. [Example as requested.](https://github.com/SpongePowered/SpongeCommon/blob/74080b0c75e59075f164f5151b08be395e2d17fc/src/main/java/org/spongepowered/common/block/SpongeBlockStateBuilder.java#L44)

src/main/java/org/spongepowered/common/registry/type/text/TextColorRegistryModule.java
SimonFlash: My point with this is that it *would* require breakages. The `Color` interface defines `getRed`, `getGreen`, `getBlue`, and other methods related to the RGB format. If there is a desire to pursue other formats like HSV in the future, then we need to consider that now and redesign the API to support that.

If the primary argument for `ofRgb` over `of` relies on future implementation that the API doesn't support, then for it to be valid we need to update the API to support multiple color formats now.

src/main/java/org/spongepowered/common/util/SpongeColor.java
SimonFlash: Perfect, thank you.

src/main/java/org/spongepowered/common/registry/type/text/TextColorRegistryModule.java
Deamon5550: An `ofHSV` function which is a simple conversion into the RGB color space would be simple and non-breaking. No need to change the backing storage, or even consider that right now. We just don't want a plain `of` function that doesn't even make clear in the simplest case between `ofRGB` and `ofBGR`.

SimonFlash: Converting HSV to RGB is not something I consider worthwhile (see [this question](https://stackoverflow.com/questions/3018313/algorithm-to-convert-rgb-to-hsv-and-hsv-to-rgb-in-range-0-255-for-both) that has the conversion formulas). For reference, the `java.awt.Color` class is defined to represent colors in [sRGB](http://www.w3.org/pub/WWW/Graphics/Color/sRGB.html) format (link from the class javadocs). If a developer would prefer to use HSV, they can use the conversion methods in `java.awt.Color`.

`Color.of(int rgb)` provides the same specification that it is in RGB format as does `new java.awt.Color(int rgb)`. I maintain that the `Color` class should be defined to use the RGB format and use `#of`, as `#ofRgb` is unnecessarily repetitive and is different from the `#of` standard used elsewhere in Sponge.

gabizou: I think I’ve said this before, the method should very well be named ofRgb to avoid confusion in the format the colors are presented. Java API is not a standard to hold us up to on making such a method name decision. I’m putting my decision as it was mentioned before.

SimonFlash: Here's what I'm tracking so far. What are we doing with the one for `java.awt.Color`?

`#of(int rgb)` -> `#ofRgb(int rgb)`
`#of(int red, int green, int blue)` -> `#ofRgb(int red, int green, int blue)`
`#of(Vector3i vector3i)` -> `#ofRgb(Vector3i vector3i)`
`#of(java.awt.Color)` -> ??? (`#ofJavaColor`?)

gabizou: Can make it ofColor since the class type is implied by the argument type. Otherwise fine.

SimonFlash:
With API 7 receiving a stable release at the end of the month, I'd like to push for this PR to be merged prior to it's release. Please see the [API](https://github.com/SpongePowered/SpongeAPI/pull/1699) PR for topics more related to the API.

In Common, there is really only one outstanding discussion point - overriding `ColoredDataProcessor#supportsColor` within SpongeForge for mods. I don't know what would be required to handle this, but if this is something that needs to be considered I could definitely use assistance from someone familiar with the Forge side of things.

src/main/java/org/spongepowered/common/util/SpongeColor.java
```diff
@@ -0,0 +1,184 @@
...
+        @Override
+        public Builder rgb(int red, int green, int blue) {
+            this.red = red & 0xFF;
+            this.green = green & 0xFF;
+            this.blue = blue & 0xFF;
+            return this;
+        }
+
+        @Override
+        public Builder mix(Color... colors) {
```
dualspiral: I'm shaky about this behaviour. A builder shouldn't do any processing like this pre-build.

My feeling is that we should offer a `mix(Color)` method that we can chain up (and this varargs method would default to chaining these up), and that we store the colours that we want to mix. Then, when `build()` is called, that's when the mixing occurs. This means that we can use the builder pattern for both mixing colours and specifiying a colour from RGB values, or both.

It allows us to offer a `Color.mix(Color...)` helper method that just calls `Color.builder().mix(...).build()` too.

To emulate the current behaviour, you could then just create the colour in a separate builder, so `Color.build().rgb(...).mix(Color.mix(...)).build()`

SimonFlash: I don't think I'm understanding what you mean by this. I believe you're suggesting that each call to `mix` should save any passed in colors to a list, which would be mixed internally when build is called.

I'm still a proponent of deprecated the mix methods entirely as per my reasoning in the API pull. One of the things for me is that the `Color` class, in my opinion, should not be represented by a builder - it's only like that so the implementation can exist in Common and use the builder registry in API. The methods don't interact in the same way as most builders as ever method essentially overrides another.

With the change to reset to have the default values be `-1`, the way this method *should* be functioning is to mix with an existing color in the builder if defined. I believe what you're getting at is that chaining the methods together should have the same effect as combining them all together into one call, which they currently do not.

However, I would still prefer deprecating these methods for removal as they are impractical.

ImMorpheus:
https://github.com/SpongePowered/SpongeAPI/pull/1699#issuecomment-461965836

PR #1629, opened 2017-12-01
Implement chorus fruit teleport type

parlough:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1709) | **SpongeCommon**

Not actually sure if this is the complete correct way to do this as I haven't worked much with this before, but with this code `TeleportTypes.CHORUS_FRUIT` was in the context for the key `EventContextKeys.TELEPORT_TYPES`. So please give feedback where you can, thanks!

Tested with: 

```Java
@Listener
public void onTeleport(MoveEntityEvent.Teleport event, @Root Player player) {
    event.getContext().get(EventContextKeys.TELEPORT_TYPE).ifPresent(t -> player.sendMessage(Text.of(TextColors.GOLD, "TeleportType: ", TextColors.GRAY, t.getId())));
}
```

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/item/MixinItemChorusFruit.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/item/MixinItemChorusFruit.java
@@ -1,0 +1,25 @@
+package org.spongepowered.common.mixin.core.item;
+
+import net.minecraft.entity.EntityLivingBase;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemChorusFruit;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.event.CauseStackManager;
+import org.spongepowered.api.event.cause.EventContextKeys;
+import org.spongepowered.api.event.cause.entity.teleport.TeleportTypes;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+@Mixin(ItemChorusFruit.class)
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/mixin/core/item/MixinItemChorusFruit.java
```diff
@@ -0,0 +1,25 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+@Mixin(ItemChorusFruit.class)
+public class MixinItemChorusFruit extends Item {
+
+    @Redirect(method = "onItemUseFinish", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/EntityLivingBase;attemptTeleport(DDD)Z"))
+    private boolean onAttemptTeleport(EntityLivingBase entity, double x, double y, double z) {
+        try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
+            Sponge.getCauseStackManager().addContext(EventContextKeys.TELEPORT_TYPE, TeleportTypes.CHORUS_FRUIT);
```
gabizou, line 20: You can replace `Sponge.getCauseStackManager()` with `frame`, shorter lines to read.

gabizou (changes requested):
Small change. I would also request a test plugin being added to have a toggle for describing teleports of the player, maybe even have some commands to provide the possible ways to teleport for the player?

parlough:
Will replace this with a more generic item teleport type when I get a chance.

PR #1632, opened 2017-12-02
Refactor Common Issue Template

phit:
Same as for Forge but for Common, not sure if we should add a section for code snippets?

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,24 +1,21 @@
-<!-- Please be as accurate as possible, and include as much information as possible.
-         Add or remove as needed to keep things informative but also relevant. -->
-
-**I am currently running...**
+**I am currently running**
 <!-- If you don't use the latest version, please tell us why. -->
 - SpongeForge version: 
 - SpongeVanilla version: 
 - Java version: 
 - Operating System:
-<!-- Replace the empty space (" ") in with an "x" within the brackets. -->
- - [ ] Linux (Ubuntu, Debian, Fedora, etc)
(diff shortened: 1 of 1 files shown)


ryantheleach:
```
<!-- 
Plugin Authors: 
  Please attach in a gist, a [Short, Self Contained, Compilable, Example](http://sscce.org/) Plugin that demonstrates the strange behavior. If you are unable to replicate it with a test plugin, how do you expect the Sponge devs to be able to?

  For examples, please see : https://github.com/SpongePowered/SpongeCommon/tree/bleeding/testplugins/src/main/java/org/spongepowered/test
-->
```

phit:
markdown doesn't do anything in issue templates, so the link part to sscce.org is useless as you can't click it
>If you are unable to replicate it with a test plugin, how do you expect the Sponge devs to be able to?

sounds really condescending, no reason to be hostile

I figured something like this replacing the last comment would work
```
<!-- If you are a developer, please attach a short and self contained code snippet to replicate your issue.
        Any relevant log files and snippets should be uploaded to gist.github.com!
        Don't use hastebin/pastebin or other similar sites, as they have a history of quickly
        deleting files before we can look at them. -->
```

ryantheleach:
I would still appreciate the link to sscce , it summarizes a lot of the points that would otherwise be too long.

ryantheleach:
LGTM

ST-DDT:
I dont seem to be able to connect to http://sscce.org properly. I get certificate warnings and 401.

phit:
stop whatever is redirecting the page to https in your browser, its certificate points to a different domain

PR #1635, opened 2017-12-03
Implementation Change for BlockChangeFlag

### LINKED ISSUE #998: Calling setBlock does not fire any ChangeBlockEvents
Simple listener

``` java
     @Listener
     public void onBlockChange(ChangeBlockEvent e) {
         System.out.println(e);
     }

```

Calling `World#setBlock` from command handler or interact event (and probably anywhere) does not fire a `ChangeBlockEvent`
Using spongeforge bleeding https://github.com/SpongePowered/SpongeForge/commit/ec945eedf4752f052f104c6c0d006f63865be1f3
This essential feature worked fine on 1.8.9

Also, it looks like `ChangeBlockEvent.Post` is fired when a player places/breaks a block which contradicts the specification

> ```
> * <p>The idea is that  a block, entity, or tile entity "ticks" in which
>  * during that "tick", they make different types of block changes. If the
>  * block change is purely one type then the corresponding event is thrown
>  * instead. For example, If the block change is purely "placing" of
>  * blocks, the {@link Place} event would be thrown instead.</p>
> ```

### LINKED ISSUE #1237: ChangeBlockEvent.Place Doesn’t Detect Block Changes Caused By Pluggin
So I'm trying to track block changes, specifically when blocks are placed, so I can later undo them. I have an executioner that creates blocks in a specified location, but those block changes are not being detected by my listener. In the listener below, I expected to receive spam messages telling me what blocks were being changed by my executioner, but the event was not fired.

```java
private final static Set<Transaction<BlockSnapshot>> snapshots = new HashSet<>();

@Listener
public void onBlockAlteration(ChangeBlockEvent.Place event) {

	snapshots.addAll(event.getTransactions());
	
	Sponge.getServer().getBroadcastChannel().send(Text.of(TextColors.GOLD,snapshots));

}
```

gabizou:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1712) | **SpongeCommon**
What this does is moves all previous plugin specific logic to flow
through the very same block physics logic that Minecraft (Sponge)
otherwise performs. With no longer needing to have the flags anded
or'ed or modified by bitwise functions, they are now boolean flags
as dictated by the BlockChangeFlag.

What this also does is fix a nasty bug that caused plugin code
performing block changes to never throw events, but get notified and
performed otherwise normally.

Fixes #1237
Fixes #998

### DIFF
--- a/src/main/java/org/spongepowered/common/registry/type/world/BlockChangeFlagRegistryModule.java
+++ b/src/main/java/org/spongepowered/common/registry/type/world/BlockChangeFlagRegistryModule.java
@@ -1,0 +1,194 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.registry.type.world;
+
+import com.google.common.collect.ImmutableList;
+import it.unimi.dsi.fastutil.ints.Int2ObjectLinkedOpenHashMap;
+import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
+import org.spongepowered.api.registry.RegistryModule;
+import org.spongepowered.api.registry.util.RegisterCatalog;
+import org.spongepowered.api.world.BlockChangeFlags;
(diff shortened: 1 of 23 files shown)


gabizou:
Prodding @simon816 @kashike @phit @ryantheleach for review. I've tested locally that this functions as it should (both in world gen, normal Minecraft, and plugin based code).

src/main/java/org/spongepowered/common/world/SpongeBlockChangeFlag.java
```diff
@@ -0,0 +1,197 @@
...
+    public int getRawFlag() {
+        return rawFlag;
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                .add("updateNeighbors", this.updateNeighbors)
+                .add("performBlockPhysics", this.performBlockPhysics)
+                .toString();
```
Cybermaxke: The other flags?

src/main/java/org/spongepowered/common/world/SpongeBlockChangeFlag.java
```diff
@@ -0,0 +1,197 @@
...
+        if (this.updateNeighbors == updateNeighbors) {
+            return this;
+        }
+        final int maskedFlag = (updateNeighbors ? BlockChangeFlagRegistryModule.Flags.NEIGHBOR_MASK : 0)
+                               | (this.performBlockPhysics ? 0 : BlockChangeFlagRegistryModule.Flags.PHYSICS_MASK)
+                               | (this.notifyObservers ? 0 : BlockChangeFlagRegistryModule.Flags.OBSERVER_MASK)
+                               | (this.notifyClients ? BlockChangeFlagRegistryModule.Flags.NOTIFY_CLIENTS : 0)
+                               | (this.ignoreRender ? BlockChangeFlagRegistryModule.Flags.IGNORE_RENDER : 0)
+                               | (this.forceReRender ? BlockChangeFlagRegistryModule.Flags.FORCE_RE_RENDER : 0);
+        for (SpongeBlockChangeFlag blockChangeFlag : BlockChangeFlagRegistryModule.getInstance().getValues()) {
```
Cybermaxke: Reuse `BlockChangeFlagRegistryModule#fromNativeInt`?

src/main/java/org/spongepowered/common/registry/type/world/BlockChangeFlagRegistryModule.java
```diff
@@ -0,0 +1,193 @@
...
+    @RegisterCatalog(BlockChangeFlags.class)
+    private final Map<String, SpongeBlockChangeFlag> flags = new HashMap<>();
+    private static BlockChangeFlagRegistryModule INSTANCE = new BlockChangeFlagRegistryModule();
+
+    public static BlockChangeFlagRegistryModule getInstance() {
+        return INSTANCE;
+    }
+
+    public static BlockChangeFlag fromNativeInt(int flag) {
+        for (SpongeBlockChangeFlag spongeBlockChangeFlag : getInstance().getValues()) {
```
Cybermaxke: Use Int2ObjectHashMap instead?

PR #1639, opened 2017-12-07
Add a check if the user is running Java 9

DoNotSpamPls:
I'd like to note that this has not been tested (I will test when I have some free time).
If someone can check for errors or things that don't work, I will be very grateful 😃

### DIFF
--- a/src/java6/java/org/spongepowered/launch/JavaVersionCheckUtils.java
+++ b/src/java6/java/org/spongepowered/launch/JavaVersionCheckUtils.java
@@ -32,10 +32,15 @@ public class JavaVersionCheckUtils {
 
     private static final String REQUIRED_VERSION = "1.8.0_20";
 
-    private static final String ERROR_MESSAGE = "We have detected that you are JRE version %s, which is out of date!\n"
+    private static final String ERROR_MESSAGE = "We have detected that you are using JRE version %s, which is out of date!\n"
                                                 + "In order to run Sponge, you **must** be running JRE version %s (or above).\n"
                                                 + "Previous builds of JRE version 1.8 will not work with Sponge.\n"
                                                 + "This can be downloaded from Oracle: http://www.oracle.com/technetwork/java/javase/downloads/jre8-downloads-2133155.html";
+    
+    private static final String UNSUPPORTED = "We have detected that you are using JRE version %s, which is currently not supported!\n"
+                                              + "In order to run Sponge, you **must** be running JRE version %s (or above).\n"
+                                              + "Previous builds of JRE version 1.8 will not work with Sponge.\n"
+                                              + "This can be downloaded from Oracle: http://www.oracle.com/technetwork/java/javase/downloads/jre8-downloads-2133155.html";
 
     public static void ensureJava8() {
         String version = getCurrentVersion();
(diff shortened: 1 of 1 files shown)


me4502:
I don’t see what you’ve done here. It seems you’ve just duplicated the previous message, and added another method to check for it (That isn’t even called anywhere)?

Java 9 isn’t supported in Sponge as a vast majority of the Minecraft Toolchain doesn’t support it at this current point in time. Work is being done, with it depending on the replacement for LaunchWrapper in the time being.

ST-DDT (changes requested):
Please follow the formatting guidelines and check that it at least compiles.

src/java6/java/org/spongepowered/launch/JavaVersionCheckUtils.java
```diff
@@ -32,10 +32,15 @@
...
     private static final String REQUIRED_VERSION = "1.8.0_20";
 
-    private static final String ERROR_MESSAGE = "We have detected that you are JRE version %s, which is out of date!\n"
+    private static final String ERROR_MESSAGE = "We have detected that you are using JRE version %s, which is out of date!\n"
                                                 + "In order to run Sponge, you **must** be running JRE version %s (or above).\n"
                                                 + "Previous builds of JRE version 1.8 will not work with Sponge.\n"
                                                 + "This can be downloaded from Oracle: http://www.oracle.com/technetwork/java/javase/downloads/jre8-downloads-2133155.html";
+    
+    private static final String UNSUPPORTED = "We have detected that you are using JRE version %s, which is currently not supported!\n"
+                                              + "In order to run Sponge, you **must** be running JRE version %s (or above).\n"
```
Deamon5550, line 41: This message does not make sense to a user running java 9.

src/java6/java/org/spongepowered/launch/JavaVersionCheckUtils.java
```diff
@@ -51,6 +56,19 @@ public static void ensureJava8() {
...
+    
+    public static void checkForNewVersions() {
+        String version = getCurrentVersion();
+        if (getVersionValue(version) > getVersionValue(REQUIRED_VERSION)) {
+            String error = String.format(UNSUPPORTED, version, REQUIRED_VERSION);
+
+                if (!GraphicsEnvironment.isHeadless()) {
+                    JOptionPane.showMessageDialog(null, error, "PEBKACException!", JOptionPane.ERROR_MESSAGE);
+                }
+            throw new RuntimeException(error);
```
Deamon5550, line 68: You've removed the if but left the print after the throw, so this doesn't compile.

src/java6/java/org/spongepowered/launch/JavaVersionCheckUtils.java
```diff
@@ -51,6 +56,19 @@ public static void ensureJava8() {
             System.out.println("You may be running an outdated version of Java. Any crashes from Sponge may require an update to Java.");
         }
     }
+    
+    public static void checkForNewVersions() {
+        String version = getCurrentVersion();
+        if (getVersionValue(version) > getVersionValue(REQUIRED_VERSION)) {
```
Deamon5550, line 62: This will cause all java 8 builds except build 20 to fail...

Deamon5550 (changes requested):
Also don't forget to actually call it.

gabizou:
Code doesn't compile, doesn't accomplish a Java 9 check, and far more to the game would fail to load, even with a sponge check.

PR #1641, opened 2017-12-10
Implements spawn load methods in MixinDimensionType

vectrixdevelops:
Before this, it caused a crash on SpongeVanilla when making a completely brand new server with a newly generated configuration and world, as it does not mixin the implemented method for `shouldLoadSpawn` like SpongeForge does. 

Original Crash: https://gist.github.com/connorhartley/dce0285cc1d8f22434e06de03463ad81

https://github.com/SpongePowered/SpongeForge/blob/bleeding/src/main/java/org/spongepowered/mod/mixin/core/world/MixinDimensionType.java - It is overriden by this for Forge (this PR fixes it for SV mainly).

I've tested this fix for Forge and Vanilla and it works.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/MixinDimensionType.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/MixinDimensionType.java
@@ -62,6 +62,7 @@ public abstract class MixinDimensionType implements IMixinDimensionType {
     private SpongeConfig<DimensionConfig> config;
     private volatile Context context;
     private boolean generateSpawnOnLoad;
+    private boolean loadSpawn;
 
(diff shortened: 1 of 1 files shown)


parlough:
@Minecrell and @Zidane I know you had some discussion about this, did you ever come to a conclusion on what we want to do? This issue is probably quite off-putting to those using SV when they start the server for the first time.

randombyte-developer:
What's the status on this? Apart from the fact that it is annoying, it may discourage some people using Sponge or makes them think that Sponge randomly crashes/is unstable idk.

Zidane:
@Meronat put this into stable-7 please.

PR #1645, opened 2017-12-11
Advancements implementation

Cybermaxke:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1554) | **Common** | [Forge](https://github.com/SpongePowered/SpongeForge/pull/1913) | [Vanilla](https://github.com/SpongePowered/SpongeVanilla/pull/363)

This is the implementation of the advancements API.


src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancement.java
```diff
@@ -66,13 +76,13 @@
 
     @Shadow @Nullable public Advancement parent;
     @Shadow @Final private ResourceLocation id;
-    @Shadow @Final private String[][] requirements;
```
gabizou: Are these no longer final? Or are you mutating them later with the added AT? In all honesty, you could still set them internally because `@Mutable` and `@Final` still allows you to set these values within code. The only time you realistically have to mark them as non final is if you want to make a method overridable. Unless I'm mistaken and @Mumfrey can correct me on the statement above.

Cybermaxke: I am not very familiar with all the available features in mixin, that's why I used the AT. I replaced most of the AT entries with `@Mutable` fields in the latest commit.

PR #1652, opened 2017-12-15
Implement DamageEntityEvent for armor stands.

### LINKED ISSUE #1644: DamageEntityEvent not working with armor stands
SpongeAPI version: stable-6
SpongeCommon version: stable-6
SpongeForge version: stable-6 

The `DamageEntityEvent` should be called when a `ArmorStand` is hit since it's an entity.

The `DamageEntityEvent` isn't fired for `ArmorStand`s.

My listener code:
```java
@Listener
public void onDeath(DamageEntityEvent event, @Getter("getTargetEntity") Entity entity) {
   Felograms.getInstance().getLogger().info(entity.getType().toString());

   if(entity.getType().equals(EntityTypes.ARMOR_STAND)) {
       logger.info("You hit an armorstand");
   }
}
```

When I hit a `Slime` it's working. Output:
`[14:20:42] [Server thread/INFO] [felograms]: SpongeEntityType{id=minecraft:slime, name=slime, translation=SpongeTranslation{id=entity.Slime.name}, modid=minecraft, class=net.minecraft.entity.monster.EntitySlime}`

but with an `ArmorStand` it isn't.

JBYoshi:
Input wanted because armor stands don't use the regular health system for damage, but it seems like the regular damage system to users.

In vanilla:
* Armor stands are immediately destroyed upon taking damage from the void or an explosion. I implement this by firing an event with a very large damage value (1000) and then changing it to fire damage if someone reduces it.
* Armor stands take regular damage from fire, but use a custom method for that damage. I pass the damage specified for that through an event and then pull the final value back from the event.
* Armor stands are usually destroyed from two punches or arrow hits from a player within 5 ticks of each other (EDIT: only one in creative), but the first push or arrow hit does not affect the armor stand's health. For the first punch, I fire a damage event with 0 damage and then ignore the resulting value. For the second, I do the same procedure as explosions/void damage.

Fixes #1644.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityArmorStand.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityArmorStand.java
@@ -27,7 +27,9 @@ package org.spongepowered.common.mixin.core.entity.item;
 import com.flowpowered.math.vector.Vector3d;
 import com.google.common.collect.Maps;
 import net.minecraft.entity.item.EntityArmorStand;
+import net.minecraft.util.DamageSource;
 import net.minecraft.util.math.Rotations;
+import org.spongepowered.api.Sponge;
 import org.spongepowered.api.data.key.Keys;
 import org.spongepowered.api.data.manipulator.DataManipulator;
 import org.spongepowered.api.data.manipulator.mutable.entity.ArmorStandData;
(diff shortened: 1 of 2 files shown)


parlough:
I suppose we should handle boats and minecarts in some similar fashion as they take multiple hits as well I believe.

JBYoshi:
I'm planning to do those entities as well once we decide on this (or something else) as precedent.

felixoi:
> Armor stands are usually destroyed from two punches

I don't know if it's important but creative players will only need one hit.

JBYoshi:
@felixoi Right... forgot to implement that one. Should be up now.

felixoi (approved):

felixoi:
Mahh, i found another bug. If I cancel the event now and try to attack the armorstand it wouldn't get any damage but with `/kill @e` it will die instantly. All other entites (Cow, Zombie, etc.) will not die if I cancel the event and use the `/kill` command.

ryantheleach:
@gabizou This is blocking @felixoi for their hologram plugin they want to release soon, as a code owner could you get a chance to review this so @JBYoshi can either close or go ahead and complete work for the other entities?

Zidane:
@JBYoshi 

Can you fix the merge conflicts so this can be merged?

JBYoshi:
@Zidane Updated. I changed the base to `stable-7` since there are no API breakages; let me know if this should be reverted.

src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityArmorStand.java
```diff
@@ -108,4 +123,87 @@ public void supplyVanillaManipulators(List<DataManipulator<?, ?>> manipulators)
...
+        }
+    }
+
+    @Inject(method = "attackEntityFrom", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/item/EntityArmorStand;playBrokenSound()V"),
+            cancellable = true)
+    private void fireDamageEventCreativePunch(DamageSource source, float amount, CallbackInfoReturnable<Boolean> cir) {
+        fireDestroyDamageEvent(source, cir);
+    }
+
+    @Inject(method = "attackEntityFrom", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/World;setEntityState(Lnet/minecraft/entity/Entity;B)V"),
```
gabizou: This injection and the next injection appear to inject at the same point, maybe use a slice if they aren't? Or make it clearer where it's injecting into?

All of the injections themselves seem out of place if we can't just overwrite the method (though that also has drawbacks). I'm somewhat keen on an idea that we should start using slices more often to define where a method we're expecting to inject into should exist.

gabizou (changes requested):

felixoi:
Could we get this done somewhen in the near future?

src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityArmorStand.java
JBYoshi: Whoops. I meant `dropBlock()` instead.

I also threw in slices for a bunch of the methods.

gabizou:
Thanks for updating, I’ll review the injections, is it possible to make some of the slices use both to and from?

PR #1658, opened 2017-12-19
Proposal to fix setting weather in ChangeWorldWeatherEvent - Fixes #1568

### LINKED ISSUE #1568: ChangeWorldWeatherEvent "Set" methods not working properly
Hi there. I'm running latest SpongeVanilla 1.12.2 and I've encountered some issues with the ChangeWorldWeatherEvent.
The "get" methods work perfectly, but the "set" methods do not. Setting the duration does nothing at all, while setting the weather will result in the event firing every tick while the weather lasts. This is due to setWeather only changing the internal weather variable and not setting the actual world weather. It then sees the new variable is different from the actual current weather and fires the event again.
The relevant code can be found at https://github.com/SpongePowered/SpongeCommon/blob/92274900f2992ecd51f98f849edfc8e82a87fb23/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java#L2242-L2243.
Could you please fix it? Thanks a lot for your time.

parlough:
Fixes https://github.com/SpongePowered/SpongeCommon/issues/1568

This is my proposal to fix setting weather `ChangeWorldWeatherEvent` and the event being called at the wrong time as well.

I was not sure the best way to go about this, as it was commented out for a reason. My solution was to simply check if the weather or duration changed as to not set the duration to its original status again.

It seems to work fine, but if you have an idea for a better solution please share.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
@@ -2268,21 +2268,22 @@ public abstract class MixinWorldServer extends MixinWorld implements IMixinWorld
 
     @Inject(method = "updateWeather", at = @At(value = "RETURN"))
     private void onUpdateWeatherReturn(CallbackInfo ci) {
-        Weather weather = getWeather();
+        final Weather weather = getWeather();
         int duration = (int) getRemainingDuration();
         if (this.prevWeather != weather && duration > 0) {
-            Sponge.getCauseStackManager().pushCause(this);
-            ChangeWorldWeatherEvent event = SpongeEventFactory.createChangeWorldWeatherEvent(Sponge.getCauseStackManager().getCurrentCause(), duration, duration,
-                    weather, weather, this.prevWeather, this);
-            SpongeImpl.postEvent(event);
-            Sponge.getCauseStackManager().popCause();
-            if (event.isCancelled()) {
-                this.setWeather(this.prevWeather);
-            } else {
-                // TODO: Rewrite this correctly so it doesn't rain 24/7
-                //this.setWeather(event.getWeather(), event.getDuration());
-                this.prevWeather = event.getWeather();
-                this.weatherStartTime = this.worldInfo.getWorldTotalTime();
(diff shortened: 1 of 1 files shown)


ryantheleach:
It was commented out in a hurried fix, to prevent it raining far too often in Sponge.

I don't think it actually fixed it, and that other changes might have happened since, but I definitely never saw evidence of how rain works, or how it is modified in Sponge.

Is it possible to create a command block statistics mechanic to determine rough averages of rain in Vanilla?

Otherwise compiling a raw MCP project could prove useful for Minecraft science in general, I'm not sure when the last complaint was, but it raining more often then expected has been a common complaint of Sponge since god knows when.

Dockter:
Would explain why some of my worlds rain so much?

parlough:
I do see a slightly better way to guarantee we don't mess with vanilla as all _maybe_, but for now, as I don't have time to explore that right now, this should fix some issues relating to the event being called when it shouldn't be and not being able to set weather.

Dockter:
I wouldn't say that the rain frequency declined any with this proposal.

ryantheleach:
~wip

I'm knocking this back to WIP @Meronat assuming Dockter's last comment was Dockter testing this locally for some time, unless anyone thinks this is better.

No rush, but when you get to try the better method do so.

phit:
Just for reference, I've been playing on a Forge only server and the rain frequency didn't seem any different from current SpongeForge.. it just rains a lot in this game.

parlough:
@phit Yeah I agree. This PR is just about us not allowing weather to be set when listening to the event.

We can handle weather better, but I feel that is out of the scope of this PR, and I won't get to that for a while.

parlough:
@ryantheleach ^

ryantheleach:
We used to have a bug where it was raining a ton. Servers that let you change the time of day or weather have traditionally reset the weather clock, causing it to rain more then expected due to the random seed raining a ton at the start.

That said, no one has done any real statistical analysis on how much it rains or found a root cause since blood's original hackfix seemed to only partially fix it.

/shrug Maybe it does just rain a ton now.

Dockter:
Damn that global warming....

PR #1659, opened 2017-12-19
Adds a simple enable/disable system for test plugins

parlough:
This is just a simple system to allow for enabling/disabling plugin's event listeners. The command is "/pluginidenable". The plugin just needs to extend `BehindCommandTestPlugin`. Not the greatest system, but it's good enough for now. 

No more pesky inventory messages :P


testplugins/src/main/java/org/spongepowered/test/BehindCommandTestPlugin.java
```diff
@@ -0,0 +1,68 @@
...
+                                src.sendMessage(Text.of(TextColors.DARK_GREEN, "You have enabled event testing for " + container.getName() + "."));
+                            } else {
+                                Sponge.getEventManager().unregisterListeners(this);
+                                src.sendMessage(Text.of(TextColors.DARK_GREEN, "You have disabled event testing for " + container.getName() + "."));
+                            }
+
+                            return CommandResult.success();
+                        })
+                        .build(),
+                this.container.getId() + "enable");
```
ryantheleach: Would love to see this behind a parent command, to aid tab completion.

testplugins/src/main/java/org/spongepowered/test/BehindCommandTestPlugin.java
```diff
@@ -0,0 +1,68 @@
...
+import org.spongepowered.api.text.format.TextColors;
+
+import javax.inject.Inject;
+
+public class BehindCommandTestPlugin {
+
+    @Inject
+    private PluginContainer container;
+
+    private boolean enabled;
```
ryantheleach: Is this supposed to be private?

ryantheleach (commented):
I started reviewing then realized this is probably still WIP

testplugins/src/main/java/org/spongepowered/test/BehindCommandTestPlugin.java
parlough: Hmm yeah as only the command registered here uses it, but protected might be better in case the sub-classes want to use it.

ryantheleach:
This is different then the approach I would have taken. 

If I had programmed this, I would have had each of these plugins extend a LoadableModule unless they needed to be run at game init time for recipes or other registries, and have a command that can configure (persistent across restarts, enabled by default, maybe 3 statuses? loaded, noisy, unloaded?) loading of the modules while the server is running (defaulting to loaded).

If a plugin needed to be run earlier, I wasn't sure what approach I would take, probably the same as what you have done, or a slightly different LoadableModule who can only be loaded by the persistent config file on start?

This way people extending it would only have to extend LoadableModule without implementing anything, or having anything gated (except maybe a 'noisy' check?) and the rest would take care of itself.

Loaded would mean running / mixing in(potentially), noisy would be spamming / taking effect, and unloaded would be completely unloaded.

testplugins/src/main/java/org/spongepowered/test/BehindCommandTestPlugin.java
parlough: That is definitely a possibility, will work on something.

parlough:
@ryantheleach I like the idea and it is something that can definitely can be explored as a larger system, I thought of this as simply a temporary solution until we come up with a better, more-fleshed out system so the random tests don't annoy us in development.

testplugins/src/main/java/org/spongepowered/test/BehindCommandTestPlugin.java
```diff
@@ -40,13 +42,13 @@
...
 
-    private boolean enabled;
+    protected boolean enabled;
 
-    @Listener
-    public void onInit(final GamePreInitializationEvent event) {
-        Sponge.getEventManager().unregisterListeners(this);
-        Sponge.getCommandManager().register(this,
-                CommandSpec.builder()
+    protected CommandSpec.Builder baseCommand;
```
ryantheleach: Static?

gabizou:
I'm definitely interested in a more robust system that would allow us to have toggle able tests. @ryantheleach's idea is the line that I've been meaning to implement at some point.

ImMorpheus:
Superseded by https://github.com/SpongePowered/SpongeCommon/pull/2195

PR #1666, opened 2017-12-23
Allow plugins to use a separate implementation class for custom data.

JBYoshi:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1723) | **SpongeCommon**

Sponge uses an interface and a separate implementing class for its `DataManipulator`s, and a few plugin developers have been [trying to use that model](https://forums.spongepowered.org/t/custom-data-not-working/21832?u=jbyoshi). However, `DataRegistration` doesn't allow multiple classes to be registered to it. This PR adds that option. (However, it remains optional for any plugins that don't currently use it, or don't want to use it.)


felixoi:
Working but throws an warning currently:
`A DataBuilder has already been registered for interface net.felixoi.felograms.api.data.HologramData. Attempted to register class net.felixoi.felograms.data.FelogramsHologramDataBuilder instead.`

PR #1670, opened 2017-12-28
Implement ChangeValueEvent

gabizou:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1728) | **SpongeCommon**
This sets up the system for implementing `ChangeValueEvent` if only for Entities at present time. `TileEntity` instances are a little different to handle as there is no data sync system short of serializing the `TileEntity` and sending it over the wire.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/datasync/entity/EntityCustomNameConverter.java
+++ b/src/main/java/org/spongepowered/common/data/datasync/entity/EntityCustomNameConverter.java
@@ -1,0 +1,50 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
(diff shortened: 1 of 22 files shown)


Zidane:
I agree with @kashike here. That is hard to see what is happening.

src/main/java/org/spongepowered/common/data/datasync/entity/EntityCustomNameConverter.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public Optional<DataTransactionResult> createTransaction(String currentValue, String value) {
+        return Optional.empty();
+    }
+
+    @Override
+    public String getValueFromEvent(String originalValue, List<ImmutableValue<?>> immutableValues) {
+        return null;
```
liach, line 48: Is this just a stub?

PR #1672, opened 2017-12-29
Add "minecraft.force-gamemode.override" permission for overriding force gamemode

dualspiral:
This allows for players to bypass the `force-gamemode` setting in server.properties on login. I've opted for `minecraft.force-gamemode.override` as the permission to try to be consistent with the spawn protection permission.

I looked for forced gamemodes when transferring worlds, couldn't see it, indeed, it seems that world gamemodes seem to be ignored for the most part. I imagine it's not there, but if it is, I'd appreciate it being pointed out! Similarly, if there is a better suggestion for the permission itself, that'd be dandy.

This came about because @Inscrutable asked me (in the Nucleus discord)

> Is it possible to bypass the forced gamemode of a server? I'd like to let admin staff and camera crews to stay in creative, but I have the server forcing survival mode. I can't see any way around it with either nucleus or sponge. Am I missing something, or is it not possible?

> it's sorta mission critical to camerabots in gamemode 3 to stay that way without interruption. I guess OP is a good enough solution. Amazing, OP is some use again at last

### DIFF
--- a/src/main/java/org/spongepowered/common/interfaces/entity/player/IMixinEntityPlayerMP.java
+++ b/src/main/java/org/spongepowered/common/interfaces/entity/player/IMixinEntityPlayerMP.java
@@ -82,4 +82,7 @@ public interface IMixinEntityPlayerMP extends IMixinEntityPlayer {
     void injectScaledHealth(Collection<IAttributeInstance> set, boolean b);
 
     void updateDataManagerForScaledHealth();
+
+    boolean hasForcedGamemodeOverridePermission();
+
 }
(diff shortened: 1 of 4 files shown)


ST-DDT:
The permission needs to be documented in SpongeDocs, if it gets merged.

gabizou (approved):
Looks good. I can't imagine this being troublesome in the future.

parlough (approved):

dualspiral:
I'll merge in the morning unless someone wants to get there first then. It'd probably be useful to get this in before SoS so there's less of an issue with the camera bots.

Inscrutable:
Actually, OP does *not* appear to be a solution, at least on the SCS (Luckperms etc). I'm OP, and it forces *my* gamemode to Survival, the same as everyone else. So I guess until this is in, the only alternative I have is to drop the ``force-gamemode``.

kashike (approved):

dualspiral:
https://github.com/SpongePowered/SpongeCommon/commit/2ef518e73c39e1a6ee3fba8761ba9a468540aed2

PR #1681, opened 2018-01-02
Fix event generics.

Cybermaxke:
**Common** | [Forge](https://github.com/SpongePowered/SpongeForge/pull/1929)

Fixes generic events that use complex generic types, like `DataRegistration<?,?>`, `Key<? extends BaseValue<?>>`, etc.
Also fixes a issue that caused the cause stack manager to be used on the client when it's not supposed to.

@gabizou

### DIFF
--- a/src/main/java/org/spongepowered/common/SpongeImpl.java
+++ b/src/main/java/org/spongepowered/common/SpongeImpl.java
@@ -197,10 +197,18 @@ public final class SpongeImpl {
         return Sponge.getEventManager().post(event);
     }
 
+    public static boolean postEvent(Event event, boolean allowClient) {
+        // TODO quick and dirty fix (cant cast in UnitTest)
+        if (Sponge.getEventManager() instanceof SpongeEventManager) {
+            return ((SpongeEventManager) Sponge.getEventManager()).post(event, allowClient);
+        }
+        return true;
+    }
+
     public static void postState(GameState state, GameStateEvent event) {
         check(game);
         game.setState(state);
-        game.getEventManager().post(event);
+        postEvent(event, true);
     }
 
     public static void postShutdownEvents() {
--- a/src/main/java/org/spongepowered/common/SpongeImpl.java
+++ b/src/main/java/org/spongepowered/common/SpongeImpl.java
@@ -197,18 +197,10 @@ public final class SpongeImpl {
         return Sponge.getEventManager().post(event);
     }
 
-    public static boolean postEvent(Event event, boolean allowClient) {
-        // TODO quick and dirty fix (cant cast in UnitTest)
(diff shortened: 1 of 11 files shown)


randombyte-developer:
:+1: I didn't bother debugging the non-working listener and kept the 'old' PreInit event for the registration of Keys. Thanks for fixing it!

testplugins/src/main/java/org/spongepowered/test/myhomes/MyHomes.java
```diff
@@ -67,61 +68,67 @@
...
-    public void onKeyRegistration(GameRegistryEvent.Register<Key<?>> evnet) {
+    public void onKeyRegistration(GameRegistryEvent.Register<Key<?>> event) {
+        this.logger.info("onKeyRegistration");
         DEFAULT_HOME = Key.builder()
             .type(new TypeToken<Value<Home>>() {})
             .id("myhomes:default_home")
             .name("Default Home")
             .query(DataQuery.of("DefaultHome"))
             .build();
+        event.register(DEFAULT_HOME);
```
gabizou, line 85: Unless you changed the `SpongeKeyBuilder` implementation, this is redundant and should throw an exception for attempting double registration.

src/main/java/org/spongepowered/common/SpongeImpl.java
```diff
@@ -197,18 +197,10 @@ public static boolean postEvent(Event event) {
...
-        if (Sponge.getEventManager() instanceof SpongeEventManager) {
-            return ((SpongeEventManager) Sponge.getEventManager()).post(event, allowClient);
-        }
-        return true;
-    }
-
     public static void postState(GameState state, GameStateEvent event) {
         check(game);
         game.setState(state);
-        ((SpongeEventManager) game.getEventManager()).post(event, true);
```
gabizou: The reason why this was added is because on clients, this will not work unless allowing a client thread event is true.

gabizou (changes requested):

gabizou:
I'm curious why you are requiring `Key`s to be registered through the event instead of through the `build()` method as it was originally designed (much like `DataRegistration` for that matter).

Cybermaxke:
@gabizou I think we should be consistent how we register stuff. And if we aren't going to use the `register` method then there is no point to register things inside event, might as well do it in the init event. `DataRegistration` would require it's `buildAndRegister` to be changed into `build` to fit into this concept.

Cybermaxke:
@gabizou This PR should also be merged into bleeding.

dualspiral:
@Cybermaxke I've pulled all the latest `stable-7` changes into `bleeding` now.

PR #1691, opened 2018-01-08
Fix skipping BlockChangeFlags.NEIGHBOR_PHYSICS

pie-flavor:
Seriously, how long has this startup message been around for?


src/main/java/org/spongepowered/common/registry/type/world/BlockChangeFlagRegistryModule.java
```diff
@@ -125,7 +125,7 @@ public void registerDefaults() {
             } else if (Flags.OBSERVER == i) {
                 register(new SpongeBlockChangeFlag("OBSERVER".toLowerCase(Locale.ENGLISH), i));
             } else if (Flags.NEIGHBOR_PHSYICS == i) {
```
dualspiral: You might as well rename this flag to be spelt correctly while you're at it. As it's in Common only, we don't need to deprecate and remove next version (i.e., this can then be backported to stable though as it's internal, not the biggest problem to leave it in stable either).

pie-flavor: Good idea, hadn't noticed it.

gabizou (approved):

PR #1698, opened 2018-01-10
Fix ride entity event not firing - Fixes #1345

### LINKED ISSUE #1345: Ride entity event not firing
**I am currently running...**

- SpongeVanilla: 1.11.2-7.0.0-BETA-263
- Java version: 1.8.0_91

Operating System:

 - [x] Linux (Ubuntu, Debian, Fedora, etc)
 - [ ] macOS
 - [x] Windows

- Plugins/Mods: ElderGuardian, WorldEdit, Precogs, Guardian, VoxelSniper

**What is the expected result?** The event to fire for mounting and dismounting entities.

**What is the current result?** The event never fires when attempting to listen for it, when mounting a horse, riding a boat and riding a pig.

**Example?** 

Of my current listener class. The first two events fire successfully. The RideEntityEvent does not.

```@Listener
    public void onEntityMove(MoveEntityEvent event, @Getter("getTargetEntity") Player player) {
        System.out.println("Player moved.");
    }

    @Listener
    public void onEntityDeath(DestructEntityEvent.Death event, @Getter("getTargetEntity") Player player) {
        System.out.println("Player died.");
    }

    @Listener
    public void onRideEntity(RideEntityEvent event) {
        System.out.println("Mount / Dismount");
    }

parlough:
Now you can listen to the base event or the sub-events just fine.

Fixes https://github.com/SpongePowered/SpongeCommon/issues/1345

### DIFF
--- a/src/main/java/org/spongepowered/common/event/ShouldFire.java
+++ b/src/main/java/org/spongepowered/common/event/ShouldFire.java
@@ -53,6 +53,7 @@ public class ShouldFire {
     public static boolean DROP_ITEM_EVENT_DESTRUCT = false;
     public static boolean DROP_ITEM_EVENT_DISPENSE = false;
 
+    public static boolean RIDE_ENTITY_EVENT = false;
     public static boolean RIDE_ENTITY_EVENT_MOUNT = false;
     public static boolean RIDE_ENTITY_EVENT_DISMOUNT = false;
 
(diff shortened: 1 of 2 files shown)


parlough:
I thought this should work fine even before this, I'll need to look into `ShouldFire` more but for now this is good.

PR #1702, opened 2018-01-10
Implement basic elytra data

parlough:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1737) | **SpongeCommon**

Implemented basic elytra data and a simple test plugin. Find more information on the API PR.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/manipulator/immutable/entity/ImmutableSpongeElytraFlyingData.java
+++ b/src/main/java/org/spongepowered/common/data/manipulator/immutable/entity/ImmutableSpongeElytraFlyingData.java
@@ -1,0 +1,45 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data.manipulator.immutable.entity;
+
+import org.spongepowered.api.data.key.Keys;
+import org.spongepowered.api.data.manipulator.immutable.entity.ImmutableElytraFlyingData;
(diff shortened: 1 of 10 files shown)


src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntityLivingBase.java
```diff
@@ -933,4 +933,8 @@ public void supplyVanillaManipulators(List<DataManipulator<?, ?>> manipulators)
         return ProjectileLauncher.launch(checkNotNull(projectileClass, "projectile class"), this, checkNotNull(velocity, "velocity"));
     }
 
+    public void setElytraFlying(boolean value) {
+        setFlag(7, value);
```
mattmess1221: :-1: for magic number

parlough: Well that's how they set it, not sure of a better way since they don't have an explicit method besides that.

mattmess1221: You could add a constant to replace it. e.g. 

`public static final int FLAG_ELYTRA = 7;`

gabizou: DataConstants!

src/main/java/org/spongepowered/common/data/manipulator/immutable/entity/ImmutableSpongeElytraFlyingData.java
```diff
@@ -0,0 +1,45 @@
...
+    public ImmutableSpongeElytraFlyingData(boolean value) {
+        super(ImmutableElytraFlyingData.class, value, Keys.IS_ELYTRA_FLYING, SpongeElytraFlyingData.class, false);
+    }
+
+    @Override
+    public ImmutableValue<Boolean> elytraFlying() {
+        return getValueGetter();
+    }
+
+}
```
gabizou: New Line.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntityLivingBase.java
```diff
@@ -933,4 +934,8 @@ public void supplyVanillaManipulators(List<DataManipulator<?, ?>> manipulators)
         return ProjectileLauncher.launch(checkNotNull(projectileClass, "projectile class"), this, checkNotNull(velocity, "velocity"));
     }
 
+    public void setElytraFlying(boolean value) {
+        setFlag(DataConstants.ELYTRA_FLYING_FLAG, value);
```
gabizou: If we're adding support for the flag, can we have a processor for the change value event?

gabizou (changes requested):
Is it possible to add implementation for the change value event?

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntityLivingBase.java
parlough: Yeah I'll do so when I get a chance, hopefully this weekend. Will need to look at how you've done it with the others.

testplugins/src/main/java/org/spongepowered/test/ElytraDataTest.java
```diff
@@ -0,0 +1,96 @@
...
+                return CommandResult.success();
+            }).build();
+
+        final CommandSpec enable = CommandSpec.builder()
+            .arguments()
+            .executor((src, args) -> {
+                src.sendMessage(Text.of(TextColors.DARK_GREEN, "You have enabled elytra listening."));
+                Keys.IS_ELYTRA_FLYING.registerEvent(Player.class, e -> {
+                    if (e.getTargetHolder() instanceof Player) {
+                        ((Player) e.getTargetHolder()).sendMessage(Text.of(TextColors.DARK_GREEN, "Changed elytra status!"));
```
parlough, line 81: @gabizou So I was working on adding support for the listening to the change value event but I couldn't get it to work for manually setting the key or changing elytra state in game. Am I missing a step necessary for implementing the event?

gabizou: It might not be possible to change for the *client* since this is a server side modification, may have to perform it on another player to test properly (Sponge human anyone?)

parlough: I mean I can change the player to elytra flying just fine, but the listener is never called. I'm not sure how to properly implement the key change listener.

Zidane:
Manually merged by me into stable-7.

PR #1704, opened 2018-01-12
Throw an exception when using Entity#remove on a player. Close #989

### LINKED ISSUE #989: Entity removal of Player doesn't kick
Currently when a Player is removed (`Entity#remove()`), they just exist in some kind of weird state where they can't interact with anything, and updates are no longer sent to them (outside of the time).

My suggestion is that `Player` override `remove()` and re-direct it to whatever handles kicking a player.

/shrug

https://forums.spongepowered.org/t/entity-removal-help/15336

ImMorpheus:
Let me know if I need to add a test plugin or a message when throwing the exception.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
@@ -1194,6 +1194,8 @@ public abstract class MixinEntityPlayerMP extends MixinEntityPlayer implements P
 
     @Override
     public void remove() {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException("This is an internal method not intended for use with Players " +
+                "as it causes the player to be placed into an undefined state. " +
+                "Consider putting them through the normal death process instead.");
     }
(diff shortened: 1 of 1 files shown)


Zidane:
@ImMorpheus 

Can you change your target to stable-7?

parlough:
Maybe add a short message to the exception explaining why?

Zidane:
@ImMorpheus @Meronat 

This needs to be on stable-7...Player entities in the MCP implementation should never be told to just remove...it causes all hell to break loose and I don't need a plugin dev blindly calling that without thinking about it first.

parlough:
@Zidane It caused problems with his PR(quick fix but still), it needs to be in bleeding too anyway, and we can easily cherry pick the change. it's going to be a pain to keep asking people to switch to stable-7 as they have different bases and the default is bleeding at the moment. Dual made a good point we should target bleeding as it is bleeding and cherry-pick to 7 not the other way around just like we normally do.

parlough (approved):

Zidane:
@Meronat 

Generally I agree but with something so minor I don't want lagtime on fixes to stable-7.

PR #1706, opened 2018-01-13
refactor to hopefully improve performance

D3-M0N:
Removed all double calls on trackedShortBlockPositions.get and trackedIntBlockPositions.get...
Hopefully that improves the performance of playertracking.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/tracking/world/MixinChunk_Tracker.java
+++ b/src/main/java/org/spongepowered/common/mixin/tracking/world/MixinChunk_Tracker.java
@@ -166,9 +166,9 @@ public abstract class MixinChunk_Tracker implements Chunk, IMixinChunk {
             return getValidatedUser(intKey, notifierIndex);
         } else {
             final short shortKey = blockPosToShort(pos);
-            if (this.trackedShortBlockPositions.get(shortKey) != null) {
-                PlayerTracker tracker = this.trackedShortBlockPositions.get(shortKey);
-                final int notifierIndex = tracker.ownerIndex;
+            final PlayerTracker shortTracker = this.trackedShortBlockPositions.get(shortKey);
+            if (shortTracker != null) {       
+                final int notifierIndex = shortTracker.ownerIndex;
                 return getValidatedUser(shortKey, notifierIndex);
             }
         }
@@ -204,9 +204,9 @@ public abstract class MixinChunk_Tracker implements Chunk, IMixinChunk {
             return getValidatedUser(intKey, notifierIndex);
         } else {
             final short shortKey = blockPosToShort(pos);
-            if (this.trackedShortBlockPositions.get(shortKey) != null) {
-                PlayerTracker tracker = this.trackedShortBlockPositions.get(shortKey);
-                final int notifierIndex = tracker.notifierIndex;
+            final PlayerTracker shortTracker = this.trackedShortBlockPositions.get(shortKey);
+            if (shortTracker != null) { 
+                final int notifierIndex = shortTracker.notifierIndex;
                 return getValidatedUser(shortKey, notifierIndex);
             }
         }
@@ -286,9 +286,9 @@ public abstract class MixinChunk_Tracker implements Chunk, IMixinChunk {
     public void setBlockNotifier(BlockPos pos, @Nullable UUID uuid) {
         if (pos.getY() <= 255) {
             short blockPos = blockPosToShort(pos);
-            if (this.trackedShortBlockPositions.get(blockPos) != null) {
-                this.trackedShortBlockPositions.get(blockPos).notifierIndex = uuid == null ? -1 :
-                                                                              ((IMixinWorldInfo) this.world.getWorldInfo()).getIndexForUniqueId(uuid);
+            final PlayerTracker shortTracker = this.trackedShortBlockPositions.get(blockPos);
+            if (shortTracker != null) {
+            	shortTracker.notifierIndex = uuid == null ? -1 : ((IMixinWorldInfo) this.world.getWorldInfo()).getIndexForUniqueId(uuid);
             } else {
                 this.trackedShortBlockPositions.put(blockPos,
                         new PlayerTracker(uuid == null ? -1 : ((IMixinWorldInfo) this.world.getWorldInfo()).getIndexForUniqueId(uuid),
@@ -296,9 +296,9 @@ public abstract class MixinChunk_Tracker implements Chunk, IMixinChunk {
             }
         } else {
             int blockPos = blockPosToInt(pos);
-            if (this.trackedIntBlockPositions.get(blockPos) != null) {
-                this.trackedIntBlockPositions.get(blockPos).notifierIndex = uuid == null ? -1 :
-                                                                            ((IMixinWorldInfo) this.world.getWorldInfo()).getIndexForUniqueId(uuid);
+            final PlayerTracker intTracker = this.trackedIntBlockPositions.get(blockPos);
+            if (intTracker != null) {
+            	intTracker.notifierIndex = uuid == null ? -1 : ((IMixinWorldInfo) this.world.getWorldInfo()).getIndexForUniqueId(uuid);
             } else {
                 this.trackedIntBlockPositions.put(blockPos,
                         new PlayerTracker(uuid == null ? -1 : ((IMixinWorldInfo) this.world.getWorldInfo()).getIndexForUniqueId(uuid),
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/tracking/world/MixinChunk_Tracker.java
```diff
@@ -166,9 +166,9 @@ public void addTrackedBlockPosition(Block block, BlockPos pos, User user, Player
             return getValidatedUser(intKey, notifierIndex);
         } else {
             final short shortKey = blockPosToShort(pos);
-            if (this.trackedShortBlockPositions.get(shortKey) != null) {
-                PlayerTracker tracker = this.trackedShortBlockPositions.get(shortKey);
-                final int notifierIndex = tracker.ownerIndex;
+            final PlayerTracker shortTracker = this.trackedShortBlockPositions.get(shortKey);
+            if (shortTracker != null) {
```
ryantheleach: Excess whitespace

src/main/java/org/spongepowered/common/mixin/tracking/world/MixinChunk_Tracker.java
```diff
@@ -204,9 +204,9 @@ public void addTrackedBlockPosition(Block block, BlockPos pos, User user, Player
             return getValidatedUser(intKey, notifierIndex);
         } else {
             final short shortKey = blockPosToShort(pos);
-            if (this.trackedShortBlockPositions.get(shortKey) != null) {
-                PlayerTracker tracker = this.trackedShortBlockPositions.get(shortKey);
-                final int notifierIndex = tracker.notifierIndex;
+            final PlayerTracker shortTracker = this.trackedShortBlockPositions.get(shortKey);
+            if (shortTracker != null) { 
+                final int notifierIndex = shortTracker.notifierIndex;
```
ryantheleach: notifierIndex could probably be eliminated here.

src/main/java/org/spongepowered/common/mixin/tracking/world/MixinChunk_Tracker.java
```diff
@@ -286,19 +286,19 @@ public void addTrackedBlockPosition(Block block, BlockPos pos, User user, Player
     public void setBlockNotifier(BlockPos pos, @Nullable UUID uuid) {
         if (pos.getY() <= 255) {
             short blockPos = blockPosToShort(pos);
-            if (this.trackedShortBlockPositions.get(blockPos) != null) {
-                this.trackedShortBlockPositions.get(blockPos).notifierIndex = uuid == null ? -1 :
-                                                                              ((IMixinWorldInfo) this.world.getWorldInfo()).getIndexForUniqueId(uuid);
+            final PlayerTracker shortTracker = this.trackedShortBlockPositions.get(blockPos);
+            if (shortTracker != null) {
+            	shortTracker.notifierIndex = uuid == null ? -1 : ((IMixinWorldInfo) this.world.getWorldInfo()).getIndexForUniqueId(uuid);
```
ryantheleach: This needs to be spaces rather then tabs.

src/main/java/org/spongepowered/common/mixin/tracking/world/MixinChunk_Tracker.java
```diff
@@ -286,19 +286,19 @@ public void addTrackedBlockPosition(Block block, BlockPos pos, User user, Player
...
                                 PlayerTracker.Type.NOTIFIER));
             }
         } else {
             int blockPos = blockPosToInt(pos);
-            if (this.trackedIntBlockPositions.get(blockPos) != null) {
-                this.trackedIntBlockPositions.get(blockPos).notifierIndex = uuid == null ? -1 :
-                                                                            ((IMixinWorldInfo) this.world.getWorldInfo()).getIndexForUniqueId(uuid);
+            final PlayerTracker intTracker = this.trackedIntBlockPositions.get(blockPos);
+            if (intTracker != null) {
+            	intTracker.notifierIndex = uuid == null ? -1 : ((IMixinWorldInfo) this.world.getWorldInfo()).getIndexForUniqueId(uuid);
```
ryantheleach: This needs to be spaces rather then tabs.

src/main/java/org/spongepowered/common/mixin/tracking/world/MixinChunk_Tracker.java
```diff
@@ -312,18 +312,18 @@ public void setBlockNotifier(BlockPos pos, @Nullable UUID uuid) {
     public void setBlockCreator(BlockPos pos, @Nullable UUID uuid) {
         if (pos.getY() <= 255) {
             short blockPos = blockPosToShort(pos);
-            if (this.trackedShortBlockPositions.get(blockPos) != null) {
-                this.trackedShortBlockPositions.get(blockPos).ownerIndex = uuid == null ? -1 : ((IMixinWorldInfo) this.world.getWorldInfo())
-                        .getIndexForUniqueId(uuid);
+            final PlayerTracker shortTracker = this.trackedShortBlockPositions.get(blockPos);
+            if (shortTracker != null) {
+            	shortTracker.ownerIndex = uuid == null ? -1 : ((IMixinWorldInfo) this.world.getWorldInfo()).getIndexForUniqueId(uuid);
```
ryantheleach: This needs to be spaces rather then tabs.

src/main/java/org/spongepowered/common/mixin/tracking/world/MixinChunk_Tracker.java
```diff
@@ -312,18 +312,18 @@ public void setBlockNotifier(BlockPos pos, @Nullable UUID uuid) {
...
                         .getIndexForUniqueId(uuid), PlayerTracker.Type.OWNER));
             }
         } else {
             int blockPos = blockPosToInt(pos);
-            if (this.trackedIntBlockPositions.get(blockPos) != null) {
-                this.trackedIntBlockPositions.get(blockPos).ownerIndex = uuid == null ? -1 : ((IMixinWorldInfo) this.world.getWorldInfo())
-                        .getIndexForUniqueId(uuid);
+            final PlayerTracker intTracker = this.trackedIntBlockPositions.get(blockPos);
+            if (intTracker != null) {
+            	intTracker.ownerIndex = uuid == null ? -1 : ((IMixinWorldInfo) this.world.getWorldInfo()).getIndexForUniqueId(uuid);
```
ryantheleach: This needs spaces rather then tabs.

ryantheleach (approved):
I havn't run this, or checked the actual perf difference, but it's a pretty obvious minor change that should improve it.

LGTM barring minor formatting issues.

D3-M0N:
Formatting should be ok now:-)

ryantheleach (approved):

gabizou:
Not to be rude, but is there much noticeable performance or any numbers that could be provided in comparison?\ of the changes?

D3-M0N:
Its hard to get a good reading for that. In theory it should cut in half the time these methods need for their track player operations.

i will try to provide a comparison later.

kashike (approved):

me4502 (approved):

ImMorpheus:
> In theory it should cut in half the time 

This is not how java works, you're ignoring any sort of compiler optimization.

TechnicianLP:
The compiler doesnt optimize out the second get-call (if you had looked at the bytecode you would have seen that @ImMorpheus )
While youre at it you (@D3-M0N ) could change the code to use the fastutil maps for primitives correctly - casting them to java.util.Map defeats the purpose of using a primitive map in the first place ...

ImMorpheus:
The bytecode is also optimized at runtime while it's being processed by the JIT compiler. 
My point was just that the "the time is cut in half because the method calls have been halved " way of reasoning is wrong. That's why Forge asks for profiling samples before merging performance PR.

EDIT:

To clarify, I'm not against this. I just wanted to point that out.

TechnicianLP:
yes it is a bad estimate but its an improvement nontheless
i dont the jit will optimize it out either as the underlying map could have chnaged betweenn those calls (unlikely but possible)

gabizou (approved):

D3-M0N:
@TechnicianLP 

> While youre at it you (@D3-M0N ) could change the code to use the fastutil maps for primitives 

> correctly  casting them to java.util.Map defeats the purpose of using a primitive map in the first place ...

im on it

Zidane:
@D3-M0N 

I'll see about getting this merged when you make the change to fastutil maps. If not, I can see about having someone else do so.

D3-M0N:
Il tell you tomorow if i can finish that:-)

TechnicianLP:
After talking back to @D3-M0N we noticed that it is pretty much impossible to change the class over to using the FastUtil maps - we basically have two options:
- Make a copy of the Map in the set()-Method (which would break if someone makes changes to the map he passed to the set()-Method as they would not be reflected into the copied Map)
- Create Wrapper classes which relays all calls to a parent Map<Integer/Double, PlayerTracker> (which would require a huge boilerplate and has terrible performance both due to the strengthened iteration-methods (as well as having excessive autoboxing))

Question now is: Which of the two solutions should be used? The old set()-Methods should be deprecated and scheduled for removal anyways ....

kashike:
...why is it impossible?

gabizou:
Why would it be impossible? We do this in several cases.

TechnicianLP:
Sorry for the bad word choice - by impossible i basically mean we are unsure wether to break functionality someone might depend on or create a lot of boilerplate that will be removed again in the near future anyways ...
Oh and could you clarify what "We do this in several cases." is supposed to mean? (That sounds bad ...)

kashike:
...it's a simple replacement of one map with another.

TechnicianLP:
Its not that easy - we have to convert the Map<Integer/Double, Playertracker> fields to Int/Double2ObjectMaps in order to be able to use its methods with primitives - thos in return breaks the setters available in the class as they still take a basic map that we can no longer directly assign to the map fields - thats why i was asking which way those (old) setters should be changed

D3-M0N:
@Zidane i will provide an update tomorrow:-)

D3-M0N:
Hello, after a long time back and forth I would have 2 suggestions:

- Make a copy of the Map in the set () - Methods (@TechnicianLP)
- change the IMixinChunk interface set () - Methods to be deprecated and
   create 2 new methods and use fastutils maps

which should i use?

Zidane:
Manually merged into stable-7 with some changes of my own. Thanks!

PR #1708, opened 2018-01-13
Implement IgniteEntityEvent (Close #970)

### LINKED ISSUE #970: Implementation Request IgniteEntityEvent
1.10.2 versions dont have the IgniteEntityEvent fired when events burst into flame from fire, lava or sunlight (skeletons/zombies).  I dont know if this event was ever implimented in 1.8.9 and didn't make it.

DamageEntityEvent only goes so far as a workaround - it only registers living entities suffereing health damage ticks due to fire.

IgniteEntityEvent will capture armor stands being ignited by lava and fire  (as well as prevent need for listening to generic damage for living entities)

ImMorpheus:
I think I'm not getting the difference between "originalfireticks" and "fireticks".

https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/event/entity/IgniteEntityEvent.java

Is there a case when they are different ?

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
@@ -39,6 +39,7 @@ import net.minecraft.entity.EntityTracker;
 import net.minecraft.entity.EntityTrackerEntry;
 import net.minecraft.entity.EnumCreatureType;
 import net.minecraft.entity.MoverType;
+import net.minecraft.entity.effect.EntityLightningBolt;
 import net.minecraft.entity.item.EntityItem;
 import net.minecraft.entity.player.EntityPlayerMP;
 import net.minecraft.init.Blocks;
@@ -51,6 +52,7 @@ import net.minecraft.network.play.server.SPacketDestroyEntities;
 import net.minecraft.network.play.server.SPacketPlayerListItem;
 import net.minecraft.network.play.server.SPacketPlayerPosLook;
 import net.minecraft.util.DamageSource;
+import net.minecraft.util.EntityDamageSource;
 import net.minecraft.util.EnumParticleTypes;
 import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.SoundCategory;
@@ -83,6 +85,7 @@ import org.spongepowered.api.event.cause.EventContextKeys;
 import org.spongepowered.api.event.cause.entity.dismount.DismountType;
 import org.spongepowered.api.event.cause.entity.dismount.DismountTypes;
 import org.spongepowered.api.event.cause.entity.teleport.TeleportTypes;
+import org.spongepowered.api.event.entity.IgniteEntityEvent;
 import org.spongepowered.api.event.entity.MoveEntityEvent;
 import org.spongepowered.api.service.user.UserStorageService;
 import org.spongepowered.api.text.Text;
@@ -105,6 +108,7 @@ import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.Redirect;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
 import org.spongepowered.common.SpongeImpl;
 import org.spongepowered.common.SpongeImplHooks;
 import org.spongepowered.common.data.manipulator.mutable.entity.SpongeGravityData;
@@ -129,6 +133,7 @@ import org.spongepowered.common.interfaces.data.IMixinCustomDataHolder;
 import org.spongepowered.common.interfaces.entity.IMixinEntity;
 import org.spongepowered.common.interfaces.entity.IMixinGriefer;
 import org.spongepowered.common.interfaces.network.IMixinNetHandlerPlayServer;
+import org.spongepowered.common.interfaces.world.IMixinWorld;
 import org.spongepowered.common.interfaces.world.IMixinWorldServer;
 import org.spongepowered.common.interfaces.world.gen.IMixinChunkProviderServer;
 import org.spongepowered.common.profile.SpongeProfileManager;
@@ -151,9 +156,6 @@ import javax.annotation.Nullable;
 @Implements(@Interface(iface = Entity.class, prefix = "entity$"))
 public abstract class MixinEntity implements IMixinEntity {
 
-    private static final String LAVA_DAMAGESOURCE_FIELD = "Lnet/minecraft/util/DamageSource;LAVA:Lnet/minecraft/util/DamageSource;";
-    private static final String ATTACK_ENTITY_FROM_METHOD = "Lnet/minecraft/entity/Entity;attackEntityFrom(Lnet/minecraft/util/DamageSource;F)Z";
-    private static final String FIRE_DAMAGESOURCE_FIELD = "Lnet/minecraft/util/DamageSource;IN_FIRE:Lnet/minecraft/util/DamageSource;";
     private static final String WORLD_SPAWN_PARTICLE = "Lnet/minecraft/world/World;spawnParticle(Lnet/minecraft/util/EnumParticleTypes;DDDDDD[I)V";
     private static final String RIDING_ENTITY_FIELD = "Lnet/minecraft/entity/Entity;ridingEntity:Lnet/minecraft/entity/Entity;";
     @SuppressWarnings("unused")
@@ -167,7 +169,6 @@ public abstract class MixinEntity implements IMixinEntity {
     private net.minecraft.entity.Entity teleportVehicle;
     private float origWidth;
     private float origHeight;
-    @Nullable private DamageSource originalLava;
     protected boolean isConstructing = true;
     @Nullable private Text displayName;
     private BlockState currentCollidingBlock;
@@ -1367,4 +1368,40 @@ public abstract class MixinEntity implements IMixinEntity {
     public void setInvulnerable(boolean value) {
         this.invulnerable = value;
     }
+
+    @Inject(method = "setFire",
+            at = @At(value = "FIELD", target = "Lnet/minecraft/entity/Entity;fire:I", opcode = Opcodes.PUTFIELD, ordinal = 0),
+            locals = LocalCapture.CAPTURE_FAILEXCEPTION,
+            cancellable = true)
+    private void onFire(int seconds, CallbackInfo ci, int i) {
+        if (((IMixinWorld) world).isFake() || !ShouldFire.IGNITE_ENTITY_EVENT) {
+            return;
+        }
+        if (this.fire < 1) {
+            try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
+
+                frame.pushCause(this.getLocation().getExtent());
+                IgniteEntityEvent event = SpongeEventFactory.
+                        createIgniteEntityEvent(frame.getCurrentCause(), i, i, this);
+
+                if (SpongeImpl.postEvent(event)) {
+                    ci.cancel();
+                }
+            }
+        }
+    }
+
+    @Redirect(method = "onStruckByLightning", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/Entity;attackEntityFrom(Lnet/minecraft/util/DamageSource;F)Z"))
+    private boolean onSpongeLightningDamage(net.minecraft.entity.Entity entity, DamageSource source, float damage, EntityLightningBolt lightningBolt) {
+        if (!this.world.isRemote) {
+            return entity.attackEntityFrom(source, damage);
+        }
+        try {
+            DamageSource.LIGHTNING_BOLT = new EntityDamageSource("lightningBolt", lightningBolt);
+            return entity.attackEntityFrom(DamageSource.LIGHTNING_BOLT, damage);
+        } finally {
+            DamageSource.LIGHTNING_BOLT = source;
+        }
+    }
+
 }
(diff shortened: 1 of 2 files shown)


parlough:
I'm assuming original means the ticks when the event was first thrown, so it won't change compared to the normal one which will when you set the ticks with the event. Event-gen-impl handles this.

parlough:
Without looking to much into this, I'm assuming you need to still handle lightning as the source for the event and you probably need a check to make sure you're only using the cause stack manager if you can. Just return early if `((IMixinWorld) world).isFake()`

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -1360,4 +1363,26 @@ public boolean shouldTick() {
...
+                }
+                if (this.originalInFire != null) {
+                    frame.pushCause(this.originalInFire);
+                }
+                frame.pushCause(this.getLocation().getExtent());
+                IgniteEntityEvent event = SpongeEventFactory.
+                        createIgniteEntityEvent(frame.getCurrentCause(), this.fire, this.fire, this);
+                SpongeImpl.postEvent(event);
+
+                if (event.isCancelled()) {
```
parlough: `if (SpongeImpl.postEvent(event))`

ImMorpheus:
Apparently, the event is not firing. 

During my tests I was injecting at HEAD, but I changed it to RETURN before commit because "this.fire" was not set at that point.

EDIT:

> I'm assuming you need to still handle lightning as the source for the event 

This is the cause when a skeleton is struck by a lightning:

```
Cause[Context=Context[], Stack={WorldServer{LevelName=world, DimensionId=0, DimensionType=minecraft:overworld}, EntityLightningBolt['unknown'/553, l='world', x=79.70, y=85.00, z=-11.30]}]
```

As of now there's no concept of "source", I tried 

```
PhaseData data = PhaseTracker.getInstance().getCurrentPhaseData();

data.context.getSource(Living.class).ifPresent(source ->
                        frame.addContext(EventContextKeys.IGNITER, source));
```

but this is only working for players (or other classes that extend Living), I can't add an EntityLightning as an IGNITER.

ryantheleach:
The source would likely be the world or weather or something, unless it was triggered by API or Command.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -1364,7 +1364,7 @@ public void setInvulnerable(boolean value) {
         this.invulnerable = value;
     }
 
-    @Inject(method = "setFire", at = @At(value = "RETURN"), cancellable = true)
+    @Inject(method = "setFire", at = @At(value = "HEAD"), cancellable = true)
```
ImMorpheus: Need help, setting this to RETURN doesn't fire the event at all.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -1360,4 +1363,39 @@ public boolean shouldTick() {
     public void setInvulnerable(boolean value) {
         this.invulnerable = value;
     }
+
+    @Inject(method = "setFire", at = @At(value = "HEAD"), cancellable = true)
```
ImMorpheus: Need help, setting this to RETURN doesn't fire the event at all.

Cybermaxke: At RETURN is `isBurning` probably already returning `true`.

You will need to implement the injection differently if you want to support the overridden fire seconds. A `ModifyVariable` should do the trick. Inject into the seconds argument and return the modified seconds from the event. `isBurning` will also work at this point.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -1360,4 +1363,39 @@ public boolean shouldTick() {
...
+                if (SpongeImpl.postEvent(event)) {
+                    ci.cancel();
+                }
+            }
+        }
+    }
+
+    private DamageSource originalLightning;
+
+    @Inject(method = "onStruckByLightning", at = @At(value = "HEAD"))
```
gabizou: Where is this coming from?

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
gabizou: You can even get around by using a `@Redirect` with a field setter. This way you can cancel the return by sending the current fire seconds, or increase it based on whatever event is thrown.

gabizou (changes requested):

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
ImMorpheus: This was created to add the DamageSource when the entity is set on fire by a lightning. There's a field called `originalLava` (set inside `setOnFireFromLava()`) for lava and a field called `originalInFire` (set inside `dealFireDamage`) for fire, but nothing for lightning.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
ImMorpheus: I was testing around and ended up with:
```
@Inject(method = "setFire",
            at = @At(value = "FIELD", target = "Lnet/minecraft/entity/Entity;fire:I", opcode = Opcodes.PUTFIELD),
            locals = LocalCapture.CAPTURE_FAILEXCEPTION,
            cancellable = true)
```
Everything is working, not sure what is better.

parlough:
I feel like this event can fire quite a lot, especially with mobs during the day, and don't feel like it is something commonly listened to, so consider using ShouldFire for this, placing the check at the beginning of the `onFire` mixin.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
gabizou: If it's coming from that, you need to be verifying where the modified lightning is being set. If it isn't being set, then this isn't doing anything special.

ryantheleach:
One feature I was looking for a while back (That I don't particularly care for now) Was knowing when a fire was extinguished on an entity. Maybe it's more applicable to a data event. At least with the ignite event I can add them to a list of entities to poll the state of.

Zidane:
@ImMorpheus 

Can you squash to one commit?

ImMorpheus:
Done

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -1360,4 +1367,52 @@ public boolean shouldTick() {
     public void setInvulnerable(boolean value) {
         this.invulnerable = value;
     }
+
+    @Inject(method = "setFire",
+            at = @At(value = "FIELD", target = "Lnet/minecraft/entity/Entity;fire:I", opcode = Opcodes.PUTFIELD),
+            locals = LocalCapture.CAPTURE_FAILEXCEPTION,
+            cancellable = true)
+    public void onFire(int seconds, CallbackInfo ci, int i) {
```
gabizou: public?

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
gabizou: Here's the thing, what I tried asking indirectly is that your'e not setting the `originalLightning` back to `DamageSource.LIGHTNING_BOLT`, nor are you making it a new `DamageSource`, so this injection is 100% useless without that logic existing to *reset* the damage sources.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -1360,4 +1367,52 @@ public boolean shouldTick() {
...
+                    ci.cancel();
+                }
+            }
+        }
+    }
+
+    private DamageSource originalLightning;
+
+    @Inject(method = "onStruckByLightning", at = @At(value = "HEAD"))
+    public void onStruck(EntityLightningBolt lightningBolt, CallbackInfo ci) {
```
gabizou: And this should be private.

gabizou (changes requested):

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -1360,4 +1368,63 @@ public boolean shouldTick() {
...
+
+                if (this.originalLava != null) {
+                    frame.addContext(EventContextKeys.DAMAGE_TYPE, DamageTypes.MAGMA);
+                    frame.pushCause(this.originalLava);
+                }
+                if (this.originalInFire != null) {
+                    frame.addContext(EventContextKeys.DAMAGE_TYPE, DamageTypes.FIRE);
+                    frame.pushCause(this.originalInFire);
+                }
+                if (this.originalLightning != null) {
```
gabizou: And technically, lightning is considered `PROJECTILE`, not `FIRE`.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -1360,4 +1368,63 @@ public boolean shouldTick() {
...
+            cancellable = true)
+    private void onFire(int seconds, CallbackInfo ci, int i) {
+        if (((IMixinWorld) world).isFake() || !ShouldFire.IGNITE_ENTITY_EVENT) {
+            return;
+        }
+        if (!this.isBurning()) {
+            try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
+
+                if (this.originalLava != null) {
+                    frame.addContext(EventContextKeys.DAMAGE_TYPE, DamageTypes.MAGMA);
```
gabizou: Lava blocks are fire, not magma (magma is specific to the magma block, not sure if we should change it).

gabizou (changes requested):
Need to verify whether the damage types need to be changed in API or not for lightning. I consider lightning to be a projectile still, while lava blocks are fire, not magma damage types.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -1360,4 +1368,63 @@ public boolean shouldTick() {
...
+    private DamageSource originalLightning;
+
+    @Inject(method = "onStruckByLightning", at = @At(value = "FIELD", target = LIGHTNING_DAMAGESOURCE_FIELD, opcode = Opcodes.GETSTATIC))
+    private void onPreStruck(EntityLightningBolt lightningBolt, CallbackInfo ci) {
+        if (!this.world.isRemote) {
+            this.originalLightning = DamageSource.LIGHTNING_BOLT;
+            AxisAlignedBB bb = this.getEntityBoundingBox().grow(-0.001D, -0.001D, -0.001D);
+            Location<World> location = DamageEventHandler.findFirstMatchingBlock((net.minecraft.entity.Entity) (Object) this, bb, block ->
+                    block.getBlock() == Blocks.FIRE);
+            DamageSource.LIGHTNING_BOLT = new MinecraftBlockDamageSource("lightningBolt", location).setFireDamage();
```
gabizou: Why a block and not the actual lightning bolt?

gabizou:
Ok, so, I just went and made some changes on how all the DamageSource re-assignments work, they should all be redirects now (they were originally implemented before we had redirects available in mixin ~oh god... the horror~ added them).

To get an idea, review [this commit](https://github.com/SpongePowered/SpongeCommon/commit/07a334c4574ab942d0cdb453d7ed685f76d8c10b).

Update to that style and I'll merge the PR!

ImMorpheus:
Oh, I see what you did. However take the `onSpongeRedirectForBlockDamageSource` method, the damagesource was assigned to this field https://github.com/SpongePowered/SpongeCommon/commit/07a334c4574ab942d0cdb453d7ed685f76d8c10b#diff-811f97d54d477fefea740d226d776a37L373 and I was using that field to establish the correct `EventContextKeys.DAMAGE_TYPE` and push the damagesource to the CauseStack https://github.com/SpongePowered/SpongeCommon/pull/1708/files#diff-811f97d54d477fefea740d226d776a37R1386.
Not sure how to do that now, am I missing something ?

gabizou:
You can still do that now, just push to the cause stack when you re-assign/create the new `DamageSource`. Though, I'm understanding that the `attackEntity` method already pushes the `DamageSource` to the cause stack (can verify [here in the massive overwrite](https://github.com/SpongePowered/SpongeCommon/blob/cf5fa8b4e5d02a5e69b4bafa63a3fe2318033c8d/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntityLivingBase.java#L364-L549)). So, ideally you shouldn't have to push those sorts of things, but I'd double check the events with a test plugin maybe (I don't think we've written one for the test).

Blayyke:
Any update on this? Would be incredibly useful to be able to use this event in my plugin

ImMorpheus:
Sorry,  I've been busy with other things. 

Squashed and rebased, however this needs testing, there are a few oddities I've experienced.

ImMorpheus:
Should be good. 
Not sure what to do with creative players: if you switch to creative mode and start swimming in a lava lake, the server spams the event (a lot). That's because it's calling `Entity#extinguish()` if you're in creative before setting you on fire again. This is not happening with invulnerable entity.

gabizou:
Had to make some changes to get the values back from plugins and setting them to the entity again. If you want to look at the commits shortly after to see how it actually works (with a very simple test plugin).

PR #1710, opened 2018-01-14
WIP Fix #1282

### LINKED ISSUE #1282: hasPlayedBefore "always" returning true, due to timestamps differing slightly.
From:

https://forums.spongepowered.org/t/hasplayedbefore-always-return-true/18311

```
 
 @Listener
 public void onLogin(ClientConnectionEvent.Join event) {
 	System.out.println(event.getTargetEntity().hasPlayedBefore());
 	System.out.println(event.getTargetEntity().firstPlayed());
 	System.out.println(event.getTargetEntity().lastPlayed());
 }
```

> Version: SpongeVanilla version 1.11.2-6.0.0-BETA-243
> 
> [16:18:56] [Server thread/INFO] [STDOUT]: [***.onLogin(***.java:32)]: true
> [16:18:56] [Server thread/INFO] [STDOUT]: [***.onLogin(***.java:33)]: SpongeValue{key=Key{Value:Value<Instant>, Query: FirstTimeJoined}, defaultValue=1970-01-01T00:00:00Z, actualValue=2017-04-04T08:18:**55.928**Z}
> [16:18:56] [Server thread/INFO] [STDOUT]: [***.onLogin(***.java:34)]: SpongeValue{key=Key{Value:Value<Instant>, Query: LastTimePlayed}, defaultValue=1970-01-01T00:00:00Z, actualValue=2017-04-04T08:18:**56.104**Z}
> 

I've filed it against Sponge Common until I've done further testing to see if it is unique to vanilla or also occurs on Sponge Forge.

ImMorpheus:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1740) | **SpongeCommon**

Fix https://github.com/SpongePowered/SpongeCommon/issues/1282

This was worse than I thought. This PR is currently WIP.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
@@ -847,7 +847,7 @@ public abstract class MixinEntityPlayerMP extends MixinEntityPlayer implements P
 
     @Override
     public Value<Instant> lastPlayed() {
-        return new SpongeValue<>(Keys.LAST_DATE_PLAYED, Instant.EPOCH, Instant.now());
+        return new SpongeValue<>(Keys.LAST_DATE_PLAYED, Instant.EPOCH, SpongePlayerDataHandler.getLastPlayed(this.getUniqueID()).get());
     }
 
     // TODO implement with contextual data


src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -847,7 +847,7 @@ public JoinData getJoinData() {
 
     @Override
     public Value<Instant> lastPlayed() {
-        return new SpongeValue<>(Keys.LAST_DATE_PLAYED, Instant.EPOCH, Instant.now());
+        return new SpongeValue<>(Keys.LAST_DATE_PLAYED, Instant.EPOCH, SpongePlayerDataHandler.getLastPlayed(this.getUniqueID()).get());
```
dualspiral, line 850: I don't think this is the way to fix it. `lastPlayed` should indicate when the player indeed last played, which in the case of an online player is _right now_. This also breaks the expectation of other plugins that might use this, so even though this isn't an API change, it's a behavioral change and we should not be doing that either without good reason.

For a simple patch, setting some flag on the player (or other) object that states a player is on its first connection would be better, which would then get passed up through the API.

From what @ryantheleach said on the linked issue (to which I agree)

> LastJoined should be when a player last connected (successfully)
> LastPlayed should be the current instant, if the player is currently online.

Thus, the better solution here would to be to add a `Keys.LAST_DATE_JOINED` key which gets populated with the last join date. That should then be used for comparisons.

dualspiral (changes requested):

ryantheleach:
I think we need to split this into 2 PR's API 7.1 which just fixes hasPlayedBefore, and API 8 which fixes all the semantic errors in the naming.

To list a few:

It's important that hasPlayedBefore works differently then comparing those Instants and works correctly, as it should only be done when the players data is going/has been created? Otherwise it's going to cause issue's with importing worlds / players / migrating from other places, incorrectly assigning kits.

LastPlayed and FirstJoined are less important to get right IMO as they are meta-data, so working for a best-case is ok.

SpongePlayerDataHandler has a getFirstJoined(UUID) and a getLastPlayed(UUID)

Yet the API uses those to expose FirstPlayed , LastPlayed and hasPlayedBefore.

LastJoined is useful to have as well, but I'm wondering if there is any benefit or not of adding a FirstJoined to represent their first join attempt or not.

I could also see it being useful to plugins that wish to reset the hasPlayedBefore state, in regards to resetting a 'new player experience' if they have played for less then x minutes, and have been away for longer then y days.


So Number 1 priority, is fixing hasPlayedBefore, and maybe exposing it as data itself, This should be done on API 7.1 (as it's a fix and not a behavior change)

Priority 2, should be adding LastJoined which could also be added on API 7.1 and adding a deprecation warning to LastPlayed warning that the behavior will change in API 8, and that current users should move to LastJoined

API: 8

Priority 3, would be to modify LastPlayed to return the current time if the player is online, or the lastPlayed if they are offline. and would need to be done on API8 as it's a behavior break.

Priority 4, would be to make all these datapoints optional except hasPlayedBefore.

Because in all cases they might not have ever joined, might not have ever played, and that they might be offline at the time the methods are called on EntityPlayerMP, or hell might be FakePlayers that have never 'played'


I believe this is a correct summary of the points covered by @dualspiral @ImMorpheus and myself on the #voice channel on Discord.

gabizou:
Superseded by https://github.com/SpongePowered/SpongeCommon/commit/802ea1d1d7a81f8cc37322f313f19935ab24ba73

Going to break the API to correctly fix this later.

PR #1712, opened 2018-01-14
Use the PotionID when creating the DataContainer (Close #1276)

### LINKED ISSUE #1276: PotionEffect Retrieval Bug
**Sponge API:** 6.0.0-20170323.180311-60
**Sponge Forge Build:** spongeforge-1.11.2-2227-6.0.0-BETA-2244
**Forge Build:** forge-1.11.2-13.20.0.2227-universal
**Java Version:** 1.8.0_121-b13

I've seem to have found a bug, which simon816 has confirmed on the forums, with the with `DataSerializable` in regards to `PotiotionEffect`s. 

**Here is the original post:**

I have the following code, which saves a player's potion effects:
```
public static void saveOriginalPotionEffects(List<PotionEffect> potionEffects, String arenaName, String userName) throws ObjectMappingException {

	int i = 1;
	
	for(PotionEffect potionEffect: potionEffects){
		
		UnversalConfigs.getConfig(ArenaConfig).getNode("Arena", arenaName, "Contestants", userName, "Potion Effects","Effect "+i).setValue(TypeToken.of(PotionEffect.class),potionEffect);
		
		UnversalConfigs.saveConfig(ArenaConfig);
		
		i++;
	}
}
```

I have no issue with the code above, but when I try to retrieve the data with the code below, I receive a `Could not translate DataSerializable of type: org.spongepowered.api.effect.potion.PotionEffect` error. This might be a programmer end error, but I coudn't get any further after troubleshooting the issue to line `PotionEffect potionEffect = node.getNode("Effect "+i).getValue(TypeToken.of(PotionEffect.class)); `Any help would be greatly appreciated. Thanks!

```
public static List<PotionEffect> fetchOriginalPotionEffects(String arenaName, String userName) throws ObjectMappingException{
	
	List<PotionEffect> potionEffects = Lists.newArrayList();
	
	CommentedConfigurationNode node = UnversalConfigs.getConfig(ArenaConfig).getNode("Arena", arenaName, "Contestants", userName, "Potion Effects");
	
	int NOPE = getNumOfPotionEffects(arenaName,userName);
	
	for(int i = 1; i<= NOPE;i++){
		
		PotionEffect potionEffect = node.getNode("Effect "+i).getValue(TypeToken.of(PotionEffect.class));
		
		potionEffects.add(potionEffect);

	}
	
	return potionEffects;
}

```
**simon816's bug replication code:**

```
try {
    PotionEffect effect = PotionEffect.of(PotionEffectTypes.JUMP_BOOST, 1, 10);
    SimpleCommentedConfigurationNode node = SimpleCommentedConfigurationNode.root();
    node.getNode("test").setValue(TypeToken.of(PotionEffect.class), effect);
    PotionEffect effect2 = node.getNode("test").getValue(TypeToken.of(PotionEffect.class));
} catch (ObjectMappingException e) {
    e.printStackTrace();
}
```

ImMorpheus:
When creating a `DataContainer` from a `PotionEffect`, we are using its name as the `DataQueries.POTION_TYPE`. However if you pass the container to the [PotionBuilder](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/effect/potion/SpongePotionBuilder.java#L73) it uses the `DataQueries.POTION_TYPE` to search for the `PotionEffect` in the registry (which requires the PotionEffect id).

Fix #1276

### DIFF
--- a/src/main/java/org/spongepowered/common/effect/potion/PotionEffectContentUpdater.java
+++ b/src/main/java/org/spongepowered/common/effect/potion/PotionEffectContentUpdater.java
@@ -1,0 +1,62 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.effect.potion;
+
+import com.google.common.collect.ImmutableMap;
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.data.DataView;
+import org.spongepowered.api.data.persistence.DataContentUpdater;
+import org.spongepowered.common.data.util.DataQueries;
+import org.spongepowered.common.data.util.DataVersions;
+import org.spongepowered.common.registry.type.effect.PotionEffectTypeRegistryModule;
+
+import java.util.stream.Collectors;
+
+public class PotionEffectContentUpdater implements DataContentUpdater {
+
+    private final static ImmutableMap<String, String> map = ImmutableMap.<String, String>builder().putAll(PotionEffectTypeRegistryModule.getInstance().getAll()
+            .stream().collect(Collectors.toMap(CatalogType::getName, CatalogType::getId))).build();
+
+    @Override
+    public int getInputVersion() {
+        return DataVersions.Potion.BROKEN_POTION_ID;
+    }
+
+    @Override
+    public int getOutputVersion() {
+        return DataVersions.Potion.POTION_V2;
+    }
+
+    @Override
+    public DataView update(DataView content) {
+        final String oldId = content.getString(DataQueries.POTION_TYPE).get();
+        final String newId = map.get(oldId);
+
+        content.remove(DataQueries.POTION_TYPE);
+        content.set(DataQueries.POTION_TYPE, newId);
+
+        return content;
+    }
+}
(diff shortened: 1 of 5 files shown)


ryantheleach:
I've marked this API8/needs updating because it is changing the contract of a DataContainer.

However, my guess is that since it's a bug fix, and it's only changing the serialization behavior, that the deserialization is currently correct so this doesn't round trip correctly?

With the assumption that it's a bug, I think it's safe for Milestone 7.1 to immediately fix the issue for people.

If it's even remotely possible that it's currently useful to someone in it's current state, I suggest bumping the content version number and making a migrator.

I suggest waiting for feedback though, but my suspicion this is safe to move to 7.1 as is.

parlough:
@ryantheleach I don't see a point in versioning as it is failing to deserialize already. I mean we could fix the current deserializer through some for loop using the name, which is probably the translation id to properly deserialize, but that's a pain, not guaranteed to work in all cases, and of course was failing already.

ryantheleach:
@Meronat good enough for me. This should be ok for merge then, pending review.

gabizou:
@Meronat @ryantheleach In my opinion, bumping the content version would resolve the issue while providing the upgrader to automatically fix the potion name -> id usage. This will maintain functionality in legacy data input without injecting a "fixer" in the deserializer. That's what I'd prefer to see happen here.

ImMorpheus:
Sorry for the delay. I missed the notification. Thanks to @Meronat for helping me out yesterday

parlough:
~qa

src/main/java/org/spongepowered/common/effect/potion/PotionEffectContentUpdater.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.stream.Collectors;
+
+public class PotionEffectContentUpdater implements DataContentUpdater {
+
+    private final static ImmutableMap<String, String> map = ImmutableMap.<String, String>builder().putAll(PotionEffectTypeRegistryModule.getInstance().getAll()
+            .stream().collect(Collectors.toMap(CatalogType::getName, CatalogType::getId))).build();
+
+    @Override
+    public int getInputVersion() {
+        return 1;
```
gabizou: Should make this a `DataVersions` constant so it's more understandable the differences between said versions.

src/main/java/org/spongepowered/common/effect/potion/PotionEffectContentUpdater.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public int getOutputVersion() {
+        return 2;
```
gabizou: Same here. Should likely be `DataVersions.Potions.CURRENT_VERSION`

src/main/java/org/spongepowered/common/effect/potion/SpongePotionBuilder.java
```diff
@@ -48,7 +48,7 @@
     private boolean showParticles;
 
     public SpongePotionBuilder() {
-        super(PotionEffect.class, 1);
+        super(PotionEffect.class, 2);
```
gabizou: And reference said constants.

gabizou (changes requested):
Minor organizational changes.

Zidane:
@gabizou 

Give me a final word on this one.

src/main/java/org/spongepowered/common/effect/potion/PotionEffectContentUpdater.java
```diff
@@ -0,0 +1,62 @@
...
+            .stream().collect(Collectors.toMap(CatalogType::getName, CatalogType::getId))).build();
+
+    @Override
+    public int getInputVersion() {
+        return DataVersions.Potion.BROKEN_POTION_ID;
+    }
+
+    @Override
+    public int getOutputVersion() {
+        return DataVersions.Potion.CURRENT_VERSION;
```
dualspiral: If we later create a version 3, won’t this break any really old data being migrated, as the migrator would then mislabel version 2 data as version 3?

This should just be version 2 or its own constant, separate from the `CURRENT_VERSION` constant to avoid this mistake later.

gabizou: Agreed, we can have two pointers, one being the constant `2` for using the proper id, and the "current version" that points to version `2`.

gabizou (approved):

gabizou:
@Zidane ready to merge. Looks good.

Zidane:
@ImMorpheus 

Squash to one commit and I can quickly merge this. Thanks!

ImMorpheus:
Done.

PR #1724, opened 2018-01-22
Move text implementation to the implementation

kashike:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1744) | **SpongeCommon** | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1971)


src/main/java/org/spongepowered/common/mixin/api/text/MixinText.java
```diff
@@ -39,13 +39,14 @@
...
+import org.spongepowered.common.text.action.ClickTextActionImpl;
+import org.spongepowered.common.text.action.HoverTextActionImpl;
 import org.spongepowered.common.text.format.SpongeTextColor;
+import org.spongepowered.common.text.impl.TextImpl;
 
 import java.util.Optional;
 
-@Mixin(value = Text.class, remap = false)
+@Mixin(value = TextImpl.class, remap = false)
 public abstract class MixinText implements IMixinText {
```
Cybermaxke: Can't the mixin code be moved to TextImpl?

src/main/java/org/spongepowered/common/mixin/api/text/MixinText.java
```diff
@@ -24,116 +24,68 @@
...
-    @Shadow @Final protected Optional<ClickAction<?>> clickAction;
-    @Shadow @Final protected Optional<HoverAction<?>> hoverAction;
-    @Shadow @Final protected Optional<ShiftClickAction<?>> shiftClickAction;
-
-    private ITextComponent component;
-    private String json;
-
-    protected TextComponentBase createComponent() {
-        throw new UnsupportedOperationException();
+public interface MixinText {
```
stephan-gh, line 37: One goal of moving Text implementation to SpongeCommon is to avoid having mixins for our own API classes, so I think the methods are at the wrong place here. This should be implemented using some kind of interface instead (even if it's default implemented in the API with those builders, like right now)

stephan-gh (changes requested):

src/main/java/org/spongepowered/common/mixin/api/text/MixinText.java
kashike: These only exist to return the constants rather than creating a new builder each time.

stephan-gh: There should be a way to provide these optimizations without mixin API classes, e.g. Lantern might also want to do this.

src/main/java/org/spongepowered/common/text/impl/LiteralTextImpl.java
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    @Override
+    public Builder toBuilder() {
+        return new Builder(this);
+    }
+
+    @Override
+    protected ITextComponent createComponent() {
+        return new TextComponentString(this.content);
```
liach, line 85: What's the difference between api text and mcp textcomponent?

src/main/java/org/spongepowered/common/text/impl/TextImpl.java
```diff
@@ -0,0 +1,535 @@
...
+
+import javax.annotation.Nullable;
+
+public abstract class TextImpl implements Text {
+
+    /**
+     * A {@link Comparator} for texts that compares the plain text of two text
+     * instances.
+     */
+    public static Comparator<Text> PLAIN_COMPARATOR = Comparator.comparing(Text::toPlain);
```
liach, line 69: Why not final?

src/main/java/org/spongepowered/common/mixin/api/text/MixinText.java
kashike: Feel free to suggest.

gabizou: @kashike As I said to use the builders, the builders themselves are fast enough if they default to returning the constant versions. No need for mixing.

liach:
:tada:

PR #1728, opened 2018-01-24
Initial work on translations

### LINKED ISSUE #659: Some block translations are off
Performing `getTranslation()` on a blocktype usually produces the correct translation (ie. if I use it on grass, I get 'Grass Block'). However, if use it on something such as sponge, I get 'tile.sponge.name'

Essentially, running the following code
https://gist.github.com/12AwsomeMan34/8023b50333d0be8707862ed51a70e72d

will produce this output
https://gist.github.com/12AwsomeMan34/753a8e191958776c5cbdd275115edb57

Note the ones such as 'tile.stone.name', 'tile.banner.name', or 'tile.sponge.name'. It would be nice if these were corrected for the usage of `getTranslation()`.

ImMorpheus:
This should fix https://github.com/SpongePowered/SpongeCommon/issues/659

- [x] tile.doorOak.name
- [x] tile.cobbleWall.name
- [x] tile.stone.name
- [x] tile.diode.name
- [x] tile.doorBirch.name
- [x] tile.comparator.name
- [x] tile.banner.name
- [x] tile.diode.name
- [x] tile.banner.name
- [x] tile.sapling.name
- [x] tile.doorDarkOak.name
- [x] tile.comparator.name
- [x] tile.prismarine.name
- [x] tile.doorJungle.name
- [x] tile.doorAcacia.name
- [x] tile.sponge.name
- [x] tile.doorSpruce.name
- [x] tile.stoneSlab2.name
- [x] tile.concretePowder.name
- [x] tile.null.name | (BlockPistonMoving) not translatable
- [x] tile.bed.name
- [x] tile.concrete.name
- [x] tile.brewingStand.name
- [x] tile.null.name | (BlockEndGateway) not translatable
- [x] tile.skull.name
- [x] tile.null.name | (BlockEndGateway) not translatable
- [x] tile.stoneSlab2.name
- [x] tile.flowerPot.name

Should also fix part of https://github.com/SpongePowered/SpongeCommon/issues/739
Broken translations:

- Handtype
- CraftingRecipe 
- Statistic 
- PopulatorType
- EntityType
- PotionEffectType
EDIT: ignore these, they require SpongeAPI changes. Another PR will follow when this get merged


ImMorpheus:
Regarding some block translations:

Every `(Mixin)Block` implements `BlockType` which implements `Translatable`, however not every `Block` (I'm referring to mojang code) is translatable (has a translation inside the .lang file).

Example:
```
[STDOUT]: ========
[STDOUT]: Mojang translation: tile.null.name
[STDOUT]: Translation: tile.null.name
[STDOUT]: Class: class net.minecraft.block.BlockEndPortal
[STDOUT]: ========
[STDOUT]: Mojang translation: tile.bed.name
[STDOUT]: Translation: tile.bed.name
[STDOUT]: Class: class net.minecraft.block.BlockBed
[STDOUT]: ========
[STDOUT]: Mojang translation: tile.concrete.name
[STDOUT]: Translation: tile.concrete.name
[STDOUT]: Class: class net.minecraft.block.BlockColored
[STDOUT]: ========
[STDOUT]: Mojang translation: tile.null.name
[STDOUT]: Translation: tile.null.name
[STDOUT]: Class: class net.minecraft.block.BlockPistonMoving
[STDOUT]: ========
[STDOUT]: Mojang translation: tile.null.name
[STDOUT]: Translation: tile.null.name
[STDOUT]: Class: class net.minecraft.block.BlockEndGateway
[STDOUT]: ========
```

How should I treat these ?

src/main/java/org/spongepowered/common/mixin/core/block/MixinBlockBanner.java
```diff
@@ -72,4 +74,9 @@ public boolean supports(Class<? extends ImmutableDataManipulator<?, ?>> immutabl
     private ImmutableAttachedData getIsAttachedFor(IBlockState blockState) {
         return ImmutableDataCachingUtil.getManipulator(ImmutableSpongeAttachedData.class, blockState.getBlock() instanceof BlockBannerHanging);
     }
+
+    @Override
+    public Translation getTranslation() {
+        return new SpongeTranslation("item.banner.white.name");
```
gabizou, line 80: Do we need the translations to be re-created every time?

ImMorpheus: Are they worth caching ?
Currently there is no "standard way" of doing the translation. Half of the blocks cache the translation (ex. https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/data/types/MixinBlockPrismarineEnumType.java )
Either way, I will probably fix all of them to be consistent in the next PR. (so, input wanted)

ST-DDT: AFAICT either the translations are not used at all or used multiple times (because a plugin actually wants them)

IMO lazy caching is a good way that does not consume that much memory.

parlough:
Due to not all translations being present and the fact the method does not return an optional. Can we fall back to their id?

ImMorpheus:
That's the current behavior.

Meronat:
~qa

ImMorpheus:
Status ? I need this for the catalog PR.

gabizou (approved):
While this works and functions as expected, a cleanup would be necessary to lazy load the translations as fields. Doesn't cost us much in memory and useful to keep re-using the same object over time.

kashike (approved):

PR #1731, opened 2018-01-27
Implement keepinventory on DestructEntityEvent.Death

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1750) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1731)

### DIFF
--- a/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
@@ -642,7 +642,7 @@ public class SpongeCommonEventFactory {
         }
     }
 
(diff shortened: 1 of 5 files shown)


ryantheleach:
I was helping Faith test this with TinkersConstruct installed.

Seems to work as intended, and allows plugin to keep items on death, and have them persist over re-spawn.

Only known bug was GameRule keepInventory=true event.setKeepInventory=false will forcefully drop soulbound items on tinkers.

Changing this requires changes to the eventFactory and the modEventManager to let deaths use the lifecycle, so that the event can fire before mods rather then after using the before mods annotation or something? was discussed between dualspire and faith.

ryantheleach (approved):

PR #1732, opened 2018-01-27
Fixes #1719

### LINKED ISSUE #1719: Bed repawning doesn't work correctly
**I am currently running**

- SpongeForge version: all latest versions
  - 1.12.2-2555-8.0.0-BETA-2841
  - 1.12.2-2555-7.1.0-BETA-2842
  - 1.11.2-2282-7.0.0-BETA-2376
  - 1.11.2-2476-6.1.0-BETA-2792
- SpongeVanilla version:
  - all versions equivalent to the above
- Java version: 8
- Operating System: Windows/Linux

- Plugins/Mods:
None

**Issue Description**

Currently if a player dies and repawns at their stored spawn point (set via sleeping on a bed), it gets reset to the server spawn after respawning at their bed once. I've checked the playerdata files and their spawn point coordinates are simply gone after respawning.

It consistently does it every time when the player dies in another dimension or returns home via End portal. I can't recall if it's ever happened after dying in the overworld. So, easy way to reproduce:
1. Set spawn point via bed
2. Navigate to another dimension (either Nether or The End)
3. Die via `/kill` (or take the End Portal back to the overworld)
4. Die via `/kill` again and the player will incorrectly respawn at the server spawn.

XakepSDK:
Fixes https://github.com/SpongePowered/SpongeCommon/issues/1719

### DIFF
--- a/src/main/java/org/spongepowered/common/entity/EntityUtil.java
+++ b/src/main/java/org/spongepowered/common/entity/EntityUtil.java
@@ -699,13 +699,12 @@ public final class EntityUtil {
 
             // First revision: Xakep_SDK, 28.01.18. MC Version: 1.12.2, API version: 7.1.0
             // START REVISION
-            // This calls
-            // MixinEntityPlayer#onSetSpawnPoint(BlockPos pos, boolean forced, CallbackInfo ci);
+            // This calls SpongeVanilla's (i don't know, hot this handled in SpongeForge)
+            /** @see org.spongepowered.server.mixin.core.entity.player.MixinEntityPlayer#onSetSpawnPoint(BlockPos pos, boolean forced, CallbackInfo ci); */
             // so bedPos will be set for target world.
             // If we don't do this, MixinPlayerList#recreatePlayerEntity(EntityPlayerMP playerIn, int targetDimension, boolean conqueredEnd)
             // will get wrong bedPos when calls playerIn.getBedLocation()
-            // I'm not removing this, because this can be a part of bigger kludge-system.
-            // I don't know, why we're doing this. Actual spawn point
+            // I'm not removing code inside revision block, because this can be a part of bigger kludge-system.
             int prevDim = playerIn.dimension;
             playerIn.dimension = targetDimensionId;
             playerIn.setSpawnPoint(bedPos, forceBedSpawn);
--- a/src/main/java/org/spongepowered/common/entity/EntityUtil.java
+++ b/src/main/java/org/spongepowered/common/entity/EntityUtil.java
@@ -696,11 +696,21 @@ public final class EntityUtil {
                 // Vanilla behaviour - Delete the known bed location if invalid
                 bedPos = null; // null = remove location
             }
-            // Set the new bed location for the new dimension
-            int prevDim = playerIn.dimension; // Temporarily for setSpawnPoint
+
+            // First revision: Xakep_SDK, 28.01.18. MC Version: 1.12.2, API version: 7.1.0
+            // START REVISION
+            // This calls
+            // MixinEntityPlayer#onSetSpawnPoint(BlockPos pos, boolean forced, CallbackInfo ci);
+            // so bedPos will be set for target world.
+            // If we don't do this, MixinPlayerList#recreatePlayerEntity(EntityPlayerMP playerIn, int targetDimension, boolean conqueredEnd)
+            // will get wrong bedPos when calls playerIn.getBedLocation()
+            // I'm not removing this, because this can be a part of bigger kludge-system.
+            // I don't know, why we're doing this. Actual spawn point
+            int prevDim = playerIn.dimension;
             playerIn.dimension = targetDimensionId;
             playerIn.setSpawnPoint(bedPos, forceBedSpawn);
             playerIn.dimension = prevDim;
+            // END REVISION
         }
         return new Location<>((World) targetWorld, targetSpawnVec);
     }
--- a/src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
@@ -529,9 +529,31 @@ public abstract class MixinPlayerList implements IMixinPlayerList {
         // Sponge - Vanilla does this before recreating the player entity. However, we need to determine the bed location
         // before respawning the player, so we know what dimension to spawn them into. This means that the bed location must be copied
         // over to the new player
+        // [Extended, Revised, 28.01.2018, MC 1.12.2, API 7.1.0] Fix for: https://github.com/SpongePowered/SpongeCommon/issues/1719
+        // First revision: Xakep_SDK, 28.01.18. MC Version: 1.12.2, API version: 7.1.0
+        // START REVISION
+        // This gets target world bed spawn location
+        // and saves original player dimension for RespawnPlayerEvent event.
+        // We're getting target world because of:
+        // 1. When we're calling EntityUtil#getPlayerRespawnLocation(EntityPlayerMP playerIn, @Nullable WorldServer targetWorld);
+        // at the end of the method we have if(bedPos != null) ...
+        // blockPos - actual bed location for target world, that we got here.
+        // At the end of the if we can see:
+        // 1. Change player dim to target world dim
+        // 2. Change player bedPos to actual one (i don't know, why this happens, maybe a bigger kludge-system)
+        // 3. Reset player dim to old one
+        // This changes bedPos for TARGET world
+        // My changes: i'm setting players dim to target one,
+        // then we can get right bed position(if present)
+        // and set it to new players bed position
+        int tmp = playerIn.dimension;
+        playerIn.dimension = targetDimension;
         if (playerIn.getBedLocation() != null) {
             newPlayer.setSpawnPoint(playerIn.getBedLocation(), playerIn.isSpawnForced());
         }
+        // here i'm returning back player dim for RespawnPlayerEvent
+        playerIn.dimension = tmp;
+        // END REVISION
 
         for (String s : playerIn.getTags()) {
             newPlayer.addTag(s);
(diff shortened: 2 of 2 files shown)


XakepSDK:
@bloodmc please review this

Faithcaio:
Sadly not that easy a fix.
Too tired to test but this should work:
https://github.com/SpongePowered/SpongeCommon/tree/fix/bedspawn

src/main/java/org/spongepowered/common/entity/EntityUtil.java
```diff
@@ -696,11 +696,21 @@ public static Profession validateProfession(int professionId) {
                 // Vanilla behaviour - Delete the known bed location if invalid
                 bedPos = null; // null = remove location
             }
-            // Set the new bed location for the new dimension
-            int prevDim = playerIn.dimension; // Temporarily for setSpawnPoint
+
+            // First revision: Xakep_SDK, 28.01.18. MC Version: 1.12.2, API version: 7.1.0
```
gabizou, line 700: Comment formatting is somewhat awkward. I'd prefer to see the date spelled out like Overwrites with the MC version and API version being in separate lines for readability.

src/main/java/org/spongepowered/common/entity/EntityUtil.java
```diff
@@ -696,11 +696,21 @@ public static Profession validateProfession(int professionId) {
...
+            // START REVISION
+            // This calls
+            // MixinEntityPlayer#onSetSpawnPoint(BlockPos pos, boolean forced, CallbackInfo ci);
```
gabizou: Where is this being called?

src/main/java/org/spongepowered/common/entity/EntityUtil.java
```diff
@@ -696,11 +696,21 @@ public static Profession validateProfession(int professionId) {
...
+            // so bedPos will be set for target world.
+            // If we don't do this, MixinPlayerList#recreatePlayerEntity(EntityPlayerMP playerIn, int targetDimension, boolean conqueredEnd)
```
gabizou: If anything, you should be able to java doc link to the method instead of a comment link, sometimes these get lost in time.

src/main/java/org/spongepowered/common/entity/EntityUtil.java
```diff
@@ -696,11 +696,21 @@ public static Profession validateProfession(int professionId) {
...
+            // will get wrong bedPos when calls playerIn.getBedLocation()
+            // I'm not removing this, because this can be a part of bigger kludge-system.
```
gabizou: you're not removing what exactly?

src/main/java/org/spongepowered/common/entity/EntityUtil.java
```diff
@@ -696,11 +696,21 @@ public static Profession validateProfession(int professionId) {
...
+            // I don't know, why we're doing this. Actual spawn point
```
gabizou: I don't know what this line adds for commenting.

src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
```diff
@@ -529,9 +529,31 @@ public EntityPlayerMP recreatePlayerEntity(EntityPlayerMP playerIn, int targetDi
         // Sponge - Vanilla does this before recreating the player entity. However, we need to determine the bed location
         // before respawning the player, so we know what dimension to spawn them into. This means that the bed location must be copied
         // over to the new player
+        // [Extended, Revised, 28.01.2018, MC 1.12.2, API 7.1.0] Fix for: https://github.com/SpongePowered/SpongeCommon/issues/1719
```
gabizou, line 533: Likewise on formatting above.

src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
```diff
@@ -529,9 +529,31 @@ public EntityPlayerMP recreatePlayerEntity(EntityPlayerMP playerIn, int targetDi
...
+        // First revision: Xakep_SDK, 28.01.18. MC Version: 1.12.2, API version: 7.1.0
+        // START REVISION
+        // This gets target world bed spawn location
+        // and saves original player dimension for RespawnPlayerEvent event.
+        // We're getting target world because of:
+        // 1. When we're calling EntityUtil#getPlayerRespawnLocation(EntityPlayerMP playerIn, @Nullable WorldServer targetWorld);
+        // at the end of the method we have if(bedPos != null) ...
+        // blockPos - actual bed location for target world, that we got here.
```
gabizou: This point doesn't make much sense. I'm not sure where the `if (bedPos != null)` comes from, unless you're referencing the next couple of lines.

src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
```diff
@@ -529,9 +529,31 @@ public EntityPlayerMP recreatePlayerEntity(EntityPlayerMP playerIn, int targetDi
...
+        // At the end of the if we can see:
+        // 1. Change player dim to target world dim
```
gabizou: Do we need to change the player dim or are we seeing if we need to change the player dim?

src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
```diff
@@ -529,9 +529,31 @@ public EntityPlayerMP recreatePlayerEntity(EntityPlayerMP playerIn, int targetDi
...
+        // 2. Change player bedPos to actual one (i don't know, why this happens, maybe a bigger kludge-system)
+        // 3. Reset player dim to old one
+        // This changes bedPos for TARGET world
+        // My changes: i'm setting players dim to target one,
```
gabizou: Maybe comment the lines themselves where you're making said changes, having paragraphs above lines of code doesn't help understand *what* lines you're changing (yes I can see the changes in this git commit, but for maintainability standpoints, this doesn't help readability).

src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
```diff
@@ -529,9 +529,31 @@ public EntityPlayerMP recreatePlayerEntity(EntityPlayerMP playerIn, int targetDi
...
+        // then we can get right bed position(if present)
+        // and set it to new players bed position
+        int tmp = playerIn.dimension;
+        playerIn.dimension = targetDimension;
         if (playerIn.getBedLocation() != null) {
             newPlayer.setSpawnPoint(playerIn.getBedLocation(), playerIn.isSpawnForced());
         }
+        // here i'm returning back player dim for RespawnPlayerEvent
```
gabizou: Eliminate ownership, just say `// Set player dim back from the RespawnPlayerEvent`

gabizou (changes requested):
Some readability changes and maintainability commenting changes.

src/main/java/org/spongepowered/common/entity/EntityUtil.java
XakepSDK: This allow's revision stacking:
First revision: Who, date, mc ver., api ver.
Second revision: Who, date, mc ver., api ver.
Third revision: Who, date, mc ver., api ver.
So we can easily search commits by dates.

src/main/java/org/spongepowered/common/entity/EntityUtil.java
XakepSDK: Mixin magic!
https://github.com/SpongePowered/SpongeVanilla/blob/bleeding/src/main/java/org/spongepowered/server/mixin/core/entity/player/MixinEntityPlayer.java#L107

src/main/java/org/spongepowered/common/entity/EntityUtil.java
XakepSDK: Fixed in next commit

src/main/java/org/spongepowered/common/entity/EntityUtil.java
XakepSDK: Fixed in next commit

src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
XakepSDK: This refers to code inside revision block in EntityUtil

src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
XakepSDK: This refers to code inside revision block in EntityUtil

src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
XakepSDK: Fixed in next commit

src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java
XakepSDK: Fixed in next commit

XakepSDK:
@Faithcaio why do you change dim 0 bed spawn loc?

Faithcaio:
I was not done testing yet but it looked like it the vanilla overworld (dim0) bedspawn went missing when a death occured in an other overworld. (with Forge)

XakepSDK:
I have tested everything in vanilla.(2 hours of debugging)
My fix should work fine for forge and vanilla. It's unnecessary to copy dim0 location, but there is hidden bug: your and mine code just copies target dim bedspawn, but there can be many bedspawns. Every dim can have specified bedspawn. These bedspawns will not be copied.
I'll make full fix today, but bit later.

Faithcaio:
I did not test vanilla. But Forge was not working with your changes.

bloodmc:
This has been fixed, check https://github.com/SpongePowered/SpongeCommon/issues/1719

PR #1737, opened 2018-01-29
Use ClassLoader from object we got to avoid NoClassDefFoundErrors

vorburger:
When trying to register a event method from a class with a ClassLoader
different from Sponge's, when we try to define the class, it throws a
NoClassDefFoundError due to the class being unaccessible from the
current ClassLoader.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
@@ -32,6 +32,7 @@ import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
 import com.google.common.reflect.TypeToken;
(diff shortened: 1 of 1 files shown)


vorburger:
This is, of course, the carry over of https://github.com/SpongePowered/SpongeCommon/pull/1090 from @Maxqia ... I've just resolved merge conflict, and undone the the *s ... I have an interest in this change for https://github.com/vorburger/ch.vorburger.minecraft.osgi and would love to see this finally (after 1 year) go in...

phit:
+qa

parlough:
@vorburger Not sure if this still applies, but Deamon mentioned this a while ago:

"It needs to handled unregistering listeners so we're not leaking the classloader"

vorburger:
@Meronat are you referring to @Deamon5550 point re. "This should be a loading cache with weak keys." in #1090 ? @Maxqia and @Deamon5550 had a conversation about this there ending in @Maxqia "That wouldn't change that because we'd still have a store of the object (and therefore it's classloader) in the event handler." and then "@Deamon5550 approved these changes on 18 Jan 2017" ... so I thought that was OK.

RedNesto (approved):

parlough:
Thanks for the PR, a variation of this has now been included in stable-7 with - 9881012e65877ec728d2994d5baa9efb398a19a8

Issue #1556, opened 2017-10-03
1.12.2 Context: XP Orbs from mob death/player death drop have no  Spawn_Type context key


theboomer:
When players die and their orbs are scattered about, or when they kill a mob and orbs are dropped to pick up, the spawn event for the orbs has no context entry , no spawn_type EventContextKey as all spawn event should.
Does not impact XP from enchantment bottles.

------
SpongeVanilla-1.12.2-beta-343
SpongeForge=1.12.2-bets-2684  Note: On my forge server version, I can not SEE any of the entities at all; I can splash-potion XP bottles and see the effects, sounds, and spawn-events in log, but not orbs.  I can summon mobs and hear them, be hurt by them, but not see them, even disconnectin/reconnecting no.  I do not have this effect on the SV version -- yet.  I will test this later today more indepth to see if I am crazy or not before reporting, but if someone can just second-glance it ...

Environment: Zero mods, zero plugins except one logging SpawnEntityEvents with missing key.

ImMorpheus:
PR merged. This can be closed.

phit:
~actually leaving this open until it's on the stable branch~

Issue #920, opened 2016-09-04
[1.12.2] Several incorrect SpawnTypes for various entity-spawn events 


theboomer:
First, thank you for fixing the few examples that were listed in my previous post as promptly as was done after the push earlier. #916
I have tested _nearly every_ spawn-event scenario possible I can imagine and discovered many are incorrect results or not assigned, and some that require a double-check to see if they are consistent with their result planned definition because I am not sure.

System: Forge-2073 + SpongeForge-2073---1.10.2-Beta-1698
No mods or plugins, vanilla clients only;  One plugin to log spawnevent information only.

```

Spawning System                             Reports As          Should be
--------------------------------------------------------------------------
Player places armor stand                    SpawnEgg            Placement
Player places item frame                     SpawnEgg            Placement
Player places pictureframe                   SpawnEgg            Placement
Player placing any minecart on track         SpawnEgg            Placement
Experience orbs from killing mobs            Custom              Experience
Experience from catching fish                Placement           Experience
Experience orbs from mining ores(coal,       Placement           Experience      
  lapis, redstone, diamond, emerald, quartz)

```

Note that XPOrbs spawned from breeding-births and xp-bottle explosions do return "**Experience**", but not from these above scenarios.  Also, we were UNABLE to get furnaces to generate XP at all in testing, something I am going to investigate further now to see if it was a human-tester error (not likely, replicating a control test) or sponge, or forge, or vanilla, so I do not know what SMELTING XP orb spawning would be returning, and will report as a separate issue pending information learned.

```
For the following, I expect/suspect that they should be throwing "Projectile" 
- Throwing a bottle of enchanting  ---- currently returns "Placement"
- Throwing a splash/linger potion ----- currently returns "Placement"
- Throwing a snowball ----------------- currently returns "Placement"

```

Placing a block of sand / gravel / anvil etc   -- currently throws **BlockSpawning**, but there is a **FallingBlock** type unused.   This may be a matter of overplanning the return-list originally, before refining the result - Is **BlockSpawning** intended for this, or was **FallingBlock** intended, or does **FallingBlock** have a different intent as a type than this?   The fact that the **BlockSpawning** contains a falling-block entity in the event makes it unambiguous either way, so this is just a sanity-check against the intent being asked of for this scenario.

[...]

PR #1740, opened 2018-01-30
Fix xpOrbs still dropping after cancelling the event

ImMorpheus:
While working on https://github.com/SpongePowered/SpongeCommon/issues/920, @RedNesto found out that Experience Orbs are still dropping even if SpawnEntityEvent is cancelled.
 
```
     @Listener
     public void onSpawn(SpawnEntityEvent event) {
         event.setCancelled(true);
     }
```
Apparently I fixed it while fixing 920.

While processing EntitySupplier in the PhaseContext, xpOrbs were not handled and processed as `normalPlacement` (thus with the default contextKey (dropped items) set at line 126).
This adds a case for the orbs and set the key to `experience`.

Note: this doesn't fix everything on 920. Just the incorrect spawntype on "Experience orbs from mining ores"

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/phase/packet/InteractionPacketState.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/phase/packet/InteractionPacketState.java
@@ -26,6 +26,7 @@ package org.spongepowered.common.event.tracking.phase.packet;
 
 import com.flowpowered.math.vector.Vector3d;
 import net.minecraft.entity.item.EntityItem;
+import net.minecraft.entity.item.EntityXPOrb;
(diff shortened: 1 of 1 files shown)


ImMorpheus:
Should fix: https://github.com/SpongePowered/SpongeCommon/issues/1556

(To be precise, that is no longer an issue: xporbs already have a `EventContextKeys.SPAWN_TYPE`, just a wrong one (`DROPPED_ITEM`). This is setting that to `EXPERIENCE`.)

gabizou (approved):

RedNesto (approved):

kashike (approved):

Faithcaio:
stable-7 too?

Issue #1523, opened 2017-09-11
CauseTracker Issues


Faithcaio:
This is a list of cause tracker issues I noticed:

- [x] MoveEntityEvent.Teleport 
    - Context TeleportType is missing sometimes (enderpearl/chorus fruit).
   - TP from End -> World is missing causes.
   - Enderpearl (no player in cause)

- [ ] SpongeCommonFactory#callChangeBlockEventModifyLiquidBreak wrong context. Should be LIQUID_BREAK or FLOW not LIQUID_MIX.

- [ ] Leaf Decay: LocatableBlock is in Cause twice. (also SPAWN_TYPE context on ChangeBlockEvent seems weird)

- [x] indirect BlockBreak/Place (e.g. Sand falling by breaking block beneath it)
   - player is not in cause nor in context as notifier or smth. else. (owner does not count)

... and more?

lesbleu:
I have already started to report CauseTracker bugs in this [issue](https://github.com/SpongePowered/SpongeCommon/issues/1521).

ryantheleach:
Leaf Decay is leaking.

gabizou:
@phit could you test some of these?

ImMorpheus:
Tested on SpongeForge bleeding branch.
I was able to reproduce:
indirect BlockBreak/Place (player not in the cause) 
LocatableBlock is in Cause twice in Leaf Decay
SpongeCommonFactory#callChangeBlockEventModifyLiquidBreak wrong context.
@RedNesto is doing the MoveEntityEvent.Teleport

phit:
https://github.com/SpongePowered/SpongeCommon/pull/1742 & https://github.com/SpongePowered/SpongeCommon/pull/1775 were merged

PR #1741, opened 2018-01-30
Add enderpearl thrower to cause and TeleportType to context

RedNesto:
All of that for the MoveEntityEvent.Teleport.
Corrects an issue of #1523 
The thrower is placed at the root of the cause, I could not figure out how to put the enderpearl at the root right now.

First time I played with the CauseStackManager, so I may have made mistakes.


ryantheleach:
Should player be root for an enderpearl? I would have expected the enderpearl.

ImMorpheus:
I'd move the try and the addContext to `handleDisplaceEntityTeleportEvent` (since the event is thrown there).

src/main/java/org/spongepowered/common/entity/EntityUtil.java
```diff
@@ -317,6 +317,8 @@ public static boolean isEntityDead(net.minecraft.entity.Entity entity) {
         final IPhaseState state = peek.state;
         final PhaseContext<?> context = peek.context;
 
+        Sponge.getCauseStackManager().pushCause(entityIn);
```
gabizou: It doesn't seem like you're popping the cause after pushing. You need to ensure that the cause stack is stack neutral after the method is completed, so either popping the cause, or wrapping with a try with resources on a stack frame.

gabizou (changes requested):

src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityEnderPearl.java
```diff
@@ -55,9 +59,13 @@ private boolean onEnderPearlImpact(EntityPlayerMP player) {
             return true;
         }
 
-        MoveEntityEvent.Teleport event = EntityUtil.handleDisplaceEntityTeleportEvent(player, this.getLocation());
-        if (event.isCancelled()) {
-            return true;
+        try(CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
```
ImMorpheus: Spacing

ImMorpheus (changes requested):

src/main/java/org/spongepowered/common/entity/EntityUtil.java
RedNesto: Where should I pop the cause ?
I don't know if I should really pop the cause since the method is called inside a try-with-resources (in fact, I'm not event sure I should do [that](https://github.com/SpongePowered/SpongeCommon/pull/1741/files/5d9823590665615f835869725b3aa037036e7478#diff-8455601503527079e2c6f540298a089eR62))

RedNesto:
I found out that the way I am adding the player to the cause adds the player another time in the Event for the /teleport command, and I don't know how to handle this.

phit:
~wip

src/main/java/org/spongepowered/common/entity/EntityUtil.java
```diff
@@ -321,6 +321,7 @@ public static boolean isEntityDead(net.minecraft.entity.Entity entity) {
 
         MoveEntityEvent.Teleport event = SpongeEventFactory.createMoveEntityEventTeleport(Sponge.getCauseStackManager().getCurrentCause(), fromTransform, toTransform, (org.spongepowered.api.entity.Entity) entityIn);
         SpongeImpl.postEvent(event);
+        Sponge.getCauseStackManager().popCause();
```
ryantheleach: If you can, you should use 1 cause stack frame instead of a push/pop

Basically, using a cause stack frame in a try catch is using the https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html to push and pop a Cause Stack Frame (similar to the Call Stack Frame)  https://stackoverflow.com/questions/10057443/explain-the-concept-of-a-stack-frame-in-a-nutshell

Also we much prefer the try style compared to manual push / pop's due to the possibility for error or cause stack corruption.

src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityEnderPearl.java
```diff
@@ -55,9 +59,14 @@ private boolean onEnderPearlImpact(EntityPlayerMP player) {
...
         }
 
-        MoveEntityEvent.Teleport event = EntityUtil.handleDisplaceEntityTeleportEvent(player, this.getLocation());
-        if (event.isCancelled()) {
-            return true;
+        try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
+            frame.addContext(EventContextKeys.TELEPORT_TYPE, TeleportTypes.ENTITY_TELEPORT);
+            frame.addContext(EventContextKeys.THROWER, (Player) player).getCurrentCause();
+
+            MoveEntityEvent.Teleport event = EntityUtil.handleDisplaceEntityTeleportEvent(player, this.getLocation());
```
ryantheleach, line 66: Cancelling the displace here, will also cancel the damage when it hits another entity. Not sure if this is intended. ( i know you didn't code it this way, and it was already like this though)

RedNesto:
Argh, seems I failed :/

PR #1745, opened 2018-01-31
Fix a Phase IllegalStateException during a teleportation

### LINKED ISSUE #1744: Phase exception during MoveEntityEvent.Teleport if closing a player's inventory while holding item in cursor
If during a `MoveEntityEvent.Teleport` event, a player is holding an item on their cursor and `player.closeInventory()` is called, a phase exception is thrown.

**Sample code to reproduce issue:**
```java
@Listener
public void onTest(MessageChannelEvent.Chat event, @Root Player player) {
	// When player writes "!test" in chat
	if (!event.getRawMessage().toPlain().equalsIgnoreCase("!test")) return;

	Inventory inventory = Inventory.builder()
			.of(InventoryArchetypes.CHEST)
			.build(plugin);

	// Put a diamond in the inventory
	inventory.offer(ItemStack.of(ItemTypes.DIAMOND, 1));

	player.openInventory(inventory);

	// Few seconds later, teleport the player to trigger the teleport event
	Task.builder()
			.delayTicks(20 * 3)
			.execute(() -> player.setLocation(player.getLocation().add(0, 2, 0)))
			.submit(plugin);
}

@Listener
public void onPlayerTeleport(MoveEntityEvent.Teleport event, @Getter("getTargetEntity") Player player) {
        // Close the inventory to trigger the phase exception
	player.closeInventory();
}
```

**Reproduce instructions:**
1. Write !test in chat
2. Quickly click on the diamond to put it on your cursor
3. After 3 seconds you will be teleported into the air and your inventory will close.
4. Diamond will disapear and an exception is thrown

**Exception log:**
https://gist.github.com/VapidLinus/d4a4226444874b1c49a18f6b3e2468d9

> spongevanilla-1.12.2-8.0.0-BETA-393

RedNesto:
It seems it's only occurring when a plugin teleports an entity under some conditions (only tested with a Player though)

Fixes #1744

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
@@ -473,7 +473,7 @@ public abstract class MixinEntity implements IMixinEntity {
             return false;
         }
 
-        try (final BasicPluginContext context = PluginPhase.State.TELEPORT.createPhaseContext().buildAndSwitch()) {
+        try (final BasicPluginContext context = PluginPhase.State.TELEPORT.createPhaseContext().addEntityCaptures().buildAndSwitch()) {
 
             // TODO Add a 'Move' plugin phase or just keep it under Teleport?
             try (final CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame();) {


ryantheleach:
~qa

Thanks for the fix!

On the other issue you mention

>I wasn't sure if no items being dropped was this issue or a separate one. I noticed but haven't confirmed that player.closeInventory() always just makes the cursor item disappear even outside of this case.

Is this still the case, or did you find a different fix?

RedNesto:
~The other issue occurs even if the player closes the inventory itself, but yes it's still an issue and I'm searching a fix, but maybe Faithcaio have a better idea of what's happening.~ Seems to be a vanilla Minecraft behaviour when in creative mode (but it works fine in survival; the item is dropped).

Faithcaio:
A similar error also occured when closing the inventory in other situations.
I instead switched to the CloseWindow Phase to properly capture the drops.
https://github.com/SpongePowered/SpongeCommon/commit/7796f59c6671ba4d34e1273a5438c7049274e1c0

Issue #1371, opened 2017-05-25
Autosave-Message showing even if saving is disabled


sleiss:
**I am currently running...**
- SpongeForge version: 2353
- Java version: 1.8
- Operating System:

 - [x] Linux (Ubuntu, Debian, Fedora, etc)
 - [ ] macOS
 - [ ] Windows

**What is the expected result?**
The auto-saving message should only be printed if saving is activated (/save-on) as only then the world gets saved.

**What is the current result?**
When the minecraft-saving-option is disabled (/save-off) and sponge performs an autosave (based on the time configured in the sponge/global.conf), the log _always_ shows the message

> Auto-saving chunks for level 'world'/Overworld

(https://github.com/SpongePowered/SpongeCommon/blob/f430565244e46ad89a219cfb1323f1492bdcdb74/src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java#L721). 
but the world does not get saved.

**Possible ways to fix:**
- Print a warning (similiar to https://github.com/SpongePowered/SpongeCommon/blob/f430565244e46ad89a219cfb1323f1492bdcdb74/src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java#L710 if saving is disabled and an autosave is performed)
- Remove the message if saving is disabled
- Always save the world (even if the saving is disabled)

johnfriedrich:
Addition to issue above:
This can happen if autosave is off and sponge autosave is trying to save:
https://pastebin.com/JL4Anp2K

Thoughts:
I would prefer sponges autosave to save even if minecraft saving is turned off.
That way you can control saving mechanics even better.

Sponges auto save should do the same as entering save-all in console.
(Saving the world, even if minecraft autosaving is off)

In my opinion it's better to have two completely independently working save systems. (Minecraft auto save and Sponge autosave)

ryantheleach:
This is clearly a bug. 
We should not claim in the log to save, if saving is disabled.

My opinion @i9hdkill is that /save-off should be the highest priority when it comes to auto-saves of any manner.

Backup systems use /save-off before triggering backups to ensure atomic writes and backups are performed. To do *anything* counter to this behavior would risk peoples backup systems.

I believe a warning is the best way to go.

ImMorpheus:
I'm able to reproduce this on SpongeForge (latest bleeding branch)

PR #1747, opened 2018-01-31
Fix Autosave-Message showing even if saving is disabled

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeCommon/issues/1371

Based on ryan's comment: 

> My opinion is that /save-off should be the highest priority when it comes to auto-saves of any manner.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java
@@ -537,13 +537,6 @@ public abstract class MixinMinecraftServer implements Server, ConsoleSource, IMi
         this.enableSaving = enabled;
     }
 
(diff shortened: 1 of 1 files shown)


ImMorpheus:
~qa

src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java
```diff
@@ -720,8 +713,11 @@ private int getSaveTickInterval(int tickInterval) {
      */
     @Overwrite
     public void saveAllWorlds(boolean dontLog) {
+        if (!enableSaving) {
```
kashike: `this.`

kashike (changes requested):
One quick change.

kashike (approved):

kashike:
~ready

kashike:
Thanks, @ImMorpheus.

Issue #1688, opened 2018-01-05
Request: Enhance the Cause of SpawnEntityEvent for ExpOrbs


Zerthick:
**Issue Description**
I'm currently working on writing a plugin that controls the amount of experience players gain for different actions. Below is a capture of the causes for the different actions I want to control:

```
ExpBottle: Cause[Context=Context["sponge:spawn_type"=SpongeSpawnType{id=sponge:experience, name=Experience}, "sponge:owner"=EntityPlayerMP['Zerthick'/1955, l='world', x=13.27, y=67.00, z=254.15]], Stack={EntityExpBottle['entity.ThrownExpBottle.name'/6348, l='world', x=10.60, y=64.25, z=250.11]}]
Fishing: Cause[Context=Context["sponge:spawn_type"=SpongeSpawnType{id=sponge:placement, name=Placement}, "sponge:owner"=EntityPlayerMP['Zerthick'/1955, l='world', x=80.87, y=63.00, z=236.84], "sponge:used_item"=SpongeItemStackSnapshot{itemType=ItemFishingRod{Name=fishingRod}, quantity=1}, "sponge:notifier"=EntityPlayerMP['Zerthick'/1955, l='world', x=80.87, y=63.00, z=236.84]], Stack={EntityPlayerMP['Zerthick'/1955, l='world', x=80.87, y=63.00, z=236.84]}]
Smelt: Cause[Context=Context[], Stack={net.minecraft.inventory.ContainerFurnace@c478eaf, EntityPlayerMP['Zerthick'/1955, l='world', x=79.76, y=64.00, z=238.36]}]
Mining: Cause[Context=Context["sponge:spawn_type"=SpongeSpawnType{id=sponge:dropped_item, name=DroppedItem}, "sponge:used_item"=SpongeItemStackSnapshot{itemType=ItemPickaxe{Name=pickaxeDiamond}, quantity=1}], Stack={EntityPlayerMP['Zerthick'/1955, l='world', x=57.23, y=65.00, z=211.30]}]
Combat: Cause[Context=Context["sponge:notifier"=EntityPlayerMP['Zerthick'/1955, l='world', x=59.64, y=65.00, z=213.15]], Stack={EntityZombie['Zombie'/10750, l='world', x=56.65, y=64.00, z=213.70]}]
[...]

PR #1749, opened 2018-02-01
Fix incorrect spawntypes

ImMorpheus:
This should fix everything on https://github.com/SpongePowered/SpongeCommon/issues/920 and https://github.com/SpongePowered/SpongeCommon/issues/1688

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketState.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketState.java
@@ -33,6 +33,7 @@ import org.spongepowered.api.Sponge;
 import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.entity.ExperienceOrb;
 import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.api.entity.projectile.Projectile;
 import org.spongepowered.api.event.CauseStackManager;
 import org.spongepowered.api.event.SpongeEventFactory;
 import org.spongepowered.api.event.cause.EventContextKeys;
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketState.java
```diff
@@ -175,10 +177,16 @@ public boolean spawnEntity(P context, Entity entity, int chunkX, int chunkZ) {
         final ArrayList<Entity> entities = new ArrayList<>(1);
         entities.add(entity);
         try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
-            Sponge.getCauseStackManager().pushCause(player);
-            Sponge.getCauseStackManager().addContext(EventContextKeys.SPAWN_TYPE, SpawnTypes.PLACEMENT);
-            Sponge.getCauseStackManager().addContext(EventContextKeys.NOTIFIER, player);
-            Sponge.getCauseStackManager().addContext(EventContextKeys.OWNER, player);
+            frame.pushCause(player);
+            if (entity instanceof ExperienceOrb) {
```
gabizou, line 181: I'm almost tempted to make this based on the entity itself. I'll merge this with some modifications through EntityUtil (so as to unity usage throughout other places the spawn type is determined based on the entity).

src/main/java/org/spongepowered/common/event/tracking/phase/tick/BlockTickPhaseState.java
```diff
@@ -145,9 +146,13 @@ public boolean spawnEntityOrCapture(BlockTickContext context, Entity entity, int
             }
             final List<Entity> nonExpEntities = new ArrayList<>(1);
             nonExpEntities.add(entity);
-            Sponge.getCauseStackManager().addContext(EventContextKeys.SPAWN_TYPE, InternalSpawnTypes.BLOCK_SPAWNING);
+            if (entity instanceof FallingBlock) {
```
gabizou, line 149: And likewise here.

gabizou (commented):
Might want to look into possibly making the spawn type deterministic for other areas as well, and reduce the duplication of code by using either the PhaseState, or the entity class itself.

PR #1752, opened 2018-02-02
Implement the new pathfinders and advanced agent data

liach:
[**SpongeAPI**](https://github.com/SpongePowered/SpongeAPI/pull/1758) | **SpongeCommon**

This overhauled the agent data to support the new value and added a mixin to implement the newly exposed `PathNavigate` (navigator).

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntityLiving.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntityLiving.java
@@ -26,8 +26,10 @@ package org.spongepowered.common.mixin.core.entity;
 
 import net.minecraft.entity.EntityLiving;
 import net.minecraft.entity.EntityLivingBase;
+import net.minecraft.entity.SharedMonsterAttributes;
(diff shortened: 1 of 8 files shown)


src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntityLiving.java
```diff
@@ -276,14 +284,19 @@ private boolean onCanGrief(GameRules gameRules, String rule) {
...
-        return new SpongeAgentData(!this.isAIDisabled());
+        return new SpongeAgentData(!this.isAIDisabled(), this.getEntityAttribute(SharedMonsterAttributes.FOLLOW_RANGE).getBaseValue());
     }
 
     @Override
     public Value<Boolean> aiEnabled() {
         return new SpongeValue<>(Keys.AI_ENABLED, true, !this.isAIDisabled());
     }
 
+    @Intrinsic
```
gabizou, line 295: Since the return type is different, intrinsic isn't needed.

gabizou (changes requested):

PR #1754, opened 2018-02-03
Fix SpongeContextCalculator NPE on Client

Faithcaio:
When running Sponge on a Client the InetSocketAddress is unresolved causing a NPE when calculating Permission Contexts.

### DIFF
--- a/src/main/java/org/spongepowered/common/service/permission/SpongeContextCalculator.java
+++ b/src/main/java/org/spongepowered/common/service/permission/SpongeContextCalculator.java
@@ -31,6 +31,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
 import org.spongepowered.api.command.CommandSource;
 import org.spongepowered.api.command.source.RemoteSource;
+import org.spongepowered.api.network.RemoteConnection;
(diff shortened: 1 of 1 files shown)


stephan-gh:
Note that we cannot expect virtual hosts to resolve correctly. This is something we're usually given by the client, and they can give us anything (even hostnames that don't resolve correctly). Not sure if throwing an exception when it cannot be resolved is a good idea.

Faithcaio:
What should we do then when its invalid?
It currently throws NPEs thats not a good idea either

stephan-gh:
I guess there are two things we can do, not sure which one is better because I don't know much about the permission API:

- Return null if the hostname can't be resolved and modify `buildAddressCache` to skip adding the context entirely if the method returns null (instead of throwing a NPE).
- Given that the virtual host should usually point to the server itself, we could use the local IP address of the server as a fallback.

Zidane:
@lucko 

Thoughts on the above?

lucko:
As long as the caches account for a null value, then that's fine.

The current fix looks good imo.

Zidane:
@Faithcaio 

Does it?

lucko:
Yes, null is handled inside of the load method :+1:

Issue #1520, opened 2017-09-09
1.12.1: Programmatic Explosions damaging blocks do not update client until relogin


theboomer:
A programmatic explosion that causes block damage and/or causes fire does not produce block damage/fire on the client until the player logs out and back again.

![explosionsv](https://user-images.githubusercontent.com/3393225/30242108-3e85abc2-955e-11e7-897b-6acb8b776230.gif)

Entire test-plugin code to demonstrate is:
```
@Plugin(id = "quickdirtytester")
public class QuickDirty {
    @Inject
    PluginContainer container;

    @Listener
    public void placeMultiblock(ChangeBlockEvent.Place event, @Root Player player) {
        if (event.getTransactions().size() > 1) { // ie Beds, Doors, tall plants
            Location<World> loc = player.getLocation();
            Explosion boom = Explosion.builder().canCauseFire(true).radius(4f).shouldBreakBlocks(true).location(loc).build();
            loc.getExtent().triggerExplosion(boom, Cause.source(this.container).build());
        }
    }
}
```

- Operating System:

 - [ ] Linux (Ubuntu, Debian, Fedora, etc)
 - [ ] macOS
 - [x] Windows

SpongeVanilla 1.12.1 beta-323, SpongeForge 1.21.1 2444-beta-2612 with Forge 1.21.1-2444

PR #1756, opened 2018-02-03
Fix Programmatic Explosions not updating the client

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeCommon/issues/1520


ImMorpheus:
~qa

Issue #1442, opened 2017-07-15
Bug : Empty block owner after falling


lesbleu:
- SpongeVanilla version: SpongeVanilla version: 1.12-7.0.0-BETA-300
- Java version: 8
- Operating System:
 - [x] Linux (Ubuntu, Debian, Fedora, etc)
 - [ ] macOS
 - [x] Windows

When player place block with gravity property, The player is the Owner of this block.
After the block falling, the block owner is empty.

[Video](https://www.youtube.com/watch?v=sNgKmmwietA)

ImMorpheus:
Same of https://github.com/SpongePowered/SpongeCommon/issues/1439

Not able to reproduce, the owner is present.

Test plugin:
```
    @Listener
    public void onPlace(ChangeBlockEvent.Place event) {
        System.out.println(event.getContext().get(EventContextKeys.OWNER));
    }
```

This can be closed.

ryantheleach:
After looking into this, I think this was the result of the `/sponge blockinfo` command, rather then the owner in the cause, which is correct as morph said. @ImMorpheus is working on a PR.

@lesbleu I know this took us some time to look into it, but it was nearly missed, The video was great, but having text instructions would have made this crystal clear.

ImMorpheus:
PR merged. This should be closed.

Issue #1439, opened 2017-07-15
[1.12] Creator : Liquid flow


lesbleu:
- SpongeVanilla version: 1.12-7.0.0-BETA-300
- Java version: 8
- Operating System:
 - [x] Linux (Ubuntu, Debian, Fedora, etc)
 - [ ] macOS
 - [x] Windows

When player place liquid, the player is the owner of liquid ![Example](https://image.noelshack.com/fichiers/2017/28/6/1500076494-2017-07-15-01-54-34.png). 
And the liquid flowing, the owner of liquid is empty ![Example](https://image.noelshack.com/fichiers/2017/28/6/1500076595-2017-07-15-01-54-43.png)

It's a bug ? Because I want to cancel this event only when player flow liquid.

JBYoshi:
@SpongePowered/developers Is this intended behavior?

ryantheleach:
@lesbleu Are you able to check on the latest SF / SV?

Additionally, when I mentioned it to blood, he says he is having no issue at listening at ChangeBlockEvent.pre and that the context is complete in GriefPrevention.

Can you supply a test plugin that demonstrates the issue in order to clarify? I assume you mean that the metadata that Sponge logs regarding player notifier/creator is incorrect, and that you are listening to ChangeBlockEvent.place or something?

bloodmc:
https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/event/block/ChangeBlockEvent.java#L97

The proper user will be in event. Just check if source is liquid or context (once updated to support liquid flow).

ryantheleach:
Reopening this. `/sponge blockinfo` was used by @ImMorpheus after his PR, the PR only fixes the sand falling not having owners, but doesn't correct the water not having owners. This still issue still needs to be addressed. @bloodmc I know we tested this ages and ages ago with GP, but since then there has been pre events and all sorts added. 

What's the correct behavior for water and owners/notifiers?

ImMorpheus:
PR merged. This should be closed.

PR #1757, opened 2018-02-03
Fix empty block owner after falling

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeCommon/issues/1442
Fix https://github.com/SpongePowered/SpongeCommon/issues/1439

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketState.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketState.java
@@ -29,6 +29,7 @@ import net.minecraft.entity.player.EntityPlayerMP;
 import net.minecraft.network.Packet;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.WorldServer;
+import net.minecraft.world.chunk.Chunk;
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.entity.living.player.Player;
@@ -94,7 +95,9 @@ public abstract class PacketState<P extends PacketContext<P>> implements IPhaseS
     public void associateNeighborStateNotifier(P unwindingContext, BlockPos sourcePos, Block block, BlockPos notifyPos, WorldServer minecraftWorld,
         PlayerTracker.Type notifier) {
         final Player player = unwindingContext.getSpongePlayer();
-        ((IMixinChunk) minecraftWorld.getChunkFromBlockCoords(notifyPos)).setBlockNotifier(notifyPos, player.getUniqueId());
+        Chunk chunk = minecraftWorld.getChunkFromBlockCoords(notifyPos);
+        ((IMixinChunk) chunk).setBlockCreator(notifyPos, player.getUniqueId());
+        ((IMixinChunk) chunk).setBlockNotifier(notifyPos, player.getUniqueId());
     }
 
     public void populateContext(EntityPlayerMP playerMP, Packet<?> packet, P context) {
--- a/src/main/java/org/spongepowered/common/event/tracking/phase/tick/BlockEventTickPhaseState.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/phase/tick/BlockEventTickPhaseState.java
@@ -123,6 +123,8 @@ class BlockEventTickPhaseState extends TickPhaseState<BlockEventTickContext> {
         final Vector3d changedPosition = changedLocation.getPosition();
         final BlockPos changedBlockPos = VecHelper.toBlockPos(changedPosition);
         final IMixinChunk changedMixinChunk = (IMixinChunk) ((WorldServer) changedLocation.getExtent()).getChunkFromBlockCoords(changedBlockPos);
+        changedMixinChunk.getBlockOwner(changedBlockPos)
+                .ifPresent(owner -> changedMixinChunk.addTrackedBlockPosition(block, changedBlockPos, owner, PlayerTracker.Type.OWNER));
         final User user = TrackingUtil.getNotifierOrOwnerFromBlock(changedLocation);
         if (user != null) {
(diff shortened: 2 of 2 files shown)


ImMorpheus:
~qa

src/main/java/org/spongepowered/common/event/tracking/phase/packet/InteractionPacketState.java
```diff
@@ -123,6 +123,7 @@ public void unwind(BasicPacketContext phaseContext) {
         final ItemStackSnapshot usedSnapshot = ItemStackUtil.snapshotOf(usedStack);
         final Entity spongePlayer = EntityUtil.fromNative(player);
         try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
+            phaseContext.addNotifierAndOwnerToCauseStack();
```
gabizou: It's an interaction packet state.... This should not be done at all since it's a *player packet being processed*...

src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketState.java
```diff
@@ -94,6 +94,7 @@ public void addNotifierToBlockEvent(P context, IMixinWorldServer mixinWorldServe
     public void associateNeighborStateNotifier(P unwindingContext, BlockPos sourcePos, Block block, BlockPos notifyPos, WorldServer minecraftWorld,
         PlayerTracker.Type notifier) {
         final Player player = unwindingContext.getSpongePlayer();
+        unwindingContext.getOwner().ifPresent(owner -> ((IMixinChunk) minecraftWorld.getChunkFromBlockCoords(notifyPos)).setBlockCreator(notifyPos, owner.getUniqueId()));
```
gabizou: Same here.... use the player......... There's no other owner for a player, is there...

src/main/java/org/spongepowered/common/event/tracking/phase/tick/BlockEventTickPhaseState.java
```diff
@@ -123,6 +123,7 @@ public void handleBlockChangeWithUser(@Nullable BlockChange blockChange,
         final Vector3d changedPosition = changedLocation.getPosition();
         final BlockPos changedBlockPos = VecHelper.toBlockPos(changedPosition);
         final IMixinChunk changedMixinChunk = (IMixinChunk) ((WorldServer) changedLocation.getExtent()).getChunkFromBlockCoords(changedBlockPos);
+        changedMixinChunk.getBlockOwner(changedBlockPos).ifPresent(owner -> changedMixinChunk.addTrackedBlockPosition(block, changedBlockPos, owner, PlayerTracker.Type.OWNER));
```
gabizou: This part actually makes sense, but move the `ifPresent` to the next line.

gabizou (changes requested):
Minor changes, needs some explanation why owners and notifiers are being used for packet states.

src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketState.java
```diff
@@ -94,6 +94,7 @@ public void addNotifierToBlockEvent(P context, IMixinWorldServer mixinWorldServe
     public void associateNeighborStateNotifier(P unwindingContext, BlockPos sourcePos, Block block, BlockPos notifyPos, WorldServer minecraftWorld,
         PlayerTracker.Type notifier) {
         final Player player = unwindingContext.getSpongePlayer();
+        ((IMixinChunk) minecraftWorld.getChunkFromBlockCoords(notifyPos)).setBlockCreator(notifyPos, player.getUniqueId());
```
Zidane: Store the chunk lookup. This is doing a double lookup now.

Zidane:
@ImMorpheus Fix my latest comment and squash to 1 commit and I'll pull.

ImMorpheus:
Done

src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketState.java
```diff
@@ -94,7 +95,9 @@ public void addNotifierToBlockEvent(P context, IMixinWorldServer mixinWorldServe
     public void associateNeighborStateNotifier(P unwindingContext, BlockPos sourcePos, Block block, BlockPos notifyPos, WorldServer minecraftWorld,
         PlayerTracker.Type notifier) {
         final Player player = unwindingContext.getSpongePlayer();
-        ((IMixinChunk) minecraftWorld.getChunkFromBlockCoords(notifyPos)).setBlockNotifier(notifyPos, player.getUniqueId());
+        Chunk chunk = minecraftWorld.getChunkFromBlockCoords(notifyPos);
+        ((IMixinChunk) chunk).setBlockCreator(notifyPos, player.getUniqueId());
```
gabizou, line 99: Why are we setting the block creator here? Just found where this came from and I'm not following why we're notifying blocks makes the player the now creator, when they haven't created the blocks in the first place (unless a block change is done on that block).

gabizou: From my understanding, what this is now doing is making any blocks notified by a player interacting with the block as *created* by the player, when that is simply not the case nor intention of such in the API.

gabizou: Ver much a bug. Will need to eliminate this line here and re-test for this PR's original intention of fixing blocks falling without an owner.

ImMorpheus: Working on it. For now, this is what I came up with https://github.com/ImMorpheus/SpongeCommon/commit/d807846c30fc5e3a49fc8590d5f3e49dcc29f600

Issue #1446, opened 2017-07-16
No event launched when lava spread fire


lesbleu:
- SpongeVanilla version: SpongeVanilla version: 1.12-7.0.0-BETA-300
- Java version: 8
- Operating System:
 - [x] Linux (Ubuntu, Debian, Fedora, etc)
 - [ ] macOS
 - [x] Windows

A player places lava around flammable blocks. When fire is spreading by lava, no event is launched.

lesbleu:
At [MixinBlock](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/block/MixinBlock.java#L155), Event is not launched with there conditions. It's normal ? If yes, why ?

ImMorpheus:
PR merged. This can be closed.

PR #1760, opened 2018-02-03
Fix no event being thrown when lava spreads fire

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeCommon/issues/1446


Zidane:
@ImMorpheus 

Want to finally make ChangeBlockEvent.Grow work :D?

Faithcaio:
Why not remove it instead? And make it a Context
(Decay too)

Zidane:
@Faithcaio 

We can do that, yes but that is an API break.

RedNesto:
I'm actually working on the removal of ChangeBlockEvent.Grow and Decay for API 8, by putting them into the context of ChangeBlockEvent.

Issue #1762, opened 2018-02-03
NBTTranslator cannot handle Enums


NeumimTo:
**I am currently running**
- SpongeVanilla version: API 7.1 beta 16
- Java version: 8
- Operating System: ubuntu 17

I'm trying to implement AbstractSingleEnumData but every time i offer the datamanipulator the action ends with  "Unable to translate object to NBTBase:". 

    at org.spongepowered.common.data.persistence.NbtTranslator.getBaseFromObject(NbtTranslator.java:164) ~[NbtTranslator.class:1.12.2-7.1.0-BETA-16]
    at org.spongepowered.common.data.persistence.NbtTranslator.getBaseFromObject(NbtTranslator.java:150) ~[NbtTranslator.class:1.12.2-7.1.0-BETA-16]
    at org.spongepowered.common.data.persistence.NbtTranslator.getBaseFromObject(NbtTranslator.java:150) ~[NbtTranslator.class:1.12.2-7.1.0-BETA-16]
    at org.spongepowered.common.data.persistence.NbtTranslator.containerToCompound(NbtTranslator.java:91) ~[NbtTranslator.class:1.12.2-7.1.0-BETA-16]
    at org.spongepowered.common.data.persistence.NbtTranslator.containerToCompound(NbtTranslator.java:72) ~[NbtTranslator.class:1.12.2-7.1.0-BETA-16]
    at org.spongepowered.common.data.persistence.NbtTranslator.translateData(NbtTranslator.java:279) ~[NbtTranslator.class:1.12.2-7.1.0-BETA-16]

see https://github.com/SpongePowered/SpongeCommon/blob/31b25e39eb8b4b2dde382de7f147b5477e054485/src/main/java/org/spongepowered/common/data/persistence/NbtTranslator.java#L164

Could it be possible to store enum as a string?

my datamanipulator code: https://pastebin.com/y3GFsxzr

Zidane:
It could be @NeumimTo . Wanna PR us a fix?

NeumimTo:
tbh I have no idea how datamanipulators internally works. i would need someone who can i harrass with bunch of questions.

NeumimTo:
eh.. ill give it a try

JBYoshi:
[...]

PR #1766, opened 2018-02-04
nbttranslator (de)serialize enum types

NeumimTo:
More discussion here some code samples to reproduce this : https://github.com/SpongePowered/SpongeCommon/issues/1762#issuecomment-362869260

sample: https://pastebin.com/y3GFsxzr

Currently NBTTranslator cannot automatically (de)serialize enum type. 
This change will serialize enum as $Enum(classname)Value

[img](https://i.gyazo.com/081fc9bc4c4413d5549bf202382f753b.png "image")

- Downsides: 
   The user will encounter problems when the enum value will change name/or change class package
If the class is not found to be present at runtime the translator simply returns serialized string as described above. Now its up to developer to handle this inconsistence.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/persistence/NbtTranslator.java
+++ b/src/main/java/org/spongepowered/common/data/persistence/NbtTranslator.java
@@ -53,12 +53,20 @@ import org.spongepowered.common.data.util.NbtDataUtil;
 
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 
 public final class NbtTranslator implements DataTranslator<NBTTagCompound> {
+
+    private static Map<String, Class<? extends Enum>> knownEnumMap = new ConcurrentHashMap<>();
 
     private static final NbtTranslator instance = new NbtTranslator();
     private static final TypeToken<NBTTagCompound> TOKEN = TypeToken.of(NBTTagCompound.class);
     public static final String BOOLEAN_IDENTIFER = "$Boolean";
+    public static final String ENUM_IDENTIFIER = "$Enum(%s)%s";
+    public static final Pattern ENUM_CLASS_PATTERN = Pattern.compile("\\$Enum\\((.*?)\\)");
 
     public static NbtTranslator getInstance() {
         return instance;
@@ -87,6 +95,9 @@ public final class NbtTranslator implements DataTranslator<NBTTagCompound> {
                 compound.setTag(key, inner);
             } else if (value instanceof Boolean) {
                 compound.setTag(key + BOOLEAN_IDENTIFER, new NBTTagByte(((Boolean) value) ? (byte) 1 : 0));
+            } else if (value instanceof Enum) {
+                Enum enum_ = (Enum) value;
+                compound.setTag(key, new NBTTagString(String.format(ENUM_IDENTIFIER, enum_.getClass().getName(), enum_.name())));
             } else {
                 compound.setTag(key, getBaseFromObject(value));
             }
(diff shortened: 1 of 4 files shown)


ryantheleach:
@NeumimTo 

**Issue/PR Management:**

I'm going to assume this is no longer WIP unless corrected. ~qa

Also can you provide a plugin or simple example of how this would be used by plugins?

**PR Feedback:**

Do you think it's absolutely necessary to include the class? I'm thinking it's making it a ton more fragile then it needs to be (it's the reason why fully qualified data id's were deprecated), Unless random enums are being stored in random maps, then generally the Data Manipulator is going to know what enum is being registered, so couldn't we limit it to the enums name (rather then package) as well as it's value? Or just the value itself?

But you could make the argument that once you have done that, then the datamanipulators could pretty easily just use name and .valueOf themselves, rendering this moot.

Additionally (and this is a place where the Data API fails in general so it's probably not too important) I like to think about what happens when the data is being serialized into a user-editable format. Could an admin with pretty advanced server administration knowledge, but no programming experience edit this given a known Schema? No, not really. They would have to know the enum names applicable, have to know the classes that the data container / manipulator is expecting, and have to know the custom format that is present to parse it to even read it.

NeumimTo:
@ryantheleach the example datamanipulator is in the linked issue. and yes its not wip

The class reference is needed. How else would you know which enum type is it supposed to map at runtime?

>Unless random enums are being stored in random maps

Im not sure what do you mean with the terms "random enums" and "random maps"

Zidane:
@gabizou Need a decision on this.

Cybermaxke:
I have to agree with @ryantheleach, it's a bad idea to reference to a `Class` since the location of the class can change. The `Enum`s should be serialized/deserialized on `DataView` level, see https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/data/DataView.java#L553
Originally classes were also used to store `DataManipulator`s, which got replaced by using ids because it was causing issues when the classes changed, and there was no way to update the location of the class.

EDIT: Seems like you fall back to using `String`s when the enum class isn't found. I would still avoid this kind of specific things when working with NBT tags, but the only right way to retrieve enum values would be through `getObject` (and implement support for it, if not done already).

gabizou:
Here’s the bigger question: why is this in the translator and not handled in DataView itself? Natively the dataview should only have primitive types after deserialization/serialization from any format, so having this specifically in the translator does not make sense. For your enum, I would recommend registering a translator for your enum type.

Cybermaxke:
Since all the values of a `Enum` are known, there is no need to register a translator and can be supported directly in sponge. Unless somebody wants to override the serialization method ofc.

gabizou:
> Since all the values of a Enum are known, there is no need to register a translator and can be supported directly in sponge. Unless somebody wants to override the serialization method ofc.

The only case I can imagine overriding is to specifically support it as a fail fast for defaults, but yes, otherwise it can be supported natively in sponge's DataView implementation.

NeumimTo:
@gabizou  umm soooo ? Should i rewrite it to dataview instead

gabizou:
You should rewrite it to be properly handled in `MemoryDataVeiw` implementation since the translators don't generally have to worry about serializing/deserializing Enums. There's other improvements that I wanted to make to `DataView` implementation where it would throw an exception if an object cannot be serialized down to primitive types or collection types.

gabizou:
Yeah, going to stand by my original decision:

> You should rewrite it to be properly handled in `MemoryDataVeiw` implementation since the translators don't generally have to worry about serializing/deserializing Enums. There's other improvements that I wanted to make to `DataView` implementation where it would throw an exception if an object cannot be serialized down to primitive types or collection types.

`NbtTranslator` should never be handling the deserialization of objects itself, just the transference of primitive types (and the limited collections) to `DataView` and back. It's the responsibility of `DataView` and to be providing an error condition for non-primitive types being added.

PR #1770, opened 2018-02-04
Fix Chunk#getHighestYAt

RedNesto:
Should do the trick.

This implements the `Chunk#getHighestYAt` method, which has the wrong signature for unknown reasons.

Ready to merge.


RedNesto:
~qa

parlough:
~reviewed

PR #1771, opened 2018-02-04
Fix a race condition with resource packs.

### LINKED ISSUE #1721: ResourcePackState unwind error
**I am currently running**

- SpongeForge version: spongeforge-1.12.2-2555-7.1.0-BETA-2842.jar
- SpongeVanilla version: 
- Java version: 1.8.0_152

- Plugins/Mods:
```
[09:29:18] [Server thread/INFO]: Mods (7): Minecraft, Minecraft Coder Pack, Forge Mod Loader, Minecraft Forge, SpongeAPI, SpongeForge, mercurius_updater
> sponge plugins
[09:29:26] [Server thread/INFO]: Plugins (6): Minecraft, Minecraft Coder Pack, SpongeAPI, SpongeForge, HuskyUI, WinterFellRP
```

**Issue Description**
Seem to have a 50/50 chance of triggering this when logging into the server
https://pastebin.ubuntu.com/26429042/

**Test Plugin:** https://pastebin.ubuntu.com/26429076/

JBYoshi:
If the client sends certain packets while the server is sending a resource pack, the server currently interprets those as "the client has closed the resource pack prompt" before the client shows the prompt, prompting it to resend the resource pack request. This prevents that lag by sending custom keep-alive packets after sending the resource pack request. Only after the client acknowledges the keep-alive packet (and, therefore, has shown the resource pack request) does the server begin to assume the resource pack prompt was closed.

And just in case a malicious/poorly-written client decides to send fake "resource pack accepted" packets to the server, this PR just logs a warning and skips processing.

Fixes #1721.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/phase/packet/ResourcePackState.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/phase/packet/ResourcePackState.java
@@ -33,7 +33,6 @@ import org.spongepowered.api.event.SpongeEventFactory;
 import org.spongepowered.api.event.entity.living.humanoid.player.ResourcePackStatusEvent;
 import org.spongepowered.api.resourcepack.ResourcePack;
 import org.spongepowered.common.SpongeImpl;
-import org.spongepowered.common.interfaces.IMixinPacketResourcePackSend;
 import org.spongepowered.common.interfaces.network.IMixinNetHandlerPlayServer;
 
 public class ResourcePackState extends BasicPacketState {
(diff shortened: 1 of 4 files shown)


ImMorpheus:
The server is still flooding the client with SPacketResourcePackSend.

```
    @Listener
    public void handleResourcePackStatusEvent(ResourcePackStatusEvent event) {
        System.out.println(event.getStatus());
    }
```
```
 [STDOUT]: [org.spongepowered.test.TestTest:handleResourcePackStatusEvent:128]: ACCEPTED
 [STDOUT]: [org.spongepowered.test.TestTest:handleResourcePackStatusEvent:128]: ACCEPTED
 [STDOUT]: [org.spongepowered.test.TestTest:handleResourcePackStatusEvent:128]: FAILED
```

should be:
```
 [STDOUT]: [org.spongepowered.test.TestTest:handleResourcePackStatusEvent:128]: ACCEPTED
 [STDOUT]: [org.spongepowered.test.TestTest:handleResourcePackStatusEvent:128]: SUCCESSFULLY_LOADED
```

Expanding my comment:
When the server sends a `SPacketResourcePackSend` NetHandlerPlayServer is [adding](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java#L263) the packet to a Deque which is [used](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java#L797) to flood the client again with the same packet pretty much ["every tick"](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java#L799). 
Quoting from the method (the reason the Deque exists):

> The vanilla client doesn't send any resource pack status if the user presses Escape to close the prompt.
> If the user moves around, they must have closed the GUI, so resend it to get a real answer.

Now, there are a lot of problems with that: `markPlayerActive()` (the method we are injecting into to resend the packet to the client) is called also inside `processChatMessage(CPacketChatMessage packetIn)` , so there's no need for you to move. This wrong assumption is causing problems like:
```
 [STDOUT]: [org.spongepowered.test.TestTest:handleResourcePackStatusEvent:128]: ACCEPTED
 [STDOUT]: [org.spongepowered.test.TestTest:handleResourcePackStatusEvent:128]: ACCEPTED
 [STDOUT]: [org.spongepowered.test.TestTest:handleResourcePackStatusEvent:128]: FAILED
```
with the test plugin.
Why this ? 

_You join the server_ 
server ----send-----> SPacketResourcePackSend (the packet is now in the deque)
_Click yes when asked to download the resourcepack_
client  ----send---->  CPacketResourcePackStatus (status: ACCEPTED (the download begins))
_a message is broadcasted in chat_ 
(...) (skipping packet exchange here for clarity)
_the server calls markPlayerActive()_
_resendLatestResourcePackRequest() is called because it was injected at head_
server ----send-----> SPacketResourcePackSend (it was in the Deque)
_The client accepts the pack_ (no prompt, when you accept a ResourcePack from a server, the setting for that server switches from Prompt to Enabled) 
client  ----send---->  CPacketResourcePackStatus (status: ACCEPTED)
_the download begi... wait oh no, I'm already downloading a ResourcePack_
client  ----send---->  CPacketResourcePackStatus (status: FAILED)

Isn't the packet removed from the Deque when the client accept the download ?

No, it is removed whenever the server receives a CPacketResourcePackStatus with a status different from ACCEPTED.
As a side note, that's the way I was going for, but there's no ResourcePack in the CPacketResourcePackStatus so the server has no way of knowing what ResourcePack the client is referring to.

This is the solution I came up with: https://github.com/ImMorpheus/SpongeCommon/commit/10eb62f9cf01abbb433c5993cccd1dc53eb6b4e2

(I was working on this because yesterday @Zidane pointed out this issue was a blocker).

PS:
I don't like my solution that much but TBH this is a [mojang bug](https://bugs.mojang.com/browse/MC-118236) and we shouln't flood the client with SPacketResourcePackSend in the first place.

JBYoshi:
@ImMorpheus I'm somewhat confused by your example.

> `markPlayerActive()` (the method we are injecting into to resend the packet to the client) is called also inside `processChatMessage(CPacketChatMessage packetIn)`, so there's no need for you to move.

`processChatMessage()` is only called when the client sends a chat message to the server, not when the server sends a chat message to the client.

> Isn't the packet removed from the Deque when the client accept the download ?
>
> No, it is removed whenever the server receives a CPacketResourcePackStatus with a status different from ACCEPTED.

I keep the resource pack in the queue because the download hasn't finished yet. It is removed once the client sends `SUCCESSFULLY_LOADED`.

ImMorpheus:
> I keep the resource pack in the queue because the download hasn't finished yet. It is removed once the client sends SUCCESSFULLY_LOADED.

That's the problem. Even if the download hasn't finished, the client sent a CPacketResourcePackStatus (status: ACCEPTED). The reason the queue exists is because, quoting

```
The vanilla client doesn't send any resource pack status if the user presses Escape to close the prompt.
If the user moves around, they must have closed the GUI, so resend it to get a real answer.
```

So it should be used to resend the packet ONLY if the client doesn't answer to the SPacketResourcePackSend (with a CPacketResourcePackStatus).

Another test case would be being hit while downloading the ResourcePack:

PlayerA join the server 
(...)
PlayerB join the server
server ----send-----> SPacketResourcePackSend (the packet is now in the deque)
Click yes when asked to download the resourcepack
client ----send----> CPacketResourcePackStatus (status: ACCEPTED (the download begins))
PlayerA hits PlayerB
(...) (skipping packet exchange here for clarity)
PlayerB moved so the server calls markPlayerActive()
resendLatestResourcePackRequest() is called because it was injected at head
server ----send-----> SPacketResourcePackSend (it was in the Deque)
The client accepts the pack (no prompt, when you accept a ResourcePack from a server, the setting for that server switches from Prompt to Enabled)
client ----send----> CPacketResourcePackStatus (status: ACCEPTED)
the download begi... wait oh no, I'm already downloading a ResourcePack
client ----send----> CPacketResourcePackStatus (status: FAILED)

Zidane:
@JBYoshi @ImMorpheus Status?

JBYoshi:
Another look at the client shows that sending a new resource pack while the client is downloading one stops the download instead of blowing up (see `ResourcePackRepository.downloadResourcePack()`, which calls `this.clearResourcePack()`). And since closing the GUI doesn't send a status message, I think it should be OK to just forward the new resource pack to the client, while using some custom keep-alive packets to determine which resource pack is being accepted. I'm still not sure how to handle the edge case where a client sends a resource pack acceptance in between receiving the resource pack request and the keep-alive packet.

I agree that this is terribly ugly, but this is the best we can do.

Zidane:
@JBYoshi Status?

JBYoshi:
I've implemented a new approach and am testing it right now.

JBYoshi:
~qa

ImMorpheus (approved):

ImMorpheus:
Everything works as expected.

kashike (approved):

kashike:
~reviewed

gabizou (approved):

parlough (approved):

PR #1772, opened 2018-02-04
Custom Data Generation

Cybermaxke:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1620) | **Common**

***Only magic in this PR!***

Implementation of the Custom Data Generation API. Still WIP. What's in the test plugin is currently working, other things are missing or untested.

Also includes extended nbt tag type support, all the data types which are supported in [LanternNBT](https://github.com/LanternPowered/LanternNBT) are now also supported in Sponge. This includes:
- `boolean[]`
- `boolean` (Was already supported)
- `String[]`
- `short[]`
- `float[]`
- `double[]`
- `char`
- `char[]`
- `DataView[]`
- `List` support for any of the default and above data types
No more losing your `double[]` (, etc.) once it got serialized into a nbt tag.


Cybermaxke:
~wip

src/main/java/org/spongepowered/common/data/SpongeDataRegistrationBuilder.java
```diff
@@ -152,20 +153,31 @@
...
         final SpongeDataRegistration<M, I> registration = new SpongeDataRegistration<>(this);
         SpongeDataManager.getInstance().registerInternally(registration);
         SpongeManipulatorRegistry.getInstance().register(registration);
         return registration;
     }
+
+    @Override
+    public DataRegistration<M, I> build() {
+        validate();
+        this.container = Sponge.getCauseStackManager().getCurrentCause().first(PluginContainer.class).get();
```
gabizou: Should make a more proper exception for this get call.

gabizou:
I'd like to see the produced class file and byte code if possible.

gabizou:
Having redone the API 8 Data stuffs, this can likely be valid for candidacy with API 7.3 if someone decides to pick it up.

PR #1776, opened 2018-02-06
Added a way to launch Arrows and SpectralArrows in the ProjectileLauncher

### LINKED ISSUE #1206: launchProjectile() method failing
Something seems to have changed between the SpongeVanilla-1.11.2-6.0.0 versions 196 and 198 where the player.launchProjectile() method is no longer working. This also affects Forge (although I'm not sure when it changed) and running the exact same plugin on 196 works whereas it does not on 198.

RedNesto:
[SpongeAPI ](https://github.com/SpongePowered/SpongeAPI/pull/1769) | SpongeCommon

This PR adds a way to launch Arrows and SpectralArrows from a ProjectileSource.

I am not sure however if we should support the launch of `Arrow`s, since they represent the abstract class `EntityArrow`. The behavior this PR adds will make possible to launch `Arrow`s as `EntityTippedArrow`s internally, and logs a warning.

The question this PR brings is: "Is this a bleeding change?" (see the SpongeAPI PR).

This fixes #1206

### DIFF
--- a/src/main/java/org/spongepowered/common/entity/projectile/ProjectileLauncher.java
+++ b/src/main/java/org/spongepowered/common/entity/projectile/ProjectileLauncher.java
@@ -41,6 +41,7 @@ import net.minecraft.entity.projectile.EntityLlamaSpit;
 import net.minecraft.entity.projectile.EntityPotion;
 import net.minecraft.entity.projectile.EntitySmallFireball;
 import net.minecraft.entity.projectile.EntitySnowball;
+import net.minecraft.entity.projectile.EntitySpectralArrow;
 import net.minecraft.entity.projectile.EntityThrowable;
 import net.minecraft.entity.projectile.EntityTippedArrow;
 import net.minecraft.entity.projectile.EntityWitherSkull;
@@ -65,6 +66,8 @@ import org.spongepowered.api.entity.projectile.Projectile;
 import org.spongepowered.api.entity.projectile.Snowball;
 import org.spongepowered.api.entity.projectile.ThrownExpBottle;
 import org.spongepowered.api.entity.projectile.ThrownPotion;
+import org.spongepowered.api.entity.projectile.arrow.Arrow;
+import org.spongepowered.api.entity.projectile.arrow.SpectralArrow;
 import org.spongepowered.api.entity.projectile.arrow.TippedArrow;
 import org.spongepowered.api.entity.projectile.explosive.DragonFireball;
 import org.spongepowered.api.entity.projectile.explosive.WitherSkull;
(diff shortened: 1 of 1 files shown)


RedNesto:
~qa

src/main/java/org/spongepowered/common/entity/projectile/ProjectileLauncher.java
```diff
@@ -183,6 +186,25 @@ private static void configureThrowable(EntityThrowable entity) {
...
+                SpectralArrow arrow = (SpectralArrow) new EntitySpectralArrow(source.world, source);
+                ((EntitySpectralArrow) arrow).shoot(source, source.rotationPitch, source.rotationYaw, 0.0F, 3.0F, 0);
+                return doLaunch(loc.getExtent(), arrow);
+            }
+        });
+        registerProjectileLogic(Arrow.class, new SimpleItemLaunchLogic<Arrow>(Arrow.class, Items.ARROW) {
+
+            @Override
+            protected Optional<Arrow> createProjectile(EntityLivingBase source, Location<?> loc) {
+                SpongeImpl.getLogger().warn("An Arrow will be launched. Arrow is not what you should use, maybe you want to use TippedArrow instead");
```
kashike, line 202: Why register `Arrow.class` at all?

RedNesto: Basically it's my big question in this PR, should we allow plugins to launch `Arrow` and launch a `TippedArrow` internally or not let plugins launch an `Arrow` at all ?

@Faithcaio says that we should support launching Arrows. To me, we should only throw a warning. I just merged the two ideas ;)

ryantheleach: On one hand you can say it removes the surprise of Arrow not launching anything as it's a super type, on the other hand it's strictly incorrect.

That said, it was a previous breaking change, that had no API contract change to cause a compile error.

I'm ok with it. but it does mask the fact that the normal arrows are actually tipped arrows, or the fact there are even multiple types of arrows, but at least it throws a warning.

I'd maybe write in the warning that "Using Arrow to launch projectiles is deprecated" instead of "Arrow is not what you should use"

Faithcaio: How is this strictly incorrect? TippedArrow extends Arrow.
Also as in the linked issue. 
As a plugin dev if you want to shoot an arrow your first idea will not be to shoot a TippedArrow instead. This could be explained by javadocs but still.
Imo the fact that there is an AbstractArrow in the impl doesn't matter and we should not expose this in the API.

RedNesto: So either we follow the Minecraft implementation and we do not allow launching  `Arrow`s, either we are maintaining the false idea people have for the today's Minecraft, which is: "A classic arrow in my inventory will launch a classic arrow entity" where there is no 'classic arrow entity' but only `TippedArrow`s

I see Sponge as an API representing the actual Minecraft, not hiding fact that the implementation is now different than it used to be.

So I would way "Let's educate people, still launch a TippedArrow when they ask for Arrow and log a warning to tell them that its not what they really want". But I'm bad for this kind of log so if we keep the warning I'll see to change the message.

parlough:
Thanks for your contribution :), handled by 86bb2c109a09664016e65cf678c4d110e89c6281

PR #1779, opened 2018-02-08
Fix the InteractBlockEvent.Secondary rounding the interaction point

### LINKED ISSUE #1778: InteractBlockEvent is not giving exact interaction point when using getInteractionPoint()
**I am currently running**

- SpongeVanilla version: 1.12.2-7.1.0-BETA-16
- Java version: 1.8.0_152
- Operating System: Win 10 Home 64-bit

- Plugins/Mods: PacketGate

**Issue Description**

The exact interaction point is available on the InteractItemEvent, but not on InteractBlockEvent, and the protocol says the interaction should be available on Player Block Placement too.

RedNesto:
This fixes the rounding issue the InteractBlockEvent.Secondary actually have when getting `InteractEvent#getInteractionPoint`.

Fixes #1778


RedNesto:
~qa

PR #1783, opened 2018-02-09
Fix InteractEntityEvent interaction point not always being absolute

### LINKED ISSUE #1782: InteractEntityEvent sometimes returns relative interaction point and sometimes not
**I am currently running**

- SpongeVanilla version: spongevanilla-1.12.2-7.1.0-BETA-0 (built from stable-7)
- Java version: 1.8.0_152
- Operating System: Win 10 Home 64-bit

- Plugins/Mods:

**Issue Description**
It's preferred to return the actual position not the relative position to the hit box , like other events
how can you test it ? try to interact an entity with right click which gives you relative position, then try to interact an entity with left click (i don't know how , but its possible) which gives you the actual position

RedNesto:
This fixes #1782 

Basically, the used Vector3d is relative to the entity (until this PR gets merged :3).


RedNesto:
~qa

PR #1792, opened 2018-02-10
Fix CatalogTypes

ImMorpheus:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1777) | **SpongeCommon**

Missing translations:
- EntityTypes: (Missing entries in the lang file)
entity.MinecartSpawner.name
entity.MinecartTNT.name
entity.ThrownEnderpearl.name 
entity.EyeOfEnderSignal.name
entity.FireworksRocketEntity.name
entity.MinecartCommandBlock.name
entity.MinecartFurnace.name
entity.LlamaSpit.name
entity.ThrownExpBottle.name
entity.ItemFrame.name
entity.SpectralArrow.name
entity.EvocationFangs.name
entity.ShulkerBullet.name
entity.MinecartRideable.name
entity.AreaEffectCloud.name
entity.LeashKnot.name
entity.WitherSkull.name
entity.EnderCrystal.name


ImMorpheus:
~wip

gabizou:
Does this fix the spam from the catalog type tests?

ImMorpheus:
All inconsistencies with mojang's translations should be fixed.

There are still a few things I don't like, I'll see if I can come up with something better. Suggestions are welcomed.

EDIT:

> Does this fix the spam from the catalog type tests?

Yes.

After this PR:

```
See edit

```

This also adds a test to make sure name and id are not the same and another one to find inconsistencies with mojang's translation

ImMorpheus:
Also, the old test had a few problems. If the translation was equal to the id, then it'd be classified as a missing translation. However minecraft is using the id as a fallback for a missing translation, thus that's not something we should fix.

Here's the situation before this PR with the new test:
```
See edit
```

src/main/java/org/spongepowered/common/entity/SpongeEntityType.java
```diff
@@ -132,4 +134,8 @@ public void setActivationRangeInitialized(boolean flag) {
                 .add("class", this.entityClass.getName());
     }
 
+    @Override
+    public Translation getTranslation() {
+        return translation;
```
ST-DDT: Add `this` here.

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemBanner.java
```diff
@@ -0,0 +1,46 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.common.interfaces.item.IMixinItem;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+@Mixin(ItemBanner.class)
+public abstract class MixinItemBanner extends MixinItem implements ItemType {
+
+    @Override
+    public Translation getTranslation() {
+        EnumDyeColor enumdyecolor = ItemBanner.getBaseColor(new ItemStack((Item) (IMixinItem) this));
```
ST-DDT: Cache the translation?

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemLingeringPotion.java
```diff
@@ -0,0 +1,44 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.common.interfaces.item.IMixinItem;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+@Mixin(ItemLingeringPotion.class)
+public abstract class MixinItemLingeringPotion extends MixinItem implements ItemType {
+
+    @Override
+    public Translation getTranslation() {
+        return new SpongeTranslation(PotionUtils.getPotionFromItem(new ItemStack((Item) (IMixinItem) this)).getNamePrefixed("lingering_potion.effect."));
```
ST-DDT: Cache the translation?

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemPotion.java
```diff
@@ -0,0 +1,45 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.common.interfaces.item.IMixinItem;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+@Mixin(ItemPotion.class)
+public abstract class MixinItemPotion extends MixinItem implements ItemType {
+
+    @Override
+    public Translation getTranslation() {
+        return new SpongeTranslation(PotionUtils.getPotionFromItem(new ItemStack((Item) (IMixinItem) this)).getNamePrefixed("potion.effect."));
```
ST-DDT: Same here

src/main/java/org/spongepowered/common/statistic/SpongeItemStatistic.java
```diff
@@ -0,0 +1,90 @@
...
+
+public class SpongeItemStatistic extends StatCrafting implements ItemStatistic, TypedSpongeStatistic {
+
+    private String spongeId;
+    private Translation translation;
+
+    public SpongeItemStatistic(String statId, String itemName, ITextComponent statName, Item item) {
+        super(statId, itemName, statName, item);
+        String args = new ItemStack(item).getDisplayName();
+        translation = new SpongeTranslation(statId.substring(0, statId.length() - 1), args);
```
ST-DDT: `this`

src/main/java/org/spongepowered/common/statistic/SpongeItemStatistic.java
```diff
@@ -0,0 +1,90 @@
...
+import org.spongepowered.common.registry.type.ItemTypeRegistryModule;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+import javax.annotation.Nullable;
+import java.util.Optional;
+
+public class SpongeItemStatistic extends StatCrafting implements ItemStatistic, TypedSpongeStatistic {
+
+    private String spongeId;
+    private Translation translation;
```
ST-DDT: `final`?

src/main/java/org/spongepowered/common/statistic/SpongeBlockStatistic.java
```diff
@@ -41,14 +43,22 @@
...
 
     public SpongeBlockStatistic(String statId, String itemName, ITextComponent statName, Item item) {
         super(statId, itemName, statName, item);
+        String args;
+        if (statId.equals("stat.mineBlock.")) {
+            args = Block.getBlockFromItem(item).getLocalizedName();
+        } else {
+            args = new ItemStack(item).getDisplayName();
+        }
+        translation = new SpongeTranslation(statId.substring(0, statId.length() - 1), args);
```
ST-DDT: `this`
also why did you initialize this in the constructor while all the others are lazy initialized in the getter?

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemTippedArrow.java
```diff
@@ -0,0 +1,44 @@
...
+import org.spongepowered.api.text.translation.Translation;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.common.interfaces.item.IMixinItem;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+@Mixin(ItemTippedArrow.class)
+public abstract class MixinItemTippedArrow extends MixinItem implements ItemType {
+
+    @Override
+    public Translation getTranslation() {
```
ST-DDT: Cache this.

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemStack.java
```diff
@@ -230,7 +241,25 @@ public DataContainer toContainer() {
     }
 
     public Translation itemstack$getTranslation() {
-        return new SpongeTranslation(shadow$getItem().getUnlocalizedName((net.minecraft.item.ItemStack) (Object) this) + ".name");
+        String id;
```
ST-DDT: Cache this.

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemSplashPotion.java
```diff
@@ -0,0 +1,45 @@
...
+import org.spongepowered.api.text.translation.Translation;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.common.interfaces.item.IMixinItem;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+@Mixin(ItemSplashPotion.class)
+public abstract class MixinItemSplashPotion extends MixinItem implements ItemType {
+
+    @Override
+    public Translation getTranslation() {
```
ST-DDT: Cache this.

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemSkull.java
```diff
@@ -0,0 +1,54 @@
...
+import org.spongepowered.common.interfaces.item.IMixinItem;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+import java.util.Optional;
+
+@Mixin(ItemSkull.class)
+public abstract class MixinItemSkull extends MixinItem implements ItemType {
+
+    @Override
+    public Translation getTranslation() {
```
ST-DDT: Cache this.

src/main/java/org/spongepowered/common/effect/record/SpongeRecordType.java
```diff
@@ -68,4 +71,9 @@ public static SPacketEffect createPacket(Vector3i position, @Nullable RecordType
         return new SPacketEffect(EFFECT_ID, pos, recordType == null ? 0 :
                 ((SpongeRecordType) recordType).getInternalId(), false);
     }
+
+    @Override
+    public Translation getTranslation() {
+        return translation;
```
ST-DDT: `this`

ST-DDT (changes requested):

ST-DDT:
Nice improvements.

src/main/java/org/spongepowered/common/statistic/SpongeBlockStatistic.java
ImMorpheus: Because I need the `statId` and the `item`

ImMorpheus:
```
    [Test worker/INFO]: FML platform manager could not load class cpw.mods.fml.relauncher.CoreModManager. Proceeding without FML support.
    [Test worker/ERROR]: MixinBootstrap.doInit() called during a tweak constructor!
    [Test worker/INFO]: Compatibility level set to JAVA_8
    [Test worker/INFO]: Loading tweak class name org.spongepowered.asm.mixin.EnvironmentStateTweaker
    [Test worker/INFO]: Calling tweak class org.spongepowered.asm.mixin.EnvironmentStateTweaker
    [Test worker/ERROR]: Classloader restrictions [PACKAGE_CLASSLOADER_EXCLUSION] encountered loading mixins.common.test.json:MixinBlock, name: org.spongepowered.common.launch.mixin.MixinBlock
    [Test worker/ERROR]: Classloader restrictions [PACKAGE_CLASSLOADER_EXCLUSION] encountered loading mixins.common.test.json:MixinEnumDifficulty, name: org.spongepowered.common.launch.mixin.MixinEnumDifficulty
    [Test worker/ERROR]: Classloader restrictions [PACKAGE_CLASSLOADER_EXCLUSION] encountered loading mixins.common.test.json:MixinGameType, name: org.spongepowered.common.launch.mixin.MixinGameType
    [Test worker/ERROR]: Classloader restrictions [PACKAGE_CLASSLOADER_EXCLUSION] encountered loading mixins.common.test.json:MixinItem, name: org.spongepowered.common.launch.mixin.MixinItem
    [Test worker/ERROR]: Classloader restrictions [PACKAGE_CLASSLOADER_EXCLUSION] encountered loading mixins.common.test.json:MixinSpongeItemStackSnapshot, name: org.spongepowered.common.launch.mixin.MixinSpongeItemStackSnapshot
    [Test worker/ERROR]: Classloader restrictions [PACKAGE_CLASSLOADER_EXCLUSION] encountered loading mixins.common.test.json:MixinStatBase, name: org.spongepowered.common.launch.mixin.MixinStatBase
    [Test worker/INFO]: Launching wrapped minecraft {org.spongepowered.common.test.TestMain}
```

I don't know what's going on here.

Everything else works as expected.

ImMorpheus:
~qa

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemSplashPotion.java
```diff
@@ -0,0 +1,48 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.potion.PotionUtils;
+import org.spongepowered.api.item.ItemType;
+import org.spongepowered.api.text.translation.Translation;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.common.interfaces.item.IMixinItem;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+@Mixin(ItemSplashPotion.class)
+public abstract class MixinItemSplashPotion extends MixinItem implements ItemType {
```
gabizou: You don't need to explicitly implement `ItemType` as it's already inherited from `MixinItem`.

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemSkull.java
```diff
@@ -0,0 +1,59 @@
...
+import org.spongepowered.api.text.translation.Translation;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.common.interfaces.item.IMixinItem;
+import org.spongepowered.common.item.inventory.util.ItemStackUtil;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+import java.util.Optional;
+
+@Mixin(ItemSkull.class)
+public abstract class MixinItemSkull extends MixinItem implements ItemType {
```
gabizou: You don't need to explicitly implement ItemType as it's already inherited from MixinItem.

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemPotion.java
```diff
@@ -0,0 +1,48 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.potion.PotionUtils;
+import org.spongepowered.api.item.ItemType;
+import org.spongepowered.api.text.translation.Translation;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.common.interfaces.item.IMixinItem;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+@Mixin(ItemPotion.class)
+public abstract class MixinItemPotion extends MixinItem implements ItemType {
```
gabizou: You don't need to explicitly implement ItemType as it's already inherited from MixinItem.

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemLingeringPotion.java
```diff
@@ -0,0 +1,47 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.potion.PotionUtils;
+import org.spongepowered.api.item.ItemType;
+import org.spongepowered.api.text.translation.Translation;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.common.interfaces.item.IMixinItem;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+@Mixin(ItemLingeringPotion.class)
+public abstract class MixinItemLingeringPotion extends MixinItem implements ItemType {
```
gabizou: You don't need to explicitly implement ItemType as it's already inherited from MixinItem.

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemBanner.java
```diff
@@ -0,0 +1,49 @@
...
+import net.minecraft.item.ItemBanner;
+import net.minecraft.item.ItemStack;
+import org.spongepowered.api.item.ItemType;
+import org.spongepowered.api.text.translation.Translation;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.common.interfaces.item.IMixinItem;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+@Mixin(ItemBanner.class)
+public abstract class MixinItemBanner extends MixinItem implements ItemType {
```
gabizou: You don't need to explicitly implement ItemType as it's already inherited from MixinItem.

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemStack.java
```diff
@@ -108,6 +119,8 @@
     @Shadow public abstract Item shadow$getItem();
 
 
+    @Shadow public abstract int getMetadata();
```
gabizou: Can you group this with the methods above? (Eliminate the extra whitespace gap)

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemStack.java
```diff
@@ -93,6 +102,8 @@
 
     private List<DataView> failedData = new ArrayList<>();
 
+    @Nullable private Translation translation;
```
gabizou, line 97: Move this one line higher to avoid too much of a gap between the added instance fields.

src/main/java/org/spongepowered/common/statistic/SpongeBlockStatistic.java
```diff
@@ -41,14 +43,22 @@
...
+        } else {
+            args = new ItemStack(item).getDisplayName();
+        }
+        this.translation = new SpongeTranslation(statId.substring(0, statId.length() - 1), args);
     }
 
     @Override
     public Translation getTranslation() {
-        return new SpongeTranslation(this.statId);
+        return this.translation;
```
gabizou, line 67: Does this have to be field initialized in the constructor? You're doing lazy loading in pretty much other places, might as well lazy load from here, no?

src/main/java/org/spongepowered/common/statistic/SpongeItemStatistic.java
```diff
@@ -0,0 +1,90 @@
...
+
+public class SpongeItemStatistic extends StatCrafting implements ItemStatistic, TypedSpongeStatistic {
+
+    private String spongeId;
+    private final Translation translation;
+
+    public SpongeItemStatistic(String statId, String itemName, ITextComponent statName, Item item) {
+        super(statId, itemName, statName, item);
+        String args = new ItemStack(item).getDisplayName();
+        this.translation = new SpongeTranslation(statId.substring(0, statId.length() - 1), args);
```
gabizou: Likewise here, lazy load maybe?

src/main/java/org/spongepowered/common/statistic/SpongeItemStatistic.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Override
+    public Translation getTranslation() {
+        return translation;
```
gabizou: Turn on your `Instance fields not qualified with 'this'` inspection. Needs to be accessed by `this.translation`.

src/main/java/org/spongepowered/common/statistic/SpongeItemStatistic.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Override
+    public ItemType getItemType() {
+        return ItemTypeRegistryModule.getInstance().getById(
```
gabizou: put the return before `.getById(` so that the next line reads:
`.getById(this.statId.substring(this.statId.lastIndexOf('.') + 1)).get();`

src/test/java/org/spongepowered/common/launch/mixin/MojangTranslatable.java
```diff
@@ -0,0 +1,30 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.launch.mixin;
+
+public interface MojangTranslatable {
```
gabizou: You can't have this in the same package as mixins. If this interface is to help you in code, you need to put it in another package where mixins are not going to be excluded, this is why you're having the warning about certain mixins not being loaded.

src/test/java/org/spongepowered/common/launch/mixin/MixinStatBase.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package org.spongepowered.common.launch.mixin;
+
+import net.minecraft.stats.StatBase;
+import net.minecraft.util.text.ITextComponent;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+
+@Mixin(value = StatBase.class, remap = false)
+public abstract class MixinStatBase implements MojangTranslatable {
```
gabizou, line 34: Why does this have to be loaded in launch?

src/test/java/org/spongepowered/common/launch/mixin/MixinSpongeItemStackSnapshot.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.common.item.inventory.SpongeItemStackSnapshot;
+import org.spongepowered.common.item.inventory.util.ItemStackUtil;
+
+@Mixin(value = SpongeItemStackSnapshot.class, remap = false)
+public abstract class MixinSpongeItemStackSnapshot implements MojangTranslatable {
```
gabizou, line 38: Why does this have to be loaded in launch? ItemStackSnapshots are usually loaded well into the game initialization phase.

src/test/java/org/spongepowered/common/launch/mixin/MixinGameType.java
```diff
@@ -0,0 +1,41 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.launch.mixin;
+
+import net.minecraft.world.GameType;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+@Mixin(value = GameType.class, remap = false)
```
gabizou, line 33: Why does this have to be used in launch?

src/test/java/org/spongepowered/common/launch/mixin/MixinEnumDifficulty.java
```diff
@@ -0,0 +1,42 @@
...
+package org.spongepowered.common.launch.mixin;
+
+
+import net.minecraft.world.EnumDifficulty;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.common.text.translation.SpongeTranslation;
+
+@Mixin(value = EnumDifficulty.class, remap = false)
+public abstract class MixinEnumDifficulty implements MojangTranslatable {
```
gabizou, line 34: Why does this have to be used in launch?

src/test/java/org/spongepowered/common/launch/mixin/MixinBlock.java
```diff
@@ -0,0 +1,40 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.launch.mixin;
+
+import net.minecraft.block.Block;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+
+@Mixin(value = Block.class, remap = false)
+public abstract class MixinBlock implements MojangTranslatable {
```
gabizou, line 33: Why is this in a separate mixin? Block is already transformed in the normal init phase, we don't need yet another `MixinBlock` class when there's two (one for common, one for sponge forge).

src/test/java/org/spongepowered/common/registry/CatalogTypeMethodTest.java
```diff
@@ -103,22 +84,13 @@
...
-                    throw e.getCause();
-                }
-                throw e;
-            }
-        } catch (Throwable t) {
-            if (ignoredFailures.contains(this.implementationClass + "#" + this.method.getName() + "()")) {
-                LOG.warn("Catalog Type: {} Id : {} has broken Method: {} ({}): {}", this.name, this.catalogId, this.methodName,
-                        this.implementationClass, t);
-                return;
+            testResult(checkNotNull(this.method.invoke(this.catalogType), "return value"));
```
gabizou, line 95: How is this changing how it used to work previously?

src/test/java/org/spongepowered/common/registry/CatalogTypeMethodTest.java
```diff
@@ -138,31 +110,47 @@ private void testResult(Object object) {
...
-                            this.name, this.catalogId, translationId, this.methodName, this.implementationClass);
-                } else {
-                    throw new RuntimeException("No translation present for " + translationId);
+                return;
+            }
+
+
+            if (object instanceof IStringSerializable || object instanceof Fish || object instanceof CookedFish) {
+                if (verbose) {
+                    LOG.info("This is probably a blockstate. Mojang doesn't directly provide translations for blockstates. Ignoring {}", translation.getId());
```
gabizou: Why is fish considered untranslatable but you're saying it's probably a block state?

gabizou (changes requested):
Some changes, clarifications.

src/main/java/org/spongepowered/common/statistic/SpongeItemStatistic.java
ImMorpheus: I took this from SpongeBlockStatistic. I suppose I have to change that as well.

src/main/java/org/spongepowered/common/statistic/SpongeBlockStatistic.java
ImMorpheus: I'm not sure there's a clean way to do this. I need `statId` and `item`.

src/main/java/org/spongepowered/common/statistic/SpongeItemStatistic.java
ImMorpheus: Same as above

src/test/java/org/spongepowered/common/registry/CatalogTypeMethodTest.java
ImMorpheus: Previously, everything inside `ignoredFailures`would be ignored (it would just log "x has broken method: etc.") but it was a workaround (see the TODO above). I've just removed the `if` and unwrapped the exception.

src/test/java/org/spongepowered/common/launch/mixin/MojangTranslatable.java
ImMorpheus: The thing is, all the other mixins load just fine. That's why I said I didn't know what is going on with the above error.

src/test/java/org/spongepowered/common/launch/mixin/MixinSpongeItemStackSnapshot.java
ImMorpheus: I'm probably missing something. These mixins should only be applied during tests and not shipped in production (Right ?).

src/main/java/org/spongepowered/common/mixin/core/item/MixinItem.java
```diff
@@ -52,11 +52,13 @@
 public abstract class MixinItem implements ItemType, IMixinItem, SpongeGameDictionaryEntry {
 
     public Optional<BlockType> blockType = Optional.empty();
+    @Nullable protected Translation translation;
 
     @Shadow private String unlocalizedName;
 
     @Shadow public abstract int getItemStackLimit();
     @Shadow public abstract String getUnlocalizedName();
```
ST-DDT: AFAICT this is no longer used.

src/main/java/org/spongepowered/common/mixin/core/item/MixinItemSkull.java
```diff
@@ -0,0 +1,59 @@
...
+@Mixin(ItemSkull.class)
+public abstract class MixinItemSkull extends MixinItem {
+
+    @Override
+    public Translation getTranslation() {
+        if (this.translation == null) {
+            ItemStack is = ItemStackUtil.fromNative(new net.minecraft.item.ItemStack((Item) (IMixinItem) this));
+            Optional<GameProfile> owner = is.get(Keys.REPRESENTED_PLAYER);
+
+            if (owner.isPresent()) {
```
ST-DDT: This cannot be present, since you just created the item or am I missing something?

src/main/java/org/spongepowered/common/statistic/SpongeBlockStatistic.java
ST-DDT: Instead of `statId.equals("stat.mineBlock.")` you could use `StatBase#statId.startsWith("stat.mineBlock.")`.

and instead of `item ` you could use `StatCrafting#Item`

src/test/java/org/spongepowered/common/interfaces/MojangTranslatable.java
```diff
@@ -0,0 +1,30 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.interfaces;
+
+public interface MojangTranslatable {
```
ST-DDT, line 30: Need javadocs that explain the purpose of this interface.

src/test/java/org/spongepowered/common/launch/mixin/MixinSpongeItemStackSnapshot.java
ST-DDT: Then it should probably be only in `src/test/java`. However for simplicity sake I would move them to `src/main/java`.

src/test/java/org/spongepowered/common/registry/CatalogTypeMethodTest.java
ST-DDT: AFAICT there are translations for fish!?

````properties
item.fish.cod.raw.name=Raw Fish
item.fish.salmon.raw.name=Raw Salmon
item.fish.pufferfish.raw.name=Pufferfish
item.fish.clownfish.raw.name=Clownfish
item.fish.cod.cooked.name=Cooked Fish
item.fish.salmon.cooked.name=Cooked Salmon
````

**Review by ST-DDT — CHANGES_REQUESTED** [review]

File: src/test/java/org/spongepowered/common/launch/mixin/MixinSpongeItemStackSnapshot.java (hunk 26 above)
Comment (ImMorpheus) [reply src/test/java/org/spongepowered/common/launch/mixin/MixinSpongeItemStackSnapshot.java:38]: They are in `src/test/java`.

> However for simplicity sake I would move them to src/main/java.

I'm not sure. The mojang translation is not something we need to expose in the API, it should only be used during tests to reveal inconsistencies.

File: src/test/java/org/spongepowered/common/registry/CatalogTypeMethodTest.java (hunk 31 above)
Comment (ImMorpheus) [reply src/test/java/org/spongepowered/common/registry/CatalogTypeMethodTest.java]: Indeed. However that's net.minecraft.item.ItemFishFood.

This check should avoid net.minecraft.item.ItemFishFood.FishType

File: src/test/java/org/spongepowered/common/launch/mixin/MixinSpongeItemStackSnapshot.java (hunk 26 above)
Comment (gabizou) [reply src/test/java/org/spongepowered/common/launch/mixin/MixinSpongeItemStackSnapshot.java:38]: oh I see, I didn't notice the source directory for these mixins, thought you were mixing into them in production and was questioning on that understanding hahah. These are fine, as long as they stay in the test sources.

**Comment by gabizou** [comment]
Sounds like it's almost ready, so with these changes the warnings are pretty much gone from all the catalog type tests? (I know I may have introduced one or two printouts with the recent PhaseTracker changes, but that I'll get to once I figure out how to beat a testing environment check without making those checks fail in production).

**Comment by ImMorpheus** [comment]
> so with these changes the warnings are pretty much gone from all the catalog type tests?

Correct, but as you can see from the travis log https://travis-ci.org/SpongePowered/SpongeCommon/jobs/403996687, it's still complaining about NotePitch having the same id and name. I think it's safe to ignore them (unless someone comes up with different names).

**Comment by Cybermaxke** [comment]
Ids shouldn't be uppercase, I noticed `C_SHARP1` for one of the NotePitches in the log. With the `CatalogKey` PR, all catalog types should also get a plugin id in the future, so maybe this is the time?
`minecraft:id`

**Comment by ImMorpheus** [comment]
That's probably a breaking change. Is it something I can do in this PR ? Input wanted

**Comment by gabizou** [comment]
> That's probably a breaking change. Is it something I can do in this PR ? Input wanted

I'd say it's a fix, but best to check with some plugin developers for input. @clienthax for your note block music plugin, do the catalog id's changing from previously unprefixed to correctly prefixed note pitches change anything for you? (first example I could think of at this moment).

**Comment by ST-DDT** [comment]
There is a solution to it though. The associated registry could still return the right entries for old ids, so after the entries get loaded and saved once, you have completed the note pitch migration. Or does the config loader only use the ids for the mapping. IIRC they use GameRegistry#get(type, id).

**Comment by ImMorpheus** [comment]
The thing is, NotePitch is not the only one having unprefixed ids. 
Prefix everything and things like https://github.com/SpongePowered/SpongeCommon/issues/1286 would arise. I could fix NotePitch and leave everything else to the catalogue key PR. Probably adding other tests to ensure we wont miss anything

**Comment by ImMorpheus** [comment]
```
org.spongepowered.common.registry.CatalogTypeMethodTest > testCatalogMethodImpl[87755 Catalog Type: Fish | Id : minecraft:raw.cod | Method: Fish#getCookedFish() | (net.minecraft.item.ItemFishFood$FishType)] STANDARD_OUT
    [Test worker/WARN]: id is not lowercase. Catalog Type: Fish | Id : cooked.COD | (net.minecraft.item.ItemFishFood$FishType)

org.spongepowered.common.registry.CatalogTypeMethodTest > testCatalogMethodImpl[87763 Catalog Type: Fish | Id : minecraft:raw.salmon | Method: Fish#getCookedFish() | (net.minecraft.item.ItemFishFood$FishType)] STANDARD_OUT
    [Test worker/WARN]: id is not lowercase. Catalog Type: Fish | Id : cooked.SALMON | (net.minecraft.item.ItemFishFood$FishType)
```
Something is not right.

EDIT: Ok, something is horribly wrong. Looks like the catalogtest was not testing every catalog type.
Thanks to @SimonFlash for making me realise that with his FireworkShape PR.

File: src/main/java/org/spongepowered/common/registry/type/item/InventoryArchetypeRegistryModule.java [hunk 35]
```diff
@@ -325,7 +325,7 @@ public void registerDefaults() {
             .with(new SpongeInventoryArchetypeBuilder()
                 .from(MENU_GRID)
                 .property(new InventoryDimensionImpl(5,3))
-                .build("horse_grid", "HorseGrid"))
+                .build("minecraft:horse_grid", "HorseGrid"))
```
ImMorpheus: @Faithcaio Please, review this one.

ST-DDT (approved):

src/main/java/org/spongepowered/common/registry/type/item/InventoryArchetypeRegistryModule.java
Faithcaio: Look good to me

ImMorpheus:
> all catalog types should also get a plugin id in the future, so maybe this is the time?

@Cybermaxke Can you clarify this ? 

Right now things are not consistent.
See
https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/registry/type/event/DismountTypeRegistryModule.java#L49

https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/registry/type/text/SelectorTypeRegistryModule.java#L59

https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/registry/type/event/SpawnTypeRegistryModule.java#L101

(The first one is not adding the prefix to the map)
What's the difference between `sponge:` and `minecraft:` (Yeah, I've seen `SpongeImplHooks#getModIdFromClass`) ? Should the prefix go in the map ? Do i have to handle it inside `getById` ?
Input wanted.

Cybermaxke:
Yeah, right now it's very inconsistent in the implementation. So maybe it's better to wait for the CatalogKey PR to do this right and avoid changing these things twice. Just fix it for the `NotePitch` and other catalog types that are currently causing errors.
The CatalogKey PR will also make the prefixing, etc. more consistent.
`minecraft` as plugin id is mostly used for things that you would be able to use in vanilla minecraft, and `sponge` for custom additions, no idea if it's properly used everywhere in the impl.

gabizou:
@Cybermaxke I'd rather we have this still fixed in stable-7 than wait for bleeding.

gabizou (approved):

src/main/java/org/spongepowered/common/registry/type/item/InventoryArchetypeRegistryModule.java
Faithcaio: actually would be sponge as the archetypes are smth. we add

ImMorpheus: I think I'll leave this as it is. I'd have to change https://github.com/ImMorpheus/SpongeCommon/blob/13cc5d531cf2437b83a009a5efff4db6c5dee703/src/main/java/org/spongepowered/common/registry/type/item/InventoryArchetypeRegistryModule.java#L214 too. Let's wait for the catalog key PR.

ImMorpheus:
Ok, the catalog key PR should make the prefixing consistent, thus fixing the warn about id and name being equal. I'll wait and do a new PR if needed. 
This should be good, just let me know if I have to silence the warnings in the catalog test for now.

gabizou:
@ImMorpheus if you can condense the commits down to a few instead of several. I know this is a lot of changes, but I'm inclined to attempt to cherry-pick a majority of these fixes to `stable-7`. But to answer your question: yes, go ahead and silence the current warnings on the tests as we'll be improving the tests come `CatalogKey` PR.

ImMorpheus:
@gabizou Done.

ImMorpheus:
~wip

src/test/java/org/spongepowered/common/registry/CatalogTypeMethodTest.java
```diff
@@ -102,68 +82,97 @@
...
+
+    private void testKeyAndName(CatalogType object) {
+        String name = object.getName();
+        CatalogKey key = object.getKey();
+
+        if (key.getValue().indexOf(':') != -1) {
+            LOG.warn("Duplicate namespace. Catalog Type: {} | Key: {} | Name: {} | ({})", object.getClass().getSimpleName(), key, name, this.implementationClass);
+        }
+
+        if (name.equals(key.toString())) { //Do we need this ?
```
ImMorpheus: Input wanted

ImMorpheus:
~qa

ImMorpheus:
Updated to reflect the CatalogKey PR.

PR #1801, opened 2018-02-13
Item TileEntity data

Faithcaio:
SpongeAPI | **SpongeCommon**

TODO

solves https://github.com/SpongePowered/SpongeAPI/issues/1781

### DIFF
--- a/src/main/java/org/spongepowered/common/data/util/NbtDataUtil.java
+++ b/src/main/java/org/spongepowered/common/data/util/NbtDataUtil.java
@@ -26,14 +26,22 @@ package org.spongepowered.common.data.util;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import net.minecraft.block.Block;
+import net.minecraft.block.ITileEntityProvider;
+import net.minecraft.block.state.IBlockState;
+import net.minecraft.item.ItemBlock;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.nbt.NBTTagDouble;
 import net.minecraft.nbt.NBTTagList;
+import net.minecraft.tileentity.TileEntity;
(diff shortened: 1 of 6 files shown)


src/main/java/org/spongepowered/common/data/util/NbtDataUtil.java
```diff
@@ -416,4 +426,37 @@ public static NBTTagList newDoubleNBTList(double... numbers) {
...
+        }
+
+        ItemBlock itemBlock = (ItemBlock) stack.getItem();
+        Block block = itemBlock.getBlock();
+        IBlockState state = block.getDefaultState();
+        if (block instanceof ITileEntityProvider) {
+            // TODO check if we really need a world here
+            TileEntity te = ((ITileEntityProvider) block).createNewTileEntity(null, block.getMetaFromState(state));
+
+            nbttagcompound.setInteger("x", -1);
```
gabizou, line 446: Instead of using the string constants, use the field constants already set up for you to avoid future portability issues.

gabizou (changes requested):
Gonna need to spend some time to debate this. While some *vanilla* `TileEntity` instances are directly inventory viewers, I don't want to blanket the idea of making these fake entities for all possible `TileEntity` instances. I'd much rather a whitelist approach, allowing the community to test the feature with modded `TE`'s, along with some GUI mods that might throw some nice exceptions on either the client or server for requesting a `TE` that doesn't exist.

That being said, there's some code here that can be used, but for the basis of representing an inventory, I'd lean on the "whitelist `TE` creation with a fallback to manual slotted inventory creation".

Zidane:
@Faithcaio Update this or close it for api-7.

Zidane:
@Faithcaio Is this still good? Needs to be redone for API-8.

Zidane:
Closing this unless a member from the community picks it back up.

PR #1808, opened 2018-02-22
Move Block Break Logic handling to after ChangeBlockEvents

gabizou:
While investigating SpongePowered/SpongeForge#1989, I discovered that we're still allowing block logic to be performed *while* tracking said logic.

Since this broke the fish net mod, I figured that we would also need to track the block break logic happening not during the phase state, but after the phase state is unwinding, or at least when block changes are being thrown as events.

This is mostly to gather some feedback on the potential resolution.


gabizou:
@Zidane @bloodmc I'll get to merging this tomorrow if no one replies. It should resolve the issue and shouldn't pose any further problems.

PR #1809, opened 2018-02-22
Fix worldarchetype builder spawnloaded npe

liach:
per zidane's request on irc.

### DIFF
--- a/src/main/java/org/spongepowered/common/interfaces/world/IMixinWorldSettings.java
+++ b/src/main/java/org/spongepowered/common/interfaces/world/IMixinWorldSettings.java
@@ -32,6 +32,8 @@ import org.spongepowered.api.world.difficulty.Difficulty;
 import org.spongepowered.api.world.PortalAgentType;
 import org.spongepowered.api.world.gen.WorldGeneratorModifier;
 
+import javax.annotation.Nullable;
+
 public interface IMixinWorldSettings {
 
     void setId(String id);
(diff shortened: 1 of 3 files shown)


phit:
https://github.com/SpongePowered/SpongeCommon/commit/008e5d5274ba435f5baac3e6b21b45b4be472203#diff-88d512170c6d39b0117572b34a5fddb2L226

this was changed for a reason, all dims besides overworld should default to false

liach:
should i change it to false then?

liach:
however, this change simply crashes plugins now.

Zidane:
I think setting it to false should be fine unless I am forgetting something that @phit  remembers.

phit:
setting it to false will also set it to false for the overworld, don't think that will not crash

why is this PR needed anyways?

Zidane:
@phit 

Well Overworld would be ignored of course. Its ignored in other areas already :p.

phit:
if what you are saying is correct, @bloodmc why was this special handling in your commit even necessary?

liach:
Ah then my fix is wrong. I will make a better fix then.

liach:
@phit I have corrected that fix now.

liach:
hmm git derped

liach:
This time it should be good @phit

liach:
Thanks blood!

liach:
@kashike Can you cherry pick this to bleeding?

PR #1811, opened 2018-02-23
move ItemStackComparators to impl

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1786) | **SpongeCommon**

### DIFF
--- a/src/main/java/org/spongepowered/common/registry/type/item/ItemStackComparatorRegistryModule.java
+++ b/src/main/java/org/spongepowered/common/registry/type/item/ItemStackComparatorRegistryModule.java
@@ -1,0 +1,114 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/registry/type/item/ItemStackComparatorRegistryModule.java
```diff
@@ -0,0 +1,132 @@
...
+
+        @Override
+        public int compare(@Nullable final ItemStack o1, @Nullable final ItemStack o2) {
+            if (o1 == null && o2 == null) {
+                return 0;
+            }
+            if (o1 == null) {
+                return 1;
+            }
+            if (o2 == null) {
```
liach: Why are nulls the biggest?

Faithcaio: copy paste
But i'll remove the null checks.
Itemstacks are not supposed to be null they are ItemStack.empty() now.

src/main/java/org/spongepowered/common/registry/type/item/ItemStackComparatorRegistryModule.java
```diff
@@ -0,0 +1,115 @@
...
+@SuppressWarnings("unchecked")
+public final class ItemStackComparatorRegistryModule implements RegistryModule {
+
+    @RegisterCatalog(ItemStackComparators.class)
+    private final Map<String, Comparator<ItemStack>> comparators = new HashMap<>();
+
+    @Override
+    public void registerDefaults() {
+
+        Comparator<ItemStack> TYPE = Comparator.comparing(i -> i.getType().getId());
```
kashike: why are these all capitals?

src/main/java/org/spongepowered/common/registry/type/item/ItemStackComparatorRegistryModule.java
```diff
@@ -0,0 +1,115 @@
...
+        this.comparators.put("ALL", type.thenComparing(size).thenComparing(properties).thenComparing(itemData));
+    }
+
+    private static final class Properties implements Comparator<ItemStack> {
+
+        @Override
+        public int compare(ItemStack o1, ItemStack o2) {
+            List<Property<?, ?>> properties = Lists.newArrayList(o2.getApplicableProperties());
+            for (Property<?, ?> property : o1.getApplicableProperties()) {
+                if (properties.contains(property)) {
```
liach: Since you call `contains` and `remove`, I'd say you should change `properties` to a `Set`.

src/main/java/org/spongepowered/common/registry/type/item/ItemStackComparatorRegistryModule.java
```diff
@@ -0,0 +1,115 @@
...
+
+        private final Class<? extends DataManipulator<?, ?>>[] ignored;
+
+        ItemDataComparator(Class<? extends DataManipulator<?, ?>>... ignored) {
+            this.ignored = ignored;
+        }
+
+        @Override
+        public int compare(ItemStack o1, ItemStack o2) {
+            List<DataManipulator<?, ?>> manipulators = Lists.newArrayList(o2.getContainers());
```
liach: Same for here. You might use a `LinkedHashSet` since you need to check everything in a set conditionally.

phit (commented):
testreview for tag oddness

gabizou (approved):

PR #1832, opened 2018-03-13
Resolve #1830

### LINKED ISSUE #1830: DataProcessorDelegate returns false for all support(EntityType)
**I am currently running**

- SpongeForge version: spongeforge-1.12.2-2611-7.1.0-BETA-2937
- SpongeVanilla version: 
- Java version: 
- Operating System:

- Plugins/Mods:

**Issue Description**

When attempting to make a snapshot of a entity, added manipulators are ignored as this always returns false.
https://github.com/SpongePowered/SpongeCommon/blob/72e33e8a59e16349b2ed74d0b247179427dc371c/src/main/java/org/spongepowered/common/data/util/DataProcessorDelegate.java#L66

This is also an issue in AbstractSingleDataSingleTargetProcessor, where this is not overridden and returns false by default.

It also seems that any standard manipulators are ignored when a snapshot is created (unless you explicatly readd them?)

clienthax:
#1830 1830

### DIFF
--- a/src/main/java/org/spongepowered/common/data/util/DataProcessorDelegate.java
+++ b/src/main/java/org/spongepowered/common/data/util/DataProcessorDelegate.java
@@ -81,8 +81,17 @@ public final class DataProcessorDelegate<M extends DataManipulator<M, I>, I exte
         return false;
     }
 
+    @SuppressWarnings("unused")
     @Override
     public boolean supports(EntityType entityType) {
+        final boolean callingFromMinecraftThread = ServerUtils.isCallingFromMainThread();
+        for (Tuple<DataProcessor<M, I>, Timing> tuple : this.processors) {
+            try (Timing timing = callingFromMinecraftThread ? tuple.getSecond().startTiming() : null) {
+                if (tuple.getFirst().supports(entityType)) {
+                    return true;
+                }
+            }
+        }
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/data/util/DataProcessorDelegate.java
```diff
@@ -83,6 +83,22 @@ public boolean supports(DataHolder dataHolder) {
 
     @Override
     public boolean supports(EntityType entityType) {
+        final boolean callingFromMinecraftThread = ServerUtils.isCallingFromMainThread();
+
+        for (Tuple<DataProcessor<M, I>, Timing> tuple : this.processors) {
+            if (callingFromMinecraftThread) {
```
gabizou: I would write it in a try like so:
```java
try (Timing timing = callingFromMainecraftThread ? tuple.getSecond().startTiming() : null) {
  if (tuple.getFirst().supports(entityType)) {
    return true;
  }
}
return false;

gabizou: This way the timing is auto-closing, in the event the processor royally breaks (there's lots of try with resources in sponge now due to their fancy handling).

gabizou (changes requested):
Itty bitty small change.

gabizou (approved):

parlough (approved):

PR #1835, opened 2018-03-17
Fix End portal platforms being created and nether portals being ignor…

Dockter:
…ed if teleporting from a nether world that isn't DIM-1.

Fixes: https://github.com/SpongePowered/SpongeForge/issues/2050

Signed-off-by: Dockter <dockter@almuramc.com>


Dockter:
Cannot assume that nether portal creations will always take place to/from DIM-1

PR #1852, opened 2018-03-30
Inventory Transformations

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1800) | **SpongeCommon**

predefined Transformations to order and filter inventories.

### DIFF
--- a/src/main/java/org/spongepowered/common/item/inventory/query/SpongeQueryTransformation.java
+++ b/src/main/java/org/spongepowered/common/item/inventory/query/SpongeQueryTransformation.java
@@ -1,0 +1,56 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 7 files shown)


src/main/java/org/spongepowered/common/item/inventory/query/SpongeQueryTransformation.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.List;
+
+public class SpongeQueryTransformation implements Transformation {
+
+    private final String id;
+    private final List<QueryOperation> operations;
+
+    public SpongeQueryTransformation(String id, List<QueryOperation> operations) {
+
```
kashike: whitespace

src/main/java/org/spongepowered/common/item/inventory/query/SpongeQueryTransformation.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public Inventory transform(Inventory inventory) {
+        if (this.operations.isEmpty()) {
+            return inventory;
+        }
+        Inventory result = new EmptyInventoryImpl(inventory);
+        for (QueryOperation operation : this.operations) {
+            result = result.union(inventory.query(operation));
+        }
+        return result;
```
SimonFlash, line 54: This should fall back to the given inventory so that slots aren't lost. If the operations only query for a portion of the inventory (or if they're not found) this will have unexpected behavior.

Faithcaio: I don't think so.
This is supposed to be able to query for parts of the inventory.
If they are not found you'll get no additional slots. Thats not unexpected.

SimonFlash: Right. So, if I transform for the Hotbar from the player's storgage inventory, I'll be given back an empty inventory because the query 'failed'. Unioning that with another empty inventory won't change that, so I'd be given back an empty inventory as a result - *loosing* all of the slots in the player's storage inventory. The only way to be sure that this can't happen is by making assumptions about the inventory your given or querying for these beforehand, which defeats the point.

I'm saying that transform, as the name implies, should only reorder the slots. If we transform for the hotbar first, it should push all slots that are in the hotbar to the front of the inventory, leaving the remaining slots at the end. This will not do that because there are slots left out if the query operations account for a subsection of the inventories slots. Your result must be unioned with the given inventory to fix this.

Faithcaio: No. This is transforming not just sorting.

The predefined HOTBAR_FIRST returns the slots of queries to
`Hotbar` and `MainPlayerInventory` in that order.

If neither is present in the inventory it will return an empty inventory. 
e.g. when transforming a ChestInventory.

Leaving the ChestInventory 
or Armor/Offhand Slots 
or in case of Container addionally the CraftingGrid and Output too 
in the resulting inventory makes no sense.

If you want that behaviour just union the transform result with the original inventory.

PR #1858, opened 2018-04-04
Fix vanilla item frame dupe - MC-124833

parlough:
Making a PR for this, so it's easier for me to track.

@clienthax


phit:
why this instead of just bumping forge to latest, https://github.com/MinecraftForge/MinecraftForge/commit/bad83a303e6770d484aad0b0d891549dc273cf05

clienthax:
@phit because this would cover spongevanilla too?

parlough:
Yeah and I also had the intention of getting it out to other users who are more willing to update Sponge than Forge, such as pack users.

PR #1869, opened 2018-04-12
delay opening/closing inventories during an InteractInventoryEvent

Faithcaio:
Opening and closing an inventory during an event prevents us from correctly handling modified transactions and event cancel.

This PR delays open/close during InteractInventoryEvent to the next tick and also prints a warning.

In API 8 we may want to instead throw an exception 
and add a method on the events so you do not have to schedule the action yourself.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
@@ -691,6 +691,7 @@ public abstract class MixinEntityPlayerMP extends MixinEntityPlayer implements P
         // Create Close_Window to capture item drops
         try (PhaseContext<?> ctx = PacketPhase.General.CLOSE_WINDOW.createPhaseContext()
                 .source(this)
+                .packetPlayer(((EntityPlayerMP)(Object) this))
                 .openContainer(this.openContainer)
                 // intentionally missing the lastCursor to not double throw close event
                 .buildAndSwitch()) {
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
@@ -244,6 +244,8 @@ public abstract class MixinEntityPlayerMP extends MixinEntityPlayer implements P
     @Shadow public int currentWindowId;
     @Shadow private void getNextWindowId() { }
 
+    @Shadow public abstract void closeContainer();
+
     public int newExperience = 0;
     public int newLevel = 0;
     public int newTotalExperience = 0;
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -649,12 +651,24 @@ public void restorePacketItem(EnumHand hand) {
 
     @Override
     public Optional<Container> openInventory(Inventory inventory) throws IllegalArgumentException {
-        return Optional.ofNullable((Container) SpongeCommonEventFactory.displayContainer((EntityPlayerMP) (Object) this,
-                inventory));
+        if (((IMixinContainer) this.openContainer).isInUse()) {
+            SpongeImpl.getLogger().warn("This player is currently modifying an open container. This action will be delayed.");
+            Sponge.getScheduler().createTaskBuilder().delayTicks(0).execute(() -> {
+                this.closeInventory(); // Cause close event first. So cursor item is not lost.
```
gabizou: Should we perform any cause stack modifications here since this is performed a tick later, and more than likely the task would be redirected as sponge causing the closure and opening of the inventory?

gabizou (commented):
Is it possible to have a test plugin for this use case to verify there aren’t any issues with closing the inventory or opening the inventory outside of the task?

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -649,12 +653,36 @@ public void restorePacketItem(EnumHand hand) {
...
     public Optional<Container> openInventory(Inventory inventory) throws IllegalArgumentException {
-        return Optional.ofNullable((Container) SpongeCommonEventFactory.displayContainer((EntityPlayerMP) (Object) this,
-                inventory));
+        if (((IMixinContainer) this.openContainer).isInUse()) {
+            Cause cause = Sponge.getCauseStackManager().getCurrentCause();
+            SpongeImpl.getLogger().warn("This player is currently modifying an open container. This action will be delayed.");
+            Sponge.getScheduler().createTaskBuilder().delayTicks(0).execute(() -> {
+                try (StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
+                    cause.all().forEach(frame::pushCause);
+                    cause.getContext().asMap().forEach((key, value) -> frame.addContext(((EventContextKey) key), value));
```
gabizou, line 667: If possible, I'd prefer to have this styled in that each stream call is a new line, only for future possible debugging possibilities (it helps being able to select a line instead of stepping over a few calls and missing it).

testplugins/src/main/java/org/spongepowered/test/InventoryOpenCloseTest.java
```diff
@@ -0,0 +1,74 @@
...
+                    } else {
+                        this.registered = true;
+                        Sponge.getEventManager().registerListeners(this, this.listener);
+                    }
+                    return CommandResult.success();
+                }).build(), "toggleInventoryopenclosetest");
+    }
+
+    public static class InventoryListener {
+
```
gabizou: Is it possible to have a test listener to print out "inventory closed by plugin" to verify the plugin is causing the closing for the event?

gabizou (changes requested):
small changes, the inventory test plugin I'm genuinely feeling stronger about being able to verify via plugin listeners that this works 100%

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
Faithcaio: I dont quite get what you mean.
There is only one forEach per line there + method reference/lambda

testplugins/src/main/java/org/spongepowered/test/InventoryOpenCloseTest.java
Faithcaio: done

gabizou (approved):

ryantheleach:
@Faithcaio Has this bug been tested against SV? https://github.com/SpongePowered/SpongeForge/issues/2116 claims that it's only a problem in SF.

BrainStone:
Definitely not happening on SV 38 (Can't test 39 as it contains the bug with the ItemStackComparators and no SV build has been pushed so far)

Anyways, I'd apprechiate the merge!

PR #1882, opened 2018-04-20
Refactor Entity Drop management for PhaseTracking

gabizou:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1808) | **SpongeCommon** | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/2110)

This is mostly to improve compatibility with forge mods, but primarily this also cleans up a lot of the management and events of item drops from entities and players. 

For the moment, just keeping the PR open for developer feedback (likely won't occur much, but still good to have some sort of discussion).


src/main/java/org/spongepowered/common/entity/EntityUtil.java
```diff
@@ -503,6 +503,27 @@ public static int getHorseInternalVariant(SpongeHorseColor color, SpongeHorseSty
         return color.getBitMask() | style.getBitMask();
     }
 
+    public static void processEntitySpawnsFromEvent(PhaseContext<?> context, SpawnEntityEvent destruct) {
```
Aaron1011: Did you mean to call this parameter `destruct`?

gabizou: It auto-populated when refactoring to use the Destruct event.

gabizou:
@kashike @Zidane @bloodmc Please review this commit: https://github.com/SpongePowered/SpongeCommon/commit/db5467ccf1900eaed02da46b051732f799e7c7ec

I'm hesitant to have this sort of hackery to accommodate DraconicEvolution (due to how it breaks blocks, checks for dropped entities from nearby blocks, then "captures them" and adds them to their special item drop. It's partially on it's way to fixing the interaction problem, but there's more to check for drop captures in this method.

PR #1887, opened 2018-04-22
Move the BlockState matcher impl to impl

Cybermaxke:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1811) | **Common**


gabizou (approved):

14mRh4X0r (changes requested):
In `SpongeBlockStateMatcherBuilder#build()`, the ` - 1` part should be removed. Currently, if no traits are specified, this will throw a `NegativeArraySizeException`. See also SpongePowered/SpongeAPI#1835.

PR #1895, opened 2018-04-29
Implement ChangeEntityEquipmentEvents

ustc-zzzz:
[SpongeAPI](//github.com/SpongePowered/SpongeAPI/pull/1819) | **SpongeCommon**

The purpose of this PR is the same as what is expressed in the title.

There are only two questions I found when I was implementing the events:
* What do the `Optional`s mean in the return values of [`getOriginalItemStack`](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/event/entity/ChangeEntityEquipmentEvent.java#L66) and [`getItemStack`](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/event/entity/ChangeEntityEquipmentEvent.java#L76) of the event interface? I didn't see the need for the existence. The two `Optional`s could never be empty.
* All the implementations of those `Lens` (`OrderedInventoryLensImpl`, `EquipmentInventoryLensImpl`) are based on the `IInventory` while general creatures (such as zombies, skeletons, etc.) never have their inventories so I cannot reuse those codes. Is it needed to extract generics in those implementations?

Besides, would this PR be merged into the `stable` branch and be available in the artifacts of stable builds?

### DIFF
--- a/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
@@ -1423,19 +1423,22 @@ public class SpongeCommonEventFactory {
 
     public static ChangeEntityEquipmentEvent callChangeEntityEquipmentEvent(EntityLivingBase entity, ItemStackSnapshot before, ItemStackSnapshot after, SlotAdapter slot) {
         ChangeEntityEquipmentEvent event;
-        Cause currentCause = Sponge.getCauseStackManager().getCurrentCause();
-        Transaction<ItemStackSnapshot> transaction = new Transaction<>(before, after);
-        if (entity instanceof EntityPlayerMP) {
-            Player player = EntityUtil.toPlayer((EntityPlayerMP) entity);
-            event = SpongeEventFactory.createChangeEntityEquipmentEventTargetPlayer(currentCause, player, slot, transaction);
-        } else if (entity instanceof Human) {
-            Human humanoid = (Human) entity;
-            event = SpongeEventFactory.createChangeEntityEquipmentEventTargetHumanoid(currentCause, humanoid, slot, transaction);
-        } else {
-            Living living = EntityUtil.fromNativeToLiving(entity);
-            event = SpongeEventFactory.createChangeEntityEquipmentEventTargetLiving(currentCause, living, slot, transaction);
+        try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
+            frame.pushCause(entity);
+            Cause cause = frame.getCurrentCause();
+            Transaction<ItemStackSnapshot> transaction = new Transaction<>(before, after);
+            if (entity instanceof EntityPlayerMP) {
+                Player player = EntityUtil.toPlayer((EntityPlayerMP) entity);
+                event = SpongeEventFactory.createChangeEntityEquipmentEventTargetPlayer(cause, player, slot, transaction);
+            } else if (entity instanceof Human) {
+                Human humanoid = (Human) entity;
+                event = SpongeEventFactory.createChangeEntityEquipmentEventTargetHumanoid(cause, humanoid, slot, transaction);
+            } else {
+                Living living = EntityUtil.fromNativeToLiving(entity);
+                event = SpongeEventFactory.createChangeEntityEquipmentEventTargetLiving(cause, living, slot, transaction);
+            }
+            SpongeImpl.postEvent(event);
+            return event;
         }
-        SpongeImpl.postEvent(event);
-        return event;
     }
 }
--- a/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
@@ -71,6 +71,7 @@ import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.entity.Item;
 import org.spongepowered.api.entity.Transform;
 import org.spongepowered.api.entity.living.Agent;
+import org.spongepowered.api.entity.living.Human;
 import org.spongepowered.api.entity.living.Living;
 import org.spongepowered.api.entity.living.player.Player;
 import org.spongepowered.api.entity.living.player.User;
@@ -86,6 +87,7 @@ import org.spongepowered.api.event.cause.Cause;
 import org.spongepowered.api.event.cause.EventContext;
 import org.spongepowered.api.event.cause.EventContextKeys;
 import org.spongepowered.api.event.cause.entity.spawn.SpawnTypes;
+import org.spongepowered.api.event.entity.ChangeEntityEquipmentEvent;
 import org.spongepowered.api.event.entity.CollideEntityEvent;
 import org.spongepowered.api.event.entity.DestructEntityEvent;
 import org.spongepowered.api.event.entity.InteractEntityEvent;
@@ -137,6 +139,7 @@ import org.spongepowered.common.interfaces.entity.player.IMixinEntityPlayerMP;
 import org.spongepowered.common.interfaces.entity.player.IMixinInventoryPlayer;
 import org.spongepowered.common.interfaces.world.IMixinWorldServer;
 import org.spongepowered.common.item.inventory.adapter.InventoryAdapter;
+import org.spongepowered.common.item.inventory.adapter.impl.slots.SlotAdapter;
 import org.spongepowered.common.item.inventory.custom.CustomInventory;
 import org.spongepowered.common.item.inventory.util.ContainerUtil;
 import org.spongepowered.common.item.inventory.util.InventoryUtil;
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -38,10 +38,7 @@
 import net.minecraft.entity.passive.AbstractHorse;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.player.EntityPlayerMP;
-import net.minecraft.inventory.Container;
-import net.minecraft.inventory.ContainerPlayer;
-import net.minecraft.inventory.IInventory;
-import net.minecraft.inventory.Slot;
+import net.minecraft.inventory.*;
```
gabizou: Don't use star imports.

gabizou:
To answer your first question, I believe we had to use `Optional`s when we [originally wrote the event almost four years ago](https://github.com/SpongePowered/SpongeAPI/commit/57d70e3ccbde63583c3ee58bbf5059a249ac8e84#diff-b89992c26f79fcabf6df2bc364ee56f8R44). Part of the problem at the time was that we did have possible `null` `ItemStack`s, until 1.11, when it was possible to only have empty item stacks. This would be an API change to implement it correctly.

parlough:
We can make the API change later in bleeding/1.13. Could you rebase this on top of stable-7 and change the target since this doesn't break anything in its current state and would be nice to get into 7.

ustc-zzzz:
I'm considering about the things related to the API and realized that it is not a problem which can be solved easily.

1. The most convenient solution is to delete these two methods directly and add a new method returning `Transaction<ItemStack>` but it will break the backward compatibility with API7 (you cannot call the two methods in the future because they are deleted). However, maybe nobody is actually using this event because it has not been implemented yet. Is breaking the backward compatibility acceptable here?

2. It sounds acceptable that we can just mark those two method as deprecated and add a method with a default implementation like this:
```java
// before
@GenerateFactoryMethod
public interface ChangeEntityEquipmentEvent extends TargetEntityEvent, TargetInventoryEvent, Cancellable {
    Optional<ItemStackSnapshot> getOriginalItemStack();
    Optional<Transaction<ItemStackSnapshot>> getItemStack();
    @Override Slot getTargetInventory();
    interface TargetLiving extends ChangeEntityEquipmentEvent, TargetLivingEvent { }
    interface TargetHumanoid extends TargetLiving, TargetHumanoidEvent { }
    interface TargetPlayer extends TargetHumanoid, TargetPlayerEvent { }
}

// after
@GenerateFactoryMethod
public interface ChangeEntityEquipmentEvent extends TargetEntityEvent, TargetInventoryEvent, Cancellable {
    @Deprecated Optional<ItemStackSnapshot> getOriginalItemStack();
    @Deprecated Optional<Transaction<ItemStackSnapshot>> getItemStack();
    @PropertySettings(generateMethods = false, requiredParameter = false)
    default Transaction<ItemStackSnapshot> getTransaction() {
        @SuppressWarnings("deprecation")
        Optional<Transaction<ItemStackSnapshot>> optional = this.getItemStack();
        return optional.orElseThrow(() -> new IllegalStateException("No transaction available!"));
    }
    @Override Slot getTargetInventory();
    interface TargetLiving extends ChangeEntityEquipmentEvent, TargetLivingEvent { }
    interface TargetHumanoid extends TargetLiving, TargetHumanoidEvent { }
    interface TargetPlayer extends TargetHumanoid, TargetPlayerEvent { }
}
```
It does not break anything, but we should still wrap two objects into `Optional`s and pass them into method `SpongeEventFactory.createChangeEntityEquipmentEventXXX` to create and fire the event, then unwrap the `Optional` in method `getTransaction`. The performance overhead is unnecessary.

3. Of course we can also temporarily mark them deprecated at API7 and remove these two methods at API8, but in that case all plugins using this event cannot be compatible with API7 and API8 at the same time. The reason is that method `SpongeEventFactory.createChangeEntityEquipmentEventXXX` would be changed. In API7 it should accept two `Optional`s but it will only accept a `Transaction<ItemStack>` in API8.

Finally, I want to paste a paragraph here from SpongeDocs:
> Most plugins developed with the Sponge API should work across several different versions of Minecraft without needing to be updated. This means that, for the most part, server owners do not need to worry about plugin incompatibilities.

API7 breaks backward compatibility in many places. I hope that API8 will not be like this again.

parlough:
You present some good ideas, the example you suggest in 2 could work well, I'd like to see what faith thinks. At some point though we will have to remove those deprecated methods and implementers of the interface will have to switch from implementing the old methods to the transaction method and the event factory will break. 

The part you pull from the docs is a real goal, but with the scope of 1.13 and other changes we plan to make, much is unfortunately going to break. It's our chance to remove all mistakes and similar from the API so we don't have to do it again in a later API release. There is pretty much a 100% chance developers will have to update their plugins. I'm thinking we implement your #2 deprecating the original methods for removal in 8. This gives people the chance to migrate to the new method while they're still using API 7.1(which will be for quite a while) and only those who are using the event factory or implementing the event manually (which is probably very very few people) will be hit in API 8.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntityLivingBase.java
```diff
@@ -845,6 +861,59 @@ private void causeTrackDeathUpdate(EntityLivingBase entityLivingBase) {
...
+                slotAdapter = slotLens.getAdapter(inventory.getFabric(), inventory);
+            } else {
+                if (this.slotLens.isEmpty()) {
+                    for (EntityEquipmentSlot slot : EntityEquipmentSlot.values()) {
+                        this.slotLens.put(slot, new SlotLensImpl<>(slot.getSlotIndex()));
+                    }
+                }
+                slotAdapter = this.slotLens.get(entityEquipmentSlot).getAdapter(this.inventory, null);
+            }
+            ChangeEntityEquipmentEvent event = SpongeCommonEventFactory.callChangeEntityEquipmentEvent(entity,
```
Faithcaio, line 945: For non-players this is missing the cause.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntityLivingBase.java
```diff
@@ -845,6 +861,59 @@ private void causeTrackDeathUpdate(EntityLivingBase entityLivingBase) {
...
+            if (entity instanceof EntityPlayerMP) {
+                SlotLens<IInventory, ItemStack> slotLens;
+                IMixinInventoryPlayer inventory = (IMixinInventoryPlayer) ((EntityPlayerMP) entity).inventory;
+                PlayerInventoryLens inventoryLens = (PlayerInventoryLens) inventory.getRootLens();
+                switch (entityEquipmentSlot) {
+                case OFFHAND:
+                    slotLens = inventoryLens.getOffhandLens();
+                    break;
+                case MAINHAND:
+                    HotbarLens<IInventory, ItemStack> hotbarLens = inventoryLens.getMainLens().getHotbar();
```
Faithcaio: We might not want an event for the player main-hand here.
All cases I can think of are covered by other events:
ChangeInventoryEvent.Held ; UseItemStackEvent ; ChangeInventoryEvent.Pickup.

In case of the held Slot changing this event will even be wrong because the item in the slot did not actually change. Only the selected Slot did.

Faithcaio (changes requested):
We should probably make those non-player Equipment Inventories available in the API at some point. (But thats stuff for another PR)

Please also add a TestPlugin (listener off by default see other testplugins:  https://github.com/SpongePowered/SpongeCommon/blob/66e7d7b7c2bd58896ba3861f6b86d5253209732d/testplugins/src/main/java/org/spongepowered/test/InventoryOpenCloseTest.java)
Just print out cause and item change or similar.

ryantheleach:
@ustc-zzzz I noticed you closed this PR, even though it was stalled and there was feedback waiting to be addressed.

Do you plan on finishing this in a different PR? 

If Not, What work remains apart from the feedback if someone were to try to pick this up and finish it?

Why was it closed (Just curious)?

Is it possible to target Stable?

ustc-zzzz:
@ryantheleach The reason is that I accidentally deleted this branch...I will recover the branch as soon as possible. Sorry for misunderstanding.

ryantheleach:
No problem, I was worried we had accidentally alienated you. But in future, I recommend using a `feature branch` when working on pull requests. It solves a lot of pain, and lets you limit contributors to Sponge working on your PR to only that branch, instead of messing with your local 'bleeding'

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntityLivingBase.java
ustc-zzzz: @Faithcaio There is another special case. None of the event mentioned above will be fired when a mod just change the `ItemStack` on main hand quietly (such as modifying NBTs, etc.), but `ChangeEntityEquipmentEvent` need to take over this change. Maybe we need to think again carefully about what to do.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntityLivingBase.java
ustc-zzzz: Solved in 210d5dc36ab35f20e858e39f1f9a64d6d5ec7886

ImMorpheus:
@ustc-zzzz Is this ready for review ?

alimtunc:
Any news for this PR ? When it will be merged ?

ustc-zzzz:
Well, it’s time to clean up these messes. @Faithcaio I have just added a testing plugin and rebased to the latest commit of the target branch. What else should I do before the PR is able to be merged?

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntityLivingBase.java
```diff
@@ -895,6 +907,59 @@ private void causeTrackDeathUpdate(EntityLivingBase entityLivingBase) {
...
+        ItemStack after = this.getItemStackFromSlot(entityEquipmentSlot);
+        EntityLivingBase entity = EntityUtil.toNative((IMixinEntityLivingBase) this);
+        if (!ItemStack.areItemStacksEqual(after, before)) {
+            InventoryAdapter slotAdapter;
+            if (entity instanceof EntityPlayerMP) {
+                SlotLens slotLens;
+                IMixinInventoryPlayer inventory = (IMixinInventoryPlayer) ((EntityPlayerMP) entity).inventory;
+                PlayerInventoryLens inventoryLens = (PlayerInventoryLens) inventory.getRootLens();
+                switch (entityEquipmentSlot) {
+                case OFFHAND:
```
ImMorpheus: Indentation

ustc-zzzz: @ImMorpheus Fixed in 5945012cb58ee83c4f235e90209ab464466d1ead

Faithcaio:
https://github.com/SpongePowered/SpongeCommon/commit/62ecb0f95577bda284e63540cb227cd3d368782a

PR #1905, opened 2018-05-08
Include both custom and vanilla DataManipulators in BlockSnapshots.

### LINKED ISSUE #2416: Getting sign lines from block break event returns null
In a break block event, I try to get the data of a sign, I check if its a wall sign, it is, so it continues, pastes all of the sign data into console:
`[13:19:50 WARN] [test]: {"ContentVersion":1,"WorldUuid":"0e8d1aa1-f7db-411b-b732-7027b0444aa9","Position":{"X":549,"Y":82,"Z":-45},"BlockState":{"ContentVersion":2,"BlockState":"minecraft:wall_sign[facing=south]"},"UnsafeData":{"x":549,"Text4":"{\"text\":\"defutyh\"}","y":82,"Text3":"{\"text\":\"dirt 10\"}","z":-45,"Text2":"{\"text\":\"stone 1\"}","id":"minecraft:sign","Text1":"{\"text\":\"[Shop]\"}"}}`

Yet when I try to retrieve the data using keys, it gives a no such element exception. I would expect it to return the first sign line to console.
```java
for ( Transaction<BlockSnapshot> block : event.getTransactions())
{
    if (!block.getOriginal().getState().getType().equals(BlockTypes.WALL_SIGN))
    {
        continue;
    }
    logger.warn(DataFormats.JSON.write(block.getOriginal().toContainer()));
    logger.warn(block.getOriginal().get(Keys.SIGN_LINES).get().get(0).toPlain());
}
```
To reproduce:
Add this code to the block break event listener,
Go into the server and break a sign, should get an error in console.

SpongeVanilla 1.12.2

JBYoshi:
> "Include ALL the manipulators!" - Quote from someone or something, or maybe no one

Fixes https://forums.spongepowered.org/t/getting-sign-lines-from-changeblockevent-break/23461.
Fixes #2416.

I'm worried about side effects, though (for example, if the custom data gets saved off and conflicts with NBT data), so this may need quite a bit of testing.

### DIFF
--- a/src/main/java/org/spongepowered/common/block/SpongeBlockSnapshotBuilder.java
+++ b/src/main/java/org/spongepowered/common/block/SpongeBlockSnapshotBuilder.java
@@ -50,7 +50,6 @@ import org.spongepowered.api.world.BlockChangeFlags;
 import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
 import org.spongepowered.api.world.storage.WorldProperties;
-import org.spongepowered.common.bridge.data.CustomDataHolderBridge;
 import org.spongepowered.common.data.persistence.NbtTranslator;
 import org.spongepowered.common.data.util.DataUtil;
 import org.spongepowered.common.util.Constants;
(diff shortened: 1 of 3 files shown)


gabizou:
The original reason why we used only custom manipulators at the time for `BlockSnapshot`s was over a discussion for performance when retrieving the manipulators for block states and tile entities and the spam that could be affected.

Can introduce some more performant code for block states such that on block state creation they already have all their manipulators and key values assigned (doesn't really increase load time since they'll be done before the game is even loaded).

gabizou:
This won't actually fix any of the issues originally requested about getting Data from a `TileEntity` as if a `BlockSnapshot` still had a `TileEntity` in place. Due to the methods being used to efficiently clone as much data into a `BlockSnapshot` for phase tracking, none of these methods will populate the DataManipulators as you'd expect. The resolution is still to implement NbtProcessors.

PR #1917, opened 2018-05-19
Fire ExpireEntityEvent for Endermites and Lightning Bolts

Zerthick:
Partial bugfix for https://github.com/SpongePowered/SpongeCommon/issues/1583.  Endermites and Lightning Bolts should now fire an ExpireEntityEvent when expiring.

To address this issue, the endermite mixin has been modified to inject a callback into the `onLivingUpDate()` method responsible for causing the endermite to expire.  The callback fires an ExpireEntityEvent when this occurs, with the endermite entity present on the cause stack.

The LightningBolt mixin has also been modified to fire an ExpireEntityEvent, I was unsure as to when in the lightning bolt lifecycle  the event should be fired.  Clearly after `LightningEvent.Pre` and `LightningEvent.Strike` as both are cancellable.  Currently I have it fire before `LightningEvent.Post`, though this might not be the appropriate location.

This fix required a modification to the Sponge API to generate a factory method for ExpireEntityEvent, the corresponding PR for that change can be found here: https://github.com/SpongePowered/SpongeAPI/pull/1833

Further investigation needs to be done into if Endermites or LightningBolts correctly fire a DestructEntityEvent when expiring as well.


src/main/java/org/spongepowered/common/mixin/core/entity/monster/MixinEntityEndermite.java
```diff
@@ -62,4 +70,15 @@ public void supplyVanillaManipulators(List<DataManipulator<?, ?>> manipulators)
         super.supplyVanillaManipulators(manipulators);
         manipulators.add(getExpirableData());
     }
+
+    @Inject(method = "onLivingUpdate",
+            at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/monster/EntityEndermite;setDead()V"))
+    private void fireExpireEventLifetime(CallbackInfo ci) {
+        try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
+            Sponge.getCauseStackManager().pushCause(this);
```
parlough: You can push this directly to the frame so you don't have to pop it manually.

Zerthick:
Attached is a plugin to test these changes. It consists of the following code:
```
package io.github.zerhtick.spongetest;

import com.google.inject.Inject;
import org.slf4j.Logger;
import org.spongepowered.api.event.entity.ExpireEntityEvent;
import org.spongepowered.api.event.game.state.GameStartedServerEvent;

@Plugin(
        id = "spongetest",
        name = "SpongeTest",
        description = "A simple testbed plugin",
        authors = {
                "Zerthick"
        }
)
public class SpongeTest {

    @Inject
    private Logger logger;

    @Listener
    public void onExpireEntity(ExpireEntityEvent event) {
        logger.info("ExpireEntityEventFired!");
        logger.info("Event Cause: " + event.getCause());
    }
}
```

[SpongeTest-1.0.0.zip](https://github.com/SpongePowered/SpongeCommon/files/2019783/SpongeTest-1.0.0.zip)

src/main/java/org/spongepowered/common/mixin/core/entity/monster/MixinEntityEndermite.java
Zerthick: Ah, that makes sense. I'll update that accordingly

parlough:
Can you update this?

Zerthick:
@Meronat What needs to be updated?

gabizou (approved):

parlough:
@Zerthick Nothing anymore, I fixed the merge conflict. Thanks for the PR!

Zerthick:
@Meronat Ah, just as well, I don't have access to a computer atm which would make fixing the merge conflict a bit difficult. Do note there is a separate issue present here where a DestructEntityEvent is not fired, which I have not yet addressed. Feels good to finally have a pr in :)

PR #1918, opened 2018-05-20
Replace absolute module reference with relative reference.

gravityfox:
Fixes issues when SF and SV are NOT the root project.


gravityfox:
actually hold this thought, need to try something first.

PR #1926, opened 2018-05-24
Change SkinData to use a custom ProfileProperty

Aaron1011:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1836) | **SpongeCommon**

This PR implements the changes in the corresponding Sponge API PR.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/util/DataQueries.java
+++ b/src/main/java/org/spongepowered/common/data/util/DataQueries.java
@@ -126,12 +126,22 @@ public final class DataQueries {
     public static final DataQuery ENTITY_TYPE = of("EntityType");
     public static final DataQuery ENTITY_ROTATION = of("Rotation");
     public static final DataQuery ENTITY_SCALE = of("Scale");
-    public static final DataQuery SKIN_UUID = of("SkinUUID");
+    //public static final DataQuery SKIN_UUID = of("SkinUUID");
+    public static final DataQuery SKIN = of("Skin");
+    public static final DataQuery UPDATE_TAB_LIST = of("UpdateTabList");
 
     // User
     public static final DataQuery USER_UUID = of("UUID");
     public static final DataQuery USER_NAME = of("Name");
     public static final DataQuery USER_SPAWNS = of("Spawns");
+
+    // GameProfile
+    public static final DataQuery PROFILE_PROPERTIES = of("ProfileProperties");
+
+    // ProfileProperty
+    public static final DataQuery PROPERTY_NAME = of("PropertyName");
+    public static final DataQuery PROPERTY_VALUE = of("PropertyValue");
+    public static final DataQuery PROPERTY_SIGNATURE = of("PropertySignature");
 
     // ItemStacks
     public static final DataQuery ITEM_COUNT = of("Count");
@@ -215,12 +225,12 @@ public final class DataQueries {
         }
 
     }
-    
+
     public static final class Schematic {
-        
+
         public static final DataQuery VERSION = of("Version");
         public static final DataQuery METADATA = of("Metadata");
-        
+
         public static final DataQuery WIDTH = of("Width");
         public static final DataQuery HEIGHT = of("Height");
         public static final DataQuery LENGTH = of("Length");
@@ -229,12 +239,12 @@ public final class DataQueries {
         public static final DataQuery PALETTE = of("Palette");
         public static final DataQuery PALETTE_MAX = of("PaletteMax");
         public static final DataQuery BLOCK_DATA = of("BlockData");
-        
+
         public static final DataQuery TILEENTITY_DATA = of("TileEntities");
         public static final DataQuery TILEENTITY_POS = of("Pos");
 
         public static final DataQuery LEGACY_MATERIALS = of("Materials");
-        
+
         public static final DataQuery LEGACY_OFFSET_X = of("WEOffsetX");
         public static final DataQuery LEGACY_OFFSET_Y = of("WEOffsetY");
         public static final DataQuery LEGACY_OFFSET_Z = of("WEOffsetZ");
@@ -244,7 +254,7 @@ public final class DataQueries {
         public static final DataQuery LEGACY_ADD_BLOCKS = of("AddBlocks");
 
         public static final DataQuery LEGACY_TILEDATA = of("TileEntities");
-        
+
         private Schematic() {
         }
     }
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
@@ -30,6 +30,8 @@ import static com.google.common.base.Preconditions.checkState;
 
 import com.flowpowered.math.vector.Vector3d;
 import com.flowpowered.math.vector.Vector3i;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 import net.minecraft.advancements.PlayerAdvancements;
 import net.minecraft.block.state.IBlockState;
@@ -37,6 +39,7 @@ import net.minecraft.command.ICommandSender;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.EntityList;
 import net.minecraft.entity.EntityLivingBase;
+import net.minecraft.entity.EntityTrackerEntry;
 import net.minecraft.entity.IMerchant;
 import net.minecraft.entity.ai.attributes.AttributeMap;
 import net.minecraft.entity.ai.attributes.IAttributeInstance;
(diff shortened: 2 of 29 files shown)


src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -1375,4 +1394,119 @@ private ITextComponent onGetDisplayName4(IMerchant villager) {
...
+        if (other != null) {
+            this.setLocation(Sponge.getServer().getWorld(other.getUniqueId()).get().getSpawnLocation());
+            this.setLocationAndRotation(loc, rotation);
+        }*/
+    }
+
+    // Adapted from MixinPlayerList#recreatePlayerEntity
+    // We want to respawn the player *only* on tts own client. Therefore,
+    // we skip most of the logic that's performed for a normal respawn.
+    private void fakeRespawn() {
```
Aaron1011, line 1775: This method is pretty hacky, but I don't believe it should cause any problems. The respawning all happens within one server tick, and should be completely transparent to plugins.

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -1375,4 +1394,119 @@ private ITextComponent onGetDisplayName4(IMerchant villager) {
...
+
+    private void updateSkin() {
+        this.updateSkinSelf();
+        this.updateSkinOthers();
+    }
+
+    private void updateSkinSelf() {
+        this.getTabList().removeEntry(this.getUniqueId());
+        this.getTabList().addEntry(TabListEntry.builder()
+                .displayName(this.getDisplayNameData().displayName().get())
```
XakepSDK: Does this will work with custom plugins?
I think, old entry values should be used.

XakepSDK: And what if self entry was removed by some plugin?

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -1375,4 +1394,119 @@ private ITextComponent onGetDisplayName4(IMerchant villager) {
...
+        for (Object potioneffect : this.getActivePotionEffects()) {
+            this.connection.sendPacket(new SPacketEntityEffect(this.getEntityId(), (PotionEffect) potioneffect));
+        }
+        this.refreshScaledHealth();
+
+    }
+
+    private void updateSkinOthers() {
+        this.offer(Keys.VANISH, true);
+        Sponge.getScheduler().createTaskBuilder().execute(() -> this.offer(Keys.VANISH, false)).delayTicks(1).submit(SpongeImpl.getPlugin());
```
XakepSDK: If player was already vanished by a plugin?

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -1375,4 +1394,119 @@ private ITextComponent onGetDisplayName4(IMerchant villager) {
...
+        //worldServer.getPlayerChunkMap().addPlayer(this);
+        //org.spongepowered.api.entity.Entity spongeEntity = (org.spongepowered.api.entity.Entity) this;
+        //((org.spongepowered.api.world.World) worldServer).spawnEntity(spongeEntity);
+        //this.playerEntityList.add(this);
+        //this.uuidToPlayerMap.put(this.getUniqueID(), this);
+        //this.addSelfToInternalCraftingInventory();
+
+        // Update reducedDebugInfo game rule
+        this.connection.sendPacket(new SPacketEntityStatus((EntityPlayerMP) (Object) this,
+                worldServer.getGameRules().getBoolean(DefaultGameRules.REDUCED_DEBUG_INFO) ? (byte) 22 : 23));
```
XakepSDK: Magic values

src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java
Aaron1011: Good point - I'll push a fix taking that into account.

src/main/java/org/spongepowered/common/data/util/DataQueries.java
```diff
@@ -126,13 +126,22 @@
...
     public static final DataQuery USER_SPAWNS = of("Spawns");
 
+    // GameProfile
+    public static final DataQuery PROFILE_PROPERTIES = of("PROFILE_PROPERTIES");
+
+    // ProfileProperty
+    public static final DataQuery PROPERTY_NAME = of("PROPERTY_NAME");
+    public static final DataQuery PROPERTY_VALUE = of("PROPERTY_VALUE");
+    public static final DataQuery PROPERTY_SIGNATURE = of("PROPERTY_SIGNATURE");
+
```
Cybermaxke: `ProfileProperties`, `PropertyName`, `PropertyValue`, etc?

Cybermaxke: Should probably be just `Name`, `Value` and `Signature`.

Cybermaxke: `Name`, `Value` and `Signature` as query path?

src/main/java/org/spongepowered/common/data/util/DataQueries.java
```diff
@@ -126,13 +126,22 @@
...
+    //public static final DataQuery SKIN_UUID = of("SkinUUID");
+    public static final DataQuery SKIN = of("Skin");
 
     // User
     public static final DataQuery USER_UUID = of("UUID");
     public static final DataQuery USER_NAME = of("Name");
     public static final DataQuery USER_SPAWNS = of("Spawns");
 
+    // GameProfile
+    public static final DataQuery PROFILE_PROPERTIES = of("PROFILE_PROPERTIES");
```
Cybermaxke: `ProfileProperties` as query path?

XakepSDK:
Can you leave old way?
upd:
Ah, i think, its possible to get skin from filled gameprofile.

src/main/java/org/spongepowered/common/data/util/DataQueries.java
Aaron1011: `Name` and `Value` are pretty generic, so I'm going to leave the `Profile` prefix. However, I'll fix the capitalization and underscore to be more consistent.

src/main/java/org/spongepowered/common/data/util/DataQueries.java
```diff
@@ -126,13 +126,23 @@
...
+
+    // ProfileProperty
+    public static final DataQuery PROPERTY_NAME = of("PROPERTY_NAME");
+    public static final DataQuery PROPERTY_VALUE = of("PROPERTY_VALUE");
+    public static final DataQuery PROPERTY_SIGNATURE = of("PROPERTY_SIGNATURE");
```
kashike: `PropertySignature`

src/main/java/org/spongepowered/common/data/util/DataQueries.java
```diff
@@ -126,13 +126,23 @@
...
     public static final DataQuery USER_UUID = of("UUID");
     public static final DataQuery USER_NAME = of("Name");
     public static final DataQuery USER_SPAWNS = of("Spawns");
 
+    // GameProfile
+    public static final DataQuery PROFILE_PROPERTIES = of("PROFILE_PROPERTIES");
+
+    // ProfileProperty
+    public static final DataQuery PROPERTY_NAME = of("PROPERTY_NAME");
+    public static final DataQuery PROPERTY_VALUE = of("PROPERTY_VALUE");
```
kashike: `PropertyValue`

src/main/java/org/spongepowered/common/data/util/DataQueries.java
```diff
@@ -126,13 +126,23 @@
...
     // User
     public static final DataQuery USER_UUID = of("UUID");
     public static final DataQuery USER_NAME = of("Name");
     public static final DataQuery USER_SPAWNS = of("Spawns");
 
+    // GameProfile
+    public static final DataQuery PROFILE_PROPERTIES = of("PROFILE_PROPERTIES");
+
+    // ProfileProperty
+    public static final DataQuery PROPERTY_NAME = of("PROPERTY_NAME");
```
kashike: `PropertyName`

src/main/java/org/spongepowered/common/data/util/DataQueries.java
kashike: +1

kashike (changes requested):

Aaron1011:
@XakepSDK I've completely rewritten the handling of previous tab list entries. There's now a new key called `UPDATE_GAME_PROFILE`, which controls whether or not changing a player's skin permanently affects `GameProfile` and tab list entries. If it's set to `false`, the previous tab list entry will always be restored as soon as possible.

Aaron1011:
@kashike Do you want any more changes?

XakepSDK:
`Keys.UPDATE_GAME_PROFILE` thing is very confusing.
Why we can't just update GameProfile internally?
And if player will be unvanished, new skin will be sent?

Aaron1011:
@XakepSDK I've added javadocs explaining how the key works.

> Why we can't just update GameProfile internally?

Setting it to `true` will update the GameProfile internally. However, plugins that wish to have more control over players can set it to `false`. This allows the player's skin to be treated as a purely visual effect, allowing plugins to manipulate the player's tab list and gameprofile separately.

> And if player will be unvanished, new skin will be sent?

Yes, the logic for properly updating and restoring the tab list is handled by the vanish code. In fact, setting a player's skin simply [vanishes and un-vanishes the player](https://github.com/SpongePowered/SpongeCommon/pull/1926/files#diff-79dbfe6bad7db36ebf78c7c7b830f87aR1678) in order to cause other players to see the new skin.

Aaron1011:
Merged in https://github.com/SpongePowered/SpongeCommon/commit/3d24556ce3c9692377758a493f1c1c6905505b87.

PR #1929, opened 2018-05-29
Fix advancement bugs

### LINKED ISSUE #1904: Sponge is soft-resetting player advancements on world load
**Tested on the following**

- SpongeForge version:
1.12.2-2611-7.1.0-BETA-3016 (forge-1.12.2-14.23.2.2645)
1.12.2-2655-7.1.0-BETA-3065 (forge-1.12.2-14.23.3.2655)
- SpongeVanilla version:
1.12.2-7.1.0-BETA-32
1.12.2-7.1.0-BETA-42
- Java version: 1.8.0_171
- Operating System:
Linux
Windows

- Plugins/Mods:
Nucleus-1.3.6-S7.0-MC1.12.2-plugin

**Issue Description**
When a world is loaded Sponge appears to do something with Advancements which results in every player currently logged in having there advancements soft-reset, relogging or running `/reload` will bring back the players advancements.

This issue was noticed on the SevTech modpack, The result of this is that players do not get the rewards from completing advancements which means they cannot progress.

**This issue is not present on a Forge only server.**

**Reproduction**
Run the following commands from a Player (Console won't work unless you are targeting an online player):

`/advancement grant @p only minecraft:story/root`
`/advancement test @p minecraft:story/root` - Make sure it says 'Player ... has completed advancement ...'
`/world unload DIM-1` - Requires Nucleus
`/world load DIM-1` - Requires Nucleus
`/advancement test @p minecraft:story/root` - This should now say 'Player ... has **not** completed advancement ...'

Extra:
Relog or run `/reload`
`/advancement test @p minecraft:story/root` - This should now say 'Player ... has completed advancement ...'

### LINKED ISSUE #2140: Twilight Forest advancements crash
- SpongeForge version: 1.12.2-2655-7.1.0-BETA-3065
- Forge version: 1.12.2-14.23.3.2655
- Java version: 1.8.0_51

- Plugins/Mods:
	
fenceoverhaul, baubles, bibliocraft, biomesoplenty, jei, blockdrops,  bloodmoon, cfm, cosmeticarmorreworked, customspawner, customnpcs, ptrmodellib, props, disenchanter, doggytalents, mocreatures, fastleafdecay, foamflower, gravestone, ichunutil, inventorytweaks, ironbackpacks, ironchest, jeiintegration, jeresources, longfallboots, malisiscore, malisisdoors, mantle, radixcore, mca, parachutemod, portalgun, roguelike, saltmod, lteleporters, spawnercraft, tails, thaumcraft, thutcore, thutcore_compat, thuttech, torohealthmod, twilightforest, vending

*I don't use sponge for plugins as of yet, only for optimization purposes.*

This ONLY happens if sponge is in my server's mods folder AND this has only happened while in the twilight forest, otherwise there are no crashes whatsoever. This has been tested many times with different circumstances, but the results are the same every time. My client hangs up, then the screen becomes 'sky colored,' then this crash happens:

https://gist.github.com/OneTrueVex/d58444706a385d38ea9aefa9a798f8cf

In some earlier instances I've been able to navigate the twilight forest up until specific areas, like the swamp, then the crash happens. In later instances where I've advanced farther, then install sponge, the crash still happens. Neither a fresh client, nor a fresh server can circumvent this issue. From what I've gathered, the advancements may be the culprit. It's impossible to login to the server once the crash has happened, as I'm placed back in the very spot that crashes my client. Deleting sponge solves the crashes, but this solution is problematic.. if sponge is removed/reinstalled not only do magic maps break, new twilight chunks become misaligned with existing chunks, so I must delete all existing dimension data for twilight if I want it to work properly afterwards, but I digress.

This seems to be a slightly similar issue, though my files are up to date:

https://github.com/SpongePowered/SpongeForge/issues/2103

Obviously the easy solution would be to just not install sponge for now, and even more so because I'm not using any plugins, but the performance boost on my server is just fantastic, so it's a real bummer.

Cybermaxke:
**Common** | [Forge](https://github.com/SpongePowered/SpongeForge/pull/2182)

Fixes #1904 
Should fix: https://github.com/SpongePowered/SpongeForge/issues/2140

- Server side advancement logic no longer applies on client constructed advancements
- Fix soft-resetting player advancements on world load

Currently WIP, I am trying to find and fix bugs related to advancements.

### DIFF
--- a/src/main/java/org/spongepowered/common/advancement/SpongeAdvancementHelper.java
+++ b/src/main/java/org/spongepowered/common/advancement/SpongeAdvancementHelper.java
@@ -26,7 +26,5 @@ package org.spongepowered.common.advancement;
 
 public final class SpongeAdvancementHelper {
 
-    public static final ThreadLocal<Boolean> CONSTRUCTING_CLIENT_ADVANCEMENTS = ThreadLocal.withInitial(() -> false);
-
     public static final ThreadLocal<Boolean> INSIDE_REGISTER_EVENT = ThreadLocal.withInitial(() -> false);
 }
--- a/src/main/java/org/spongepowered/common/advancement/SpongeAdvancementHelper.java
+++ b/src/main/java/org/spongepowered/common/advancement/SpongeAdvancementHelper.java
@@ -1,0 +1,32 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.advancement;
+
+public final class SpongeAdvancementHelper {
+
+    public static final ThreadLocal<Boolean> CONSTRUCTING_CLIENT_ADVANCEMENTS = ThreadLocal.withInitial(() -> false);
+
+    public static final ThreadLocal<Boolean> INSIDE_REGISTER_EVENT = ThreadLocal.withInitial(() -> false);
+}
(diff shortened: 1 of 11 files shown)


src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancement.java
```diff
@@ -89,10 +92,22 @@
...
+
+    private void checkServer() {
+        checkState(!this.client);
+    }
+
     @SuppressWarnings("ConstantConditions")
     @Inject(method = "<init>", at = @At("RETURN"))
     private void onInit(ResourceLocation id, @Nullable Advancement parentIn, @Nullable DisplayInfo displayIn,
             AdvancementRewards rewardsIn, Map<String, Criterion> criteriaIn, String[][] requirementsIn, CallbackInfo ci) {
+        this.client = SpongeAdvancementHelper.CONSTRUCTING_CLIENT_ADVANCEMENTS.get();
```
gabizou: No..... Should be able to rely on other methods than just relying on thread local booleans.

src/main/java/org/spongepowered/common/advancement/SpongeAdvancementHelper.java
```diff
@@ -0,0 +1,32 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.advancement;
+
+public final class SpongeAdvancementHelper {
+
+    public static final ThreadLocal<Boolean> CONSTRUCTING_CLIENT_ADVANCEMENTS = ThreadLocal.withInitial(() -> false);
+
+    public static final ThreadLocal<Boolean> INSIDE_REGISTER_EVENT = ThreadLocal.withInitial(() -> false);
```
gabizou: You can use the phase tracker to figure out if you're in an event or not.

src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancement.java
```diff
@@ -89,10 +92,22 @@
 
     @Nullable private Advancement tempParent;
 
+    // Whether the Advancement is constructed for the client
+    private boolean client;
+
+    private void checkServer() {
+        checkState(!this.client);
```
gabizou: Use `ServerUtils.isMainThread()` etc.

gabizou (changes requested):

Cybermaxke:
@gabizou I replaced the thread local to check the client/server thread. But I have no clue how to use phase tracking (never used it before) to replace the other one.

PR #1932, opened 2018-06-05
Implementation Code Cleanup

gabizou:
Overall code cleanup for a bunch of compiler warnings. This is a WIP and of course, the cleanup is intended only for cleanup, no actual functionality changes.

This should improve however some corner cases that can occur as nullability and validation checks can be better written for certain areas of code.

### DIFF
--- a/src/main/java/org/spongepowered/common/entity/EntityUtil.java
+++ b/src/main/java/org/spongepowered/common/entity/EntityUtil.java
@@ -1007,7 +1007,7 @@ public final class EntityUtil {
      * @param offsetY The offset y coordinate
      * @return The item entity
      */
-    @SuppressWarnings({"unchecked", "rawType"})
+    @SuppressWarnings({"unchecked", "rawType", "rawtypes"})
     @Nullable
     public static EntityItem entityOnDropItem(Entity entity, ItemStack itemStack, float offsetY, double xPos, double zPos) {
         if (itemStack.isEmpty()) {
@@ -1028,7 +1028,7 @@ public final class EntityUtil {
 
         // Gather phase states to determine whether we're merging or capturing later
         final PhaseData peek = PhaseTracker.getInstance().getCurrentPhaseData();
-        final IPhaseState currentState = peek.state;
+        final IPhaseState<?> currentState = peek.state;
         final PhaseContext<?> phaseContext = peek.context;
 
         // We want to frame ourselves here, because of the two events we have to throw, first for the drop item event, then the constructentityevent.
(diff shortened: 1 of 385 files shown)


phit:
nice, there's a few little things here that could be nice to include https://github.com/SpongePowered/SpongeCommon/issues/23

src/main/java/org/spongepowered/common/entity/EntityUtil.java
```diff
@@ -1007,7 +1007,7 @@ public static EntityItem entityOnDropItem(Entity entity, ItemStack itemStack, fl
      * @param offsetY The offset y coordinate
      * @return The item entity
      */
-    @SuppressWarnings({"unchecked", "rawType"})
+    @SuppressWarnings({"unchecked", "rawType", "rawtypes"})
```
ST-DDT: Isnt the old `rawType` a typo and thus can be removed?

src/main/java/org/spongepowered/common/entity/EntityUtil.java
```diff
@@ -1205,6 +1205,17 @@ public static ItemStack throwDropItemAndConstructEvent(IMixinEntity entity, doub
...
+     * This is used to create the "dropping" motion for items caused by players. This
+     * specifically was being used (and should be the correct math) to drop from the
+     * player, when we do item stack captures preventing entity items being created.
+     *
+     * @param dropAround True if it's being "dropped around the player like dying"
+     * @param player The player to drop around from
+     * @param random The random instance
+     * @return The motion vector
+     */
+    @SuppressWarnings("unused")
```
ST-DDT, line 1219: Why should we keep the unused?

src/main/java/org/spongepowered/common/event/SpongeEventManager.java
```diff
@@ -411,16 +414,18 @@ private boolean post(Event event, List<RegisteredListener<?>> handlers) {
             try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame();
                  final PhaseContext<?> context = PluginPhase.Listener.GENERAL_LISTENER.createPhaseContext()
                             .source(handler.getPlugin())
-                            .buildAndSwitch()) {
-                handler.getTimingsHandler().startTimingIfSync();
+                            .buildAndSwitch();
+                 final Timing timings = handler.getTimingsHandler()) {
```
phit: extra space?

src/main/java/org/spongepowered/common/event/SpongeEventManager.java
```diff
@@ -411,16 +414,18 @@ private boolean post(Event event, List<RegisteredListener<?>> handlers) {
             try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame();
                  final PhaseContext<?> context = PluginPhase.Listener.GENERAL_LISTENER.createPhaseContext()
```
phit, line 414: extra leading space?

gabizou: That's the proper indentation since it's a multiple try with resources.

src/main/java/org/spongepowered/common/event/SpongeEventManager.java
gabizou: Same as here, note how it lines up with `CauseStackManager`

src/main/java/org/spongepowered/common/entity/EntityUtil.java
gabizou: Item pre-merging will be reintroduced, just not right now.

src/main/java/org/spongepowered/common/event/SpongeEventManager.java
ST-DDT: I guess he referred to it being 5 spaces instead of 4 or 8. IIRC 8 is a common standard for this. Aligning it as it is here is also used but less common. As always: styles are just personal preferences.

phit: yeah, looked really odd to me

PR #1935, opened 2018-06-09
Inventory properties to Implementation

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1841) | **SpongeCommon**

### DIFF
--- a/testplugins/src/main/java/org/spongepowered/test/inventory/CustomInventoryTest.java
+++ b/testplugins/src/main/java/org/spongepowered/test/inventory/CustomInventoryTest.java
@@ -22,12 +22,13 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.test;
+package org.spongepowered.test.inventory;
 
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.block.BlockTypes;
 import org.spongepowered.api.command.CommandResult;
 import org.spongepowered.api.command.spec.CommandSpec;
+import org.spongepowered.api.data.Property;
 import org.spongepowered.api.data.key.Keys;
 import org.spongepowered.api.entity.living.animal.Horse;
 import org.spongepowered.api.entity.living.animal.Llama;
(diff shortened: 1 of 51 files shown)


Faithcaio:
I dont know. I'm not the Mixin expert

Cybermaxke:
Well, I think I have an idea how it could be implemented. Not sure if it will work though, but worth trying.

Create a Mixin with higher priority then MixinPropertyHolder (e.g. 1010), so that you can overwrite that `getProperty` method. Within the overwritten method, check if it's a inventory property class, if that's the case, do inventory related stuff. Else, if `this` is a PropertyHolder, use the original non inventory behavior. So basically the two methods are being merged into one, not sure how this will affect the generics of the method.

Well, this will only be possible if `ContainerMinecart` is no longer a Inventory in the API. This kind of ruins the point of it being a `Carrier`. If `ContainerMinecart` no longer implements Inventory in the API, this could work.

Cybermaxke:
Not even `FurnaceMinecart` extends `ContainerMinecart`, so that's already inconsistent.

Cybermaxke:
A possible solution (if it works):
`MixinPropertyHolder` gets priority lower then `MixinTraitInventory`
In `MixinTraitInventory`:
```java
@Override
public <T extends InventoryProperty<?, ?>> Optional<T> getProperty(Class<T> propertyClass) {
    // Try the inventory property behavior first
    if (InventoryProperty.class.isAssignableFrom(propertyClass)) {
        Object key = AbstractInventoryProperty.getDefaultKey(propertyClass);
        return getProperty(propertyClass, key);
    }
    // Fall back to PropertyHolder method behavior, if this is a property holder
    if (this instanceof PropertyHolder) {
        final Optional<PropertyStore<T>> optional = SpongeImpl.getPropertyRegistry().getStore(propertyClass);
        return optional.flatMap(store -> store.getFor((PropertyHolder) this));
    }
    return Optional.empty();
}
```

Faithcaio:
FurnaceMinecart has no inventory

Cybermaxke:
Ah yes, they are not actual furnaces

testplugins/src/main/java/org/spongepowered/test/inventory/CustomInventoryTest.java
```diff
@@ -22,7 +22,9 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.test;
+package org.spongepowered.test.inventory;
+
+import static org.spongepowered.api.data.Property.Operator.DELEGATE;
```
gabizou: Don't static import this

PR #1947, opened 2018-06-17
Implements @ContextValue annotation

MinusKube:
The `@ContextValue` annotation for event parameters is now fully implemented, example code:

```java
@Listener
public void onEntitySpawn(SpawnEntityEvent event,
                          @ContextValue("SPAWN_TYPE") SpawnType type) {

    logger.info(type.getName());
}
```

### DIFF
--- a/src/main/java/org/spongepowered/common/event/filter/delegate/ContextValueFilterSourceDelegate.java
+++ b/src/main/java/org/spongepowered/common/event/filter/delegate/ContextValueFilterSourceDelegate.java
@@ -27,19 +27,26 @@ package org.spongepowered.common.event.filter.delegate;
 import org.objectweb.asm.Label;
 import org.objectweb.asm.MethodVisitor;
 import org.objectweb.asm.Type;
+import org.spongepowered.api.GameRegistry;
+import org.spongepowered.api.Sponge;
 import org.spongepowered.api.event.cause.EventContext;
+import org.spongepowered.api.event.cause.EventContextKey;
 import org.spongepowered.api.event.filter.cause.ContextValue;
 
 import java.lang.reflect.Parameter;
+import java.util.Optional;
 
 import static org.objectweb.asm.Opcodes.ACONST_NULL;
 import static org.objectweb.asm.Opcodes.ALOAD;
 import static org.objectweb.asm.Opcodes.ARETURN;
+import static org.objectweb.asm.Opcodes.CHECKCAST;
 import static org.objectweb.asm.Opcodes.DUP;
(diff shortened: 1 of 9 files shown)


src/main/java/org/spongepowered/common/event/filter/delegate/ContextValueFilterSourceDelegate.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.Optional;
+
+public class ContextValueFilterSourceDelegate extends ContextFilterSourceDelegate {
+
+    private final ContextValue anno;
+
+    public ContextValueFilterSourceDelegate(ContextValue anno) {
+        this.anno = anno;
+
+        Optional<EventContextKey> opKey = SpongeImpl.getRegistry().getType(EventContextKey.class, this.anno.value());
```
gabizou, line 64: Does this even work? I don't believe the context keys are I'd with the way that the annotation is intending. It was intended to modify the annotation to improve it, since the class type filter is easily remedied by the declaration type of the method parameter of the listener.

Does the test pass? Can you verify a test plugin will function given the test?

gabizou (changes requested):
test plugins could do with some of this PR for testing and verification that the implementation is working as intended.

I don't believe the value string is valid anymore.

src/main/java/org/spongepowered/common/event/filter/delegate/ContextValueFilterSourceDelegate.java
MinusKube: Yeah it works and the test passes, I don't really understand the problem, how was it intended to be improved?

gabizou: Because, the id's are not the same as the field names found in `ContextValueKeys` class.

MinusKube: I don't really understand why wouldn't the ids be the same, but well, how am I supposed to change that?

MinusKube: ? 🤔

gabizou: You can't. We have to change the API as it's currently unimplementable. If you want to change the API to do it, by all means, you'd need some sort of class reference (to look up where the thing comes from) and the field name. That would be a way to do it.

PR #1951, opened 2018-06-21
Add tracker config to control block, entity, and TE tracking.

gabizou:
Just for verifying the changes proposed for improving capturing compatibility and allowing server administrators to configure for better compatibility with mod entities.


src/main/java/org/spongepowered/common/event/tracking/phase/tick/NoCaptureBlockTickPhaseState.java
```diff
@@ -51,4 +58,35 @@ public boolean requiresBlockCapturing() {
...
+
+    @Override
+    public boolean alreadyCapturingEntitySpawns() {
+        return true;
+    }
+
+    @Override
+    public void handleBlockChangeWithUser(@Nullable BlockChange blockChange, Transaction<BlockSnapshot> snapshotTransaction,
+        BlockTickContext context) {
+        super.handleBlockChangeWithUser(blockChange, snapshotTransaction, context);
```
gabizou: What's the point of overriding if it's calling super?

src/main/java/org/spongepowered/common/event/tracking/phase/tick/NoCaptureEntityTickPhaseState.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Override
+    public boolean alreadyCapturingEntitySpawns() {
+        return true;
+    }
+
+    @Override
+    public void handleBlockChangeWithUser(@Nullable BlockChange blockChange, Transaction<BlockSnapshot> snapshotTransaction,
+        EntityTickContext context) {
+        super.handleBlockChangeWithUser(blockChange, snapshotTransaction, context);
```
gabizou: And here.

phit:
got multiple gb of this https://gist.github.com/phit/d61a72b67d227b3057721dd133bec330
related or separate issue?

gabizou:
It looks like there is a tile entity that is causing chunk loads, but I don't think ti's directly caused by this branch. I'll need to get the Unwinding state to provide the nested context to get some better understanding. I'll push some changes to better debug this sort of issue once I finish documenting `IPhaseState`.

PR #1972, opened 2018-07-03
Improve ManipulatorTest

Aaron1011:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1859) | **SpongeCommon**

Currently, `ManipulatorTest` attempts to construct, manipulate, and serialize all registered `DataManipulators`. While this is able to catch many bugs, it suffers from a significant limitation: it is unable to set any new values. Whenever a `DataManipulator` is constructed, the default values for every associated `Key` are always (implicitly) used. When a `Key` represents a wrapper or container type (e.g. `Optional` or `List`), the corresponding default value will always be 'empty' (e.g. `Optional.empty()`, `Lists.newArrayList()`).

This often results in code paths not being exercised at all by `ManipulatorTest`, leaving bugs like https://github.com/SpongePowered/SpongeCommon/issues/1276 uncaught.

This PR greatly expands the amount of code that `ManipulatorTest` is able to cover, by creating and setting values for each of a `DataManipulator`'s `Key`s whenever possible. Several different strategies are used to accomplish this:

* `CatalogType`s are creating by retrieving the first instance from `GameRegistry#getAllOf`
* Types with a `builder` static method  are created by walking the methods of the `Builder` class. Any setter methods (defined as methods taking a single parameter and returning the `Builder` instance) are called with parameters created recursively through this process.
* If all else fails, types are created using `SpongeEventFactoryTest.mockParam`. This returns default values for a number of Java and Sponge types (`int`, `Text`, `Color`, etc). Otherwise, it falls back to a special 'recursive mock', which use `SpongeEventFactoryTest.mockParam` to create return values for all methods of the target type.

With these three value-creation strategies, I was able to uncover quite a few new bugs, which are fixed in this PR. These include:

* A missing registration for `SpongeTradeOfferBuilder`
* Broken serialization for `SpongeTradeOfferData`
* Invalid deserialization of lists of `DataContainers` in `DataUtil`
* Missing `equals` overrides in classes used in data values,
* An incorrect `TypeToken` registered in `KeyRegistryModule.java`

To ensure that dynamically creating values does not reduce our test coverage, all tests in `ManipulatorTest` are run twice - once with the previous behavior (default values only), and once with the new behavior (dynamically created values for applicable `Key`s).

There's still room to improve and add to the value-creation strategies included in this PR. However, these strategies will always be implemented on a 'best-effort' basis. It's simply impossible to create an instance of an arbitrary Java type, and have things work out sanely every time. As new `DataManipulator`s are added, it will likely be necessary to add rules for instantiating new types (`final` API types in particular)

### DIFF
--- a/src/main/java/org/spongepowered/common/util/TypeTokenHelper.java
+++ b/src/main/java/org/spongepowered/common/util/TypeTokenHelper.java
@@ -39,6 +39,10 @@ import javax.annotation.Nullable;
 @SuppressWarnings({"unchecked", "rawtypes", "WeakerAccess"})
 public final class TypeTokenHelper {
 
+    public static Class<?>  getGenericParam(TypeToken<?> token, int typeIndex) {
+        return (Class) ((ParameterizedType) token.getType()).getActualTypeArguments()[typeIndex];
+    }
+
     public static boolean isAssignable(TypeToken<?> type, TypeToken<?> toType) {
         return isAssignable(type.getType(), toType.getType());
     }
--- a/build.gradle
+++ b/build.gradle
@@ -62,6 +62,8 @@ dependencies {
     }
 
     testCompile 'org.spongepowered:lwts:1.0.0'
+    // Taken from https://stackoverflow.com/a/7969292
+    testCompile api.sourceSets.test.output
 }
 
 uploadArchives {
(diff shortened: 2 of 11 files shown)


Aaron1011:
@XakepSDK This should address your concern about data builder registration issues not being caught.

gabizou (approved):

src/main/java/org/spongepowered/common/util/TypeTokenHelper.java
```diff
@@ -39,6 +39,10 @@
 @SuppressWarnings({"unchecked", "rawtypes", "WeakerAccess"})
 public final class TypeTokenHelper {
 
+    public static Class<?>  getGenericParam(TypeToken<?> token, int typeIndex) {
```
ST-DDT, line 42: Duplicate whitespace

PR #1974, opened 2018-07-03
Add --tracker/-t flag for tracker config reload/save

dualspiral:
Adds `-t` for:

* `/sponge config`
* `/sponge reload`
* `/sponge save`

This simply uses the existing infrastructure for configs in the command structure. I couldn't verify this works because my tracker.conf didn't seem to need a reload to save changes made by the user, but I haven't had much time to test properly (though it doesn't seem to breaking anything, the current sponge commands work fine, a key thing for putting this PR up!)

@phit, mind giving this a go?


phit:
seems to work perfectly
https://phit.link/s/2018-07-04_00-14-31.mp4
https://phit.link/s/2018-07-04_00-15-38.mp4

PR #1977, opened 2018-07-04
Ensure that all PhaseContext captures are fully processed

Aaron1011:
This commit checks that all PhaseContext capture suppliers are either
null or empty when a phase completes. This has been responsible for a
large number of bugs in the past, such as the one fixed by e8d394f3.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/PhaseContext.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/PhaseContext.java
@@ -51,6 +51,7 @@ import org.spongepowered.common.event.tracking.context.CapturedSupplier;
 import org.spongepowered.common.event.tracking.context.EntityItemDropsSupplier;
 import org.spongepowered.common.event.tracking.context.EntityItemEntityDropsSupplier;
 import org.spongepowered.common.event.tracking.context.GeneralizedContext;
+import org.spongepowered.common.event.tracking.context.ICaptureSupplier;
 import org.spongepowered.common.event.tracking.context.ItemDropData;
 import org.spongepowered.common.event.tracking.phase.general.GeneralPhase;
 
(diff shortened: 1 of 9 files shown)


src/main/java/org/spongepowered/common/event/tracking/PhaseContext.java
```diff
@@ -296,6 +297,27 @@ public PrettyPrinter printCustom(PrettyPrinter printer, int indent) {
...
+                || isNonEmpty(this.capturedItemsSupplier)
+                || isNonEmpty(this.capturedEntitiesSupplier)
+                || isNonEmpty(this.capturedItemStackSupplier)
+                || isNonEmpty(this.entityItemDropsSupplier)
+                || isNonEmpty(this.entityItemEntityDropsSupplier)
+                || isNonEmpty(this.blockEntitySpawnSupplier);
+    }
+
+    private boolean isNonEmpty(@Nullable ICaptureSupplier supplier) {
+        if (supplier != null && !supplier.isEmpty()) {
```
XakepSDK: just `return supplier != null && !supplier.isEmpty()`

src/main/java/org/spongepowered/common/event/tracking/PhaseTracker.java
```diff
@@ -422,6 +425,18 @@ public void printExceptionFromPhase(Throwable e, PhaseContext<?> context) {
         }
     }
 
+    private void checkPhaseContextProcessed(IPhaseState<?> state, PhaseContext<?> context) {
+        if (context.notAllCapturesProcessed()) {
```
gabizou: Make sure to make this toggle able for verbosity. We don't want to spam server owners if we missed some captures.

src/main/java/org/spongepowered/common/event/tracking/PhaseTracker.java
```diff
@@ -224,6 +224,7 @@ public void completePhase(IPhaseState<?> prevState) {
         } catch (Exception e) {
             this.printMessageWithCaughtException("Exception Post Dispatching Phase", "Something happened when trying to post dispatch state", state, context, e);
         }
+        this.checkPhaseContextProcessed(state, context);
```
gabizou, line 231: Maybe this could be flagged for validation? If we have fixed all the bugs where the check is no longer necessary, it'd be extra processing that wouldn't be necessary.

gabizou (changes requested):

src/main/java/org/spongepowered/common/event/tracking/PhaseTracker.java
Aaron1011: I don't see this check adding much overhead at all. I think it's useful to always have on, so that we can catch any regressions that might occur (like the TNT bug).

PR #1979, opened 2018-07-05
Include all captures in capture supplier toString(), not just counts

Aaron1011:
Now that we have config settings to limit phase tracker error spam,
including the full contents of a capture supplier shouldn't make much of
a difference. This can be valuable for debugging, especially when
dealing with recursive capture processing issues (where the captured
blocks may provide a hint as to the cause of the issue).

@gabizou Does this look reasonable?

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/context/CapturedMultiMapSupplier.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/context/CapturedMultiMapSupplier.java
@@ -220,7 +220,7 @@ public abstract class CapturedMultiMapSupplier<K, V> implements Supplier<ArrayLi
     @Override
     public String toString() {
         return com.google.common.base.MoreObjects.toStringHelper(this)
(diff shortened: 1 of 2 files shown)


gabizou:
I was thinking of doing something as far as exposing the captured objects in the testing for pretty printing in a specific case, maybe the first 5 elements?

Aaron1011:
@gabizou: I could limit it to cases like a block capture processing overflow. However, I don't think printing out all of the captures should cause much of a problem.

gabizou:
> @gabizou: I could limit it to cases like a block capture processing overflow. However, I don't think printing out all of the captures should cause much of a problem.

It's more of cases where we don't want to be printing a massive list of captures in rare cases that they exceed something like 20 blocks. Also, may need to reduce the `toString()` for `SpongeBlockSnapshot` to avoid printing redundant information (like world, or block state + extended).

gabizou:
Closing since issues with what is being captured is now transactional based on the state to appropriately print BlockTransactions when necessary.

PR #1980, opened 2018-07-07
Merge StackFrame usage with IPhaseStates

gabizou:
Currently, we have issues where within an `IPhaseState`'s target's processing, various `StackFrame`s are pushed and popped, leading to potential loss of contextual information when unwinding captures. To solve for this, we're aiming to further link the `StackFrame` usage with `IPhaseState` switches so that during a single `IPhaseState`, we can safely push to contexts and causes without losing said information outside the state.

This is very much WIP, but will resolve a few issues that @bloodmc has mentioned. This PR is being opened for posterity and feedback once no longer WIP (likewise to serve as common discussion for comparisons when we get some better testing as the system is refactored).


gabizou:
Closing as this was superseded by #1987.

PR #1992, opened 2018-07-14
Add simplistic frame pooling

dualspiral:
As we potentially create hundreds of frames a tick, this can mean that a lot of objects are created. To reduce the number of objects that we create and destroy during cause creation, we pool the frames themselves. We pool up to 50.

This isn't yet ready for pulling, I need to do some profiling on this first. This might be slower! Just looking for any feedback - if this is no good, it's not good!

This also cleans up some of the calls for the CSM to reduce the number of calls needed. The included tests verify that the functionality is the same.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/SpongeCauseStackManager.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeCauseStackManager.java
@@ -36,7 +36,6 @@ import org.spongepowered.api.event.CauseStackManager;
 import org.spongepowered.api.event.cause.Cause;
 import org.spongepowered.api.event.cause.EventContext;
 import org.spongepowered.api.event.cause.EventContextKey;
-import org.spongepowered.api.util.Tuple;
 import org.spongepowered.asm.util.PrettyPrinter;
 import org.spongepowered.common.SpongeImpl;
 import org.spongepowered.common.SpongeImplHooks;
@@ -48,7 +47,6 @@ import java.util.ArrayDeque;
 import java.util.Arrays;
 import java.util.Deque;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Optional;
@@ -61,12 +59,29 @@ import javax.annotation.Nullable;
 @Singleton
 public final class SpongeCauseStackManager implements CauseStackManager {
 
-    public static final boolean DEBUG_CAUSE_FRAMES = Boolean.valueOf(System.getProperty("sponge.debugcauseframes", "false"));
+    private static final boolean DEBUG_CAUSE_FRAMES = Boolean.parseBoolean(System.getProperty("sponge.debugcauseframes", "false"));
+    public static final int MAX_POOL_SIZE;
+
+    static {
+        int maxPoolSize = 50;
+        try {
+            maxPoolSize = Integer.parseInt(System.getProperty("sponge.cause.framePoolSize", "50"));
+        } catch (NumberFormatException ex) {
+            SpongeImpl.getLogger().warn("sponge.cause.framePoolSize must be an integer, was set to {}. Defaulting to 50.",
+                    System.getProperty("sponge.cause.framePoolSize"));
+        }
+        MAX_POOL_SIZE = Math.max(0, maxPoolSize);
+    }
 
     private final Deque<Object> cause = Queues.newArrayDeque();
-    private final Deque<CauseStackFrameImpl> frames = Queues.newArrayDeque();
-    private Map<EventContextKey<?>, Object> ctx = Maps.newHashMap();
 
+    // Frames in use
+    private final Deque<CauseStackFrameImpl> frames = Queues.newArrayDeque();
+
+    // Frames not currently in use
+    private final Deque<CauseStackFrameImpl> framePool = new ArrayDeque<>(MAX_POOL_SIZE);
+
+    private Map<EventContextKey<?>, Object> ctx = Maps.newHashMap();
     private int min_depth = 0;
     private int[] duplicateCauses = new int[100];
     @Nullable private Cause cached_cause;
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/event/SpongeCauseStackManager.java
```diff
@@ -354,56 +363,22 @@ public void popFrameMutator(PhaseContext<?> context) {
 
     }
 
-    // TODO could pool these for more fasts
```
Deamon5550, line 401: More fasts!

Deamon5550 (approved):

src/main/java/org/spongepowered/common/event/SpongeCauseStackManager.java
```diff
@@ -178,7 +187,14 @@ public Object peekCause() {
     @Override
     public StackFrame pushCauseFrame() {
         enforceMainThread();
-        CauseStackFrameImpl frame = new CauseStackFrameImpl(this.min_depth);
+        CauseStackFrameImpl frame;
+        if (this.framePool.isEmpty()) {
+            frame = new CauseStackFrameImpl();
+        } else {
+            frame = this.framePool.pop();
```
gabizou, line 251: How does this interact with the debugging of stack frames with the exception built in?

dualspiral:
I've done some limited profiling on what I have so far and... speeds are actually worse for push/pop with this pooling adding about 7ms for approximately 2000 calls over 20s when compared for non-pooled (handwavy 57ms with pooling, 50ms without on my local machine) This is, of course, me on my own pootling about placing and breaking blocks - super scientific I'm sure you can agree.

To check whether the pooling was the major problem, I removed the pooling but kept the simplification of the frame context rollback methods. The speeds are comparable there, little more time adding context, less time popping. Again, not very scientific, will need to hammer it a bit more for more info, but it indicates the time lost is to the pooling.

Using a pooling library may be better, but pushing/popping a stack can't be that expensive and we don't really need to expire objects. My only other thought is to use a fixed size array and an int to indicate where the last object was taken from (we don't need to worry about thread safety as we enforce a single thread), treat the array as a stack without adding/removing items. I'll give that a go now, see where we get to.

While it might seem like this is a bad idea based on the timings, what isn't included here is where garbage collection is involved. Over a 20 second period, I counted 2000 frames being pushed/popped - but this isn't a production server, just me going around placing/breaking sponges. This might be better for larger servers if they experience GC spikes, but less of an issue on smaller servers.

It may be that pooling is just not worth it, all things considered.

Regardless, this is dwarved by the number of frame mutators that are created - necessarily though the number of <s>`BiConsumers` and</s> `Tuple`s that are created doing this dwarfs the number for frames created - it actually increases the time loss to frames as the mutators might create multiple frames. I do think that this could be improved a fair bit, though it's complicated by the fact that each phase context is a singleton (again, correctly so), but that's for another PR/discussion.

src/main/java/org/spongepowered/common/event/SpongeCauseStackManager.java
dualspiral: Yep - it basically doesn't, mistake on my part. I'd need to clear it before reusing the frame.

dualspiral:
Access only arrays (even with filling the array with 50 frames on manager creation) made this a bit better, but still a little slower, still a few milliseconds in it. So, pooling is going to be slower, GC not withstanding.

I think we probably want to park this unless we think the savings from not GC-ing the objects is worth it. The object creation is just faster than access with the bounds checking we need to do, probably thanks to the JVM JIT. It would probably be faster if we could just choose the element, but we need to increment an index and check it's not above the max pool size - it looks like this kills any performance advantage.

src/main/java/org/spongepowered/common/event/SpongeCauseStackManager.java
```diff
@@ -76,7 +80,12 @@
     private Deque<Tuple<PhaseContext<?>, BiConsumer<StackFrame, PhaseContext<?>>>> phaseContextProviders = new ArrayDeque<>();
 
     @Inject
-    private SpongeCauseStackManager() { }
+    private SpongeCauseStackManager() {
+        // start with an initial frame pool of 10 - this may need to be tweaked
```
kashike: Expose this as a system property, perhaps? `-Dsponge.cause.framePoolSize=50`

dualspiral:
So, I came back to this and decided to create a timing plugin that allows us to measure the performance of creating and destroying the frames, vs this pooling. The results show little difference, though I only repeated the high volume tests.

https://gist.github.com/dualspiral/5d79086118d5021e8e7ffb381089def1

The averages for 20000 repeats of the test, putting 5 objects in each of the 50 frames (so, pushing and popping 100,000 frames and half a million objects), gets us:

Without pooling: 269ms
With pooling: 290ms

The timings can be highly variable and this is being tested on my desktop, a Linux machine running a full GUI. It might be interesting to run this on a server OS - I suspect the results will be more uniform. However, it looks like pooling is slower. The only thing I don't know is what the impact of the garbage collector is - will the GCing potentially hundreds of thousands of objects offset this slight hit in performance?

I'll leave this open for now if anyone has an opinion or has a good way to measure the GCing effectively, but I'm happy for this to be closed if this pooling is not desired in light of this.

dualspiral:
@gabizou 

So, after our discussion a couple of days ago about this and frame/context pooling, I came back to this to give it another shot. This time, I'm running on a Windows 10 machine with the latest Java 8 version, so the comparison with the previous test cannot be considered to be equal.

5 objects per frame, 100 frames pushed then popped, repeated 20000 times, test itself repeated 3 times:

With pooling (ascending order):

563522700ns (563.5ms)
616426000ns (616.4ms)
739853200ns (739.9ms)

Without:

807430600ns (807.4ms)
864693000ns (864.6ms)
873630900ns (873.6ms)

At lower volumes, 50 frames, 500 times, 

With: 8174600ns (8.1ms)
Without: 11500800ns (11.50ms)

And 100 frames, 500 times:

With: 13388200ns (13.3ms)
Without: 15654400ns (15.65ms)

---

The times with the pooling are a lot more variable but generally lower than non-pooled, though negligible as we get to fewer frames. The memory usage was also that little bit lower. GC pauses should be lower with the pooling too. So, I think this bodes well for looking at doing this with the phase tracker too, especially if some of the data in the states can also remain within the pooled/cached objects.

dualspiral:
https://github.com/SpongePowered/SpongeCommon/commit/ce6928471b1a88515a405210fc6637b6a9028ef0

PR #1995, opened 2018-07-14
Add feature to reload individual plugins

SimonFlash:
This PR allows the `/sponge plugins reload` command to specify a plugin to be reloaded.

 - This requires a way to send an event to listeners registered by a particular plugin. To do this, I added `SpongeEventManager#post(Event, PluginContainer)` that runs a filter operation on the available listeners.
 - The command syntax already supports `/sponge plugins reload <pluginid>`, however it does not verify that the plugin id is valid (if it's not, all plugins will be reloaded). Changing this requires a redesign to Sponge's commands and is beyond the scope of this PR.

Please let me know if there's a better way to handle posting an event to listeners for a specific plugin.

### DIFF
--- a/src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
+++ b/src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
@@ -91,6 +91,7 @@ import org.spongepowered.common.config.type.GlobalConfig;
 import org.spongepowered.common.config.type.TrackerConfig;
 import org.spongepowered.common.config.type.WorldConfig;
 import org.spongepowered.common.entity.EntityUtil;
+import org.spongepowered.common.event.SpongeEventManager;
 import org.spongepowered.common.interfaces.IMixinChunk;
 import org.spongepowered.common.interfaces.IMixinMinecraftServer;
 import org.spongepowered.common.interfaces.entity.IMixinEntity;
@@ -560,9 +561,15 @@ public class SpongeCommandFactory {
                 .arguments(optionalWeak(literal(Text.of("reload"), "reload")), optional(plugin(Text.of("plugin"))))
                 .executor((src, args) -> {
                     if (args.hasAny("reload") && src.hasPermission("sponge.command.plugins.reload")) {
-                        src.sendMessage(Text.of("Sending reload event to all plugins. Please wait."));
                         Sponge.getCauseStackManager().pushCause(src);
-                        SpongeImpl.postEvent(SpongeEventFactory.createGameReloadEvent(Sponge.getCauseStackManager().getCurrentCause()));
+                        if (args.hasAny("plugin")) {
+                            PluginContainer plugin = args.<PluginContainer>getOne("plugin").get();
+                            src.sendMessage(Text.of("Sending reload event to " + plugin.getId() + ". Please wait."));
+                            ((SpongeEventManager) Sponge.getEventManager()).post(SpongeEventFactory.createGameReloadEvent(Sponge.getCauseStackManager().getCurrentCause()), plugin);
+                        } else {
+                            src.sendMessage(Text.of("Sending reload event to all plugins. Please wait."));
+                            SpongeImpl.postEvent(SpongeEventFactory.createGameReloadEvent(Sponge.getCauseStackManager().getCurrentCause()));
+                        }
                         Sponge.getCauseStackManager().popCause();
                         src.sendMessage(Text.of("Reload complete!"));
                     } else if (args.hasAny("plugin")) {
--- a/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
@@ -461,6 +461,8 @@ public class SpongeEventManager implements EventManager {
     }
 
     public boolean post(Event event, PluginContainer plugin) {
-        return post(event, getHandlerCache(event).getListeners().stream().filter(l -> l.getPlugin().equals(plugin)).collect(Collectors.toList()));
+        return post(event, getHandlerCache(event).getListeners().stream()
+                .filter(l -> l.getPlugin().equals(plugin))
+                .collect(Collectors.toList()));
     }
 }
--- a/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeEventManager.java
@@ -76,6 +76,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 import javax.annotation.Nullable;
 import javax.inject.Inject;
@@ -457,7 +458,9 @@ public class SpongeEventManager implements EventManager {
 
     public boolean post(Event event, boolean allowClientThread) {
         return post(event, getHandlerCache(event).getListeners());
+    }
 
-
+    public boolean post(Event event, PluginContainer plugin) {
+        return post(event, getHandlerCache(event).getListeners().stream().filter(l -> l.getPlugin().equals(plugin)).collect(Collectors.toList()));
     }
 }
--- a/testplugins/src/main/java/org/spongepowered/test/IndividualPluginReloadTest.java
+++ b/testplugins/src/main/java/org/spongepowered/test/IndividualPluginReloadTest.java
@@ -1,0 +1,22 @@
+package org.spongepowered.test;
+
+import com.google.inject.Inject;
+import org.slf4j.Logger;
+import org.spongepowered.api.event.Listener;
+import org.spongepowered.api.event.game.GameReloadEvent;
+import org.spongepowered.api.plugin.Plugin;
+
+@Plugin(id="individual-plugin-reload-test",
+        name="Individual Plugin Reload Test",
+        description = "Tests sending GameReloadEvent to specific plugins.",
+        version = "1.0.0")
+public class IndividualPluginReloadTest {
+
+    @Inject private Logger logger;
+
+    @Listener
+    public void onReload(GameReloadEvent event) {
+        this.logger.info("GameReloadEvent: " + event);
+    }
+
+}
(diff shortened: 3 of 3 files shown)


src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
```diff
@@ -560,9 +561,15 @@ private static CommandSpec createSpongePluginsCommand() {
                 .arguments(optionalWeak(literal(Text.of("reload"), "reload")), optional(plugin(Text.of("plugin"))))
                 .executor((src, args) -> {
                     if (args.hasAny("reload") && src.hasPermission("sponge.command.plugins.reload")) {
-                        src.sendMessage(Text.of("Sending reload event to all plugins. Please wait."));
                         Sponge.getCauseStackManager().pushCause(src);
-                        SpongeImpl.postEvent(SpongeEventFactory.createGameReloadEvent(Sponge.getCauseStackManager().getCurrentCause()));
+                        if (args.hasAny("plugin")) {
+                            PluginContainer plugin = args.<PluginContainer>getOne("plugin").get();
+                            src.sendMessage(Text.of("Sending reload event to " + plugin.getId() + ". Please wait."));
+                            ((SpongeEventManager) Sponge.getEventManager()).post(SpongeEventFactory.createGameReloadEvent(Sponge.getCauseStackManager().getCurrentCause()), plugin);
```
gabizou, line 568: You can use `SpongeImpl.getEventManager()` to get the implementation class variant, instead of casting here.

src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
```diff
@@ -560,9 +561,15 @@ private static CommandSpec createSpongePluginsCommand() {
                 .arguments(optionalWeak(literal(Text.of("reload"), "reload")), optional(plugin(Text.of("plugin"))))
                 .executor((src, args) -> {
                     if (args.hasAny("reload") && src.hasPermission("sponge.command.plugins.reload")) {
-                        src.sendMessage(Text.of("Sending reload event to all plugins. Please wait."));
                         Sponge.getCauseStackManager().pushCause(src);
-                        SpongeImpl.postEvent(SpongeEventFactory.createGameReloadEvent(Sponge.getCauseStackManager().getCurrentCause()));
+                        if (args.hasAny("plugin")) {
+                            PluginContainer plugin = args.<PluginContainer>getOne("plugin").get();
```
gabizou, line 566: This can crash the command, can you check and if it's not, return an appropriate response to the sender?

src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
```diff
@@ -560,9 +561,15 @@ private static CommandSpec createSpongePluginsCommand() {
...
+                            src.sendMessage(Text.of("Sending reload event to " + plugin.getId() + ". Please wait."));
+                            ((SpongeEventManager) Sponge.getEventManager()).post(SpongeEventFactory.createGameReloadEvent(Sponge.getCauseStackManager().getCurrentCause()), plugin);
+                        } else {
+                            src.sendMessage(Text.of("Sending reload event to all plugins. Please wait."));
+                            SpongeImpl.postEvent(SpongeEventFactory.createGameReloadEvent(Sponge.getCauseStackManager().getCurrentCause()));
```
gabizou, line 571: Could you potentially provide a test plugin to verify the causes used for this event in this case?

src/main/java/org/spongepowered/common/event/SpongeEventManager.java
```diff
@@ -453,7 +454,9 @@ public boolean post(Event event) {
 
     public boolean post(Event event, boolean allowClientThread) {
         return post(event, getHandlerCache(event).getListeners());
+    }
 
-
+    public boolean post(Event event, PluginContainer plugin) {
+        return post(event, getHandlerCache(event).getListeners().stream().filter(l -> l.getPlugin().equals(plugin)).collect(Collectors.toList()));
```
gabizou: Lots of nested methods here, can you multi-line them so if any one of them (I know, highly unlikely) are on a new line? Helps with line numbers in stack traces for possible errors in the future.

gabizou (changes requested):

src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
SimonFlash: I checked for something like that, but `SpongeImpl` doesn't have anything.

src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
SimonFlash: This shouldn't be crashing the command, as the `plugin` argument is parsed through `GenericArguments.plugin`. If the type isn't a `PluginContainer` then there are other issues at play.

src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
SimonFlash: No, sorry. I don't know enough about it or what I would even do to test that. The calls to push/pop are in the same places so there shouldn't be a difference here.

src/main/java/org/spongepowered/common/event/SpongeEventManager.java
SimonFlash: I can break the stream call up onto separate lines.

src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
gabizou: Maybe write one then? 😉

src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
gabizou: Oh, duh, it's behind the `hasAny` call. But, given that it's not checking if the plugin is actually there, would that still pose an issue?

src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
gabizou: Write a plugin in the [testplugins directory](https://github.com/SpongePowered/SpongeCommon/tree/bleeding/testplugins/src/main/java/org/spongepowered/test) just as you would, to implement the listener and maybe just print out the cause from the event? 
```java
@Listener
public void onReload(GameReloadEvent event) {
  this.logger.log(event.getCause());
}
```

src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
SimonFlash: That opens up a lot of questions into this PR, as if something like that is added than the rest of the codebase should be updated as well. Would this just be a method with an `instanceof` check, or should it be caching it the same as the other managers? At that point, this would need to be done by updating Guice injections which is definitely beyond where I want to go here.

src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
SimonFlash: No. They use the same key. I believe dual has a PR open for an API method to get it directly for cases when you know the argument is present, so it's a standard thing here.

src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
SimonFlash: What would be the point of this? It doesn't seem like there's actually anything that would really be being tested here.

dualspiral: Real world testing. What it allows us to do is run the server in a test scenario and check things are working with a known set of plugins easily in the IDE. In this case, have the reload event in one allows us to check that the reload event fires as expected, to all or to a target, and makes sure nothing has regressed.

SimonFlash: Gotcha, I wasn't thinking of testing the event being sent to specifically the test plugin and was mainly confused at what the point of just logging the cause would be as it hasn't changed. That makes a lot more sense now and I can add that in.

SimonFlash:
Still looking for guidance on how to handle `SpongeImpl.getEventManager()` if that's something to be included in this PR.

src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
gabizou: Then don't, we'll do a rewrite to provide the appropriately casted event manager at some point in the future.

gabizou:
Also, any chance of basing this off stable-7 instead of bleeding?

SimonFlash:
This should be ready to merge into stable-7. Thank you to @gabizou and @dualspiral for assistance with switching the target branch and git stuff, and apologies to anyone accidentally requested for review in the process.

src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java
dualspiral: This is fine with `hasAny` then `getOne`, if it crashes, it's indicative of an issue.

It may be slightly better to do `getOne` then use the optional for performance reasons, rather than having two calls to the context, but that's splitting hairs. It's fine as is.

testplugins/src/main/java/org/spongepowered/test/IndividualPluginReloadTest.java
```diff
@@ -0,0 +1,22 @@
...
+        name="Individual Plugin Reload Test",
+        description = "Tests sending GameReloadEvent to specific plugins.",
+        version = "1.0.0")
+public class IndividualPluginReloadTest {
+
+    @Inject private Logger logger;
+
+    @Listener
+    public void onReload(GameReloadEvent event) {
+        this.logger.info("GameReloadEvent: " + event);
```
ryantheleach, line 43: Being **incredibly** picky, aka I don't expect you to do this.

2 test plugins would be a ton better.

/sponge reload 

will trigger this message, and unless any of the other test plugins have messages on reload, or startup (I didn't check) you would never know it was reloading everything.

2 would let you pick between them.

But I'm sure it's fine as-is since you could issue the reload command to a different testplugin anyway.

PR #1999, opened 2018-07-15
Register/track all the advancements properly.

Cybermaxke:
Better catch them all before they slip into more exceptions.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancement.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancement.java
@@ -76,7 +76,6 @@ import javax.annotation.Nullable;
 public class MixinAdvancement implements org.spongepowered.api.advancement.Advancement, IMixinAdvancement {
 
     @Shadow @Final @Mutable @Nullable private Advancement parent;
-    @Shadow @Final private ResourceLocation id;
     @Shadow @Final @Mutable private String[][] requirements;
     @Shadow @Final @Mutable private Map<String, Criterion> criteria;
     @Shadow @Final @Nullable private DisplayInfo display;
@@ -97,7 +96,7 @@ public class MixinAdvancement implements org.spongepowered.api.advancement.Advan
     }
 
     private boolean isMainThread() {
-        return SpongeImplHooks.isMainThread() || AdvancementRegistryModule.getInstance().isInInitialRegistration();
+        return SpongeImplHooks.isMainThread();
     }
 
     @SuppressWarnings("ConstantConditions")
@@ -117,9 +116,7 @@ public class MixinAdvancement implements org.spongepowered.api.advancement.Advan
         if (displayIn != null) {
             this.name = SpongeTexts.toPlain(displayIn.getTitle());
         }
-        if (!PhaseTracker.getInstance().getCurrentState().isEvent()) {
-            AdvancementRegistryModule.getInstance().registerAdditionalCatalog(this);
-        } else {
+        if (PhaseTracker.getInstance().getCurrentState().isEvent()) {
             // Wait to set the parent until the advancement is registered
             this.tempParent = parentIn;
             this.parent = AdvancementRegistryModule.DUMMY_ROOT_ADVANCEMENT;
--- a/src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancementList.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancementList.java
@@ -30,10 +30,9 @@ import net.minecraft.util.ResourceLocation;
 import org.apache.logging.log4j.Logger;
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.advancement.AdvancementTree;
-import org.spongepowered.api.event.cause.Cause;
-import org.spongepowered.api.event.cause.EventContext;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Mutable;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
@@ -45,8 +44,10 @@ import org.spongepowered.common.event.registry.SpongeGameRegistryRegisterEvent;
 import org.spongepowered.common.event.tracking.phase.plugin.ListenerPhaseContext;
 import org.spongepowered.common.event.tracking.phase.plugin.PluginPhase;
 import org.spongepowered.common.interfaces.advancement.IMixinAdvancementList;
+import org.spongepowered.common.registry.type.advancement.AdvancementMap;
 import org.spongepowered.common.registry.type.advancement.AdvancementRegistryModule;
 import org.spongepowered.common.registry.type.advancement.AdvancementTreeRegistryModule;
+import org.spongepowered.common.registry.type.advancement.RootAdvancementSet;
 
 import java.util.Map;
 import java.util.Set;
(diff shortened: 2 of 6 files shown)


src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancement.java
```diff
@@ -93,19 +93,19 @@
...
-        return SpongeImplHooks.isMainThread() || AdvancementRegistryModule.getInstance().isInInitialRegistration();
+        return SpongeImplHooks.isMainThread();
     }
 
     @SuppressWarnings("ConstantConditions")
     @Inject(method = "<init>", at = @At("RETURN"))
     private void onInit(ResourceLocation id, @Nullable Advancement parentIn, @Nullable DisplayInfo displayIn,
             AdvancementRewards rewardsIn, Map<String, Criterion> criteriaIn, String[][] requirementsIn, CallbackInfo ci) {
         // Don't do anything on the client, unless we're performing registry initialization
-        if (!this.isMainThread()) {
```
Zidane: Why did you remove the this.?

src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancementList.java
```diff
@@ -58,8 +60,8 @@
 
     @Shadow @Final private static Logger LOGGER;
 
-    @Shadow @Final private Map<ResourceLocation, Advancement> advancements;
-    @Shadow @Final private Set<Advancement> roots;
+    @Shadow @Final private Map<ResourceLocation, Advancement> advancements = new AdvancementMap();
```
Zidane: This literally makes no sense. You're shadowing a field that is finalized in the actual class then initializing it here. Can't do that.

src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancementList.java
```diff
@@ -58,8 +60,8 @@
...
+    @Shadow @Final private final Set<Advancement> roots = new RootAdvancementSet();
```
Zidane: Same as above.

src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancementList.java
Zidane: As well as you'll run into mod compatibility issues with replacing the map reference that Vanilla is using..

Cybermaxke: It's working fine like this. But I can move it if you want.

Cybermaxke: Same is done in: https://github.com/SpongePowered/SpongeCommon/blob/df3266d7b2923018f111df3a4db442e6a4882841/src/main/java/org/spongepowered/common/mixin/core/world/gen/MixinChunkProviderServer.java#L90

Cybermaxke: `@Mutable` is only used for setting fields after it's initialized

src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancement.java
```diff
@@ -93,19 +93,19 @@
     @Nullable private Advancement tempParent;
 
     private void checkServer() {
-        checkState(this.isMainThread());
```
gabizou: Formatting change? I'd rather keep `this.`

src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancementList.java
gabizou: The only reason why we do that for the chunk provider server is to provide better caching of the previously requested chunk, reducing lookups. If it's doing the same with the `AdvancementMap`, you should name it more appropriately.

src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancementList.java
```diff
@@ -70,18 +72,13 @@ private void onLoadAdvancements(Map<ResourceLocation, Advancement.Builder> advan
         if (!SpongeImplHooks.isMainThread() || !ShouldFire.GAME_REGISTRY_EVENT_REGISTER) {
             return;
         }
-
+        final SpongeGameRegistryRegisterEvent<org.spongepowered.api.advancement.Advancement> event =
+                new SpongeGameRegistryRegisterEvent<>(Cause.of(EventContext.empty(), SpongeImpl.getRegistry()),
```
gabizou: Why are you creating the cause yourself and not delegating to the cause stack manager?

gabizou: Which, in this case, you'd still want to put it in the try statement so the ListenerPhaseContext has a chance to provide some additional context

src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancementList.java
Cybermaxke: I am just making sure that all the Advancements are getting captured and registered using the map.

src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancementList.java
Cybermaxke: It's a mistake, so I will fix it

src/main/java/org/spongepowered/common/mixin/core/advancement/MixinAdvancementList.java
```diff
@@ -58,8 +58,8 @@
 
     @Shadow @Final private static Logger LOGGER;
 
-    @Shadow @Final private Map<ResourceLocation, Advancement> advancements;
-    @Shadow @Final private Set<Advancement> roots;
+    @Shadow @Final private final Map<ResourceLocation, Advancement> advancements = new AdvancementMap();
```
gabizou: Double check what the AP output is when compiling this mixin, I'm not sure whether `@Mutable` would be required. Pinging @Aaron1011 @Mumfrey for clarifying input?

src/main/java/org/spongepowered/common/registry/type/advancement/AdvancementMap.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+package org.spongepowered.common.registry.type.advancement;
+
+import net.minecraft.advancements.Advancement;
+import net.minecraft.util.ResourceLocation;
+import org.spongepowered.common.SpongeImplHooks;
+
+import java.util.HashMap;
+
+public final class AdvancementMap extends HashMap<ResourceLocation, Advancement> {
```
gabizou, line 33: Can you javadoc what this class is doing and why and how we're using it?

bloodmc:
@Cybermaxke @gabizou is this PR ready?

Cybermaxke:
@bloodmc Don't forget to merge this before RB ;)

XakepSDK:
Advancement loading errors
https://gist.github.com/XakepSDK/249e9ca265f5c6b5c1ccb069ea6d4aa2
```
Mods (50): Minecraft, Minecraft Coder Pack, Forge Mod Loader, Minecraft Forge, SpongeAPI, SpongeForge, AI Improvements, AbyssalCraft, AbyssalCraft Integration, AppleCore, AppleSkin, Applied Energistics 2, BetterBedrockGen, Biomes O' Plenty, CoFH Core, CoFH World, CodeChicken Lib, CosmeticArmorReworked, Diet Hopper, FoamFix, FoamFixCore, InstantUnify, Inventory Tweaks, JourneyMap, Just Enough Items, Just Enough Pattern Banners, LagGoggles, LegibleNei, Mantle, Not Enough Items, NuclearCraft, Pam's Cookables, Pam's HarvestCraft, Quick Leaf Decay, Reborn Core, Redstone Flux, Roguelike Dungeons, Simply Jetpacks 2, Tech Reborn, The Betweenlands, Thermal Dynamics, Thermal Expansion, Thermal Foundation, Tinker's JEI, Tinkers' Construct, UniDict, WAIM, Waila, WanionLib, What Are We Looking At

Plugins (10): Minecraft, Minecraft Coder Pack, SpongeAPI, SpongeForge, GriefPrevention, LuckPerms, MagiBridge, Nucleus, Total Economy, UltimateChat

SpongeForge
    Minecraft: 1.12.2
    SpongeAPI: 7.1.0-SNAPSHOT-51caddf
    SpongeForge: 1.12.2-2705-7.1.0-BETA-3355
Minecraft Forge: 14.23.4.2739
```

PR #2002, opened 2018-07-18
Fix for FireworkShape id being... really messed up.

SimonFlash:
The id for `SpongeFireworkShape` is currently stored as a `byte`, and the name is returned instead. This makes the following changes to fix this behavior:

 - Switches the id to a `String` consistent with the `CatalogType` field.
 - Changes the name to be a 'normal' name, not upper case.
 - Adds the `minecraft` namespace to the id
 - Adds the `minecraft` namespace (if necessary) when looking up the id.

### DIFF
--- a/src/main/java/org/spongepowered/common/registry/type/item/FireworkShapeRegistryModule.java
+++ b/src/main/java/org/spongepowered/common/registry/type/item/FireworkShapeRegistryModule.java
@@ -32,6 +32,7 @@ import org.spongepowered.api.item.FireworkShapes;
 import org.spongepowered.api.registry.CatalogRegistryModule;
 import org.spongepowered.api.registry.util.RegisterCatalog;
 import org.spongepowered.common.data.processor.common.FireworkUtils;
+import org.spongepowered.common.item.SpongeFireworkShape;
 
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/registry/type/item/FireworkShapeRegistryModule.java
```diff
@@ -58,8 +63,7 @@
 
     @Override
     public void registerDefaults() {
-        this.fireworkShapeMap.putAll(FireworkUtils.shapeMapping.values()
-        .stream()
-        .collect(Collectors.toMap(shape -> shape.getId().toLowerCase(Locale.ENGLISH), Function.identity())));
+        this.fireworkShapeMap.putAll(FireworkUtils.shapeMapping.values().stream()
+                .collect(Collectors.toMap(SpongeFireworkShape::getId, Function.identity())));
     }
```
ST-DDT, line 68: A simple foreach would be much easier to read.
Also please also register the old ids to ease migration. `shape.getName().toUppercase().replace(' ', '_')`

ImMorpheus:
I'm partially against this. This is the same situation NotePitch is in (See https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/data/processor/common/NoteUtils.java). I think Sponge is trying to minimize layout disparities with mojang code. That's how they identify fireworkshapes, we should remain consistent with their code (when possible).

I'd prefer to see a solution to get rid of `shapeMapping`.

src/main/java/org/spongepowered/common/registry/type/item/FireworkShapeRegistryModule.java
SimonFlash: I'd rather do a foreach or something here, but I've been told previously to avoid making any unnecessary changes as it doesn't affect the outcome. Adding the old ids is a good idea as well to mitigate the below problem.

SimonFlash:
While it's generally good to keep the implementation as close to Minecraft as possible, the problem is that in this case it makes the API incorrect. `CatalogType`s are intended to be uniquely identifiable by their id, and while it's not guaranteed to be lowercase the majority of them are. However, one thing that *must* be handled properly is the `minecraft` namespace, as currently they don't which can cause lookup problems.

I agree that the whole `shapeMapping` thing is a bit unwieldy, but for now the point of this PR is to ensure that `FireworkShape`s can be retrieved by id in the same way as other `CatalogType`s.

randombyte-developer:
Maybe related? #2040

Issue #723, opened 2016-05-16
Missing Datamanipulators when recreating EntitySnapshot from DataContainer


Faithcaio:
When serializing and deserializing an EntitySnapshot it looses the DataManipulators.
`EntitySnapshot.builder().build(event.getTargetEntity().createSnapshot().toContainer()).get()`
(Data is not completly lost as restoring the Snapshot gets an entity with the Manipulators again) 

because ->
`AbstractDataBuilder#build:86`
`SpongeEntitySnapshotBuilder#buildContent:279`
`DataUtil#deserializeImmutableManipulatorList:146`
`SpongeDataManager#getImmutableManipulatorBuilder:301`
`this.immutableBuilderMap` only contain Builders for custom data.

I found the easiest way to test was to check for `Keys.REPRESENTED_ITEM` on `DestructEntityEvent` by picking up items.

(tested running latest SpongeForge)

Faithcaio:
The same happens for BlockSnapshots

Zidane:
@gabizou 

Status?

ImMorpheus:
@ryantheleach "pr pending" please

JBYoshi:
Has this been fixed? I saw #2333 was merged, but the description said there would be another part.

PR #2023, opened 2018-07-27
Fix missing vanilla manipulators

ImMorpheus:
This should fix https://github.com/SpongePowered/SpongeCommon/issues/723 and a couple of other issues.

Step to reproduce:
kill a pig.

Code:

```
    @Listener
    public void onBlockBreak(DestructEntityEvent event) {
        if (event.getTargetEntity().getType() == EntityTypes.PIG) {
            System.out.println("1: " + event.getTargetEntity().get(Keys.HEALTH).orElse(12345678.9));
            System.out.println("2: " + event.getTargetEntity().createSnapshot().get(Keys.HEALTH).orElse(12345678.9));
            EntitySnapshot e = EntitySnapshot.builder().build(event.getTargetEntity().createSnapshot().toContainer()).get();
            System.out.println("3: " + e.get(Keys.HEALTH).orElse(12345678.9));
            EntitySnapshot e1 = EntitySnapshot.builder().build(event.getTargetEntity().toContainer()).get();
            System.out.println("4: " + e1.get(Keys.HEALTH).orElse(12345678.9));
        }
    }
```
This is wip but I was looking for input.

Is it possible for an entity to have null vanillaManipulators ? 

While this was born as a fix for #723 I have noticed the problem is present also on ItemStackSnapshot and BlockSnapshot. There was an old PR for the latter (https://github.com/SpongePowered/SpongeCommon/pull/1905). @gabizou Is the concern about performances still valid ?

### DIFF
--- a/src/main/java/org/spongepowered/common/entity/SpongeEntitySnapshotBuilder.java
+++ b/src/main/java/org/spongepowered/common/entity/SpongeEntitySnapshotBuilder.java
@@ -198,8 +198,11 @@ public class SpongeEntitySnapshotBuilder extends AbstractDataBuilder<EntitySnaps
             this.rotation = optional.get().getRotation();
             this.scale = optional.get().getScale();
         }
-        for (ImmutableDataManipulator<?, ?> manipulator : holder.getContainers()) {
-            add(manipulator);
+        if (!holder.getContainers().isEmpty()) {
+            if (this.customManipulators == null) {
+                this.customManipulators = new ArrayList<>();
+            }
+            this.customManipulators.addAll(holder.getContainers());
         }
         if (holder instanceof SpongeEntitySnapshot) {
             this.compound = ((SpongeEntitySnapshot) holder).getCompound().orElse(null);
--- a/src/main/java/org/spongepowered/common/entity/SpongeEntitySnapshotBuilder.java
+++ b/src/main/java/org/spongepowered/common/entity/SpongeEntitySnapshotBuilder.java
@@ -29,6 +29,7 @@ import static com.google.common.base.Preconditions.checkState;
 
 import com.flowpowered.math.vector.Vector3d;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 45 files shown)


src/main/java/org/spongepowered/common/entity/SpongeEntitySnapshotBuilder.java
```diff
@@ -272,10 +256,20 @@ public EntitySnapshot build() {
         final String entityTypeId = container.getString(DataQueries.ENTITY_TYPE).get();
         this.entityType = SpongeImpl.getRegistry().getType(EntityType.class, entityTypeId).get();
 
+        ImmutableList.Builder<ImmutableDataManipulator<?, ?>> list = ImmutableList.builder();
+
+        Entity entity = Sponge.getServer().getWorld(this.worldId).get().createEntity(this.entityType, this.position);
```
ImMorpheus: Suggestions ? I'd really like not to do this

ryantheleach:
Is this an issue only on bleeding?

ImMorpheus:
~qa

gabizou:
> While this was born as a fix for #723 I have noticed the problem is present also on ItemStackSnapshot and BlockSnapshot. There was an old PR for the latter (#1905). @gabizou Is the concern about performances still valid ?

The concern is still valid. All manipulators need to be lazily loaded, preferably read from NBT (expansion of the nbt data processors) because block snapshots need to be fast for our block captures. It's on a long todo list of things to write.

ImMorpheus:
I see, I will leave blocksnapshot manipulators as they are then. Just a couple of questions:

https://github.com/SpongePowered/SpongeCommon/blob/dcbf53c72268c79e65280f920d67c19bb71b4e90/src/main/java/org/spongepowered/common/entity/SpongeEntitySnapshot.java#L82-L84

Keys and value are stored in sets but manipulators are stored in a list. Is it possible to have duplicate manipulators ? It doesn't look like it's supported but I want to be sure.

Manipulators and keys are already lazy loaded for BlockSnapshots, do I have to do the same for EntitySnapshots ?

PR #2026, opened 2018-07-29
SpongeEntityType invalid cast fix

SimonFlash:
Changes an always true `!= null` check to use `instanceof`. Unsure if this is a proper fix, but comparisons with existing code show `!= null` definitely did not have the intended behavior.

From LemADEC#4620, when executing `sponge -g reload`.
https://gist.github.com/LemADEC/1799a9e8370a423e34715aeba6f8e40f
>>
    Minecraft: 1.12.2
    SpongeAPI: 7.1.0-SNAPSHOT-51caddf
    SpongeForge: 1.12.2-2705-7.1.0-BETA-3355
    Minecraft Forge: 14.23.4.2739

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
@@ -1397,7 +1397,7 @@ public abstract class MixinEntity implements org.spongepowered.api.entity.Entity
 
     @Override
     public void refreshCache() {
-        if (this.entityType != null) {
+        if (this.entityType instanceof SpongeEntityType) {
             this.allowsBlockBulkCapture = ((SpongeEntityType) this.entityType).allowsBlockBulkCapture;
             this.allowsEntityBulkCapture = ((SpongeEntityType) this.entityType).allowsEntityBulkCapture;
             this.allowsBlockEventCreation = ((SpongeEntityType) this.entityType).allowsBlockEventCreation;


src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
```diff
@@ -1397,7 +1397,7 @@ public boolean allowsEntityEventCreation() {
 
     @Override
     public void refreshCache() {
-        if (this.entityType != null) {
+        if (this.entityType instanceof SpongeEntityType) {
```
ImMorpheus, line 1400: Quoting gabizou `EntityTypes.UNKNOWN should never be found or used, even when mods are present.`

IMO `this.entityType != null` is fine, it avoids a NPE.

ImMorpheus (commented):
The problem is that there's an entity loaded of type UNKNOWN.

src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java
SimonFlash: It does not avoid an NPE because `entityType` is never null, as said in the description. `this.entityType` is initialized through `Registry#getTranslation`, which *never* returns null and instead returns `EntityTypes.UNKNOWN`, which does not extend from `SpongeEntityType`.

If the `entityType` should not be unknown at this point, then this should be replaced with a `Preconditions#checkState` verification.

ImMorpheus: > this.entityType is initialized through Registry#getTranslation

Oh I see, I got entity and entityType mixed up.

> If the entityType should not be unknown at this point, then this should be replaced with a Preconditions#checkState verification.

What about logging more information about the entity (like position, etc) ? Do we really have to crash the server ?

ryantheleach: > What about logging more information about the entity (like position, etc) ? Do we really have to crash the server ?

dump the class name, it will name and shame the entity / mod doing registration incorrectly (or reveal a flaw in our system)

You will want to make sure it only shows the message once though, otherwise it's going to spam.

I believe we have caches and rate limiting elsewhere?

Issue #1896, opened 2018-04-30
Performance drop after upgrading from SF 3029 to 3064


Lemmmy:
**I am currently running**
- SpongeForge version: 3064
- Java version: 1.8.0_151
- Operating System: Ubuntu 16.04.4 LTS

- Plugins/Mods: Plugins (15): Minecraft, Minecraft Coder Pack, SpongeAPI, SpongeForge, ActiveTime, AutoSpy, Buycraft, Dynmap, GriefPrevention, LuckPerms, Nucleus, SpongePrometheusExporter, SpongeSlack, TabManager, kristpay
Mods (29): Minecraft, Minecraft Coder Pack, Forge Mod Loader, Minecraft Forge, SpongeAPI, SpongeForge, AutoRegLib, Baubles, BiblioCraft, CC Prometheus, CC: Tweaked, Chisel, Computronics, Dynmap, DynmapBlockScan, Iron Chest, Just Enough Items, MinimalPeripherals, OpenComputers, OpenComputers (Core), Plethora, Plethora Core, Quark, Sampling Performance Profiler, The Weirding Gadget, TickProfiler, WorldEdit, asielib, mercurius_updater

**Issue Description**

Last night we upgraded our server from SpongeForge 3029 to 3064, and have noticed a stable decline in TPS since. This 24-hour graph shows the TPS before upgrading, and since upgrading:

![](https://you.wouldnot.download/a-textbook.png)

Here is a snippet of a 1-minute Sampler timing: ![](https://you.wouldnot.download/a-hemp.png)
And the full NPS: https://crashes.switchcraft.pw/2018-04-30-19-13.nps

Additionally, a timings report: https://timings.aikar.co/?id=42dd995e9fa042ce9db99a8f1a35e9a6

Phit suggests this may be related to #1706.

ProsperCraft:
If it helps, here is some more data using 3064.

https://timings.aikar.co/?id=934635a70aa549adab990405becbd642

Warm Roast-
[sponge3064.zip](https://github.com/SpongePowered/SpongeCommon/files/1962058/sponge3064.zip)

Sampler-
[samplersponge3064.zip](https://github.com/SpongePowered/SpongeCommon/files/1962072/samplersponge3064.zip)

Tickprofiler- 
https://gist.github.com/ProsperCraft/dd334abea4d170a861387460af3a360a

LagGoggles-
[...]

PR #2031, opened 2018-07-31
Fix/improve performance

ImMorpheus:
>  block snapshots need to be fast for our block captures

@gabizou Can you suggest a way to stress-test the system ? I've been igniting chunks of tnt for hours.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/TrackingUtil.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/TrackingUtil.java
@@ -789,7 +789,9 @@ public final class TrackingUtil {
                     }
                 }
                 final ChangeBlockEvent.Post post = ((IPhaseState) state).createChangeBlockPostEvent(context, transactions);
-                SpongeImpl.postEvent(post);
+                if (ShouldFire.CHANGE_BLOCK_EVENT_POST) {
+                    SpongeImpl.postEvent(post);
+                }
                 return post;
             }
         }
--- a/src/main/java/org/spongepowered/common/mixin/core/network/datasync/MixinEntityDataManager.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/network/datasync/MixinEntityDataManager.java
@@ -25,9 +25,12 @@
 package org.spongepowered.common.mixin.core.network.datasync;
 
 import it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap;
+import net.minecraft.crash.CrashReport;
+import net.minecraft.crash.CrashReportCategory;
 import net.minecraft.entity.Entity;
 import net.minecraft.network.datasync.DataParameter;
 import net.minecraft.network.datasync.EntityDataManager;
+import net.minecraft.util.ReportedException;
 import org.apache.commons.lang3.ObjectUtils;
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.data.DataHolder;
@@ -46,6 +49,7 @@ import org.spongepowered.common.registry.type.data.KeyRegistryModule;
 
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.locks.ReadWriteLock;
 
 @Mixin(EntityDataManager.class)
 public abstract class MixinEntityDataManager {
@@ -58,7 +62,7 @@ public abstract class MixinEntityDataManager {
     @Shadow @Final protected Entity entity;
     @Shadow private boolean dirty;
 
-    @Shadow protected abstract <T> EntityDataManager.DataEntry<T> getEntry(DataParameter<T> key);
+    @Shadow @Final private ReadWriteLock lock;
 
     /**
      * @author gabizou December 27th, 2017
@@ -77,7 +81,7 @@ public abstract class MixinEntityDataManager {
         // Sponge Start - set up the current value, so we don't have to retrieve it multiple times
         final T currentValue = dataentry.getValue();
         final T incomingValue = value;
-        if (ObjectUtils.notEqual(value, currentValue)) { // Sponge - change dataentry.getValue() to use local variable
+        if (!value.equals(currentValue)) { // Sponge - change dataentry.getValue() to use local variable
             // Sponge Start - retrieve the associated key, if available
             // Client side can have an entity, because reasons.......
             // Really silly reasons......
@@ -115,5 +119,23 @@ public abstract class MixinEntityDataManager {
             dataentry.setDirty(true);
             this.dirty = true;
         }
+    }
+
+    @Overwrite
+    private <T> EntityDataManager.DataEntry<T> getEntry(DataParameter<T> key) {
+        this.lock.readLock().lock();
+        EntityDataManager.DataEntry<T> dataentry;
+
+        try {
+            dataentry = (EntityDataManager.DataEntry)((Int2ObjectOpenHashMap) this.entries).get(key.getId());
+        } catch (Throwable throwable) {
+            CrashReport crashreport = CrashReport.makeCrashReport(throwable, "Getting synched entity data");
+            CrashReportCategory crashreportcategory = crashreport.makeCategory("Synched entity data");
+            crashreportcategory.addCrashSection("Data ID", key);
+            throw new ReportedException(crashreport);
+        }
+
+        this.lock.readLock().unlock();
+        return dataentry;
     }
 }
--- a/src/main/java/org/spongepowered/common/mixin/core/network/datasync/MixinEntityDataManager.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/network/datasync/MixinEntityDataManager.java
@@ -40,6 +40,7 @@ import org.spongepowered.asm.mixin.Mutable;
 import org.spongepowered.asm.mixin.Overwrite;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.common.data.datasync.DataParameterConverter;
+import org.spongepowered.common.event.ShouldFire;
 import org.spongepowered.common.interfaces.network.datasync.IMixinDataParameter;
 import org.spongepowered.common.registry.type.data.KeyRegistryModule;
 
@@ -81,7 +82,7 @@ public abstract class MixinEntityDataManager {
             // Client side can have an entity, because reasons.......
             // Really silly reasons......
             // I don't know, ask Grum....
-            if (this.entity != null && this.entity.world != null && !this.entity.world.isRemote) { // We only want to spam the server world ;)
+            if (ShouldFire.CHANGE_DATA_HOLDER_EVENT_VALUE_CHANGE && this.entity != null && this.entity.world != null && !this.entity.world.isRemote) { // We only want to spam the server world ;)
                 final Optional<DataParameterConverter<T>> converter = ((IMixinDataParameter) key).getConverter();
                 // At this point it is changing
                 if (converter.isPresent()) {
--- a/src/main/java/org/spongepowered/common/mixin/core/world/MixinExplosion.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/MixinExplosion.java
@@ -26,6 +26,7 @@ package org.spongepowered.common.mixin.core.world;
 
 import com.flowpowered.math.vector.Vector3d;
 import com.google.common.collect.Sets;
+import it.unimi.dsi.fastutil.objects.ObjectOpenHashSet;
 import net.minecraft.block.Block;
 import net.minecraft.block.material.Material;
 import net.minecraft.block.state.IBlockState;
@@ -43,6 +44,7 @@ import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.MathHelper;
 import net.minecraft.util.math.Vec3d;
 import net.minecraft.world.WorldServer;
+import net.minecraft.world.chunk.Chunk;
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.entity.explosive.Explosive;
 import org.spongepowered.api.event.SpongeEventFactory;
@@ -63,6 +65,7 @@ import org.spongepowered.common.SpongeImplHooks;
 import org.spongepowered.common.entity.EntityUtil;
 import org.spongepowered.common.event.tracking.PhaseTracker;
 import org.spongepowered.common.event.tracking.PhaseData;
+import org.spongepowered.common.interfaces.util.math.IMixinBlockPos;
 import org.spongepowered.common.interfaces.world.IMixinExplosion;
 import org.spongepowered.common.interfaces.world.IMixinLocation;
 import org.spongepowered.common.interfaces.world.IMixinWorldServer;
@@ -171,9 +174,10 @@ public abstract class MixinExplosion implements Explosion, IMixinExplosion {
         // Sponge Start - If the explosion should not break blocks, don't bother calculating it
         if (this.shouldBreakBlocks) {
             // Sponge End
-            Set<BlockPos> set = Sets.<BlockPos>newHashSet();
+            Set<BlockPos> set = new ObjectOpenHashSet<>(300);
             int i = 16;
 
+            Chunk chunk = null;
             for (int j = 0; j < 16; ++j) {
                 for (int k = 0; k < 16; ++k) {
                     for (int l = 0; l < 16; ++l) {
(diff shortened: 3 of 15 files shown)


ImMorpheus:
~wip

src/main/java/org/spongepowered/common/config/category/WorldCategory.java
```diff
@@ -77,6 +77,9 @@
                                                     + "then we recommend disabling it.")
     private boolean denyChunkRequests = false;
 
+    @Setting(value = "deny-te-chunk-requests", comment = "If 'true', TileEntities are not allowed to load chunks when calling getBlockState.")
```
gabizou: This needs to be mapped to tile entity id's because certain tile entities do need to be allowed to load chunks (otherwise you break some modded tile entities that depend on random stuff)

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -676,16 +677,18 @@ public static NotifyNeighborBlockEvent callNotifyNeighborEvent(World world, Bloc
             return null;
         }
         try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
-            final BlockState blockstate = (BlockState) ((net.minecraft.world.World) world).getBlockState(sourcePos);
+            final net.minecraft.world.chunk.Chunk chunk = ((WorldServer) world).getChunkFromBlockCoords(sourcePos);
```
gabizou: Why the difference here?

src/main/java/org/spongepowered/common/event/tracking/TrackingUtil.java
```diff
@@ -255,9 +255,10 @@ public static void updateTickBlock(IMixinWorldServer mixinWorld, Block block, Bl
         }
 
         final LocatableBlock locatable = LocatableBlock.builder()
-            .location(new Location<>(apiWorld, pos.getX(), pos.getY(), pos.getZ()))
```
gabizou: Why the difference?

src/main/java/org/spongepowered/common/event/tracking/phase/packet/PlaceBlockPacketState.java
```diff
@@ -96,7 +96,6 @@ public void appendNotifierToBlockEvent(BasicPacketContext context, IMixinWorldSe
         final Location<World> location = new Location<>(player.getWorld(), pos.getX(), pos.getY(), pos.getZ());
         final LocatableBlock locatableBlock = LocatableBlock.builder()
                 .location(location)
```
gabizou: It doesn't make sense that you're changing the builders usage in the file change above but not changing it here. Why?

src/main/java/org/spongepowered/common/mixin/core/world/MixinExplosion.java
```diff
@@ -383,7 +383,7 @@ public void doExplosionB(boolean spawnParticles) {
                         if (peek.state.requiresBlockPosTracking()) {
                             peek.context.getCaptureBlockPos().setPos(blockpos);
                         }
-                        block.dropBlockAsItemWithChance(this.world, blockpos, this.world.getBlockState(blockpos), 1.0F / this.size, 0);
+                        block.dropBlockAsItemWithChance(this.world, blockpos, iblockstate, 1.0F / this.size, 0);
```
gabizou: You're potentially changing the passed in block state, and therefor changing vanilla mechanics. I get that the block state is already retrieved, but should comment that you're making that change since it's an overwrite.

src/main/java/org/spongepowered/common/mixin/core/world/MixinExplosion.java
```diff
@@ -410,8 +410,8 @@ public void doExplosionB(boolean spawnParticles) {
 
         if (this.causesFire) {
             for (BlockPos blockpos1 : this.affectedBlockPositions) {
-                if (this.world.getBlockState(blockpos1).getMaterial() == Material.AIR && this.world.getBlockState(blockpos1.down()).isFullBlock()
-                    && this.random.nextInt(3) == 0) {
+                if (this.random.nextInt(3) == 0 && this.world.getBlockState(blockpos1).getMaterial() == Material.AIR
```
gabizou: And here. you should be commenting why you're changing it (it's failing faster than checking the block state).

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldEntitySpawner.java
```diff
@@ -208,7 +208,7 @@ public int findChunksForSpawning(WorldServer worldServerIn, boolean spawnHostile
                         int k1 = blockpos.getX();
                         int l1 = blockpos.getY();
                         int i2 = blockpos.getZ();
-                        IBlockState iblockstate = worldServerIn.getBlockState(blockpos);
+                        IBlockState iblockstate = chunk.getBlockState(blockpos);
```
gabizou: Why bypass the buildheight checks? You could be getting block states that are outside the build height, which vanilla will technically attempt to perform.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -944,7 +944,6 @@ private boolean onAddBlockEvent(WorldServer.ServerBlockEventList list, Object ob
 
         final LocatableBlock locatable = LocatableBlock.builder()
             .location(new Location<>(this, pos.getX(), pos.getY(), pos.getZ()))
-            .state(this.getBlock(pos.getX(), pos.getY(), pos.getZ()))
```
gabizou: Same with here, either change all of them to use `location` or the chained `pos`, and `world` methods.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -1891,16 +1890,20 @@ public IBlockState getBlockState(BlockPos pos) {
...
         } else {
+            net.minecraft.world.chunk.Chunk chunk;
             // ExtraUtilities 2 expects to get the proper chunk while mining or it gets stuck in infinite loop
-            // TODO add TE config to disable/enable chunk loads
-            final boolean forceChunkRequests = this.mixinChunkProviderServer.getForceChunkRequests();
-            final PhaseTracker phaseTracker = PhaseTracker.getInstance();
-            final IPhaseState<?> currentState = phaseTracker.getCurrentState();
-            if (currentState == TickPhase.Tick.TILE_ENTITY) {
-                ((IMixinChunkProviderServer) this.getChunkProvider()).setForceChunkRequests(true);
+            if (!this.activeConfig.getConfig().getWorld().getDenyTEChunkRequests()) {
```
gabizou: As mentioned before, you'd want to check on the `TileEntityType` instead of a global flag. Global flags are something we've moved away from for configs for things like this.

src/main/java/org/spongepowered/common/world/SpongeProxyBlockAccess.java
```diff
@@ -72,8 +72,9 @@ public TileEntity getTileEntity(BlockPos pos) {
 
     @Override
     public IBlockState getBlockState(BlockPos pos) {
-        if (this.processed.containsKey(pos)) { // first just check if there's already a pos list built.
-            return this.processed.get(pos);
+        IBlockState value = this.processed.get(pos);
```
gabizou: Fluff that isn't necessary. We're not actually using this class since it's been rendered unnecessary by the PhaseTracker system. It used to be required but was eliminated from use since the bugs that were introduced are now gone.

gabizou (changes requested):

ImMorpheus:
I've been inconsistent with SpongeLocatableBlockBuilder. 

Here's the problem:

https://github.com/SpongePowered/SpongeCommon/blob/d1b8572f639d71267955becc8e2e263fa4f34770/src/main/java/org/spongepowered/common/world/SpongeLocatableBlockBuilder.java#L67-L74 

`SpongeLocatableBlockBuilder#location` is setting the blockstate, by using the one of the block at the given location. Thus calling builder().state(..).location(..) is just overriding the blockstate, adding the overhead of World#getBlockState().

Before:
```
final LocatableBlock locatable = LocatableBlock.builder()
            .location(new Location<>(apiWorld, pos.getX(), pos.getY(), pos.getZ()))
            .state((BlockState) state)
            .build();
```

Solution 1:
```
final LocatableBlock locatable = LocatableBlock.builder()
            .state((BlockState) state)
            .position(pos.getX(), pos.getY(), pos.getZ())
            .world(apiWorld)
            .build();
```

Solution 2:
```
final LocatableBlock locatable = LocatableBlock.builder()
            .location(new Location<World>((World) this.world, this.pos.getX(), this.pos.getY(), this.pos.getZ()))
            .build();
```

I'm leaning towards the first one (It looks clearer). Suggestions are welcomed.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
ImMorpheus: Yep, I've just commited other local changes I was testing. I will add this in a few commits.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldEntitySpawner.java
ImMorpheus: Doesn't `getRandomChunkPosition` return a valid blockpos ?

src/main/java/org/spongepowered/common/mixin/core/world/MixinExplosion.java
ImMorpheus: Because there's no need to check the blockstate if `this.random.nextInt(3)` is not 0. I've moved the fastest condition first. I will add a comment.

src/main/java/org/spongepowered/common/mixin/core/world/MixinExplosion.java
ImMorpheus: > You're potentially changing the passed in block state, and therefor changing vanilla mechanics.

I hope not, because `iblockstate` and `block` (`Block block = iblockstate.getBlock();`) are used here https://github.com/SpongePowered/SpongeCommon/blob/d1b8572f639d71267955becc8e2e263fa4f34770/src/main/java/org/spongepowered/common/mixin/core/world/MixinExplosion.java#L377-L378

to check if the block can be dropped. Either we can reuse iblockstate or we need to update iblockstate before L377 because that would be a bug in vanilla mc.

> but should comment that you're making that change since it's an overwrite.

Whoops, I missed that

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
ImMorpheus: I've just moved this from 
https://github.com/ImMorpheus/SpongeCommon/blob/115f88f4245b912e07b3040d0a6463c02858889c/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java#L699

src/main/java/org/spongepowered/common/world/SpongeProxyBlockAccess.java
Zidane: @gabizou can he just remove it then?

gabizou:
> I'm leaning towards the first one (It looks clearer). Suggestions are welcomed.

That's the better one to use to be honest. Since we already have the block state instance for the method to process, we should be using the already provided values in these cases.

src/main/java/org/spongepowered/common/world/SpongeProxyBlockAccess.java
gabizou: Yep, can be removed.

ImMorpheus: There's also a Mixin of this class (thus an entry in mixins.forge.core.json). I think I will leave this for the OCD issue.

gabizou:
Also, why is this targeting bleeding? Performance should be targeting stable-7.

ImMorpheus:
Trying to improve `MixinChunk_Tracker` caused https://github.com/SpongePowered/SpongeCommon/issues/1896 in the past.
I've tested it a lot, but I'm looking for feedback.

Before this PR:


before


This PR:

post

ImMorpheus:
~qa

src/main/java/org/spongepowered/common/event/tracking/TrackingUtil.java
```diff
@@ -792,7 +792,9 @@ private static void spawnEntitiesForBlock(Collection<net.minecraft.entity.Entity
                     }
                 }
                 final ChangeBlockEvent.Post post = ((IPhaseState) state).createChangeBlockPostEvent(context, transactions);
-                SpongeImpl.postEvent(post);
+                if (ShouldFire.CHANGE_BLOCK_EVENT_POST) {
```
Zidane, line 792: Only create the event within the ShouldFire

src/main/java/org/spongepowered/common/mixin/core/network/datasync/MixinEntityDataManager.java
```diff
@@ -76,12 +81,12 @@
...
         final T currentValue = dataentry.getValue();
         final T incomingValue = value;
-        if (ObjectUtils.notEqual(value, currentValue)) { // Sponge - change dataentry.getValue() to use local variable
+        if (!value.equals(currentValue)) { // Sponge - change dataentry.getValue() to use local variable
             // Sponge Start - retrieve the associated key, if available
             // Client side can have an entity, because reasons.......
             // Really silly reasons......
             // I don't know, ask Grum....
-            if (this.entity != null && this.entity.world != null && !this.entity.world.isRemote) { // We only want to spam the server world ;)
+            if (ShouldFire.CHANGE_DATA_HOLDER_EVENT_VALUE_CHANGE && this.entity != null && this.entity.world != null && !this.entity.world.isRemote) { // We only want to spam the server world ;)
```
Zidane, line 89: We should only do the code within this method if this is true. Otherwise skip all of this to be as fast as the engine was.

ImMorpheus: This is an Overwrite.

Even if the event is not fired, what is after //Sponge End need to be executed

src/main/java/org/spongepowered/common/event/tracking/TrackingUtil.java
ImMorpheus: Then I would've to return null if the event is not fired.

That would mean that `//Means that we have had no actual block changes apparently?`

https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/event/tracking/TrackingUtil.java#L496-L500

src/main/java/org/spongepowered/common/mixin/core/world/storage/MixinWorldInfo.java
```diff
@@ -158,7 +162,8 @@
     private boolean generateBonusChest;
     private NBTTagCompound spongeRootLevelNbt = new NBTTagCompound(), spongeNbt = new NBTTagCompound();
     private final NBTTagList playerUniqueIdNbt = new NBTTagList();
-    private final BiMap<Integer, UUID> playerUniqueIdMap = HashBiMap.create();
+    private final Object2IntOpenHashMap<UUID> playerIdMap = new Object2IntOpenHashMap<>();
```
Cybermaxke, line 165: Default value should be set to `-1`, `0` is a valid uuid index (which is the default).

gabizou:
I'm curious about the moving from traditional maps to fastutil maps. Some of the map designs are great for large data sets, and others I've seen cause performance degradation due to how the maps are used. I understand this has been tested extensively, but I'm curious whether there's enough gain in certain map usages that would benefit from having the map changes.

src/main/java/org/spongepowered/common/mixin/core/network/datasync/MixinEntityDataManager.java
```diff
@@ -115,4 +120,22 @@
...
+        try {
+            dataentry = (EntityDataManager.DataEntry)((Int2ObjectOpenHashMap) this.entries).get(key.getId());
+        } catch (Throwable throwable) {
+            CrashReport crashreport = CrashReport.makeCrashReport(throwable, "Getting synched entity data");
+            CrashReportCategory crashreportcategory = crashreport.makeCategory("Synched entity data");
+            crashreportcategory.addCrashSection("Data ID", key);
+            throw new ReportedException(crashreport);
+        }
+
+        this.lock.readLock().unlock();
```
Deamon5550, line 138: this should be in a finally block

Deamon5550 (changes requested):

ImMorpheus:
I'm going to split this into separate PRs.

PR #2033, opened 2018-08-01
Selector updates and improvements

dualspiral:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1883) | **SpongeCommon**

Previous PRs: [Original API PR](https://github.com/SpongePowered/SpongeAPI/pull/1270) | [Original Common PR](https://github.com/SpongePowered/SpongeCommon/pull/793) | [Revised API PR](https://github.com/SpongePowered/SpongeAPI/pull/1645) | [Revised Common PR](https://github.com/SpongePowered/SpongeCommon/pull/1519)

With thanks to @kenzierocks and @rexbut for their contributions, however, as both of these PRs went stale (sorry!) and I've had no response on https://github.com/SpongePowered/SpongeCommon/pull/1519, I figured it best to open the PR, get it reviewed and in as there are issues with selectors selecting the wrong (or too many) things as it stands.

This PR does the following from previous PRs:

> * Adds new selector 1.12
> * Improvements and fixes for selectors based on 1.9 and my own review of the code.
> * <s>Selectors now use a List</s>, the resolver returns the elements ordered properly as well as the right number of elements. Removed the force argument as I don't remember what it was really useful for.

I've done a review of the code and written a test plugin with a command (`/test-selector`) to check selectors. I have fixed the following issues:

* API is compatible with API 7, selectors use the fact `ImmutableSet` is ordered. This should be changed to a list or some sort of sorted set for API 8.
* Fixed dimension check from checking that its bounding box intersects itself, rather than checking it intersects the entity's.
* Fixed class cast exception when using `@e` on the console

The test plugin lists the entities selected by the provided selector, along with their location and distance from the origin. Players are highlighted in yellow in the list.

### DIFF
--- a/src/main/java/org/spongepowered/common/text/selector/SelectorResolver.java
+++ b/src/main/java/org/spongepowered/common/text/selector/SelectorResolver.java
@@ -28,9 +28,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.flowpowered.math.vector.Vector3d;
 import com.flowpowered.math.vector.Vector3i;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
-
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.command.CommandSource;
 import org.spongepowered.api.data.manipulator.mutable.DisplayNameData;
@@ -97,7 +95,8 @@ public class SelectorResolver {
         return null;
     }
 
-    private final CommandSource origin;
+    @Nullable private final CommandSource origin;
+    @Nullable private final Entity entityOrigin;
     private final Collection<Extent> extents;
     private final Vector3d position;
     private final Selector selector;
@@ -115,6 +114,11 @@ public class SelectorResolver {
         this.selector = checkNotNull(selector);
         this.extents = ImmutableSet.copyOf(extents);
         this.origin = origin;
+        if (this.origin instanceof Entity) {
+            this.entityOrigin = (Entity) origin;
+        } else {
+            this.entityOrigin = null;
+        }
         this.position = position == null ? Vector3d.ZERO : position;
         this.selectorFilter = makeFilter();
     }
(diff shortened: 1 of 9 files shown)


src/main/java/org/spongepowered/common/util/OptionalUtils.java
```diff
@@ -24,17 +24,15 @@
...
+import com.google.common.collect.ImmutableSet;
 
 public class OptionalUtils {
 
-    public static <E> Collection<E> asSet(Optional<E> opt) {
-        if (opt.isPresent()) {
-            return Collections.singleton(opt.get());
-        }
-        return Collections.emptySet();
+    public static <E> Set<E> asSet(Optional<E> opt) {
```
kashike: Why are you changing this to `ImmutableSet`?

dualspiral: It was in the original PR - 
https://github.com/SpongePowered/SpongeCommon/pull/793/files#diff-c2ae1593cc5f5d014b6d276969856f4cR35

That said, it does make sense that a method labelled “asSet” returns a set rather than a collection.

kashike: https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#singleton-T-

dualspiral: I was referring to the signature change.

kashike: I seem to have clicked on the wrong line when commenting - the signature change is fine, I'm referring to the change from `Collections` to `ImmutableSet`.

dualspiral: I know you were, considering the signature didn’t change to an immutable set. My second comment was about the signature, not about the contents, I just wasn’t clear.

kashike: :tada:

parlough (approved):

dualspiral:
Before this gets merged, any other comments? Specifically, @kashike, would you like the asSet change reverted? Either way is fine!

dualspiral:
(the answer was yes, revert the change!)

kashike:
x

dualspiral:
The change was made, but it turns out that class isn't even used, at least not in SF/SC. I've left it there in case it's used in SV.

I have also squashed commits. For API, in order to ensure there is no breakages on an individual commit, I've merged them all into one, but included credit in the commit message.

Ready to go.

src/main/java/org/spongepowered/common/text/selector/SelectorResolver.java
```diff
@@ -241,147 +205,184 @@ private void addNameFilters(List<Predicate<Entity>> filters) {
...
+            return ImmutableSet.of(((Locatable) this.origin).getWorld());
         }
         return ImmutableSet.copyOf(this.extents);
     }
 
-    private Collection<ArgumentType<?>> getArgumentTypes(Collection<Argument<?>> arguments) {
-        Collection<ArgumentType<?>> types = Sets.newHashSet();
-        types.addAll(arguments.stream().map(Argument::getType).collect(Collectors.toList()));
-        return types;
+    private static AABB getAABB(Vector3i pos, int x, int y, int z) {
```
kashike, line 376: The local variables in this method should be named.

kashike (changes requested):

src/main/java/org/spongepowered/common/text/selector/SelectorResolver.java
dualspiral: All sorted.

kashike (approved):

kashike:
~ready

frank89k:
so what i need to do to let it work on my server ?

PR #2054, opened 2018-08-08
 Fix the bug that you can't read custom data in data class before postInit event

### LINKED ISSUE #2047: Access custom data inside custom smelting recipe failed
**I am currently running**

- SpongeForge version: 3374
- Java version: OpenJDK 64-Bit Server VM, version 1.8.0_171
- Operating System: Linux

- Plugins/Mods:

**Issue Description**
Use custom data inside custom smelting recipe will fail

```
[19:17:22] [Server thread/ERROR] [Sponge]: Could not pass FMLPreInitializationEvent to Plugin{id=example, name=Example Project, version=1.0, source=./mods/g.jar}
        at com.google.common.collect.ImmutableList.copyOf(ImmutableList.java:257) ~[minecraft_server.1.12.2.jar:?]
        at org.spongepowered.common.data.SpongeManipulatorRegistry.getDelegate(SpongeManipulatorRegistry.java:330) ~[SpongeManipulatorRegistry.class:1.12.2-2705-7.1.0-BETA-3374]
        at org.spongepowered.common.data.util.DataUtil.getBaseValueProcessor(DataUtil.java:437) ~[DataUtil.class:1.12.2-2705-7.1.0-BETA-3374]
        at net.minecraft.item.ItemStack.get(ItemStack.java:1639) ~[aip.class:?]
        at org.spongepowered.common.item.inventory.SpongeItemStackSnapshot.get(SpongeItemStackSnapshot.java:313) ~[SpongeItemStackSnapshot.class:1.12.2-2705-7.1.0-BETA-3374]
        at me.mmis1000.test.Main$onPreInit$$inlined$apply$lambda$1.getResult(Main.kt:115) ~[Main$onPreInit$$inlined$apply$lambda$1.class:?]
        at net.minecraft.item.crafting.FurnaceRecipes.register(FurnaceRecipes.java:657) ~[akp.class:?]
        at net.minecraft.item.crafting.FurnaceRecipes.register(FurnaceRecipes.java:555) ~[akp.class:?]
        at me.mmis1000.test.Main.onPreInit(Main.kt:101) ~[Main.class:?]
        at org.spongepowered.common.event.listener.GamePreInitializationEventListener_Main_onPreInit2.handle(Unknown Source) ~[?:?]
```

It seems [getDelegate(Key<?> key)](https://github.com/SpongePowered/SpongeCommon/blob/9ee7075d2842763be7993afc7e7c67d9c9d591da/src/main/java/org/spongepowered/common/data/SpongeManipulatorRegistry.java#L326-L333) try to read data from a map in temp registry and alway assume it can get data from valueProcessorMap, but it is actully nullable, and copy `null` cause crash.

[Minimal setup to reproduce issue](https://gist.github.com/mmis1000/85701bbadc5e9703edeafbbc88e8e2a1)

mmis1000:
`this.tempRegistry.valueProcessorMap.get(key)` is nullable and should be checked.  
Not every custom data class is associated with data processor.  
So, `this.tempRegistry.valueProcessorMap.get(key)` may return null.  
And `ImmutableList.copyOf(null)` will throw `NullPointerException` and panic the plugin.

This PR fix the problem that custom data is unreadable before the `SpongeManipulatorRegistry` being completely initialized.

Fixes #2047

### DIFF
--- a/src/main/java/org/spongepowered/common/data/SpongeManipulatorRegistry.java
+++ b/src/main/java/org/spongepowered/common/data/SpongeManipulatorRegistry.java
@@ -326,13 +326,13 @@ public class SpongeManipulatorRegistry implements SpongeAdditionalCatalogRegistr
     @Nullable
     public ValueProcessor<?, ?> getDelegate(Key<?> key) {
         if (this.tempRegistry != null) {
+            // During soft registrations
             List<ValueProcessor<?, ?>> list = this.tempRegistry.valueProcessorMap.get(key);
 
             if (list == null) {
                 return null;
             }
 
-            // During soft registrations
             return new ValueProcessorDelegate(key, ImmutableList.copyOf(list));
         }
         return this.valueDelegates.get(key);
--- a/src/main/java/org/spongepowered/common/data/SpongeManipulatorRegistry.java
+++ b/src/main/java/org/spongepowered/common/data/SpongeManipulatorRegistry.java
@@ -326,8 +326,14 @@ public class SpongeManipulatorRegistry implements SpongeAdditionalCatalogRegistr
     @Nullable
     public ValueProcessor<?, ?> getDelegate(Key<?> key) {
         if (this.tempRegistry != null) {
+            List<ValueProcessor<?, ?>> list = this.tempRegistry.valueProcessorMap.get(key);
+
+            if (list == null) {
+                return null;
+            }
+
             // During soft registrations
-            return new ValueProcessorDelegate(key, ImmutableList.copyOf(this.tempRegistry.valueProcessorMap.get(key)));
+            return new ValueProcessorDelegate(key, ImmutableList.copyOf(list));
         }
         return this.valueDelegates.get(key);
     }


src/main/java/org/spongepowered/common/data/SpongeManipulatorRegistry.java
```diff
@@ -327,7 +327,11 @@ private SpongeManipulatorRegistry() {
     public ValueProcessor<?, ?> getDelegate(Key<?> key) {
         if (this.tempRegistry != null) {
             // During soft registrations
-            return new ValueProcessorDelegate(key, ImmutableList.copyOf(this.tempRegistry.valueProcessorMap.get(key)));
+            if (this.tempRegistry.valueProcessorMap.containsKey(key)) {
```
gabizou: Make it faster simply by calling `get` once and checking if the processor is null.

gabizou (changes requested):
minor optimization, but worthy to keep it working in startup stages.

src/main/java/org/spongepowered/common/data/SpongeManipulatorRegistry.java
mmis1000: fixed

XakepSDK:
This is breaking-change in impl. Old impl. throws exception, if list is null.

mmis1000:
But this method is nullable by design, since valueProcessor does not always exist for every custom type.
Why you should make the temporary method throw exception when the formal one never does?

```
 return this.valueDelegates.get(key);
```

The formal one always return something and never throws

So instead break old behaviour, it actually the behaviour consistent between different stage.

mmis1000:
@gabizou Hello, I fixed the problem you mentioned, is there still any problem in this PR?

ImMorpheus (approved):
Working as intented

ImMorpheus:
How to reproduce:

```
@Listener
public void onPreInit(GamePreInitializationEvent event) {
    ItemStack.of(ItemTypes.SPONGE, 1).createSnapshot().createStack()
            .offer(MyHomes.FRIENDS, Arrays.asList(UUID.randomUUID(), UUID.randomUUID()));
}
```
(MyHomes is part of the myhomes testplugin)

gabizou (approved):

PR #2055, opened 2018-08-09
Optimize Vanilla map items

Aaron1011:
Background:

Vanilla map items are *absurdly* inefficient. Consider this Spark profile: https://sparkprofiler.github.io/?ITi5VNQSFC. 36% of the server time was spent calling `InvemtoryPlayer#hasItemStack` directly as a result of map items.

Several factors contribute to the extraordinary inefficiency of Vanilla maps:

* The number of calls per second to `hasItemStack` is proprotional to `N^2 * F`, where `N` is the number of online players, and `F` is the number of ItemFrames. In `EntityTrackerEntry`, `MapData#updateVisiblePlayers` is called for each online player, for every `ItemFrame` entity in the game. Inside `Map#updateVisiblePlayers`, `hasItemStack` is called for the passed in player, and then *for every other player in the world*
* Vanilla runs `MapData#updateVisiblePlayers` for every map `ItemStack`, not for every unique `MapData` stored in `MapStorage`.

This PR makes a number of changes to improve map performance:

* All calls to `hasItemStack` made from `updateVisiblePlayers` are eliminated. In Vanilla, there are only two places where `updateVisiblePlayers` is callled:
  * From an `ItemMap`, during the ticking of the items in a player's inventory. At this point, we know for certain that the player's inventory contains the itemstack
  * From an `EntityItemFrame`. We don't care whether or not any players contain the `ItemStack`, since we can update them separately.

Map update logic now works like this:
 
1. When an `ItemMap` is ticked from a player's inventory, we set a flag in the corresponding `MapData`, indicating that the player has an active map.
2. When the entity tracker sends updates for an `EntityItemFrame`, we update its decoration in its corresponding `MapData`, if it currently holds a filled map.
2. Once per tick, we call a new method - MixinMapData#tickMap. This checks and clears the flags set in the previous step. If a flag is set, then we know the player has a map in their inventory. If the map remains in their inventory, the flag will be set back to `true` during the next tick, when the map item ticks again.

For the overwhelming majority of servers, these changes should be completely transparent. However, on the off chance that another coremod may try to modify these classes, I've made these mixins part of the `Optimization` category. Should an issue arise, they can be turned off, restoring Vanilla behavior.


src/main/java/org/spongepowered/common/mixin/core/world/storage/MixinMapStorage.java
```diff
@@ -0,0 +1,46 @@
...
+import net.minecraft.world.storage.WorldSavedData;
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.common.interfaces.IMixinMapStorage;
+
+import java.util.List;
+
+@Mixin(MapStorage.class)
+public abstract class MixinMapStorage implements IMixinMapStorage {
```
gabizou: Could this just be an accessor mixin?

src/main/java/org/spongepowered/common/mixin/optimization/mapoptimization/MixinEntityItemFrame.java
```diff
@@ -0,0 +1,58 @@
...
+import org.spongepowered.common.mixin.core.entity.MixinEntityHanging;
+
+@Mixin(EntityItemFrame.class)
+public abstract class MixinEntityItemFrame extends MixinEntityHanging {
+
+    @Shadow public abstract ItemStack getDisplayedItem();
+
+    @Inject(method = "setDisplayedItemWithUpdate", at = @At(value = "HEAD"))
+    private void onSetItem(ItemStack stack, boolean p_174864_2_, CallbackInfo ci) {
+        if (this.world.isRemote) {
```
gabizou: Should be `((IMixinWorld) this.world).isFake()`

src/main/java/org/spongepowered/common/mixin/optimization/mapoptimization/MixinEntityTrackerEntry.java
```diff
@@ -0,0 +1,55 @@
...
+public abstract class MixinEntityTrackerEntry {
+
+    @Shadow @Final private Entity trackedEntity;
+
+    @Redirect(method = "updatePlayerList", at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;getItem()Lnet/minecraft/item/Item;", ordinal = 0))
+    private Item onGetItem(ItemStack itemStack) {
+        if (itemStack.getItem() instanceof ItemMap) {
+            ((IMixinMapData) ((ItemMap) itemStack.getItem()).getMapData(itemStack, this.trackedEntity.world)).updateItemFrameDecoration((EntityItemFrame) this.trackedEntity);
+        }
+        // We return 'null' to prevent the 'if' block from executing
```
gabizou: Can you javadoc this comment instead of just the comment?

src/main/java/org/spongepowered/common/mixin/optimization/mapoptimization/MixinMapInfo.java
```diff
@@ -0,0 +1,46 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.mixin.optimization.mapoptimization;
+
+import net.minecraft.world.storage.MapData;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.common.interfaces.world.IMixinMapInfo;
+
+@Mixin(MapData.MapInfo.class)
+public class MixinMapInfo implements IMixinMapInfo {
```
gabizou: Likewise this could be an accessor mixin.

src/main/resources/mixins.common.optimization.json
```diff
@@ -11,6 +11,12 @@
         "block.MixinBlockRedstoneWire",
         "enchantment.MixinEnchantmentHelper_No_Source_Leak",
         "entity.MixinEntityTameable_Cached_Owner",
+        "mapoptimization.MixinEntityItemFrame",
```
gabizou: I personally prefer the suffixed names so open class lookups aren't so confusing or annoying to have to read the fine print to know which one is which.

gabizou (changes requested):
Good optimizations. I look forward to seeing how this plays out with some stress testing.

src/main/java/org/spongepowered/common/mixin/optimization/mapoptimization/MixinMapInfo.java
Aaron1011: This one actually can't be, since I'm adding an new field.

Aaron1011:
@gabizou I've made the changes you requested.

Aaron1011:
Merged in https://github.com/SpongePowered/SpongeCommon/commit/74573a2d0af6facdabfa58973edf1ad2eb181024

PR #2062, opened 2018-08-17
Fix DropItemEvent.Dispense being fired instead of DropItemEvent.Destruct

RedNesto:
Fixes SpongePowered/SpongeForge#2236

Tested with:
- SpongeVanilla stable-7 and bleeding
- SpongeForge bleeding


RedNesto:
The commit 4cb23b2cd1b0fd179cb8b569c0dfd12c9e74fc9b silently fixed the issue this PR was meant to fix so it is useless now.

PR #2074, opened 2018-08-25
Fix that mod entities are inactive at some ticks

ustc-zzzz:
Please refer to SpongePowered/SpongeForge#2374 and ChaosBuffalo/MKUltra#69.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/plugin/entityactivation/EntityActivationRange.java
+++ b/src/main/java/org/spongepowered/common/mixin/plugin/entityactivation/EntityActivationRange.java
@@ -65,6 +65,7 @@ import org.spongepowered.common.config.SpongeConfig;
 import org.spongepowered.common.config.category.EntityActivationModCategory;
 import org.spongepowered.common.config.category.EntityActivationRangeCategory;
 import org.spongepowered.common.config.type.GeneralConfigBase;
+import org.spongepowered.common.entity.EntityUtil;
 import org.spongepowered.common.entity.SpongeEntityType;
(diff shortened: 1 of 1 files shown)


phit:
am I reading this right, always tick modded entities and making entity activation useless for them? if so thats a big please no for me just to fix compat with some mods, if anything this should be behind a config thats disabled by default

ustc-zzzz:
@phit There is a comment near the modified code of my Pull Request:

> Should never happen but just in case for mods, always tick

Therefore, it seem that sponge has never been thinking about skipping update ticks for mod entities. Force skipping some specific ticks seems to be weird and also not good for mod compatibility. Vanilla Minecraft is under the control but mods are not obligated to decide whether an update is important or not just because sponge says so.

If such a requirement does exist, it should be implemented by a mod or a sponge plugin. It is not necessary that a mod should be compatible with another, but sponge has such a mission.

dualspiral:
**We already have a config section for this**, if you set `entity-activation-range.auto-populate` to `true`, play a bit, the entity types that that are affected should be listed in the `entity-activation-range.mods`, where you can then disable this on a per entity setting. If that doesn't work, then we need to fix that.

> but mods are not obligated to decide whether an update is important or not just because sponge says so

No, this is why EAR is a mixin plugin which can be disabled _and_ we provide a config entry to tweak this. We're well aware that some mods may have trouble with this.

Doing this is like using a sledgehammer to crack a walnut. We shouldn't be destroying this feature because one mod is having issues, the majority do not.

ustc-zzzz:
I am so sorry for misunderstanding this. It's my fault.

PR #2079, opened 2018-08-28
Fix ProjectileSource being UNKNOWN when launching with a dispenser

RedNesto:
Tested with a dispenser throwing fire charges and a ghast.
All of those cases work as intended.

I hope the method is cleaner too now.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/projectile/MixinEntityFireball.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/projectile/MixinEntityFireball.java
@@ -38,24 +38,28 @@ import org.spongepowered.common.entity.projectile.ProjectileSourceSerializer;
 import org.spongepowered.common.event.SpongeCommonEventFactory;
 import org.spongepowered.common.mixin.core.entity.MixinEntity;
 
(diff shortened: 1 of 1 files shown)


ImMorpheus:
~qa

src/main/java/org/spongepowered/common/mixin/core/entity/projectile/MixinEntityFireball.java
```diff
@@ -38,24 +38,28 @@
...
-                this.projectileSource = (ProjectileSource) this.shootingEntity;
-            } else {
-                this.projectileSource = ProjectileSource.UNKNOWN;
-            }
+        if (this.shootingEntity instanceof ProjectileSource) {
+            return (ProjectileSource) this.shootingEntity;
+        }
+
+        if (this.projectileSource != null) {
+            return projectileSource;
```
ImMorpheus: `this`

ImMorpheus (changes requested):

ImMorpheus (approved):

ImMorpheus (approved):
Working as intended

kashike (approved):

PR #2080, opened 2018-08-28
Make SignSource great again!

RedNesto:
The actual Mixin target was only the first anonymous implementation, and not the most important one, for players clicking signs.
Meaning the CommandSender will never be a SignSource, but a WrapperCommandSource.Located, making it impossible to get the original CommandSource (the Player who clicked the sign).

This commit now makes all references target the two anonymous implementations,
and makes possible to have a SignSource as source of a command.

I found this issue while investigating an issue someone was having in the dev Discord channel.

You can get your own [test plugin here](https://gist.github.com/RedNesto/8d138fd1e62ef90e7d912f847992f376)!

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/api/command/BlockCommandSourceMixin_API.java
+++ b/src/main/java/org/spongepowered/common/mixin/api/command/BlockCommandSourceMixin_API.java
@@ -27,16 +27,13 @@ package org.spongepowered.common.mixin.api.command;
 import net.minecraft.entity.item.EntityMinecartCommandBlock;
 import net.minecraft.tileentity.TileEntityCommandBlock;
 import org.spongepowered.api.command.source.CommandBlockSource;
-import org.spongepowered.api.service.permission.PermissionService;
-import org.spongepowered.api.util.Tristate;
 import org.spongepowered.api.util.annotation.NonnullByDefault;
 import org.spongepowered.asm.mixin.Mixin;
-import org.spongepowered.common.bridge.command.CommandSenderBridge;
 import org.spongepowered.common.bridge.command.CommandSourceBridge;
-import org.spongepowered.common.bridge.permissions.SubjectBridge;
 
 @NonnullByDefault
-@Mixin(value = {TileEntityCommandBlock.class, EntityMinecartCommandBlock.class}, targets = "net/minecraft/tileentity/TileEntitySign$1")
+@Mixin(value = {TileEntityCommandBlock.class, EntityMinecartCommandBlock.class},
+        targets = {"net/minecraft/tileentity/TileEntitySign$1", "net/minecraft/tileentity/TileEntitySign$2"})
 public abstract class BlockCommandSourceMixin_API implements CommandBlockSource {
 
     @Override
--- a/src/main/java/org/spongepowered/common/mixin/api/command/CommandSourceMixin_API.java
+++ b/src/main/java/org/spongepowered/common/mixin/api/command/CommandSourceMixin_API.java
@@ -35,16 +35,15 @@ import org.spongepowered.api.command.CommandSource;
 import org.spongepowered.api.entity.living.player.User;
 import org.spongepowered.api.text.Text;
 import org.spongepowered.api.text.channel.MessageChannel;
-import org.spongepowered.asm.mixin.Intrinsic;
 import org.spongepowered.asm.mixin.Mixin;
-import org.spongepowered.common.bridge.command.CommandSenderBridge;
 import org.spongepowered.common.bridge.command.CommandSourceBridge;
 import org.spongepowered.common.text.SpongeTexts;
 
 import java.util.Optional;
 
 @Mixin(value = {EntityPlayerMP.class, TileEntityCommandBlock.class, EntityMinecartCommandBlock.class, MinecraftServer.class, RConConsoleSource.class},
-        targets = "net/minecraft/tileentity/TileEntitySign$1", priority = 899) // Allows for other implementations of CommandSource to override/write
+        targets = {"net/minecraft/tileentity/TileEntitySign$1", "net/minecraft/tileentity/TileEntitySign$2"},
+        priority = 899) // Allows for other implementations of CommandSource to override/write
 public abstract class CommandSourceMixin_API implements CommandSource {
 
     private MessageChannel channel = MessageChannel.TO_ALL;
--- a/src/main/java/org/spongepowered/common/mixin/api/mcp/tileentity/TileEntitySignCommandSendersMixin_API.java
+++ b/src/main/java/org/spongepowered/common/mixin/api/mcp/tileentity/TileEntitySignCommandSendersMixin_API.java
@@ -1,0 +1,66 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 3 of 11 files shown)


src/main/java/org/spongepowered/common/mixin/core/command/MixinCommandBlockSource.java
```diff
@@ -36,7 +36,7 @@
 import org.spongepowered.common.interfaces.IMixinSubject;
 
 @NonnullByDefault
-@Mixin(value = {TileEntityCommandBlock.class, EntityMinecartCommandBlock.class}, targets = IMixinCommandSender.SIGN_CLICK_SENDER)
+@Mixin(value = {TileEntityCommandBlock.class, EntityMinecartCommandBlock.class}, targets = {IMixinCommandSender.SIGN_CLICK_SENDER_1, IMixinCommandSender.SIGN_CLICK_SENDER_2})
```
ImMorpheus: Move `targets` to a new line like in `MixinCommandSource`

src/main/java/org/spongepowered/common/mixin/core/command/MixinICommandSender.java
```diff
@@ -36,7 +36,7 @@
 import org.spongepowered.common.command.CommandPermissions;
 import org.spongepowered.common.interfaces.IMixinCommandSender;
 
-@Mixin(value = {EntityPlayerMP.class, MinecraftServer.class, RConConsoleSource.class, CommandBlockBaseLogic.class}, targets = IMixinCommandSender.SIGN_CLICK_SENDER)
+@Mixin(value = {EntityPlayerMP.class, MinecraftServer.class, RConConsoleSource.class, CommandBlockBaseLogic.class}, targets = {IMixinCommandSender.SIGN_CLICK_SENDER_1, IMixinCommandSender.SIGN_CLICK_SENDER_2})
```
ImMorpheus: Move `targets` to a new line like in `MixinCommandSource`

ImMorpheus (changes requested):

src/main/java/org/spongepowered/common/mixin/core/command/MixinCommandBlockSource.java
```diff
@@ -36,7 +36,8 @@
 import org.spongepowered.common.interfaces.IMixinSubject;
 
 @NonnullByDefault
-@Mixin(value = {TileEntityCommandBlock.class, EntityMinecartCommandBlock.class}, targets = IMixinCommandSender.SIGN_CLICK_SENDER)
+@Mixin(value = {TileEntityCommandBlock.class, EntityMinecartCommandBlock.class},
+       targets = {IMixinCommandSender.SIGN_CLICK_SENDER_1, IMixinCommandSender.SIGN_CLICK_SENDER_2})
```
ImMorpheus: `MixinCommandBlockSource` is a `CommandBlockSource` which 
```
Represents a CommandBlock source, either a placed block or a CommandBlockMinecart.
```


They should not be targeting IMixinCommandSender#SIGN_CLICK_SENDER_1 (and 2)

ImMorpheus (dismissed):

RedNesto:
~wip

Morph and I found strange things with CommandSources, will check if everything works as it should.

RedNesto:
Okay so there are some odds as to what should be a `CommandSource`: A `CommandBlock` is a `CommandSource` (its `ICommandSender` is `CommandBlockBaseLogic`), but not `Sign` (whose `ICommandSender` is `IMixinCommandSender#SIGN_CLICK_SENDER_2`, an anonymous class created when it have a command and is clicked by a player).

RedNesto:
~qa 
I guess it's fine now

gabizou:
I'd recommend adding the test plugin source as a proper test plugin into the PR (so we can do testing in the future to ensure it still works).

src/main/java/org/spongepowered/common/mixin/core/command/MixinSubject.java
```diff
@@ -63,7 +63,7 @@
  */
 @NonnullByDefault
 @Mixin(value = {EntityPlayerMP.class, TileEntityCommandBlock.class, EntityMinecartCommandBlock.class, MinecraftServer.class, RConConsoleSource.class,
```
ImMorpheus: Why is `TileEntityCommandBlock` (and EntityMinecartCommandBlock, I guess) a Subject ?

RedNesto: I'd like to know too, but I don't know enough about permissions to know what should be a Subject. 
I don't want to make this PR go beyond what it was meant for when I opened it - I don't want to make (breaking?) changes in this one, especially since I don't have enough knowledge about the Permissions API.

ImMorpheus: As discussed the other day, `TileEntityCommandBlock` and `EntityMinecartCommandBlock` should not be here. This should mixin into `CommandBlockBaseLogic.class`. Just like TileEntitySign is not the IMixinSubject.

src/main/java/org/spongepowered/common/mixin/api/command/SignCommandSourceMixin_API.java
```diff
@@ -34,15 +34,14 @@
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
-import org.spongepowered.common.bridge.command.CommandSenderBridge;
 import org.spongepowered.common.command.WrapperCommandSource;
 
-@Mixin(targets = "net/minecraft/tileentity/TileEntitySign$1")
-public abstract class TileEntitySign_1Mixin_API implements SignSource {
+@Mixin(targets = {"net/minecraft/tileentity/TileEntitySign$1", "net/minecraft/tileentity/TileEntitySign$2"})
+public abstract class SignCommandSourceMixin_API implements SignSource {
```
dualspiral: @Zidane Would appreciate your guidance on naming this one - how are we naming multi-target mixins? 

Also, I think this shouldn't have moved package as both **targets** are in `tilentity` so they were in the right place.

Zidane: Multi-targets are odd. It is on a case by case basis and basically I name them given a theme (or name them in reference to their bridge).

Also if all the targets exist in one package, leave it in that package.

RedNesto: I moved this mixin because it only implements SignSource, thought this was a better place based on its role.

dualspiral: It's not so much about its role, more the classes it mixes into. In this case, they are all in the `n.m.tileentity` package, so where it was before is the right place (`o.s.c.m.api.mcp.tileentity`).

The name? Probably `TileEntitySignAnonMixin_API` - it's only targetting anonymous classes in `TileEntitySign`. Alternatively, `SignSourceMixin_API`, naming after the `SignSource` interface you are implementing.

Naming is hard.

RedNesto: Since both targets are `ICommandSender` implementations the mixin could be named `TileEntitySignCommandSendersMixin_API`.

I'll move it back to `mcp.tileentity` once a name is decided.

dualspiral: That name would work for me.

Zidane: This looks fine to me too.

dualspiral (approved):
We finally settled on the naming!

dualspiral:
Thanks - merged in manually - f650b2238f7cf2e356450b3d43993543f6b8de9e

PR #2084, opened 2018-08-29
Implement UpdateAnvilEvent

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1898) | **SpongeCommon**

### DIFF
--- a/src/main/java/org/spongepowered/common/event/inventory/UpdateAnvilEventCost.java
+++ b/src/main/java/org/spongepowered/common/event/inventory/UpdateAnvilEventCost.java
@@ -1,0 +1,69 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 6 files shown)


src/main/java/org/spongepowered/common/mixin/core/item/inventory/MixinContainerRepair.java
```diff
@@ -63,4 +85,25 @@ public SlotProvider slotProvider(Fabric fabric, InventoryAdapter adapter) {
                 .add(36);
         return builder.build();
     }
+
+    @Inject(method = "updateRepairOutput",
+            locals = LocalCapture.CAPTURE_FAILSOFT,
+            cancellable = true,
+            at = @At(value = "INVOKE_ASSIGN", target = "Lnet/minecraft/item/ItemEnchantedBook;getEnchantments(Lnet/minecraft/item/ItemStack;)Lnet/minecraft/nbt/NBTTagList;"))
```
gabizou: I'm confused as to where this is being injected to. It sounds like it's injecting right after `getEnchantments(ItemStack)` is called, but before the `NBTTagList.isEmpty()` is called. I get that the map is not being modified here, but what's the point of the map if it's not being used? Couldn't this cancellable injection occur after the isEmpty and value assignment is made?

src/main/java/org/spongepowered/common/event/inventory/UpdateAnvilEventCost.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public int getContentVersion() {
+        return 1;
+    }
+
+    @Override
+    public DataContainer toContainer() {
+        return DataContainer.createNew().set(Queries.CONTENT_VERSION, this.getContentVersion())
+                .set(DataQuery.of("materialcost"), this.getMaterialCost())
```
gabizou: Make these queries static in `DataQueries`.

src/main/java/org/spongepowered/common/mixin/core/item/inventory/MixinContainerRepair.java
```diff
@@ -63,4 +87,28 @@ public SlotProvider slotProvider(Fabric fabric, InventoryAdapter adapter) {
                 .add(36);
         return builder.build();
     }
+
+    @Inject(method = "updateRepairOutput", at = @At(value = "RETURN"))
+    private void onUpdateRepairOutput(CallbackInfo ci) {
+        ItemStack itemstack = this.inputSlots.getStackInSlot(0);
+        ItemStack itemstack2 = this.inputSlots.getStackInSlot(1);
+        ItemStack result = this.outputSlot.getStackInSlot(0);
+        UpdateAnvilEvent event = SpongeCommonEventFactory.callUpdateAnvilEvent((ContainerRepair) (Object) this, itemstack, itemstack2, result, this.repairedItemName, this.maximumCost, this.materialCost);
```
gabizou, line 99: Maybe do a check if the event is being listened to with `ShouldFire`?

gabizou (changes requested):
Small changes.

PR #2110, opened 2018-09-27
Fix lastCauseSize getting set on the wrong index

LXGaming:
SpongeCauseStackManager::popCauseFrame uses old_min_depth variable from the StackFrame in order to reset the duplicate counter for the last cause.

However this is not guaranteed to be the last index as SpongeCauseStackManager::pushCauseFrame passes the min_depth variable to the StackFrame before it is incremented to the size of the cause meaning the cause can actually have objects in it prior to a stackframe being created.

The end result is duplicateCauses reporting an index having duplicate objects when it actually doesn't


gabizou:
Huh. Can you provide a test plugin perhaps to make sure this doesn't get reintroduced in the future?

LXGaming:
I've added the test plugin (In case the commits didn't notify)

Issue #2111, opened 2018-10-05
Cannot setup world Gamemodes [World Gamemode doesnt save - World Manager Problem?]


HaniXP:
**I am currently running**

- SpongeForge version: 1.12.2-2705-7.1.0-BETA-3477
- SpongeVanilla version: 
- Java version: Version 8 - 181
- Operating System: Windows 7 Ultimate (64 - bit)

- Plugins/Mods:
Plugins : 
Minecraft, Minecraft Coder Pack, SpongeAPI, SpongeForge, CatClearLag, EconomyLite, FlexibleLogin, GriefPrevention, LuckPerms, MMCRestrict, MagiBridge, NPCs, Nations, Nucleus, PlaceholderAPI, UltimateCore

Mods : 
Minecraft, Minecraft Coder Pack,
[...]

PR #2113, opened 2018-10-07
Fix GameMode not updating

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeCommon/issues/2111

### DIFF
--- a/src/main/java/org/spongepowered/common/registry/type/entity/GameModeRegistryModule.java
+++ b/src/main/java/org/spongepowered/common/registry/type/entity/GameModeRegistryModule.java
@@ -45,13 +45,4 @@ public final class GameModeRegistryModule extends MinecraftEnumBasedCatalogTypeM
         }
     }
 
-    public static GameType toGameType(GameMode gamemode) {
-        for (GameType gameType : GameType.values()) {
-            if (((GameMode) (Object) gameType).getId().equals(gamemode.getId())) {
-                return gameType;
-            }
-        }
-        return GameType.SURVIVAL;
-    }
-
 }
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/registry/type/entity/GameModeRegistryModule.java
```diff
@@ -47,7 +47,7 @@ protected void generateInitialMap() {
 
     public static GameType toGameType(GameMode gamemode) {
         for (GameType gameType : GameType.values()) {
-            if (gameType.getName().equalsIgnoreCase(gamemode.getId())) {
+            if (gameType.getName().equals(gamemode.getName())) {
```
dualspiral: Good spot.

Given that we guarantee IDs to be unique, and not names, should this not perform an equality check on the ID instead?

gabizou (approved):
Good catch. Fine to merge. Before RB.

src/main/java/org/spongepowered/common/registry/type/entity/GameModeRegistryModule.java
```diff
@@ -47,7 +47,7 @@ protected void generateInitialMap() {
 
     public static GameType toGameType(GameMode gamemode) {
         for (GameType gameType : GameType.values()) {
-            if (gameType.getName().equalsIgnoreCase(gamemode.getId())) {
+            if (((GameMode) (Object) gameType).getId().equals(gamemode.getId())) {
```
JBYoshi: Since Sponge implements `GameMode` as a [mixin](https://github.com/SpongePowered/SpongeCommon/blob/9d594880e16905f61784f37d0bc890da05981caf/src/main/java/org/spongepowered/common/mixin/core/data/types/MixinGameType.java#L43) to `GameType`, this method could actually be rewritten to simply `return (GameType) (Object) gamemode;`

JBYoshi (changes requested):

ImMorpheus:
Replaced the method with a simple cast (consistency with other setter, O(1) and looks cleaner)

JBYoshi (approved):

kashike (approved):

kashike:
Thanks, @ImMorpheus.

Issue #2120, opened 2018-10-15
Invalid return value of World#getBlockMax()


Lignium:
**I am currently running**

- SpongeVanilla version: spongevanilla-1.12.2-7.1.0-BETA-121
- Java version: 1.8.0_181
- Operating System: Kubuntu 18.04 Bionic
- Plugins/Mods: None

**Issue Description**

The `World#getBlockMax()` method returns incorrect value. Instead of (29999999, 255, 29999999), it returns (30000000, 256, 30000000), that is, at one larger. Along with that breaks the `World#getBlockSize()` method, which returns (60000001, 257, 60000001) instead of (60000000, 256, 60000000).

NamStalker:
From what I can tell, the World#getBlockMax() should return (30000000, 256, 30000000), but the World#getBlockSize() is actually wrong and is returning with the added 1 to all values. I can submit a PR to fix this, but it seems like the first method is returning as intended.

Let me know if I may misunderstand what it should be returning.

Lignium:
The default world length is 60M. 30M blocks for negative values (from -30M to -1 inclusive), and remaining 30M blocks for positive values with zero (from 0 to 29,999,999 inclusive). The fact is that we confuse coordinates of blocks and coordinates of points in the world. Coordinates of a block are given by the coordinate of it center + Vector3d(-0.5, -0.5, -0.5). This means that at a block with coordinates (29999999, 255, 29999999) the opposite vertex will be just at the coordinates (30M, 256, 30M). I hope clearly written. :)

NamStalker:
Ok, I see what happened there. I'll look into this once I get home. If anyone else wants to look into this or has more info, it is related to the change in #1958

PR #1958, opened 2018-06-26
Fix issues with world limit

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeCommon/issues/168
Fix https://github.com/SpongePowered/SpongeCommon/issues/651
Supersedes https://github.com/SpongePowered/SpongeCommon/pull/1612

PR #2123, opened 2018-10-19
Partially revert #1958 for world limit fix

dualspiral:
@ImMorpheus - I'm PRing this because I want to check that I've got this right before it lands on stable - I suspect (as others stated in #2120) that there was some confusion between block position and decimal position. Can you confirm whether this was an error, or if there is a reason for the changes that this reverts? Thanks!

---

Commit message:

A common source of confusion is the decimal co-ordinates vs block co-ordinates. It is stated that the minecraft X/Z co-ordinates have a range of +/-30,000,000, which is true, however, the block coordinates range from -30,000,000 to +29,999,999.

This is because the edges of blocks line up with the unit co-ordinates, such that a block spans from n to n+1 (where n is an integer). Thus, a block that we class as being at (29,999,999, 0, 29,999,999) actually spans to (30,000,000, 0, 30,000,000).

The block limit should be within the 30M radius around 0, meaning that the last possible block that can be interacted with is at 30M-1, as the block with block co-ordinate  X=30M actually strays past the Minecraft border to decimal co-ordinate 30M+1 - which is out of bounds.

This commit reverts the max block limit change.

Fixes https://github.com/SpongePowered/SpongeCommon/issues/2120

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
@@ -38,7 +38,6 @@ import com.google.common.collect.Lists;
 import net.minecraft.block.Block;
 import net.minecraft.block.BlockEventData;
 import net.minecraft.block.BlockPistonBase;
-import net.minecraft.block.ITileEntityProvider;
 import net.minecraft.block.material.Material;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.crash.CrashReportCategory;
@@ -170,8 +169,6 @@ import org.spongepowered.common.event.tracking.PhaseContext;
 import org.spongepowered.common.event.tracking.PhaseData;
 import org.spongepowered.common.event.tracking.PhaseTracker;
 import org.spongepowered.common.event.tracking.TrackingUtil;
-import org.spongepowered.common.event.tracking.phase.entity.BasicEntityContext;
-import org.spongepowered.common.event.tracking.phase.entity.EntityPhase;
 import org.spongepowered.common.event.tracking.phase.general.ExplosionContext;
 import org.spongepowered.common.event.tracking.phase.general.GeneralPhase;
 import org.spongepowered.common.event.tracking.phase.generation.GenerationPhase;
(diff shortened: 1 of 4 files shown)


gabizou (approved):

ImMorpheus:
This is breaking vanilla behaviour. See https://github.com/SpongePowered/SpongeCommon/pull/1958#issuecomment-401635540

dualspiral:
Sure, I can see that, but also considering that you can only place blocks one block outside the "invisible wall" on the positive edges, but not the negative edges makes me wonder what the real intention is. Regardless, that's a moot point.

Are we restricting block placements to within the "invisible wall" or do we strictly follow the vanilla behaviour, though it may be incorrect? I am happy either way - but are there any other consequences of having the max block one outside the invisible edge? I know I use it for some teleportation limits, and other plugin developers might use it for the edge of the _playable_ world. If so, I would rather see this reversion happen, possibly with a way to support the erroneous behaviour of this command in some other way.

src/main/java/org/spongepowered/common/mixin/core/world/MixinWorldServer.java
```diff
@@ -237,7 +233,7 @@
     private static final String PROFILER_ESS = "Lnet/minecraft/profiler/Profiler;endStartSection(Ljava/lang/String;)V";
 
     private static final Vector3i BLOCK_MIN = new Vector3i(-30000000, 0, -30000000);
-    private static final Vector3i BLOCK_MAX = new Vector3i(30000000, 256, 30000000);
+    private static final Vector3i BLOCK_MAX = new Vector3i(30000000, 256, 30000000).sub(Vector3i.ONE);
```
ImMorpheus: You can safely remove these and switch to the getter of the superclass. 

(They are only used inside checkBlockBounds)

ImMorpheus:
> Are we restricting block placements to within the "invisible wall" or do we strictly follow the vanilla behaviour, though it may be incorrect?

IMO this is a bug with vanilla minecraft (`CommandBase#parseCoordinate` passing 30000000, Y, 30000000 as a valid maxima for blockpos). 

`World#getBlockMax()` should be changed back to 29,999,999 or you need to modify the API as well

https://github.com/SpongePowered/SpongeAPI/blob/4b0b4024bc569ab0daacbea18d618baae23d7d19/src/main/java/org/spongepowered/api/world/extent/BlockVolume.java#L55-L61

ImMorpheus (approved):

PR #2124, opened 2018-10-19
Use Filesystem watchers to cache game profiles/users in the USS

dualspiral:
This is very WIP - based on some tinkering I've been doing. Not ready to merge yet, may not merge, but is currently at a proof of concept stage to try to improve server performance when thousands of game profiles exist - usually when tab completing a user argument on a command! I wanted to put this out now to get any early feedback on the idea - especially if I've missed something.

Basically, the idea behind this is after the first, initial, user file scan, we cache the stored UUIDs. We then setup a file system watcher to determine which UUIDs have had files added/removed/changed and therefore only perform operations on _those_ UUIDs - before, though we had the user cache, we still operated on every file.

I did consider the fact that I could have just made the line `final GameProfile profile = profileCache.getProfileByUUID(uuid);` only execute if we don't have a user in the cache (at the moment, we basically ignored this cache), that would have saved some performance. This is intended to save time otherwise spend on I/O and UUID parsing too. I also figured that if a user's profile changed, it might be that their username has changed with Mojang and has changed in the profile cache - so the lookup shouldn't require a call to Mojang servers in the vast majority of cases.

The elephant in the room is that for this cache to work, it needs to have run once! This needs thinking about - when do we populate the cache such that the first tab complete doesn't kill a server!

With one player (me) on my machine, getting all profiles was an 11ms operation the first time around, and went down to 1ms after the cache was built. I hope to get a bigger dataset and play about and see what further gains we can get.

### DIFF
--- a/src/main/java/org/spongepowered/common/service/user/UserDiscoverer.java
+++ b/src/main/java/org/spongepowered/common/service/user/UserDiscoverer.java
@@ -27,29 +27,43 @@ package org.spongepowered.common.service.user;
 import com.google.common.base.Preconditions;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.RemovalCause;
+import com.google.common.cache.RemovalNotification;
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
 import com.mojang.authlib.GameProfile;
 import net.minecraft.server.management.PlayerList;
 import net.minecraft.server.management.PlayerProfileCache;
 import net.minecraft.server.management.UserListBans;
+import net.minecraft.server.management.UserListBansEntry;
 import net.minecraft.server.management.UserListEntry;
-import net.minecraft.server.management.UserListEntryBan;
 import net.minecraft.server.management.UserListWhitelist;
 import net.minecraft.server.management.UserListWhitelistEntry;
 import net.minecraft.world.WorldServer;
-import net.minecraft.world.storage.SaveHandler;
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.entity.living.player.User;
+import org.spongepowered.api.profile.GameProfileCache;
 import org.spongepowered.api.profile.ProfileNotFoundException;
 import org.spongepowered.common.SpongeImpl;
-import org.spongepowered.common.entity.player.SpongeUser;
 import org.spongepowered.common.bridge.entity.player.EntityPlayerMPBridge;
+import org.spongepowered.common.bridge.world.storage.SaveHandlerBridge;
+import org.spongepowered.common.entity.player.SpongeUser;
 import org.spongepowered.common.mixin.core.server.management.UserLIstEntryAccessor;
 import org.spongepowered.common.mixin.core.server.management.UserListAccessor;
 import org.spongepowered.common.mixin.core.world.storage.SaveHandlerAccessor;
 import org.spongepowered.common.world.WorldManager;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.DirectoryStream;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardWatchEventKinds;
+import java.nio.file.WatchEvent;
+import java.nio.file.WatchKey;
+import java.nio.file.WatchService;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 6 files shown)


GDVE:
What about this?

dualspiral:
With further improvements, I got the first scan to 2ms, and the subsequent check to 0ms. What I need now is a set of test data with large numbers of players to test how fast, and stable, this has become.

If anyone wants to test this with a larger data set, please feel free and let me know how it goes. I'll try to find a good test set to run with myself.

src/main/java/org/spongepowered/common/service/user/UserDiscoverer.java
```diff
@@ -131,8 +164,9 @@ static User findByProfile(org.spongepowered.api.profile.GameProfile profile) {
         return user;
     }
 
+    @Nullable
     static User findByUsername(String username) {
-        User user = userByNameCache.getIfPresent(username);
+        User user = userByNameCache.getIfPresent(username.toLowerCase());
```
ImMorpheus: Why ?
You seem to lowercase the player name, but Dualspiral and dualspiral are not the same GameProfile.

dualspiral: I get that, but it's because the checks that we perform from the `UserStorageService` are case insensitive, as per the javadocs.

https://github.com/SpongePowered/SpongeAPI/blob/stable-7/src/main/java/org/spongepowered/api/service/user/UserStorageService.java#L50-L51

The key to the cache is effectively used as a search index, so this allows for faster matching - we don't have to iterate over all keys and do a case insensitive check, we just do hash matching.

ImMorpheus: > The key to the cache is effectively used as a search index, so this allows for faster matching - we don't have to iterate over all keys and do a case insensitive check, we just do hash matching.

This is perfectly fine. I have no problem with that.

I'm more concerned about the case-insensitive part.
To clarify: if we try to lookup dualspiral, the name is cached and the `User` pointing to dualspiral is returned. Then we try to lookup Dualspiral, the system will see the cached entry "dualspiral" and return the `User` pointing to dualspiral (which is wrong).
Is this correct ?
If yes, I think this needs more discussion unless I'm missing something.

dualspiral: Yes, this is what it does, but I must stress that I'm simply making the impl conform to what our Javadocs say. The only issue may arise when you use a `GameProfile` to lookup a user, but at that point, I use the UUID anyway.

dualspiral: Thinking about this some more, I'd be happy to update this to add a cache that has the case-sensitive username, and then a case insensitive cache (or multimap) that contains a list of users. I'm concerned it's a _little_ overkill, but it should alleviate the issue here that you raise.

ImMorpheus: > Yes, this is what it does, but I must stress that I'm simply making the impl conform to what our Javadocs say. 

I can see that, my point is that the implementation was "fine" as it was. 
Even if the implementation was not respecting the API (case-insensitive lookup) I think the way to go is to fix the javadocs.
This would result in no breaking-change since the implementation was already "broken".

dualspiral: > This would result in no breaking-change since the implementation was already "broken".

It would be. There were case insensitive checks below the line you pointed out which I hadn't removed. I'm just gonna go for the two cache approach at the moment.

dualspiral:
If anyone wants to help me test this, I have created a plugin version that can be downloaded at ~~https://github.com/dualspiral/USSCaching/releases/download/v1/USSSponge-1.0-SNAPSHOT.jar~~
~~https://github.com/dualspiral/USSCaching/releases/download/v2/USSSponge-1.0-SNAPSHOT.jar~~
https://github.com/dualspiral/USSCaching/releases/tag/v3

Remember, this is for testing purposes. If things go wrong (and they could), don't say I didn't warn you! 

This should work on the latest and recommended builds of Sponge 7.1.x, both SF and SV, allowing those not wanting to try the latest changes in the mainline builds to give this a go.

After starting your server, you should see the following messages: 

```
Starting init of user discoverer async

User discoverer init complete - time taken: xx.yyy ms
```

At that point, USS lookups should be much faster than now. I am also interested if people want to post the time taken - along with the number of users in your player data directory (I'll add the count in a future update to the plugin as I get this sorted).

I haven't yet solved the first cache lookup issue, but I have an idea - updates will follow.

dualspiral:
From previous discussion with blood, I've gone for using the `GameProfileCache` to provide the profiles while waiting for the USS to populate its cache. It's not idea, as the cache will include users who have no data stored, but it is the best I can come up with - they should really only be a corner case.

As the population of users only takes 7ms on my machine thanks to there being only one user, some data with the above (v2) plugin would be useful to get an idea as to how long the filesystem scans really are taking.

dualspiral:
Anyone using v2, there is a v3 that should actually work better...

ImMorpheus:
I've been playing with this.

Sadly, this is making worse (the issue was already there) a pretty bad issue with `SpongeUser` manipulators.
Manipulators are manually read from nbt, but that's not the case in the `UserDiscoverer`
https://github.com/SpongePowered/SpongeCommon/blob/a8e09ddb506632d688c532535550e37088a0ee85/src/main/java/org/spongepowered/common/service/user/UserDiscoverer.java#L128 

Empty `User`s are dangerous, they can lead to desync with Player object.
Gabi confirmed we can just lazy-load manipulators, I'm working on PR.

src/main/java/org/spongepowered/common/service/user/UserDiscoverer.java
```diff
@@ -164,51 +216,184 @@ static User findByUsername(final String username) {
...
 
-            // it exists, so we make sure to remove the uuid from the map (it may have been added manually in the meantime)
-            nonExistentUsers.remove(uuid);
+    // This method is potentially slow and should be run as few times as possible (thus, the filesystem scan)
+    private static Set<UUID> getAvailablePlayerUUIDs(Path playersDirectory) {
+        Set<UUID> ret = new HashSet<>();
+        try (DirectoryStream<Path> stream = Files.newDirectoryStream(playersDirectory, "*.dat")) {
+            for (Path entry : stream) {
+                try {
+                    String name = entry.getFileName().toString().replaceAll("\\.dat$", "");
```
ImMorpheus: `"\\.dat$"` is compiled each time. It's better to compile the regex once and use `Matcher#replaceAll`

src/main/java/org/spongepowered/common/service/user/UserDiscoverer.java
```diff
@@ -377,4 +561,57 @@ private static boolean deleteBanlistEntry(final UUID uniqueId) {
         return true;
     }
 
+    private static void createCacheEntry(User user) {
+        invalidateEntry(user.getProfile());
```
ImMorpheus, line 568: Is this needed given how the value are replaced ? (put v remove+put)

ImMorpheus (changes requested):

src/main/java/org/spongepowered/common/service/user/UserDiscoverer.java
dualspiral: Yes, given how some values may not be replaced (look at the ifPresent call in invalidateEntry and the if statement in this call). It also means that the cache is invalidated for everything before we add things - I'd rather make sure everything is removed before we start re-adding here.

I bet the cost is minimal and so I'd rather remove everything then add everything, then go piecemeal for the sake of a couple of cycles.

gabizou (approved):
LGTM

ImMorpheus (approved):

Issue #2013, opened 2018-07-22
InteractItemEvent and InteractBlockEvent behave incorrectly


Lignium:
**I am currently running**

- SpongeVanilla version: spongevanilla-1.12.2-7.1.0-BETA-93
- Java version: OpenJDK Version "1.8.0_171"
- Operating System: Linux Mint 18.3 Sylvia (Kernel: x86_64 Linux 4.13.0-45-generic)
- Plugins/Mods: Empty (Only one my experimental plugin)

**Issue Description**

~~At the last build for version 1.12.2, the following bugs are observed:~~
~~1. InteractBlockEvent is not called when the right mouse button is clicked (this is a SECONDARY event for both hands).~~
~~2. InteractBlockEvent for the left hand (PRIMARY event) does not contain data about the used item when the player clicked into the air. EventContext is empty.~~

My experiment conditions:

- Main hand: Stick
- Off hand: Empty

Listener code:

```
@Listener
public final void onInteractBlock(InteractBlockEvent event) {
	Objects.requireNonNull(event, "event");
	
	System.out.println(event.getClass().getName());
	System.out.println(event);
}
```

Console output:

**Left button to air**

```
org.spongepowered.api.event.InteractBlockEvent$Primary$MainHand$Impl
MainHand{cancelled=false, cause=Cause[Context=Context[], Stack={EntityPlayerMP['Lignium'/0, l='world', x=-0,53, y=64,00, z=1,22]}], context=Context[], handType=MAIN_HAND, interactionPoint=Optional.empty, source=EntityPlayerMP['Lignium'/0, l='world', x=-0,53, y=64,00, z=1,22], targetBlock=SpongeBlockSnapshot{worldUniqueId=00000000-0000-0000-0000-000000000000, position=(0, 0, 0), blockState=minecraft:air, extendedState=minecraft:air}, targetSide=NONE}
```

**Right button to air**

```
<empty>
```

**Left Button to block**

```
org.spongepowered.api.event.InteractBlockEvent$Primary$MainHand$Impl
[...]

Issue #1799, opened 2018-02-13
InteractBlockEvent.Primary.MainHand getTargetBlock not accurate (sometimes invalid)


jd1378:
**I am currently running**

- SpongeVanilla version: 1.12.2-7.1.0-BETA-18
- Java version: 1.8.0_152
- Operating System: Windows 10 Home 64-bit

**Issue Description**
`InteractBlockEvent.Primary.MainHand getTargetBlock` doesn't return interacted block and returns the block the player is currently looking at, which wont be accurate on the move (like in minecart , since its moving forward , if the player sends a hit , it will be received with 1-2 tick + X ms ping delay , so when you ray trace the player , its already past the block, especially on higher velocity movements) .
on the other hand , if player is using freecam hack , the block interacted wont be same as the block in the raytracing.
i [checked here](https://github.com/SpongePowered/SpongeCommon/blob/6496eb268e25a752d5dffdd77ad001e05c9112bf/src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java#L644) to see how its fired.

an idea to fix it is using [digging packet](http://wiki.vg/Protocol#Player_Digging)

PR #2125, opened 2018-10-20
Standardize interact events

ImMorpheus:
**SpongeCommon** | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/392) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/2483) 

Fix https://github.com/SpongePowered/SpongeCommon/issues/2013

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
@@ -602,10 +602,11 @@ public abstract class MixinNetHandlerPlayServer implements PlayerConnection, IMi
     @Redirect(method = "processTryUseItemOnBlock", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/management/PlayerInteractionManager;processRightClickBlock(Lnet/minecraft/entity/player/EntityPlayer;Lnet/minecraft/world/World;Lnet/minecraft/item/ItemStack;Lnet/minecraft/util/EnumHand;Lnet/minecraft/util/math/BlockPos;Lnet/minecraft/util/EnumFacing;FFF)Lnet/minecraft/util/EnumActionResult;"))
     public EnumActionResult onProcessRightClickBlock(PlayerInteractionManager interactionManager, EntityPlayer player, net.minecraft.world.World worldIn, @Nullable ItemStack stack, EnumHand hand, BlockPos pos, EnumFacing facing, float hitX, float hitY, float hitZ) {
         EnumActionResult actionResult = interactionManager.processRightClickBlock(this.player, worldIn, stack, hand, pos, facing, hitX, hitY, hitZ);
+
+        final PhaseData peek = PhaseTracker.getInstance().getCurrentPhaseData();
+
         // If a plugin or mod has changed the item, avoid restoring
-        if (!SpongeCommonEventFactory.playerInteractItemChanged) {
-            final PhaseTracker phaseTracker = PhaseTracker.getInstance();
-            final PhaseData peek = phaseTracker.getCurrentPhaseData();
+        if (!((PacketContext<?>) peek.context).getInteractItemChanged()) {
             final ItemStack itemStack = ItemStackUtil.toNative(((PacketContext<?>) peek.context).getItemUsed());
 
             // Only do a restore if something actually changed. The client does an identity check ('==')
@@ -615,7 +616,29 @@ public abstract class MixinNetHandlerPlayServer implements PlayerConnection, IMi
                 PacketPhaseUtil.handlePlayerSlotRestore((EntityPlayerMP) player, itemStack, hand);
             }
         }
-        SpongeCommonEventFactory.playerInteractItemChanged = false;
+        ((PacketContext<?>) peek.context).interactItemChanged(false);
+        SpongeCommonEventFactory.interactBlockRightClickEventCancelled = false;
+        return actionResult;
+    }
+
+    @Redirect(method = "processTryUseItem", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/management/PlayerInteractionManager;processRightClick(Lnet/minecraft/entity/player/EntityPlayer;Lnet/minecraft/world/World;Lnet/minecraft/item/ItemStack;Lnet/minecraft/util/EnumHand;)Lnet/minecraft/util/EnumActionResult;"))
+    public EnumActionResult onProcessRightClick(PlayerInteractionManager interactionManager, EntityPlayer player, net.minecraft.world.World worldIn, @Nullable ItemStack stack, EnumHand hand) {
+        EnumActionResult actionResult = interactionManager.processRightClick(this.player, worldIn, stack, hand);
+
+        final PhaseData peek = PhaseTracker.getInstance().getCurrentPhaseData();
+
+        // If a plugin or mod has changed the item, avoid restoring
+        if (!((PacketContext<?>) peek.context).getInteractItemChanged()) {
+            final ItemStack itemStack = ItemStackUtil.toNative(((PacketContext<?>) peek.context).getItemUsed());
+
+            // Only do a restore if something actually changed. The client does an identity check ('==')
+            // to determine if it should continue using an itemstack. If we always resend the itemstack, we end up
+            // cancelling item usage (e.g. eating food) that occurs while targeting a block
+            if (!ItemStack.areItemStacksEqual(itemStack, player.getHeldItem(hand)) && SpongeCommonEventFactory.interactBlockRightClickEventCancelled) {
+                PacketPhaseUtil.handlePlayerSlotRestore((EntityPlayerMP) player, itemStack, hand);
+            }
+        }
+        ((PacketContext<?>) peek.context).interactItemChanged(false);
         SpongeCommonEventFactory.interactBlockRightClickEventCancelled = false;
         return actionResult;
     }
@@ -790,7 +813,6 @@ public abstract class MixinNetHandlerPlayServer implements PlayerConnection, IMi
                     try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
                         EnumHand hand = packetIn.getHand();
                         ItemStack itemstack = hand != null ? this.player.getHeldItem(hand) : ItemStack.EMPTY;
-                        frame.addContext(EventContextKeys.USED_ITEM, ItemStackUtil.snapshotOf(itemstack));
 
                         SpongeCommonEventFactory.lastSecondaryPacketTick = this.server.getTickCounter();
 
--- a/src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java
@@ -578,29 +578,22 @@ public abstract class MixinMinecraftServer implements Server, ConsoleSource, IMi
             EntityPlayerMP player = SpongeCommonEventFactory.lastAnimationPlayer.get();
             if (player != null && lastAnimTick != lastPrimaryTick && lastAnimTick != lastSecondaryTick && lastAnimTick != 0 && lastAnimTick - lastPrimaryTick > 3 && lastAnimTick - lastSecondaryTick > 3) {
                 BlockSnapshot blockSnapshot = BlockSnapshot.NONE;
-                EnumFacing side = null;
+
                 final RayTraceResult result = SpongeImplHooks.rayTraceEyes(player, SpongeImplHooks.getBlockReachDistance(player) + 1);
                 // Hit non-air block
                 if (result != null && result.getBlockPos() != null) {
-                    blockSnapshot = new Location<>((World) player.world, VecHelper.toVector3d(result.getBlockPos())).createSnapshot();
-                    side = result.sideHit;
+                    return;
                 }
 
                 Sponge.getCauseStackManager().pushCause(player);
-                if (SpongeCommonEventFactory.callInteractItemEventPrimary(player, player.getHeldItemMainhand(), EnumHand.MAIN_HAND,
-                        result == null ? null : VecHelper.toVector3d(result.hitVec), blockSnapshot).isCancelled()) {
+                if (!player.getHeldItemMainhand().isEmpty() && SpongeCommonEventFactory.callInteractItemEventPrimary(player, player.getHeldItemMainhand(), EnumHand.MAIN_HAND, null, blockSnapshot).isCancelled()) {
                     SpongeCommonEventFactory.lastAnimationPacketTick = 0;
                     Sponge.getCauseStackManager().popCause();
                     return;
                 }
 
-                if (side != null) {
-                    SpongeCommonEventFactory.callInteractBlockEventPrimary(player, blockSnapshot, EnumHand.MAIN_HAND, side, VecHelper.toVector3d
-                            (result.hitVec));
-                } else {
-                    SpongeCommonEventFactory.callInteractBlockEventPrimary(player, EnumHand.MAIN_HAND, result == null ? null : VecHelper.toVector3d
-                            (result.hitVec));
-                }
+                SpongeCommonEventFactory.callInteractBlockEventPrimary(player, player.getHeldItemMainhand(), EnumHand.MAIN_HAND, null);
+
                 Sponge.getCauseStackManager().popCause();
             }
         }
--- a/src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
@@ -30,6 +30,7 @@ import net.minecraft.block.BlockChest;
 import net.minecraft.block.BlockCommandBlock;
 import net.minecraft.block.BlockDoor;
 import net.minecraft.block.BlockStructure;
+import net.minecraft.block.material.Material;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.player.EntityPlayerMP;
@@ -52,6 +53,7 @@ import net.minecraft.util.EnumActionResult;
 import net.minecraft.util.EnumFacing;
 import net.minecraft.util.EnumHand;
 import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.RayTraceResult;
 import net.minecraft.world.GameType;
 import net.minecraft.world.ILockableContainer;
 import org.spongepowered.api.Sponge;
@@ -59,8 +61,11 @@ import org.spongepowered.api.block.BlockSnapshot;
 import org.spongepowered.api.event.CauseStackManager;
 import org.spongepowered.api.event.block.InteractBlockEvent;
 import org.spongepowered.api.event.cause.EventContextKeys;
+import org.spongepowered.api.event.item.inventory.InteractItemEvent;
 import org.spongepowered.api.item.inventory.slot.EquipmentSlot;
+import org.spongepowered.api.util.Direction;
 import org.spongepowered.api.util.Tristate;
+import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Overwrite;
@@ -71,12 +76,18 @@ import org.spongepowered.asm.mixin.injection.Redirect;
 import org.spongepowered.asm.mixin.injection.Slice;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 import org.spongepowered.common.SpongeImpl;
+import org.spongepowered.common.SpongeImplHooks;
 import org.spongepowered.common.event.SpongeCommonEventFactory;
+import org.spongepowered.common.event.tracking.PhaseData;
+import org.spongepowered.common.event.tracking.PhaseTracker;
+import org.spongepowered.common.event.tracking.phase.packet.PacketContext;
 import org.spongepowered.common.interfaces.IMixinContainer;
+import org.spongepowered.common.interfaces.entity.player.IMixinEntityPlayerMP;
 import org.spongepowered.common.interfaces.server.management.IMixinPlayerInteractionManager;
 import org.spongepowered.common.interfaces.world.IMixinWorld;
 import org.spongepowered.common.item.inventory.util.ItemStackUtil;
 import org.spongepowered.common.registry.provider.DirectionFacingProvider;
+import org.spongepowered.common.util.VecHelper;
 
 @Mixin(value = PlayerInteractionManager.class)
 public abstract class MixinPlayerInteractionManager implements IMixinPlayerInteractionManager {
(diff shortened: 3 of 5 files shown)


Lignium:
Please don't forget my full analysis of these events. https://github.com/SpongePowered/SpongeCommon/issues/2013#issuecomment-416403984 Thank!

ImMorpheus:
I'm not. As a matter of fact, I'm using your table to double-check everything is ok.

`EventContextKeys.USED_ITEM`and `EventContextKeys.BLOCK_HIT` have been fixed.

I'm trying to see what's going on with owner/notifier and interaction points.

ImMorpheus:
Sponge is breaking forge behaviour. Found a lot of inconsistencies with Interact events.

_See edit_

ImMorpheus:
Given the following piece of code:

```
@SubscribeEvent
public void onTest(PlayerInteractEvent event) {
    event.setCanceled(true);
    event.setCancellationResult(EnumActionResult.PASS);
}
```

the return value of `PlayerInteractionManager#processRightClickBlock` is not consistent

forge: EnumActionResult.PASS
sponge: EnumActionResult.FAIL

Is there a reason why we're doing this ? (always returning EnumActionResult.FAIL when the event is cancelled)
Input wanted.

Relevant javadocs: https://github.com/MinecraftForge/MinecraftForge/blob/a5f29570a39508afe96870e7e665ff4bbf964aa8/src/main/java/net/minecraftforge/event/entity/player/PlayerInteractEvent.java#L390

src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
```diff
@@ -95,80 +98,6 @@ public void onBlockRemoving(final BlockPos pos, final CallbackInfo ci) {
         }
     }
 
-    /**
```
gabizou: And this has been tested for cancellations? This was added for bucket cancelations causing clients to think they had changed itemstacks.

src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
```diff
@@ -326,6 +258,117 @@ public EnumActionResult processRightClickBlock(EntityPlayer player, net.minecraf
         // } // Sponge - Remove unecessary else bracket
     }
 
+    /**
+     * @reason Due to the way that buckets and the like can be handled
```
gabizou, line 354: If there’s an overwrite, author who wrote it, with updates dated etc.

src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
```diff
@@ -326,6 +258,117 @@ public EnumActionResult processRightClickBlock(EntityPlayer player, net.minecraf
...
+     * effect, and therefor requires telling the client to set back the item
+     * in hand.
+     * @reason Fire interact item event.
+     */
+    @Overwrite
+    public EnumActionResult processRightClick(EntityPlayer player, net.minecraft.world.World worldIn, ItemStack stack, EnumHand hand) {
+        if (this.gameType == GameType.SPECTATOR) {
+            return EnumActionResult.PASS;
+        }
+
```
gabizou, line 370: Can you comment where forge patches to this method occur, so we don’t have to look out for that on the side, but in the overwrite itself?

src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
```diff
@@ -253,27 +179,33 @@ public EnumActionResult processRightClickBlock(EntityPlayer player, net.minecraf
         }
         // Sponge End
 
-        if (!player.isSneaking() || player.getHeldItemMainhand().isEmpty() && player.getHeldItemOffhand().isEmpty()) {
+        EnumActionResult result = EnumActionResult.PASS;
+
+        boolean bypass = true;
+        final ItemStack[] itemStacks = {player.getHeldItemMainhand(), player.getHeldItemOffhand()};
```
gabizou: For a “hot” method, this array creation and for loop are unnecessary. The advantage of the old if is that it’s fail fast, this has to actually check for both and store, which is more expensive.

gabizou (changes requested):

gabizou:
> Given the following piece of code:

>@SubscribeEvent
public void onTest(PlayerInteractEvent event) {
    event.setCanceled(true);
    event.setCancellationResult(EnumActionResult.PASS);
}
the return value of PlayerInteractionManager#processRightClickBlock is not consistent

>forge: EnumActionResult.PASS
sponge: EnumActionResult.FAIL

>Is there a reason why we're doing this ? (always returning EnumActionResult.FAIL when the event is cancelled)
Input wanted.

You were correct in saying sponge is breaking contract here. By design, we’ve made cancel as a stop all process, but for interactions, it should be considered a pass result, since this is server side, fail results will not have the same effect compared to running on the client.

src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
ImMorpheus: They don't. 
Forge only adds its event and check for cancellation 
https://github.com/MinecraftForge/MinecraftForge/blob/a5f29570a39508afe96870e7e665ff4bbf964aa8/patches/minecraft/net/minecraft/server/management/PlayerInteractionManager.java.patch#L199

ImMorpheus:
This is ready for review.

I will continue to test event cancellation but interact events are now consistent with Forge events

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -620,6 +620,27 @@ public EnumActionResult onProcessRightClickBlock(PlayerInteractionManager intera
         return actionResult;
     }
 
+    @Redirect(method = "processTryUseItem", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/management/PlayerInteractionManager;processRightClick(Lnet/minecraft/entity/player/EntityPlayer;Lnet/minecraft/world/World;Lnet/minecraft/item/ItemStack;Lnet/minecraft/util/EnumHand;)Lnet/minecraft/util/EnumActionResult;"))
+    public EnumActionResult onProcessRightClick(PlayerInteractionManager interactionManager, EntityPlayer player, net.minecraft.world.World worldIn, @Nullable ItemStack stack, EnumHand hand) {
+        EnumActionResult actionResult = interactionManager.processRightClick(this.player, worldIn, stack, hand);
+        // If a plugin or mod has changed the item, avoid restoring
+        if (!SpongeCommonEventFactory.playerInteractItemChanged) {
```
gabizou: Is it possible to use some phase context flag that gets flipped here instead of checking a static variable? While I do want to standardize the interact events, I do want to see some of the legacy style static variable nonsense going the way of the dodo.

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -620,6 +620,27 @@ public EnumActionResult onProcessRightClickBlock(PlayerInteractionManager intera
...
+            final PhaseTracker phaseTracker = PhaseTracker.getInstance();
```
gabizou: Redundant variable?

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -620,6 +620,27 @@ public EnumActionResult onProcessRightClickBlock(PlayerInteractionManager intera
...
+            final PhaseData peek = phaseTracker.getCurrentPhaseData();
+            final ItemStack itemStack = ItemStackUtil.toNative(((PacketContext<?>) peek.context).getItemUsed());
+
+            // Only do a restore if something actually changed. The client does an identity check ('==')
+            // to determine if it should continue using an itemstack. If we always resend the itemstack, we end up
+            // cancelling item usage (e.g. eating food) that occurs while targeting a block
+            if (!ItemStack.areItemStacksEqual(itemStack, player.getHeldItem(hand)) && SpongeCommonEventFactory.interactBlockRightClickEventCancelled) {
+                PacketPhaseUtil.handlePlayerSlotRestore((EntityPlayerMP) player, itemStack, hand);
```
gabizou, line 638: We're always sure that the player is a player mp? (I know... it should never, but what are the cases of single player this code getting called?

src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
```diff
@@ -96,77 +110,78 @@ public void onBlockRemoving(final BlockPos pos, final CallbackInfo ci) {
...
-        )
-    )
-    private EnumActionResult spongeGetResultCheckForFailure(ActionResult<ItemStack> actionResult, EntityPlayer player, net.minecraft.world.World worldIn, ItemStack stack, EnumHand hand) {
-        // Sanity checks on the world being used (hey, i don't know the rules about clients...
-        // and if the world is in fact a responsible server world.
-        final EnumActionResult result = actionResult.getType();
-        if (!(worldIn instanceof IMixinWorld) || ((IMixinWorld) worldIn).isFake()) {
-            return result;
+    @Overwrite
+    public void onBlockClicked(BlockPos pos, EnumFacing side) {
```
gabizou, line 121: If this is an overwrite, please comment existing method and why things are being done like so:
```java
@Overwrite
public void onBlockClicked(BlockPos pos, EnumFacing side) {
  // Sponge Start - throw a pre-event for block clicks to verify if they need to be cancelled
  final Blocksnapshot ........
  // Sponge End
  if (this.isCreative()) {
    vanilla code;
  }
}
```

gabizou (changes requested):
Overall fine, just need documentation on overwrites for future maintenance.

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
ImMorpheus: None.
The `@redirect` takes place before (`BeforeInvoke`) `this.player.interactionManager.processRightClickBlock(this.player, worldserver, itemstack, enumhand, blockpos, enumfacing, packetIn.getFacingX(), packetIn.getFacingY(), packetIn.getFacingZ());` 
`this.player` (inside `NetHandlerPlayServer`) is of type `EntityPlayerMP`, thus this is a safe cast.
Mixin requires to match the signature of the method (`PlayerInteractionManager#processRightClickBlock`) which is why the `player` parameter is of type `Player`

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
ImMorpheus: Any chance it was done for performance reasons since this is a "hot" method ?

gabizou: It's blindingly fast to still have the check through the phase tracker, especially since it's still another boolean field, just a few field accesses. If it were say adding extra processing costs to mass convert and cast, then sure. If you want to talk about hot methods, phase tracker is 100% integrated into setting blocks, and performing thousands of notifications, far less than handling processing right clicks.

src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
```diff
@@ -96,77 +110,80 @@ public void onBlockRemoving(final BlockPos pos, final CallbackInfo ci) {
...
+
+        // Sponge start - Fire interact block event
```
gabizou, line 122: If this is in fact just a pre-check, why don't we make this a cancellable injection?

ImMorpheus: That's not possible. 
This method is also modified by Forge to add its event: https://github.com/MinecraftForge/MinecraftForge/blob/a5f29570a39508afe96870e7e665ff4bbf964aa8/patches/minecraft/net/minecraft/server/management/PlayerInteractionManager.java.patch#L34

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
ImMorpheus: Fair enough, I was playing devil's advocate

src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
gabizou: Then you need to communicate that in the javadoc or comments, if I did a blind sweep of overwrites, this would’ve broken because of that important fact not being documented ;)

src/main/java/org/spongepowered/common/mixin/core/server/MixinMinecraftServer.java
```diff
@@ -571,40 +571,6 @@ public void onServerTickStart(CallbackInfo ci) {
 
     @Inject(method = "tick", at = @At(value = "RETURN"))
     public void onServerTickEnd(CallbackInfo ci) {
-        int lastAnimTick = SpongeCommonEventFactory.lastAnimationPacketTick;
-        int lastPrimaryTick = SpongeCommonEventFactory.lastPrimaryPacketTick;
```
bloodmc: The whole purpose of this injection is to fire interact events while left-clicking AIR. If you remove this, there needs to be an alternative to handle these events or many plugins will break.

ImMorpheus:
@bloodmc I will address your comment(s) here and request an input from other developers since my message on the sponge-dev channel has been disregarded. 

> The whole purpose of this injection is to fire interact events while left-clicking AIR.

Correct.

> If you remove this, there needs to be an alternative to handle these events 

I disagree, but let's put that aside for later since this is not the point.

---

Forge design on Interact events is clear and a lot of dicussion was going on back in the days.
https://github.com/MinecraftForge/MinecraftForge/pull/2604#issuecomment-199319370

Thus the reason they decided to fire PlayerInteractEvent.LeftClickEmpty on the client

_The server is not aware of when the client left clicks empty space_, however 
we decided it was a good idea to fire Interact events based on the ticks between animation packets, making the server aware of it (kind of).
IMO this is wrong. We already have https://github.com/SpongePowered/SpongeAPI/blob/c4c8ce6cc1d902e1f9514ba23686683bf5046891/src/main/java/org/spongepowered/api/event/entity/living/humanoid/AnimateHandEvent.java#L38, which should be enough for simple cases. Do you want to take a guess and raytrace the block the player is looking at ? Good, that's on you.

> many plugins will break

I can see that. However, everyone seems to be missing that Interact events are converted to their Forge counterpart, which means the current implementation is breaking _a shit ton of things cit._ 

I'm against it and I've asked for input before doing the change. However (again) I'm open to discussion whether we should keep this hack to support left-clicking AIR, but there are a few things I'd like to point out:

- It's by no means accurate since this is not reliable by design. Issues like https://github.com/SpongePowered/SpongeCommon/issues/1799 should be considered as `won't fix`
- If we want to support this, that's on us. We should not enforce this hacky behaviour on Forge. (Only the sponge counterpart of the event should be fired).

bloodmc:
> Forge design on Interact events is clear and a lot of dicussion was going on back in the days.
> [MinecraftForge/MinecraftForge#2604 (comment)](https://github.com/MinecraftForge/MinecraftForge/pull/2604#issuecomment-199319370)
> 
> Thus the reason they decided to fire PlayerInteractEvent.LeftClickEmpty on the client

Client events are only a solution for SpongeForge. This would not be accessible by SpongeVanilla. Plus, even if we did introduce a client event, that would require every player to run SpongeForge on client-side.

> 
> _The server is not aware of when the client left clicks empty space_, however
> we decided it was a good idea to fire Interact events based on the ticks between animation packets, making the server aware of it (kind of).
> IMO this is wrong. We already have https://github.com/SpongePowered/SpongeAPI/blob/c4c8ce6cc1d902e1f9514ba23686683bf5046891/src/main/java/org/spongepowered/api/event/entity/living/humanoid/AnimateHandEvent.java#L38, which should be enough for simple cases. Do you want to take a guess and raytrace the block the player is looking at ? Good, that's on you.
> 

https://github.com/SpongePowered/SpongeAPI/blob/stable-7/src/main/java/org/spongepowered/api/event/block/InteractBlockEvent.java#L56-L62

Our API states otherwise. The javadocs would have to be updated stating to use the AnimationEvent instead but even that would be troublesome as a plugin would have no idea if the animation occurred during an actual dig/place event. We would have to provide some type of context to make it easier for them.

Also, `InteractBlockEvent.Secondary` is fired for AIR (because Mojang actually sends this packet) so it would be really awkward to have Secondary to fire for AIR but not Primary.

> > many plugins will break
> 
> I can see that. However, everyone seems to be missing that Interact events are converted to their Forge counterpart, which means the current implementation is breaking _a shit ton of things cit._
> 

> I'm against it and I've asked for input before doing the change. However (again) I'm open to discussion whether we should keep this hack to support left-clicking AIR, but there are a few things I'd like to point out:
> 
> * It's by no means accurate since this is not reliable by design. Issues like #1799 should be considered as `won't fix`
> * If we want to support this, that's on us. We should not enforce this hacky behaviour on Forge. (Only the sponge counterpart of the event should be fired).

You can simply modify the factory to avoid firing to forge when required.
https://github.com/SpongePowered/SpongeForge/blob/stable-7/src/main/java/org/spongepowered/mod/event/SpongeToForgeEventFactory.java#L411

We can look at trying to come up with a better solution to address issue 1799.

bloodmc:
> Forge design on Interact events is clear and a lot of dicussion was going on back in the days.
> [MinecraftForge/MinecraftForge#2604 (comment)](https://github.com/MinecraftForge/MinecraftForge/pull/2604#issuecomment-199319370)
> 
> Thus the reason they decided to fire PlayerInteractEvent.LeftClickEmpty on the client

Forge design has the advantage of client-side events. For us, client events would only work with SpongeForge and would require SpongeForge be installed on all clients which is not always wanted. It also doesn't address SpongeVanilla.

> 
> _The server is not aware of when the client left clicks empty space_, however
> we decided it was a good idea to fire Interact events based on the ticks between animation packets, making the server aware of it (kind of).
> IMO this is wrong. We already have https://github.com/SpongePowered/SpongeAPI/blob/c4c8ce6cc1d902e1f9514ba23686683bf5046891/src/main/java/org/spongepowered/api/event/entity/living/humanoid/AnimateHandEvent.java#L38, which should be enough for simple cases. Do you want to take a guess and raytrace the block the player is looking at ? Good, that's on you.
> 

Animation events are fired in all cases where the hand moves. We would have to provide context to let plugins know when the click was indeed in AIR. Otherwise a plugin would have to raytrace always and could see mixed results.

As per our API

https://github.com/SpongePowered/SpongeAPI/blob/stable-7/src/main/java/org/spongepowered/api/event/block/InteractBlockEvent.java#L56-L62

It states the event will be fired whenever the client clicks the primary button. If we opt-out and decide to never fire it then the javadocs would have to be updated. However, such a change could never go into 1.12 as it is a breaking change and would have to be pushed to 1.13. 

There is also the issue with `InteractBlockEvent.Secondary` which would always be fired when right-clicking AIR as mojang sends a packet for this. If we only fire the secondary but not the primary, this would be awkward.


> > many plugins will break
> 
> I can see that. However, everyone seems to be missing that Interact events are converted to their Forge counterpart, which means the current implementation is breaking _a shit ton of things cit._
> 

This can be addressed in the SpongeToForgeEventFactory 

https://github.com/SpongePowered/SpongeForge/blob/stable-7/src/main/java/org/spongepowered/mod/event/SpongeToForgeEventFactory.java#L411

> I'm against it and I've asked for input before doing the change. However (again) I'm open to discussion whether we should keep this hack to support left-clicking AIR, but there are a few things I'd like to point out:
> 
> * It's by no means accurate since this is not reliable by design. Issues like #1799 should be considered as `won't fix`

We can discuss an alternative to try to solve this issue.

> * If we want to support this, that's on us. We should not enforce this hacky behaviour on Forge. (Only the sponge counterpart of the event should be fired).

See above.

ImMorpheus:
> Forge design has the advantage of client-side events. For us, client events would only work with SpongeForge and would require SpongeForge be installed on all clients which is not always wanted. It also doesn't address SpongeVanilla.

PlayerInteractEvent.LeftClickEmpty is already converted to its sponge counterpart, so there'd be no need to change things on our side.
My point was that (as the old issue I linked states): _the server does not know when a player left click empty space because it just calls swingArm which sends an animation packet to the server to swing the arm_
Forge decided not to fire an event on the server since there was not enough information to do so.

> Animation events are fired in all cases where the hand moves. We would have to provide context to let plugins know when the click was indeed in AIR. Otherwise a plugin would have to raytrace always and could see mixed results.

Point taken. I missed that.

> There is also the issue with InteractBlockEvent.Secondary which would always be fired when right-clicking AIR as mojang sends a packet for this. If we only fire the secondary but not the primary, this would be awkward.

Minor inconsistency, but point taken. 

I think we're leaning toward not firing the event to Forge.

bloodmc:
> I think we're leaning toward not firing the event to Forge.

Yes, I think this is the best solution for now.

I also came across another bug while testing. It seems there are no secondary interact events fired to plugins when right-clicking on AIR with an item in hand.

Thanks!

ImMorpheus:
I'd like to explain the new event flow.
Note: This is only about right-click (RC)

Sponge design right now:

> basically our interact flow is as follows : Item -> Block -> Break/Place/None (something like this)
> interact always begins with item event

Here's a overview of the current situation:

- RC on air with an empty hand = no event at all 
- RC on air with an itemstack = 1 InteractItemEvent + 1 InteractBlockEvent
- RC on a block with an itemstack = 2 InteractItemEvent + 2 InteractBlockEvent*
- RC on a block with an empty hand = 2 InteractItemEvent + 2 InteractBlockEvent*

\* One for each hand

This was **arguably** confusing and it was breaking a whole lot of things (at least for modders/Forge).

The current design (same of Forge) of this PR aims to fire events based on the packets received by the server. 
CPacketPlayerTryUseItem -> InteractItemEvent 
CPacketPlayerTryUseItemOnBlock -> InteractBlockEvent 

After this PR:

- RC on air with an empty hand = no event at all (unchanged) 
- RC on air with an itemstack = 1 InteractItemEvent 
- RC on a block with an itemstack = 1 InteractItemEvent + 2 InteractBlockEvent*
- RC on a block with an empty hand = 2 InteractBlockEvent*

\* One for each hand

effects on (plugins/mods):

### mods 

everything would work as intended since our flow would be consistent with Forge. A lot less mod issues, happier modders and the world would probably be a better place.

### plugins

well, there are two things to keep in mind: 

InteractBlockEvent would be fired **only** if you're right-clicking on a block. 
InteractItemEvent would be fired **only** if you have an item in your hand. 

Every plugin exploiting the following ""features"" would then be broken: 

1. listening for InteractBlockEvent to catch a right-click on air with an itemstack. 
2. listening for InteractItemEvent to catch a right-click on a block with an empty hand. 
 
If you're a plugin developer a feedback would be appreciated.

bloodmc:
@ImMorpheus This looks good, thanks for the changes. Would you mind updating the API javadocs to reflect your changes?

ImMorpheus:
@bloodmc 
API looks good to me.

InteractItemEvent: `Base event for all interactions with an {@link ItemStack} in hand.`
InteractBlockEvent: `Base event for all interactions involving a {@link BlockSnapshot} at a {@link Location}.`

If anything, the implementation was breaking the contract by firing an InteractBlockEvent when right-clicking on a block with an empty hand (since there's no ItemStack in the hand).

https://github.com/SpongePowered/SpongeAPI/blob/859d917605f71aeca934c8a6095c533e1a104d2a/src/main/java/org/spongepowered/api/event/item/inventory/InteractItemEvent.java#L33

https://github.com/SpongePowered/SpongeAPI/blob/859d917605f71aeca934c8a6095c533e1a104d2a/src/main/java/org/spongepowered/api/event/block/InteractBlockEvent.java#L39

ryantheleach:
> catch a right-click on air with an itemstack.
> catch a right-click on a block with an empty hand.

Are these still possible regardless of method used?

ImMorpheus:
> Are these still possible regardless of method used?
---

> After this PR:
> 
> RC on air with an itemstack = 1 InteractItemEvent
> RC on a block with an empty hand = 2 InteractBlockEvent*

PR #2136, opened 2018-11-15
Fix SlotAdapter iterator

### LINKED ISSUE #2135: Inventory query for ItemType does not work if inventory contains only 1 stack
**I am currently running**

- SpongeVanilla version: spongevanilla-1.12.2-7.1.4.jar

- Plugins/Mods: test plugin

**Issue Description**
Inventory query for item type doesn't work if inventory contains only 1 stack.

Test code:
```java
@Listener
public void onPistonMove(MessageChannelEvent.Chat event, @Root Player player) {
    int quantity = 0;
    Inventory query = player.getInventory().query(QueryOperationTypes.ITEM_TYPE.of(ItemTypes.GOLD_BLOCK));
    for (Inventory inventory : query) {
        quantity += inventory.peek().get().getQuantity();
    }
    player.sendMessage(Text.of(quantity));
}
```

1. Take stack of gold blocks
2. Inventory must contain only one stack of ItemType
2. Write something in chat
3. You got 0 output
4. Divide stack, run again - works fine.

XakepSDK:
Fix behavior, fixes https://github.com/SpongePowered/SpongeCommon/issues/2135
@Faithcaio is fix correct?
Test code from https://github.com/SpongePowered/SpongeCommon/issues/2135 works fine with this fix

### DIFF
--- a/src/main/java/org/spongepowered/common/item/inventory/adapter/impl/slots/SlotAdapter.java
+++ b/src/main/java/org/spongepowered/common/item/inventory/adapter/impl/slots/SlotAdapter.java
@@ -42,6 +42,8 @@ import org.spongepowered.common.item.inventory.lens.impl.slots.SlotLensImpl;
 import org.spongepowered.common.item.inventory.lens.slots.SlotLens;
 import org.spongepowered.common.item.inventory.util.ItemStackUtil;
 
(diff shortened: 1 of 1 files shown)


Faithcaio (approved):
Looks correct to me.

PR #2139, opened 2018-11-18
Allow modify item MaxStackSize more then 64

PorPit:
I created a CoreMod to modify ItemStack MaxStackSize (more then 64)
it success running in Forge Server
but running in SpongeForge Server has some problem

### DIFF
--- a/src/main/java/org/spongepowered/common/item/inventory/util/ContainerUtil.java
+++ b/src/main/java/org/spongepowered/common/item/inventory/util/ContainerUtil.java
@@ -169,7 +169,7 @@ public final class ContainerUtil {
 
                     while (!itemStack.isEmpty())
                     {
-                        int i = RANDOM.nextInt(21) + 10;
+                        int i = RANDOM.nextInt(itemStack.getMaxStackSize()/3) + 10;
 
                         EntityItem entityitem = new EntityItem(worldServer, x + f, y + f1, z + f2, itemStack.splitStack(i));
 
--- a/src/main/java/org/spongepowered/common/item/inventory/util/ContainerUtil.java
+++ b/src/main/java/org/spongepowered/common/item/inventory/util/ContainerUtil.java
@@ -169,7 +169,7 @@ public final class ContainerUtil {
 
                     while (!itemStack.isEmpty())
                     {
-                        int i = RANDOM.nextInt(itemStack.getMaxStackSize()/3) + 10;
+                        int i = RANDOM.nextInt(itemStack.getMaxStackSize() / 3) + 10;
 
                         EntityItem entityitem = new EntityItem(worldServer, x + f, y + f1, z + f2, itemStack.splitStack(i));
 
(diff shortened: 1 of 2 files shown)


XakepSDK:
Here too, i think
https://github.com/SpongePowered/SpongeCommon/blob/1bbcb8297c7fdf5f2c949bc4326959a5b08ac7c8/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java#L383

PorPit:
@XakepSDK  yes

src/main/java/org/spongepowered/common/item/inventory/util/ContainerUtil.java
```diff
@@ -169,7 +169,7 @@ public static void performBlockInventoryDrops(WorldServer worldServer, double x,
 
                     while (!itemStack.isEmpty())
                     {
-                        int i = RANDOM.nextInt(21) + 10;
+                        int i = RANDOM.nextInt(itemStack.getMaxStackSize()/3) + 10;
```
kashike: `itemStack.getMaxStackSize() / 3` (spaces)

kashike (changes requested):

src/main/java/org/spongepowered/common/item/inventory/util/ContainerUtil.java
JBYoshi:
```suggestion
                        int i = RANDOM.nextInt(itemStack.getMaxStackSize() / 3) + 10;
```

JBYoshi:
```suggestion
                        int i = RANDOM.nextInt(itemStack.getMaxStackSize() / 3) + 10;
```

PorPit: https://github.com/SpongePowered/SpongeCommon/pull/2139/commits/a96ecc7a1257121d3cd877e56d2c29369ef1edfe
here also need

JBYoshi: Your change on that section looks fine. It's only this one that needs to be fixed before the pull request can be merged.

src/main/java/org/spongepowered/common/item/inventory/util/ContainerUtil.java
```diff
@@ -169,7 +169,7 @@ public static void performBlockInventoryDrops(WorldServer worldServer, double x,
 
                     while (!itemStack.isEmpty())
                     {
-                        int i = RANDOM.nextInt(21) + 10;
```
JBYoshi, line 172: I think this is supposed to stay as-is. It was copied from `InventoryHelper.spawnItemStack()`.

JBYoshi: This shouldn't be breaking your code because it loops until the original item stack is empty.

PR #2149, opened 2018-12-01
Fixes MySQL 8.0 compatibility with current SqlService, updates Hikari…

ghost:
This is not intended for merging as-is if MariaDB compatibility is to remain a feature of SpongeCommon. The MariaDB connector can co-exist with the MySQL 8.0.x connector if the SqlService implementation is adjusted to allow for options in the JDBC URL. 
 
Info on this: https://mariadb.com/kb/en/library/about-mariadb-connector-j/#having-mariadb-and-mysql-drivers-in-the-same-classpath

The regex would need to be updated for options (e.g. useSSL, disableMariaDbDriver, etc.). The changes in this PR simply allow the current SqlService to connect to a MySQL 8.0.x database without any invasive changes. The driver name has been updated to the proper 8.0 name.

Tested and working properly **for MySQL 8.0.13**. No other tests have been performed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,8 +67,9 @@ dependencies {
     runtime 'org.apache.logging.log4j:log4j-slf4j-impl:2.8.1'
 
     // Database connectors
-    compile 'com.zaxxer:HikariCP:2.6.3'
-    runtime 'org.mariadb.jdbc:mariadb-java-client:2.0.3'
+    compile 'com.zaxxer:HikariCP:3.2.0'
+    //runtime 'org.mariadb.jdbc:mariadb-java-client:2.0.3'
+    runtime 'mysql:mysql-connector-java:8.0.13'
     compile 'com.h2database:h2:1.4.196'
     runtime 'org.xerial:sqlite-jdbc:3.20.0'
 
--- a/gradle/implementation.gradle
+++ b/gradle/implementation.gradle
@@ -146,7 +146,8 @@ shadowJar ext.shadowConfiguration = {
 
         include dependency('com.zaxxer:HikariCP')
 
-        include dependency('org.mariadb.jdbc:mariadb-java-client')
+        //include dependency('org.mariadb.jdbc:mariadb-java-client')
+        include dependency('mysql:mysql-connector-java')
         include dependency('com.h2database:h2')
         include dependency('org.xerial:sqlite-jdbc')
 
--- a/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
+++ b/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
@@ -85,8 +85,8 @@ public class SqlServiceImpl implements SqlService, Closeable {
         mySqlProps.setProperty("useConfigs",
                 "maxPerformance"); // Config options based on http://assets.en.oreilly
                 // .com/1/event/21/Connector_J%20Performance%20Gems%20Presentation.pdf
-        build.put("com.mysql.jdbc.Driver", mySqlProps);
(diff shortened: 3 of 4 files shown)


build.gradle
```diff
@@ -67,8 +67,9 @@ dependencies {
     runtime 'org.apache.logging.log4j:log4j-slf4j-impl:2.8.1'
 
     // Database connectors
-    compile 'com.zaxxer:HikariCP:2.6.3'
-    runtime 'org.mariadb.jdbc:mariadb-java-client:2.0.3'
+    compile 'com.zaxxer:HikariCP:3.2.0'
+    //runtime 'org.mariadb.jdbc:mariadb-java-client:2.0.3'
```
ryantheleach, line 71: 1. This removes mariadb support, which would be a breaking change. additionally it's commented.

gradle/implementation.gradle
```diff
@@ -146,7 +146,8 @@ shadowJar ext.shadowConfiguration = {
 
         include dependency('com.zaxxer:HikariCP')
 
-        include dependency('org.mariadb.jdbc:mariadb-java-client')
+        //include dependency('org.mariadb.jdbc:mariadb-java-client')
```
ryantheleach, line 149: See 1.

src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
```diff
@@ -85,8 +85,8 @@
         mySqlProps.setProperty("useConfigs",
                 "maxPerformance"); // Config options based on http://assets.en.oreilly
                 // .com/1/event/21/Connector_J%20Performance%20Gems%20Presentation.pdf
-        build.put("com.mysql.jdbc.Driver", mySqlProps);
-        build.put("org.mariadb.jdbc.Driver", mySqlProps);
+        build.put("com.mysql.cj.jdbc.Driver", mySqlProps);
+        //build.put("org.mariadb.jdbc.Driver", mySqlProps);
```
ryantheleach, line 89: see 1.

src/test/java/org/spongepowered/common/service/sql/SqlServiceImplTest.java
```diff
@@ -40,7 +40,8 @@ public void testMysqlConnectionInfo() throws SQLException {
         assertEquals("zml", subject.getUser());
         assertEquals("totallymypassword", subject.getPassword());
         assertEquals("jdbc:mysql://localhost/sponge", subject.getAuthlessUrl());
-        assertEquals("org.mariadb.jdbc.Driver", subject.getDriverClassName());
+        //assertEquals("org.mariadb.jdbc.Driver", subject.getDriverClassName());
```
ryantheleach, line 43: see 1.

ryantheleach (changes requested):
Thanks for the Pull Request!

It's appreciated when community members like yourself have put effort by creating a pull request. 

However **If** we need to add mysql explicitly, we need to do so without dropping mariadb support.

The mariadb connector has explicit support for this, and can let mysql connector handle mysql connectors and let mariadb handle mariadb connections.

Whilst researching your issue, yesterday, I came across https://mariadb.com/kb/en/library/about-mariadb-connector-j/#having-mariadb-and-mysql-drivers-in-the-same-classpath

Which has explicit notes on how to have both loaded.

I'd prefer someone more familiar with databases in Java to confirm if this is actually the route we want to take, As I believe with additional testing it may show that all we need to do is update the mariadb connector which we already use, which would cut down on the amount of dependencies needed.

ghost:
New PR #2152 with completed feature impl.

Issue #2148, opened 2018-12-01
SqlService incompatible with MySQL 8.0.x (failing to read password)


ghost:
**I am currently running**

- SpongeForge version: 
- SpongeVanilla version: 1.12.2-7.1.5-RC139
- Java version: 1.8.0_192
- Operating System: Windows 10
- SQL Server: mysql 8.0.13

- Plugins/Mods: Irrelevant.

**Issue Description**
Despite proper URL syntax (checked against SqlService regex), a password is not being read when it is provided. 
`Access denied for user 'user'@'localhost' (using password: NO)`

`aliases {
    myalias="jdbc:mysql://user:password@localhost/test"
}`

Notably, HikariCP is outdated (latest is `com.zaxxer:HikariCP:3.2.0`) and MariaDB Java Client is outdated (latest is `org.mariadb.jdbc:mariadb-java-client:2.3.0`).

Updating the dependencies to their latest respective versions does **not** fix the issue. 

**Associated PR:** https://github.com/SpongePowered/SpongeCommon/pull/2152

phit:
Duplicate of #2008

phit:
I know its technically not the same symptoms, but we really don't need two issues to track that the libs are outdated.

ryantheleach:
I'm not sure that's the root cause, after investigation it was found that updating it didn't help.

phit:
@brjcarbone server version?

ryantheleach:
Edited in the mysql server version from information in discord.

ryantheleach:
Testing with the following plugin code:

https://gist.github.com/ryantheleach/ed688fb6510c32d93785a72a0752da13

With the following Sponge config

```
# Configuration options related to the Sql service, including connection aliases etc
    sql {
        # Aliases for SQL connections, in the format jdbc:protocol://[username[:password]@]host/database
        aliases {
            frank="jdbc:mysql://ryan:SomethingSecret@localhost/sponge"
        }
    }
```

Resulted in the following exception:

```
02:10:29] [main/INFO] [minecraft/GuiNewChat]: [CHAT] Leaking mysql connection string: jdbc:mysql://ryan:SomethingSecret@localhost/sponge
[02:10:29] [Server thread/INFO] [com.zaxxer.hikari.HikariDataSource]: HikariPool-1 - Starting...
[02:10:30] [Server thread/ERROR] [com.zaxxer.hikari.pool.HikariPool]: HikariPool-1 - Exception during pool initialization.
[...]

Issue #2008, opened 2018-07-19
Add support for mysql 8.0.3+


onyn:
**I am currently running**
- SpongeForge version: 1.12.2-2705-7.1.0-BETA-3206
- Java version: HotSpot 1.8.0_131-b11 64-bit
- Operating System: Ubuntu Xenial
- Plugins/Mods: none

**Issue Description**

When using sponge database service with mysql 8.0.11, db initialization in my plugin crashed with exception:

```
com.zaxxer.hikari.pool.HikariPool$PoolInitializationException: Failed to initialize pool: (conn:10) Unknown system variable 'tx_isolation'
```

Code example:

```java
SqlService service = Sponge.getServiceManager().provide(SqlService.class).get();
String aliasKey = conf.getNode("db", "mysql", "spongeSqlAlias").getString();
db = service.getDataSource(service.getConnectionUrlFromAlias(aliasKey).orElse(defaultDsn));
```

This issue related to maria db jdbc driver and was resolved in release 2.2.5. Corresponding ticket: [CONJ-604](https://jira.mariadb.org/browse/CONJ-604).

**Solution**

Update mariadb-java-client dependency version to 2.2.5.

**Incompatibilities**

Not found. According to MariaDB Connector/J release notes from 2.0.3 up to 2.2.5 there is no BC breaking changes.

**Possible workarounds**

Clone sponge repo, change dependency version and build. Or repack existing sponge jar with new mariadb driver. This is bad and not recommended.

ryantheleach:
@gabizou What amount of testing would be required to get this relatively simple dependency change into the next minor/major (assuming has binary changes, and that the connector follows semver) SpongeAPI builds.

ryantheleach:
I've done a minimal amount of testing against `org.mariadb.jdbc:mariadb-java-client:2.3.0` . 

Using that connector you can indeed make a connection, and execute a query (as long as you are using an authentication method mariadb-java-client supports, see comments on #2148).

I'm no expert of the advanced usages of jdbc connectors, so I'm not sure what else would need testing, other then running with some plugins that use jdbc connections.

dualspiral:
Okay, my apologies for not acting upon this all that time ago.

There was another issue whereby the default MySQL 8.0.11+ authentication mode wasn't supported and from what I see, that was resolved in version 2.5. So we're going to see if the 2.6 release will work with what we expose and, if so, we'll push that out.

BrainStone:
[...]

PR #2152, opened 2018-12-03
MySQL 8.0 SHA-2 auth support with proper conflict avoidance for MariaDB connector in same classpath

ghost:
Tested on MySQL 8.0.13 with SHA-2 auth only. Please test using MariaDB/MySQL 5 if you are able.

Includes compilation test for ensuring that the proper driver is selected.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,8 +67,9 @@ dependencies {
     runtime 'org.apache.logging.log4j:log4j-slf4j-impl:2.8.1'
 
     // Database connectors
-    compile 'com.zaxxer:HikariCP:2.6.3'
-    runtime 'org.mariadb.jdbc:mariadb-java-client:2.0.3'
+    compile 'com.zaxxer:HikariCP:3.2.0'
+    runtime 'mysql:mysql-connector-java:8.0.13'
+    runtime 'org.mariadb.jdbc:mariadb-java-client:2.3.0'
     compile 'com.h2database:h2:1.4.196'
     runtime 'org.xerial:sqlite-jdbc:3.20.0'
 
--- a/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
+++ b/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
@@ -48,6 +48,8 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.sql.DriverManager;
 import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
@@ -175,9 +177,10 @@ public class SqlServiceImpl implements SqlService, Closeable {
 
     public static class ConnectionInfo {
 
-        private static final Pattern URL_REGEX = Pattern.compile("(?:jdbc:)?([^:]+):(//)?(?:([^:]+)(?::([^@]+))?@)?(.*)");
+        private static final Pattern URL_REGEX = Pattern.compile("(?:jdbc:)?([^:]+):(//)?(?:([^:]+)(?::([^@]+))?@)?([^?]+)(?:(\\?)(?=.)(.+))?");
         @Nullable private final String user;
         @Nullable private final String password;
+        @Nullable private final String[] options;
         private final String driverClassName;
         private final String authlessUrl;
         private final String fullUrl;
@@ -190,9 +193,10 @@ public class SqlServiceImpl implements SqlService, Closeable {
          * @param authlessUrl A JDBC url for this driver not containing authentication information
          * @param fullUrl The full jdbc url containing user, password, and database info
          */
-        public ConnectionInfo(@Nullable String user, @Nullable String password, String driverClassName, String authlessUrl, String fullUrl) {
+        public ConnectionInfo(@Nullable String user, @Nullable String password, String[] options, String driverClassName, String authlessUrl, String fullUrl) {
             this.user = user;
             this.password = password;
+            this.options = options;
             this.driverClassName = driverClassName;
             this.authlessUrl = authlessUrl;
             this.fullUrl = fullUrl;
@@ -206,6 +210,11 @@ public class SqlServiceImpl implements SqlService, Closeable {
         @Nullable
         public String getPassword() {
             return this.password;
+        }
+
+        @Nullable
+        public String[] getOptions() {
+            return options;
         }
 
         public String getDriverClassName() {
@@ -266,13 +275,11 @@ public class SqlServiceImpl implements SqlService, Closeable {
             if (container != null && derelativizer != null) {
                 serverDatabaseSpecifier = derelativizer.apply(container, serverDatabaseSpecifier);
             }
-            final String unauthedUrl = "jdbc:" + protocol + (hasSlashes ? "://" : ":") + serverDatabaseSpecifier;
+            final boolean hasOptions = match.group(6) != null;
+            final String[] options = hasOptions ? match.group(7).split("&") : new String[0];
+            final String unauthedUrl = hasOptions ? "jdbc:" + protocol + (hasSlashes ? "://" : ":") + serverDatabaseSpecifier + "?" + match.group(7) : "jdbc:" + protocol + (hasSlashes ? "://" : ":") + serverDatabaseSpecifier;;
             final String driverClass = DriverManager.getDriver(unauthedUrl).getClass().getCanonicalName();
-            if (protocol.equals("mysql")) {
-                return new ConnectionInfo(user, pass, driverClass, unauthedUrl + "?disableMariaDbDriver", fullUrl);
-            } else {
-                return new ConnectionInfo(user, pass, driverClass, unauthedUrl, fullUrl);
-            }
+            return new ConnectionInfo(user, pass, options, driverClass, unauthedUrl, fullUrl);
         }
     }
 
--- a/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
+++ b/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
@@ -24,14 +24,11 @@
  */
 package org.spongepowered.common.service.sql;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-
 import com.google.common.base.Objects;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import com.google.common.cache.RemovalListener;
-import com.google.common.cache.RemovalNotification;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Sets;
 import com.zaxxer.hikari.HikariConfig;
@@ -43,15 +40,16 @@ import org.spongepowered.api.util.annotation.NonnullByDefault;
 import org.spongepowered.common.SpongeImpl;
 import org.spongepowered.common.config.SpongeConfigManager;
 
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import javax.sql.DataSource;
 import java.io.Closeable;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.sql.DriverManager;
 import java.sql.SQLException;
-import java.util.ArrayList;
 import java.util.Collections;
-import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
@@ -61,9 +59,7 @@ import java.util.function.BiFunction;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
-import javax.sql.DataSource;
+import static com.google.common.base.Preconditions.checkNotNull;
 
 
 /**
@@ -246,6 +242,7 @@ public class SqlServiceImpl implements SqlService, Closeable {
             ConnectionInfo that = (ConnectionInfo) o;
             return Objects.equal(this.user, that.user)
                     && Objects.equal(this.password, that.password)
+                    && Objects.equal(this.options, that.options)
                     && Objects.equal(this.driverClassName, that.driverClassName)
                     && Objects.equal(this.authlessUrl, that.authlessUrl)
                     && Objects.equal(this.fullUrl, that.fullUrl);
@@ -253,7 +250,7 @@ public class SqlServiceImpl implements SqlService, Closeable {
 
         @Override
         public int hashCode() {
-            return Objects.hashCode(this.user, this.password, this.driverClassName, this.authlessUrl, this.fullUrl);
+            return Objects.hashCode(this.user, this.password, this.options, this.driverClassName, this.authlessUrl, this.fullUrl);
         }
 
         /**
@@ -282,6 +279,7 @@ public class SqlServiceImpl implements SqlService, Closeable {
             }
             final boolean hasOptions = match.group(6) != null;
             final Set<String> rawOptions = hasOptions ? Sets.newHashSet(match.group(7).split("&")) : Collections.EMPTY_SET;
+            // Potential to modify flags here.
             final String options = !rawOptions.isEmpty() ? "?" + String.join("&", rawOptions) : "";
             final String unauthedUrl = "jdbc:" + protocol + (hasSlashes ? "://" : ":") + serverDatabaseSpecifier;
             final String driverClass;
--- a/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
+++ b/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
@@ -134,11 +134,7 @@ public class SqlServiceImpl implements SqlService, Closeable {
                     if (driverSpecificProperties != null) {
                         config.setDataSourceProperties(driverSpecificProperties);
                     }
-                    if (key.getDriverClassName().equals("com.mysql.cj.jdbc.Driver")) {
-                        config.setJdbcUrl(key.getAuthlessUrl() + "?disableMariaDbDriver");
-                    } else {
-                        config.setJdbcUrl(key.getAuthlessUrl());
-                    }
+                    config.setJdbcUrl(key.getAuthlessUrl());
                     return new HikariDataSource(config);
                 }
             });
--- a/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
+++ b/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
@@ -134,7 +134,11 @@ public class SqlServiceImpl implements SqlService, Closeable {
                     if (driverSpecificProperties != null) {
                         config.setDataSourceProperties(driverSpecificProperties);
                     }
-                    config.setJdbcUrl(key.getAuthlessUrl());
+                    if (key.getDriverClassName().equals("com.mysql.cj.jdbc.Driver")) {
+                        config.setJdbcUrl(key.getAuthlessUrl() + "?disableMariaDbDriver");
+                    } else {
+                        config.setJdbcUrl(key.getAuthlessUrl());
+                    }
                     return new HikariDataSource(config);
                 }
             });
@@ -268,13 +272,11 @@ public class SqlServiceImpl implements SqlService, Closeable {
             }
             final String unauthedUrl = "jdbc:" + protocol + (hasSlashes ? "://" : ":") + serverDatabaseSpecifier;
             final String driverClass = DriverManager.getDriver(unauthedUrl).getClass().getCanonicalName();
-            final String modifiedUnauthedUrl;
             if (protocol.equals("mysql")) {
-                modifiedUnauthedUrl = unauthedUrl + "?disableMariaDbDriver";
+                return new ConnectionInfo(user, pass, driverClass, unauthedUrl + "?disableMariaDbDriver", fullUrl);
             } else {
-                modifiedUnauthedUrl = unauthedUrl;
+                return new ConnectionInfo(user, pass, driverClass, unauthedUrl, fullUrl);
             }
-            return new ConnectionInfo(user, pass, driverClass, modifiedUnauthedUrl, fullUrl);
         }
     }
 
--- a/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
+++ b/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
@@ -85,7 +85,7 @@ public class SqlServiceImpl implements SqlService, Closeable {
         mySqlProps.setProperty("useConfigs",
                 "maxPerformance"); // Config options based on http://assets.en.oreilly
                 // .com/1/event/21/Connector_J%20Performance%20Gems%20Presentation.pdf
-        build.put("com.mysql.jdbc.Driver", mySqlProps);
+        build.put("com.mysql.cj.jdbc.Driver", mySqlProps);
         build.put("org.mariadb.jdbc.Driver", mySqlProps);
 
         PROTOCOL_SPECIFIC_PROPS = build.build();
@@ -268,7 +268,13 @@ public class SqlServiceImpl implements SqlService, Closeable {
             }
             final String unauthedUrl = "jdbc:" + protocol + (hasSlashes ? "://" : ":") + serverDatabaseSpecifier;
             final String driverClass = DriverManager.getDriver(unauthedUrl).getClass().getCanonicalName();
-            return new ConnectionInfo(user, pass, driverClass, unauthedUrl, fullUrl);
+            final String modifiedUnauthedUrl;
+            if (protocol.equals("mysql")) {
+                modifiedUnauthedUrl = unauthedUrl + "?disableMariaDbDriver";
+            } else {
+                modifiedUnauthedUrl = unauthedUrl;
+            }
+            return new ConnectionInfo(user, pass, driverClass, modifiedUnauthedUrl, fullUrl);
         }
     }
 
--- a/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
+++ b/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
@@ -85,7 +85,7 @@ public class SqlServiceImpl implements SqlService, Closeable {
         mySqlProps.setProperty("useConfigs",
                 "maxPerformance"); // Config options based on http://assets.en.oreilly
                 // .com/1/event/21/Connector_J%20Performance%20Gems%20Presentation.pdf
-        build.put("com.mysql.jdbc.Driver", mySqlProps);
+        build.put("com.mysql.cj.jdbc.Driver", mySqlProps);
         build.put("org.mariadb.jdbc.Driver", mySqlProps);
 
         PROTOCOL_SPECIFIC_PROPS = build.build();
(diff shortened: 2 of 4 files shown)


src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
```diff
@@ -134,7 +134,11 @@ public HikariDataSource load(@Nonnull ConnectionInfo key) throws Exception {
                     if (driverSpecificProperties != null) {
                         config.setDataSourceProperties(driverSpecificProperties);
                     }
-                    config.setJdbcUrl(key.getAuthlessUrl());
+                    if (key.getDriverClassName().equals("com.mysql.cj.jdbc.Driver")) {
+                        config.setJdbcUrl(key.getAuthlessUrl() + "?disableMariaDbDriver");
```
ryantheleach: Info for anyone reviewing / testing: https://mariadb.com/kb/en/library/about-mariadb-connector-j/#having-mariadb-and-mysql-drivers-in-the-same-classpath

XakepSDK:
Oh god, why you just don't bump mariadb driver version?

ryantheleach:
**Arguments for having the mysql adapter directly:**

1. It prevents (other potential) compatibility problems with the MariaDB connector in the future.
2. We can presumably fix issues by updating the mysql connector faster then MariaDB can by updating MariaDB Connector.
3. We can fix #2148 on the next minor? major? RB? rather then later.
4. It saves people from having to bundle in the adapter into their own plugins, creating larger file sizes on Ore.
5. Fixing the related issue provides a better out of the box experience for users using mysql 8 using the default options.

**Possible arguments against:**

1. Additional dependencies
2. potential to have mismatching versions with plugins or mods?
3. Filesize of jars. Alternative: Potential to have 'official' plugins that provide the jdbc connectors that people need, shrinking the filesize, but making a greater barrier to entry.
4. There is an `easy` workaround on the mysql server, by using a compatible authentication method with mariadb connector. This was the default before mysql 8, and the change was driven by security concerns. However this means documenting the change, and getting it in front of the users it will affect, before they give up.

src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
```diff
@@ -268,7 +268,11 @@ public static ConnectionInfo fromUrl(@Nullable PluginContainer container, String
             }
             final String unauthedUrl = "jdbc:" + protocol + (hasSlashes ? "://" : ":") + serverDatabaseSpecifier;
             final String driverClass = DriverManager.getDriver(unauthedUrl).getClass().getCanonicalName();
-            return new ConnectionInfo(user, pass, driverClass, unauthedUrl, fullUrl);
+            if (protocol.equals("mysql")) {
+                return new ConnectionInfo(user, pass, driverClass, unauthedUrl + "?disableMariaDbDriver", fullUrl);
```
onyn: This hack produce invalid result when jdbc url already contains some parameters.

```java
String jdbcUrl = "jdbc:mysql://zml:totallymypassword@localhost/sponge?useSSL=false";
SqlServiceImpl.ConnectionInfo subject = SqlServiceImpl.ConnectionInfo.fromUrl(null, jdbcUrl);
System.out.println(subject.getAuthlessUrl());
// outputs: jdbc:mysql://localhost/sponge?useSSL=false?disableMariaDbDriver
```

Checking for question mark and appending `&disableMariaDbDriver` is not a solution. Possibility of setting custom data source properties by user as key-value map in sponge configs needed. Also there is need to maintain backward compatibility with already existing configurations which put properties in jdbc url. This involves parsing of query string in fromUrl factory method as well.

onyn:
I think there is no need for dependency on mysql connector just for providing `caching_sha2_password` support - this problem have excellent workaround at server side.

ghost:
“Just for providing support”... except for the fact that it’s the default authentication method on the most popular relational database management system.

XakepSDK:
Do not touch this. Possible side effects are bigger, than benefits. Having 2 drivers for 1 database in one classpath is bad idea.
Its really easy to change auth type in mysql 8. Just change it(auth method in your db user) until mariadb connector will support cached sha2 pluggable auth.

ghost:
It’s funny that someone suggested this can cause problems with JDBC URLs that already contain parameters when the SqlService obtains information by dissecting a rigid regex-based syntax that doesn’t allow parameters. 

It’s also funny that someone suggested that this can have “bigger side effects” without reading the ‘official’ suggestion to do this from MariaDB.

gabizou:
> It’s funny that someone suggested this can cause problems with JDBC URLs that already contain parameters when the SqlService obtains information by dissecting a rigid regex-based syntax that doesn’t allow parameters.
> 
> It’s also funny that someone suggested that this can have “bigger side effects” without reading the ‘official’ suggestion to do this from MariaDB. You are a side effect. Change my mind.

Take the tone down a notch. People are trying to ensure that by updating dependencies, we don't break any workflows, nor would we want to have issues after the update that requires more handling and configuration than necessary. Don't be a dick.

ghost:
@gabizou This is an issue that's important to me and so I put in the research and work (regardless of how minimal) to resolve the issue without creating any incompatibilities with the current implementation. It's been about a year since MySQL 8.0 introduced cached sha-2 as the default authentication method and MariaDB isn't embracing it. **Do not touch this** without any testing or review of the actual code is not constructive or helpful and deserves an equally as constructive response.

FWIW I am actively developing plugins that leverage a MySQL 8.0.13 database and my changes are working perfectly.

XakepSDK:
Is it worth, increasing sponge jar file for just adding `caching_sha2_password` feature?
How much admins use MySQL 8?
What real profit gives us `caching_sha2_password`?
Why we should add/replace current driver with your one?

ghost:
It's not "replacing" anything and the difference in jar size is negligible. It spares plugin developers the need to shade HikariCP, slf4j, and a connector into each of their plugins (obvious benefit). Moreover, what is the benefit in not supporting the default authentication method of MySQL? Insofar as Sponge's barriers to entry go, there's no need to alienate its audience further by refusing to accommodate **the default auth mechanism for the most prevalent database management system in existence**.

brjcarbone:
Screw it, I'll make my own SqlService plugin.

ryantheleach:
Please don't confuse the comments and opinions of the members of the community, with that of the Sponge team. I've been trying to work closely with you to get this through, please don't let this effort on both our parts go to waste.

It's nearly **always** easier to create your own plugins / mods / nms hooks to solve a problem, then getting **any** community to agree to changes / PR's from outside contributors. That doesn't mean it's not worth the effort, worth doing right, and worth amplifying the effects of the platform for others. Without people like yourself contributing, causing discussions, and even becoming members of the team, Sponge wouldn't be the platform it is today.

All that was asked was the tone to remain civil.

src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
ryantheleach: The following unit test 

```
@Test
    public void testMysqlConnectionInfoAdvanced() throws SQLException {
        final String jdbcUrl = "jdbc:mysql://zml:totallymypassword@localhost/sponge?useSSL=false";
        final SqlServiceImpl.ConnectionInfo subject = SqlServiceImpl.ConnectionInfo.fromUrl(null, jdbcUrl);

        assertEquals("zml", subject.getUser());
        assertEquals("totallymypassword", subject.getPassword());
        assertEquals("jdbc:mysql://localhost/sponge?useSSL=false", subject.getAuthlessUrl());
        assertEquals("org.mariadb.jdbc.Driver", subject.getDriverClassName());
    }
```
Shows that (without this PR) the url parameters are just passed straight through.

> Checking for question mark and appending &disableMariaDbDriver is not a solution. Possibility of setting custom data source properties by user as key-value map in sponge configs needed. Also there is need to maintain backward compatibility with already existing configurations which put properties in jdbc url. This involves parsing of query string in fromUrl factory method as well.

I don't think this is strictly necessary for this PR. Is there a good reason why we should explode those properties in the Sponge config, rather then just following the existing standard of jdbc urls?

Alternate things that I had discussed with @brjcarbone in voice, was parsing and manipulating the URL using existing URL building / manipulation. But I havn't had time yet to investigate if there is something simple in the standard libraries, or any of the utility libraries that we use in the Sponge implementation.

If someone has a reliable fix for this, I'm all ears.

I hadn't yet commented on the string concatenation yet, because I only just tested that the parameters indeed do pass through the regex.

Before this I was told the regex would need updating, and that the parameters were stripped.

ghost: You're actually correct, I misinterpreted the regex and everything after `@` is Group 5. In hindsight, I can remove the MySQL logic and put disabling the MariaDB driver into the hands of the user. Or I can update the regex to consider a "Group 6" that distinguishes options from the database name.

ryantheleach: Yeah, I'm not sure which would be preferred.

Letting the user handle it gives more control, but complicates it for a first time user, meaning documentation etc, which is one of the arguments for accepting this PR in the first place.

Manipulating the URL ourselves, wouldn't be too hard, but manipulating complicated schemes such as emails and URL's has bitten me in the ass before without using a well tested library to do so, If we have something in the classpath that can do that, It'd be a nobrainer to me to add it for mysql.

onyn: All sponge supported databases accept login and password inside jdbc url.

H2:

```
jdbc:h2:<url>[;USER=<username>][;PASSWORD=<value>]
jdbc:h2:file:~/sample;USER=sa;PASSWORD=123
```

MySQL/MariaDB:

```
jdbc:mysql://localhost:3306/DB?user=root&password=myPassword
```

Sqlite doesn't have any athentication, so there in no syntax for passing credentials.

So, any jdbc url parsing is not needed at all. Introducing custom auth syntax in jdbc url is sponge's major flaw, which must be supported forever. This does not mean that you should touch it too.

> I don't think this is strictly necessary for this PR. Is there a good reason why we should explode those properties in the Sponge config, rather then just following the existing standard of jdbc urls?

You think right. This is not necessary. I just forgot that options in jdbc url may be intermixed with `setDataSourceProperties` method. Hence there is no reasons to introduce new syntax construct in sponge configs and change `SqlServiceImpl` or `ConnectionInfo` signature.

> Letting the user handle it gives more control, but complicates it for a first time user, meaning documentation etc, which is one of the arguments for accepting this PR in the first place.

Exposing this implementation detail to user makes impossible to get rid of mysql connector in the future, when mariadb connector finally implements this new auth scheme. 

The only way to properly inject `disableMariaDbDriver` without affecting user experience - is adding it in `PROTOCOL_SPECIFIC_PROPS` collection. Need additional testing to ensure that other supported jdbc libraries ignores this option if they do not understand it.

src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
```diff
@@ -190,9 +193,10 @@ public void close() throws IOException {
          * @param authlessUrl A JDBC url for this driver not containing authentication information
          * @param fullUrl The full jdbc url containing user, password, and database info
          */
-        public ConnectionInfo(@Nullable String user, @Nullable String password, String driverClassName, String authlessUrl, String fullUrl) {
+        public ConnectionInfo(@Nullable String user, @Nullable String password, String[] options, String driverClassName, String authlessUrl, String fullUrl) {
             this.user = user;
```
onyn, line 199: Changing signature of public method leads to change of MAJOR version due to semver. You need to introduce new constructor if you do not want affect version number of the whole project.

Also you forgot to change `equals` and `hashCode` methods.

XakepSDK: That's ok, this is implementation change, not API.
But `equals` and `hashCode` should be implemented, that's flaw.

onyn: Anyway, this is breaking change. You can't be sure that no one using this class as it has public visibility.

Also I think, that introducing new field is unnecessary. My explanations about this, listed in another [review comment](https://github.com/SpongePowered/SpongeCommon/pull/2152#discussion_r238593182).

NeumimTo: You can simply make it the change unbreaking by passing the options as last parameter and using syntax for varargs instead of []

ghost: WIP, it’ll be adjusted again tm.

dualspiral: > Anyway, this is breaking change.

> You can't be sure that no one using this class as it has public visibility.

It doesn't matter here in SpongeCommon. We don't make _any_ guarantees on the implementation. Anyone building against Common is doing so with the understanding that this might break build to build and is therefore at their own risk. We only guarantee that the Sponge API (and the associated dependencies with that) will not have any API breaks.

As long as it doesn't break Sponge API compatibility, do as you need to here.

onyn: Public and protected modifiers always means that they are part of public exported api all over the world, but not in SpongeCommon :-/

Why not make the absence of guarantees explicit by making this class package-private? This make intentions more clear and less confusing for newcomers like me.

src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
```diff
@@ -266,9 +278,15 @@ public static ConnectionInfo fromUrl(@Nullable PluginContainer container, String
             if (container != null && derelativizer != null) {
                 serverDatabaseSpecifier = derelativizer.apply(container, serverDatabaseSpecifier);
             }
+            final boolean hasOptions = match.group(6) != null;
+            final Set<String> rawOptions = hasOptions ? Sets.newHashSet(match.group(7).split("&")) : Collections.EMPTY_SET;
+            if (protocol.equals("mysql")) {
```
dualspiral: Rather than messing about with options, could you not just set the name of the driver class to the MySQL driver so that we explicity set it, rather than pull apart all the options and force the MariaDB driver to parse it first? It seems to me that you're doing all this work to get a specific JDBC driver, without specifying that specific driver. Why say "MariaDB driver, reject this protocol and send it off to someone else" when you can say "I know I want MySQL to handle this, this is the class it needs to be"? 

I'd think all you need to do is remove all this option parsing stuff and set the driverClass explicitly (also extracting the Driver class name into its own constant):

```java
private final static String MYSQL_DRIVER_CLASS_NAME = "com.mysql.cj.jdbc.Driver";
...

final String driverClass;
if (protocol.equals("mysql")) {
    driverClass = MYSQL_DRIVER_CLASS_NAME; // set explicitly for MySQL 8 compatibility
} else {
    driverClass = DriverManager.getDriver(unauthedUrl).getClass().getCanonicalName(); // let JDBC decide
}
```

I know that the MariaDB pages talk about the `disableMariaDbDriver` option, but that's for people who write the connection URLs to gain control without changing source code. As you're taking that control away by appending it the way you are, you might as well just explicitly select the MySQL driver, pass the connection string to that driver, and be done with it.

dualspiral (commented):
Am I missing something here?

src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
ghost: Yeah, that's a slightly cleaner way to do it--I'll push that change and get rid of the options modification. 

I'm _not_ entirely opposed to leaving the part which distinguishes options from other groups in the regex, though. It leaves a proper method to interfere with flags in the future, if necessary.

ghost: Disregarding the erroneous certificate error, this PR should be complete. The latest push includes your suggested changes and maintains the increased modularity of the new regex.

src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
JBYoshi: If we were to do that, that would mean making every class in SpongeCommon package-private. However, since we use many different packages, that would mean SpongeCommon wouldn't be able to talk to itself.

The reason this isn't part of the API is that we already have an explicit API - SpongeAPI. Anything that isn't part of SpongeAPI is not part of the API. People can use the internal code if they want to, but we don't provide any expectations of compatibility.

(Most of the major Minecraft projects do it the same way. Spigot has lots of public classes, but it explicitly specifies which ones are part of its API, so everything else is not part of the API. Same thing with BungeeCord, Waterfall, PaperSpigot, Bukkit/CraftBukkit, Canary, and others.)

ghost:
Final review of changes to fully clarify what is happening here:
- The JDBC URL regex has been expanded upon to allow for direct interaction with user-specified flags should the need arise (nearly the case in this instance). This breaks the "database" group into two groups: `database` and `options`. The implementation in this PR prevents accidental duplicate flags and detects a syntactical error in which the URL ends in a trailing `?` without options.

- The MySQL 8.0 connector (`mysql:mysql-connector-java:8.0.13`) has been shaded and jurisdiction over MySQL JDBC URLs is delegated to this connector rather than the MariaDB connector. This provides support for SHA-2 authentication, removing the need for end users to modify the default authentication method of their MySQL users in order to accommodate for MariaDB connector's lack of support. The MySQL driver name has been properly updated to reflect the name required in 8.0 (`com.mysql.cj.jdbc.Driver`).

- The following dependencies have been updated to their respective latest versions given the current methods have not changed and incompatibilities have not arisen: `com.zaxxer:HikariCP:3.2.0`, `org.mariadb.jdbc:mariadb-java-client:2.3.0`

- Additional unit tests have been added to confirm proper parsing of JDBC URLs under the new regex standard.

If you have MySQL 5.x or MySQL 8.0.x available for testing and you're willing to contribute to testing, please do so! On MySQL 8.0.x, attempt to test both Native and SHA-2 authentication if possible. Of course, ensure that MariaDB compatibility remains.

src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
onyn: Yeah, a lot of refactoring and project architecture revision required.

I'm upset, that all the projects you listed ignores such common and widespread programming practices.

ghost: Assuming that public methods are "public" in the sense that developers are expected to interact with them is unfair, really. To design a project perfectly such that this is the case seems unlikely. Often, non-static methods from one package may need to be used in another (non-sub) package.

onyn: It's not hard to make good design with adequate encapsulation from the ground up (look at the java standard library, for example). It is much more difficult to put in order the design of an already formed project. It's like putting a skeleton in meat. And finally, it is truly impossible to convince the community of the usefulness and necessity of such fundamental changes.

ghost: It's definitely sensible, but often you'll have a catch-all class or package for handlers, registration of sorts, etc. and multiple packages will hook into those "nodes." Of course, it's possible to implement an abstract class of these types when you need one, but you won't find much of that in Minecraftia.

dualspiral: I don't particularly want to silence this conversation, because I think it's important that we take the time to discuss things with others and explain our thinking if people would like to do so. That said, this is off topic for this PR now, so I'd like to ask that futher conversation regarding this, if any is needed, including asking about why we chose this direction, is directed to our forums or other channels, so we can focus on getting this PR reviewed. 

Thanks! :)

onyn:
Found one more problem. When someone uses mariadb with specific mariadb options and `jdbc:mysql` prefix - he is in trouble.

Server may fail to start or crash at some undetermined time in the future (which is a lot worse).

Example of such url:

```
jdbc:mysql:aurora://host1,host2/database
```

Aurora feature is not supported by mysql connector.

I don't know an easy way how to avoid this problem within this PR.

I still think that the best way to fix #2148 is wait until mariadb team fixes [CONJ-663](https://jira.mariadb.org/browse/CONJ-663) and update just mariadb connector.

dualspiral:
> Found one more problem. When someone uses mariadb with specific mariadb options and jdbc:mysql prefix - he is in trouble.

Then they should be using the `mariadb` protocol, not `mysql`. We cannot catch everything, if you're using a MariaDB specific feature, use the MariaDB specific protocol.

onyn:
If user interaction required in any circumstances, may be better just change auth plugin on the server side? Having two competing connectors is really awful.

dualspiral:
Honestly, yes, I would much rather see MariaDB update their connector which would allow this to all go away - but when is that going to be? Tomorrow, next week, next month, in a year? We cannot predict their timelines, but we can consider what _we_ want to do to mitigate this.

We have to keep in mind what this is all about, why this PR has come about. Some users would rather have more secure authentication with their MySQL database - that's honestly a good thing. However, many of our users are those that have no programming or server admin, experience and just want things to work. They've heard of MySQL, there are guides to setting that up, people use the latest versions and end up with auth errors that they don't understand.

It might be that this PR does not get accepted. While I've been helping it along to try to ensure that we can serve most users in a reasonable way, my opinion is by no means the loudest in the Sponge dev team - though I don't advocate either way. What I do care about is working out what's best for the community as a whole.

There is a decision that the Sponge team has to make here, whether we explicity include the MySQL connector for all time (in which case, we drop all the compatibility stuff and tell users to use the `disableMariaDbDriver` option), include it temporarily and educate users that they should only use MariaDB specific options in a string that expliticly targets the `mariadb` protocol (which is good practice anyway), we make the MySQL 8 support a plugin or optional mixin instead, or we don't include it at all and tell users that they have to downgrade the auth security on their database for the time being - yes, educating users in being DBAs for their servers isn't a terrible thing, but we have to consider our product's target.

This is not a decision I will be making alone - or at all, others in Sponge might do so. 

However, I am greatful that someone, in this case, @brjcarbone, _has_ taken the time to do this to show it can be done, and regardless of the outcome, I'm thankful for him sticking with it. He had an issue, and he dived in and did what he thought was right - even if we end up going a different direction. I want to make sure that's recognised too.

ghost:
Quite frankly, we can encourage users to enter the proper protocol in the comment above the alias node. “Enter ‘jdbc:mysql’ for MySQL only. Enter ‘jdbc:mariadb’ for MariaDB.” I predict that this change will affect a minority of users, but there’s not much else we can do aside from providing people with information.

Oh, and the “wait for Maria to update their connector” option is in poor taste; MySQL 8.0 has been around for a very long time. This PR is very late to the game.

ryantheleach:
Related discussion to the jdbc url's being non-standard in Sponge.

https://github.com/SpongePowered/SpongeAPI/issues/1726

I would request that if those review points on this PR are to be maintained in the future, that discussion relating it be posted there, and this PR reserved for the **sole** issue of MySQL 8, caching_sha2 auth support, and the additional inclusion of MySQL 8 jdbc drivers.

brjcarbone:
I would gladly address special parsing for file-based database setups in addition to my current PR, but I would really prefer to work on it atop my current changes than create two parallel PRs in a “choose one” fashion.

build.gradle
```diff
@@ -67,8 +67,9 @@ dependencies {
     runtime 'org.apache.logging.log4j:log4j-slf4j-impl:2.8.1'
 
     // Database connectors
-    compile 'com.zaxxer:HikariCP:2.6.3'
-    runtime 'org.mariadb.jdbc:mariadb-java-client:2.0.3'
+    compile 'com.zaxxer:HikariCP:3.2.0'
+    runtime 'mysql:mysql-connector-java:8.0.13'
+    runtime 'org.mariadb.jdbc:mariadb-java-client:2.3.0'
```
ryantheleach, line 72: This isn't a criticism of this PR (So please don't make changes without one of the devs chiming in)

But is updates to HikariCP / mariadb required for MySQL 8?

The question comes down to, is the fact that the drivers / pooling is provided by (these particular drivers / pool) an implementation specific thing that we can change major versions on fine, or is there potential issues with bumping it?

(https://github.com/brettwooldridge/HikariCP/blob/dev/CHANGES#L29) shows that deprecated methods were removed, so there *are* breaking changes if plugins were using HikariCP directly.

At first glance, the mariadb java-client bump seems fine (https://mariadb.com/kb/en/library/mariadb-connector-j-release-notes/).

ghost: It's certainly not necessary, but insofar as Sponge's SqlService is concerned, nothing in use is at risk of deprecation. HikariConfig setup methods will remain the same and there's really no reason not to update the dependency. Obviously, no MariaDB methods are in use and the driver name remains the same. Version 2.x is for Java 8 and 9, and Sponge already shaded version 2.x.

FWIW, I've been using SV with this PR merged and it's been perfect.

ryantheleach: I'm more worried about leaked implementation / Lantern.

ghost: Honestly, I'm fine with the PR getting merged without Hikari/Maria dependency updates if a proper risk assessment can't be made. I'm unfamiliar with anything not to do with Sponge. Given the impending Java 11 upgrade coming to Minecraft (and thus Forge), the dependencies will need to be updated at some point.

onyn: I agree that's important to stay with up to date dependencies. But this PR is not a good place to update HikariCP because this is not concerned original issue and because of potential compatibility issues, despite that leaked HikaryCP is completely not sponge problem. If we may maintain compatibility, why won't do that? There is nothing critical in HikariCP changelog that causes to update it asap thus breaking potential 0.000001% of users. It's better to move this update to the next major version or even when minecraft finally would support java 11.

> But is updates to HikariCP / mariadb required for MySQL 8?

Mariadb update is ok, because sponge's [contribution guidline](https://docs.spongepowered.org/stable/en/contributing/guidelines.html) notices that small changes (under 20 lines) must be merged with other smaller changes. That's why I don't make PR dedicated only to #2008.

src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
Zidane: The method signature break is fine. Anyone who builds against the implementation does so under their own peril.

gabizou:
Seeing as 1.12.2 is LTS, I’m making the decision that for the short term, this PR has had plenty of feedback and I want to commend the work put in. Long term however, I’d like to explore the split of database connectors as “core plugins” that aren’t included by default. This way the plugins themselves can be uploaded to something like Ore, and be updated more frequently without issue of breaking compatibility for several server owners. 

I’m inclined on approving this merge with the understanding that the core plugin idea will be implemented. @Zidane thoughts?

ghost:
My name isn't Zidane, but modularity for selecting a relevant connector plugin is a great idea. It doesn't change UX with regard to utilizing the SqlService and curbs concerns of increasing file size.

gabizou:
> My name isn't Zidane

You're right, I was asking for @Zidane's input since he's my fellow leader of the sponge project. 



> It doesn't change UX with regard to utilizing the SqlService and curbs concerns of increasing file size.

It most definitely does change the UX, since the service and it's adapters would be split up out of the sponge implementation.

ghost:
From a developer's perspective, not from a server owner's perspective. Server owners don't really use the SqlService implementation insofar as inputting an alias is concerned.

ryantheleach:
@brjcarbone "Server owners don't really use the SqlService implementation insofar as inputting an alias is concerned."

Splitting them into 'core plugins' would probably mean that end users would need to download the SpongeMySQL adapter off of Ore or something, before using MySQL databases.

Most of the team are very against automatic downloading of jar's and running them. 

However, for the common plugins, a helpful error could be displayed in the log.

Anyway, this is getting off-topic

ghost:
@Zidane I just wanted to send you a reminder to visit this PR since it's been over 2 weeks and you may have missed gazibou's request. Thank you!

PR #2161, opened 2018-12-14
Implement Text compact methods

Katrix:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1424) | **SpongeCommon**

Implements Text compaction in a similar way to how I did it in Minejson


gabizou (approved):

gabizou:

PR #2163, opened 2018-12-20
Add Multiple Changes per Block Position Capabilities

gabizou:
This is an overarching fix for adding the ability to properly track block changes from things like pistons by exposing internal intermediary transactions of block changes to positions. TLDR is at the bottom.

##### Ok, but why?

To understand the why, you need to get a little background. Currently, the [`PhaseTracker`](https://github.com/SpongePowered/SpongeCommon/blob/198e8ce33c40adb86e79d707e8d58f29a607f1e3/src/main/java/org/spongepowered/common/event/tracking/TrackingUtil.java#L336) will capture blocks via a [`CaptureListSupplier`](https://github.com/SpongePowered/SpongeCommon/blob/198e8ce33c40adb86e79d707e8d58f29a607f1e3/src/main/java/org/spongepowered/common/event/tracking/context/CapturedBlocksSupplier.java) object, which is pretty much a wrapper around a lazily initialized `List` of `BlockSnapshot`s. This is all fine and dandy when you consider most things happening in the game will be one-off or multiple placement transactions, but for things like Pistons, they perform a lot of operations involving block changes, tile entity replacements, and individualized neighbor notifications after blocks and tiles have been moved/placed. What ends up happening at this point is that the `List` will have duplicated block positions (read: multiple `BlockSnapshot`s and therefor, multiple [`Transaction<BlockSnapshot>`s](https://github.com/SpongePowered/SpongeCommon/blob/198e8ce33c40adb86e79d707e8d58f29a607f1e3/src/main/java/org/spongepowered/common/event/tracking/TrackingUtil.java#L468) to exist, and worst yet, they will only understand the *original* block change, not the "final" change.

##### I guess, but why is that a problem?

It's a problem because take this situation: A Piston is going to move. What does it do?
- Iterates over all blocks it needs to destroy
  - Tells the block to drop itself as items (snow drops snowballs)
  - Sets the block to air with no notifications
- Iterates over all blocks it needs to move
  - Sets the original position to air with no notifications
  - Offsets to the facing direction
  - Sets the offsetPosition to a Piston Extension block with no notifications
  - Sets a TileEntity for moving piston on the offset position
- If the piston is extending
  - Sets the "piston head" piston extension block with no notifications
  - Sets the piston tile entity
- Iterates in *reverse order* destroyed blocks first, then moved blocks
  - Sends a neighbor notification of the original blocks that were destroyed/moved
- If extending, sends a neighbor notification to the extended piston head

Now, this may be a little confusing, but, with the multiple blocks being moved, you can have a position potentially having **two** transactions! Let's explain:
1) Block being moved gets set to air
2) Block being moved gets set to Piston Extension
Respectively, this would be represented incorrectly as two different `Transaction`s and worst yet, because of the core logic with how snapshots are generated for these transactions, [the snapshot only gets a change flag assigned when the block change is captured](https://github.com/SpongePowered/SpongeCommon/blob/198e8ce33c40adb86e79d707e8d58f29a607f1e3/src/main/java/org/spongepowered/common/event/tracking/TrackingUtil.java#L346), so the first transaction will have a `BREAK`, and the second would get a `PLACE`, whereas we both know in reality, it's a `MODIFY` in cases where there was an existing block in place!

##### Huh, I guess... So, how does this PR fix that?

Welcome to my world... The few hundreds of lines being changed involve specifying certain `IPhaseState`s to selectively capture not just block changes, but also `TileEntity` changes, and neighbor notification requests. This got a little [messy in the later stages](https://github.com/SpongePowered/SpongeCommon/blob/e6e0e53042c88838d527ab464b64a7725ddce747/src/main/java/org/spongepowered/common/event/tracking/context/MultiBlockCaptureSupplier.java#L571-L622) especially for transaction application; however, it is properly capturing and playing back the physics that would otherwise have not been captured had we disabled capturing entirely.... For the most part.... What this does now is builds a singular `Transaction<BlockSnapshot>` only providing the *updated* change according to [the capture supplier having multiple changes for the desired position](https://github.com/SpongePowered/SpongeCommon/blob/9906e5ef315ba70301bd8257677f84f410b96bd1/src/main/java/org/spongepowered/common/event/tracking/context/MultiBlockCaptureSupplier.java#L90-L168) while still being lazily evaluated. This neat change will apply globally to all `IPhaseState`s that end up potentially capturing multiple changes to a single block position, but the advantage is that for the specific `IPhaseState`s that need more refined tracking (read: tile entity changes and neighbor notifications), that is added as a possible [tracking hook to `IPhaseState`](https://github.com/SpongePowered/SpongeCommon/blob/9906e5ef315ba70301bd8257677f84f410b96bd1/src/main/java/org/spongepowered/common/event/tracking/phase/tick/BlockEventTickPhaseState.java#L100-L118).

##### Why did you say "For the most part...."?

Well... Here's the catch.... Tracking is hard. The way that Sponge handles tracked block changes is by allowing block changes to pass through to the [`Chunk without making explicit calls to performing block physics`](https://github.com/SpongePowered/SpongeCommon/blob/9906e5ef315ba70301bd8257677f84f410b96bd1/src/main/java/org/spongepowered/common/mixin/core/world/MixinChunk.java#L602-L737). Because of that, there's potential cases where [played back notifications](https://github.com/SpongePowered/SpongeCommon/blob/9906e5ef315ba70301bd8257677f84f410b96bd1/src/main/java/org/spongepowered/common/event/tracking/context/BlockTransaction.java#L265-L272) or [`onBlockAdded` calls](https://github.com/SpongePowered/SpongeCommon/blob/9906e5ef315ba70301bd8257677f84f410b96bd1/src/main/java/org/spongepowered/common/event/tracking/context/BlockTransaction.java#L198-L228) can have different BlockStates than what currently exists on the `Chunk`, leading to vanilla logic to validate itself to setting a block to Air instead of performing the desired operation that it originally intended to perform. As much as I'd like to explain the depth of that logic, it comes down to this:
Because tracking injects itself to capture changes as transactions, but allows the Chunk to still get the block change, without playing any physics, and then plays back the "physics" after all of those changes to the target `Chunk`, we are slightly modifying the order of operations that ends up causing (as an example) [T-flop pistons](https://minecraft.gamepedia.com/Tutorials/Piston_circuits#T-Flip_Flops) to potentially break as the tile entity being processed at the time will see an invalid state and set itself to air... I'm working on [re-applying a systematic way to provide in-progress `IBlockState`s to the World to alleviate this issue](https://github.com/SpongePowered/SpongeCommon/blob/9906e5ef315ba70301bd8257677f84f410b96bd1/src/main/java/org/spongepowered/common/event/tracking/context/SpongeProxyBlockAccess.java#L37), but that is an experiment I'm currently working on.

#### So, it's complicated, I get it. I don't fully understand what or how this all works, but I guess you do... My question is, is it going to cause lag?

Maybe? From what I've seen, it's not causing any increase in lag from my limited testing. Once I've ~~called it quits to resolving the 1 tick piston issue~~ decided this branch is satisfactory with how it functions, I'll be reviewing performance comparisons with `stable-7` and sending out some test builds on discord. For the most part, this entire system was designed with minimal memory footprint additions as possible. Multi-block changes are not adding any new objects until it is detected a duplicate transaction is being added for a position, and even then, when that happens, the actual `BlockTransaction` bits aren't used until certain `IPhaseState`s decide it needs to use them. AND EVEN THEN, it's optimized in that it only stores references to objects that otherwise we'd need to be able to play back (process) or cancel (restore). Restores are working beautifully btw, but playback is the only issue at current writing.

#### TLDR; Previous tracking system doesn't know how to handle two block changes to a single position. This fixes it, and allows using ChangeBlockEvent to cancel pistons. It's complicated.


src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayer.java
```diff
@@ -189,6 +189,8 @@ public void construct(World worldIn, GameProfile gameProfileIn, CallbackInfo ci)
         this.targetedLocation = VecHelper.toVector3d(worldIn.getSpawnPoint());
     }
 
+
```
Aaron1011, line 192: These lines can be removed

src/main/java/org/spongepowered/common/event/tracking/context/BlockPosMultiMap.java
```diff
@@ -0,0 +1,268 @@
...
+     * {@link BlockPos} being changed will have an "original" snapshot to compare
+     * to what currently exists within a {@link World}. This list is intended not
+     * to be added to and will result in unmodifiable exceptions.
+     * @return
+     */
+    public final List<SpongeBlockSnapshot> get() {
+        return Collections.unmodifiableList(this.snapshots == null ? Collections.emptyList() : this.snapshots);
+    }
+
+    public final void prune(BlockSnapshot snapshot) {
```
Aaron1011: Should there be a method body, or can this be removed?

src/main/java/org/spongepowered/common/event/tracking/context/BlockPosMultiMap.java
```diff
@@ -0,0 +1,268 @@
...
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.BiConsumer;
+
+import javax.annotation.Nullable;
+
+public final class BlockPosMultiMap implements ICaptureSupplier {
+
+    @Nullable private ListMultimap<BlockPos, SpongeBlockSnapshot> multimap;
+    @Nullable private List<SpongeBlockSnapshot> snapshots;
```
Aaron1011: Based on how this field is used, it looks like you could just initialize it here - none of the other methods ascribe any special meaning to it being `null`.

Aaron1011 (changes requested):

src/main/java/org/spongepowered/common/event/tracking/context/BlockPosMultiMap.java
gabizou: There's plenty of meaning for it to be `null` by default:
- No blocks captured? No wasted memory
- `isEmpty()` checks are blazing fast
- `put(BlockSnapshot)` is more fasts on initialization due to fail fast if statements on nullability
- Because there's thousands of these objects being created per tick, they need to construct and be as lightweight as possible.

Remember, a lot of the capture suppliers have had nullable fields for these reasons to lower the average memory cost of the PhaseTracker in general.

src/main/java/org/spongepowered/common/event/tracking/context/BlockPosMultiMap.java
```diff
@@ -0,0 +1,317 @@
...
+     *     exposed the {@link Transaction#getOriginal()} as the first
+     *     {@link BlockSnapshot} that would exist in this list. Intermediary
+     *     {@link BlockSnapshot} changes for that postiion are internally
+     *     utilized to process physics, but are not exposed to the event.</li>
+     * </ul>
+     *
+     * @return An unmodifiable list of first block originals being changed
+     */
+    public final List<SpongeBlockSnapshot> get() {
+        return Collections.unmodifiableList(this.snapshots == null ? Collections.emptyList() : this.snapshots);
```
liach: The empty list is already unmodifiable, so
```java
return this.snapshots == null ? Collections.emptyList() : Collections.unmodifiableList(this.snapshots);
```

src/main/java/org/spongepowered/common/event/tracking/context/BlockPosMultiMap.java
liach: guess this is still wip.

src/main/java/org/spongepowered/common/event/tracking/phase/tick/BlockTickPhaseState.java
```diff
@@ -116,9 +117,9 @@ public boolean doesCaptureEntitySpawns() {
 
     @Override
     public void postTrackBlock(BlockSnapshot snapshot, BlockTickContext context) {
-        if (context.shouldProcessImmediately()) {
-            TrackingUtil.processBlockCaptures(context.getCapturedBlocks(), this, context);
-            context.getCapturedBlockSupplier().get().remove(snapshot);
+        if (false) {
```
liach: add todo?

gabizou:
Just to clarify, this *is* still a work in progress for implementation.

gabizou:
For those curious, I've updated the PR description explaining the original issue, and how this solves it, along with some side effects (and improvements).

src/main/java/org/spongepowered/common/event/tracking/PhaseTracker.java
```diff
@@ -604,31 +609,45 @@ public void notifyBlockOfStateChange(final IMixinWorldServer mixinWorld, final B
...
-        final IBlockState iblockstate = ((WorldServer) mixinWorld).getBlockState(notifyPos);
+        final IBlockState notifyState = ((WorldServer) mixinWorld).getBlockState(notifyPos);
+
+        performNeighborNotificationOnTarget(mixinWorld, notifyPos, sourceBlock, sourcePos, notifyState);
+    }
 
+    @SuppressWarnings("rawtypes")
+    public void performNeighborNotificationOnTarget(IMixinWorldServer mixinWorld, BlockPos notifyPos, Block sourceBlock, BlockPos sourcePos,
+        IBlockState iblockstate) {
+        if (notifyPos.getX() == -1100 && notifyPos.getY() == 58 && notifyPos.getZ() == -496) {
```
Zidane: Tsk.

src/main/java/org/spongepowered/common/event/tracking/context/BlockTransaction.java
```diff
@@ -0,0 +1,275 @@
...
+
+    abstract void process(Transaction<BlockSnapshot> eventTransaction, IPhaseState phaseState, PhaseContext<?> phaseContext,
+        Map<BlockPos, IBlockState> processingBlocks, int currentDepth);
+
+    @Nullable
+    public IBlockState getBlockState(BlockPos pos) {
+        return null;
+    }
+
+    static class TileEntityAdd extends BlockTransaction {
```
Zidane, line 80: AddTileEntity

liach:
finally!

PR #2164, opened 2018-12-22
ParticleType Prefix fix

adam6806:
Refactored Registry Module name for consistency. Implemented prefix type registry module to so as to preclude the use of minecraft: in front of the name.

### DIFF
--- a/src/main/java/org/spongepowered/common/registry/type/effect/ParticleTypeRegistryModule.java
+++ b/src/main/java/org/spongepowered/common/registry/type/effect/ParticleTypeRegistryModule.java
@@ -47,24 +47,22 @@ import org.spongepowered.api.util.Color;
 import org.spongepowered.api.util.Direction;
 import org.spongepowered.common.effect.particle.SpongeParticleType;
 import org.spongepowered.common.item.inventory.SpongeItemStackSnapshot;
-import org.spongepowered.common.registry.AbstractCatalogRegistryModule;
+import org.spongepowered.common.registry.type.AbstractPrefixAlternateCatalogTypeRegistryModule;
 import org.spongepowered.common.registry.type.BlockTypeRegistryModule;
 import org.spongepowered.common.registry.type.ItemTypeRegistryModule;
 import org.spongepowered.common.registry.type.NotePitchRegistryModule;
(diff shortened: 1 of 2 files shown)


adam6806:
Realized after the fact that this probably qualifies as a change under 20 lines of code but I figured since this is my first contribution that it might make sense to talk to someone about it.

gabizou:
It might make sense to have this ported to `stable-7` as a fix more so than having this done in `bleeding`. Basically what you'd need to do is cherry-pick the commit onto `stable-7`. Once that's done, I can merge. Otherwise, great changes! Can't complain about it.

src/main/java/org/spongepowered/common/registry/type/effect/ParticleTypeRegistryModule.java
```diff
@@ -47,24 +47,22 @@
...
+import org.spongepowered.common.registry.type.AbstractPrefixAlternateCatalogTypeRegistryModule;
 import org.spongepowered.common.registry.type.BlockTypeRegistryModule;
 import org.spongepowered.common.registry.type.ItemTypeRegistryModule;
 import org.spongepowered.common.registry.type.NotePitchRegistryModule;
 import org.spongepowered.common.registry.type.item.FireworkShapeRegistryModule;
 
-import java.util.Collections;
-import java.util.Locale;
-import java.util.Map;
+import java.util.*;
```
phit, line 56: please revert this, no wildcard imports
https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

adam6806: Thanks. IntelliJ sniped me I need to fix my import settings ;)

adam6806:
I just figured out that common and api are setup underneath vanilla and forge. I'm going to re setup my projects for this and create a new PR from a different fork. @gabizou is there some documentation somewhere on a good testing setup for something like this? I was thinking that since this was in regard to calling this from a plugin I could just setup a simple plugin to do a quick test on and debug through to make sure. If there's a unit test area that would be better suited I would be happy to write up a unit test for this.

adam6806:
closed after resetting my development environment.

PR #2165, opened 2018-12-23
Prevent having to use "minecraft:" when getting particle types

adam6806:
#1286 
As stated in the referenced issue the desire was to be able to get the particle type without needing to use "minecraft:" prepended for example "minecraft:heart".

`Optional<ParticleType> heart = Sponge.getRegistry().getType(ParticleType.class, "heart");`

When calling getType and passing only "heart" you would get an empty optional. Looking for other classes that allow you to pass in just the name I saw that they were extending AbstractPrefixAlternateCatalogTypeRegistryModule instead of implementing CatalogRegistryModule. This requires you to set the default string to prepend to the id's. Now in the getById method we're referencing the default string that was set in the constructor and prepending it for the caller when we attempt to find the ParticleType in the map.

`Optional.ofNullable(this.particleByName.get(defaultModIdToPrepend + ":" + checkNotNull(id).toLowerCase(Locale.ENGLISH)));`

I tested this by creating a plugin, calling getType and passing only "heart" and confirmed that you will get the appropriate Optional for the heart particle type.

I have also renamed ParticleRegistryModule to ParticleTypeRegistryModule in keeping with perceived naming conventions.

### DIFF
--- a/src/main/java/org/spongepowered/common/registry/CommonModuleRegistry.java
+++ b/src/main/java/org/spongepowered/common/registry/CommonModuleRegistry.java
@@ -212,7 +212,7 @@ import org.spongepowered.common.registry.type.data.InstrumentTypeRegistryModule;
 import org.spongepowered.common.registry.type.data.KeyRegistryModule;
 import org.spongepowered.common.registry.type.economy.TransactionTypeRegistryModule;
 import org.spongepowered.common.registry.type.effect.ParticleOptionRegistryModule;
-import org.spongepowered.common.registry.type.effect.ParticleRegistryModule;
+import org.spongepowered.common.registry.type.effect.ParticleTypeRegistryModule;
 import org.spongepowered.common.registry.type.effect.PotionEffectTypeRegistryModule;
 import org.spongepowered.common.registry.type.effect.RecordTypeRegistryModule;
 import org.spongepowered.common.registry.type.effect.SoundCategoryRegistryModule;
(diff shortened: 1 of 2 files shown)


adam6806:
Looks like there were unit tests to be updated. I'm guessing the run configuration doesn't execute these tests hence why I was able to test with the plugin without the build failing. @gabizou if you're good with this change let me know and I'll go update all the tests trying to make calls with minecraft:whatever to whatever.

ImMorpheus:
Your analysis is correct. 
However, the registries are not consistent. Some of them support both. 
The correct way is to use the prefixed id.

A quick fix would be to do what `RecordTypeRegistryModule` is doing:

https://github.com/SpongePowered/SpongeCommon/blob/c288d39af4f07bb5da7fbc0f96d7648c4c691170/src/main/java/org/spongepowered/common/registry/type/effect/RecordTypeRegistryModule.java#L67-L70

Requesting input for future reference. Do we want to support the call with an unprefixed id ?

gabizou (approved):

ImMorpheus:
Asked around, prefixed id is the way to go. However if you try to lookup a catalog without a namespace, it should default to `minecraft:`

@adam6806 This PR is breaking `Sponge.getRegistry().getType(ParticleType.class, "minecraft:heart");` because at this line it would try to look for 
`minecraft:minecraft:heart`
https://github.com/SpongePowered/SpongeCommon/blob/45eec25b08cfdebee9d82e141dbd4e0e9eb6a27b/src/main/java/org/spongepowered/common/registry/type/effect/ParticleTypeRegistryModule.java#L80

You should prepend the defaultModId only if there's no namespace.
(`if (!id.contains(":")) `)

Everything else looks good.

PR #2175, opened 2019-01-25
Rate-limiting InteractBlockEvent

ImMorpheus:
[SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/2579) | **SpongeCommon**

Introduce the same rate-limiting policy for InteractBlock events.

`// If the time between packets is small enough, use the last result.`

and use the appropriate flag based on the event.


gabizou:
> Cancelling InteractBlockEvent won't duplicate the fluid but it WON'T stop you from picking up the fluid.

I'm confused, how is that fixing the issue if fluids are not duping but still being filled/picked up?

ImMorpheus:
Okay, I didn't notice there was a test plugin named bucket_cancel_test.

It completely messed up my tests.

This PR is still valid, it's just not fixing that issue.

Issue #2185, opened 2019-02-10
Container minecart lens are constructed without a size check


liach:
**I am currently running**

- SpongeForge version: 1.12.2-2768-7.1.5-RC3542
- SpongeVanilla version: not applicable
- Java version: 1.8.0_191
- Operating System: Windows 10 (amd64) version 10.0

- Plugins/Mods: See the crash report mod list section at https://pastebin.com/5SVgA4Xj

**Issue Description**
When a minecart's lens is constructed, there is no check for the size of the lens. In contrast, there is a check for basic inventories. As a result, when a minecart with an inventory size of zero is created, the game errors and the minecart won't spawn.

Example log: https://pastebin.com/VQLtKWWX

Guess I can make a pr.

Simple analysis: https://github.com/Railcraft/Railcraft/issues/1692#issuecomment-462115902

> Compare:
https://github.com/SpongePowered/SpongeCommon/blob/a9423c7ec9f2b423b52c273f7e7a23818a8f3e41/src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityMinecartContainer.java#L51
>
> versus:
https://github.com/SpongePowered/SpongeCommon/blob/a9423c7ec9f2b423b52c273f7e7a23818a8f3e41/src/main/java/org/spongepowered/common/mixin/core/inventory/MixinInventoryBasic.java#L49-L52

PR #2186, opened 2019-02-10
Make sure lens don't throw errors when a container minecart with empt…

liach:
…y inventory is created

Also fixes a todo for empty lens iterator (to prevent possible crash)

@Faithcaio Please review. Thanks!

Signed-off-by: liach <liach@users.noreply.github.com>

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityMinecartContainer.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityMinecartContainer.java
@@ -32,6 +32,9 @@ import org.spongepowered.api.item.inventory.type.CarriedInventory;
 import org.spongepowered.asm.mixin.Implements;
 import org.spongepowered.asm.mixin.Interface;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 import org.spongepowered.common.item.inventory.adapter.InventoryAdapter;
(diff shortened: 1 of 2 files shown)


liach:
Moreover, this fixes #2185 and Railcraft/Railcraft#1692

Faithcaio:
Looks fine

src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityMinecartContainer.java
```diff
@@ -48,7 +53,13 @@
 
     protected Fabric fabric = new IInventoryFabric(this);
     protected SlotCollection slots = new SlotCollection.Builder().add(this.getSizeInventory()).build();
-    protected Lens lens = new OrderedInventoryLensImpl(0, this.getSizeInventory(), 1, this.slots);
+    protected Lens lens;
+
+    @Inject(method = "<init>", at = @At("RETURN"))
```
ImMorpheus: This will only affect the default constructor.
Is it intended ?

ImMorpheus (changes requested):

src/main/java/org/spongepowered/common/mixin/core/entity/item/MixinEntityMinecartContainer.java
liach: So add a `*`?

ImMorpheus: yep, `<init>*`

liach: Did it on github.

ImMorpheus (approved):

PR #2195, opened 2019-02-22
Add enable/disable system for testplugins

ImMorpheus:
```
/test enable|disable <plugin1> <plugin2>
/test status
```
TestPlugins are disabled by default.

### DIFF
--- a/testplugins/src/main/java/org/spongepowered/test/CopyPasta.java
+++ b/testplugins/src/main/java/org/spongepowered/test/CopyPasta.java
@@ -34,6 +34,7 @@ import org.slf4j.Logger;
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.block.BlockSnapshot;
 import org.spongepowered.api.command.CommandResult;
+import org.spongepowered.api.command.CommandSource;
 import org.spongepowered.api.command.spec.CommandSpec;
 import org.spongepowered.api.config.ConfigDir;
 import org.spongepowered.api.data.DataContainer;
@@ -66,7 +67,7 @@ import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
 
 @Plugin(id = "copypasta", name = "CopyPasta", description = CopyPasta.DESCRIPTION, version = "0.0.0")
-public class CopyPasta {
+public class CopyPasta implements LoadableModule {
 
     public static final String PLUGIN_ID = "copypasta";
     public static final String DESCRIPTION = "A simple plugin that allows creating, saving, and pasting schematics with SpongeAPI";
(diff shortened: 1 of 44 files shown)


testplugins/src/main/java/org/spongepowered/test/CopyPasta.java
```diff
@@ -225,23 +227,37 @@ public void onGamePreInitialization(GamePreInitializationEvent event) {
...
-    public void onInteract(InteractBlockEvent.Secondary.MainHand event, @Root Player player) {
-        Optional<ItemStack> item = player.getItemInHand(HandTypes.MAIN_HAND);
-        if (item.isPresent() && item.get().getType().equals(ItemTypes.WOODEN_AXE) && event.getTargetBlock() != BlockSnapshot.NONE) {
-            get(player).setPos2(event.getTargetBlock().getPosition());
-            player.sendMessage(Text.of(TextColors.LIGHT_PURPLE, "Position 2 set to " + event.getTargetBlock().getPosition()));
-            event.setCancelled(true);
-        }
+    @Override
+    public void disable(CommandSource src) {
+        Sponge.getEventManager().unregisterListeners(this.plugin);
```
gabizou: Could this be streamlined/default implemented so only the plugins that have more than just the listeners will need to override?

gabizou (changes requested):

testplugins/src/main/java/org/spongepowered/test/CopyPasta.java
ImMorpheus: I would need another method to retrieve the class handling the listeners to default implement this. 
Is it worth adding more methods ?

gabizou: I'd say it's worth adding a method returning an object as "listener".

testplugins/src/main/java/org/spongepowered/test/MainTestPlugin.java
```diff
@@ -0,0 +1,94 @@
...
+                                        Sponge.getServer().getBroadcastChannel().send(Text.of(pc.getId(), " enabled"));
+                                        ((LoadableModule) plugin).enable(src);
+                                    } else {
+                                        Sponge.getServer().getBroadcastChannel().send(Text.of(pc.getId(), " is not a LoadableModule"));
+                                    }
+                                });
+                            }
+                            return CommandResult.success();
+                        }))
+                        .build(), "enableplugins");
```
ImMorpheus: `ep` and `dp`: shorter alias, faster to type 
`test enable|disable <plugin1> <plugin2>`: cleaner

Thoughts ?

phit: im fine with dropping the short alias, if you can make `/test` without any arguments a quick toggle to enable/disable

ImMorpheus: I would need to track the status (enabled/disabled) of individual plugins then. That would become uglier than the alias IMO.

Unless you want it to be just a global switch.

gabizou: I don't see why the test plugins suite couldn't have an enabled/disabled container, would be rather interesting to see as a centralized debugging spot to visualize what plugins are enabled in the event of some listeners being active or inactive. Granted, a majority of these plugins are for testing functionality, and/or bug fixes, so it's not common to have this sort of interaction being preformed. My two cents are to have a global container (specific to the test plugins of course) to keep track of that.

gabizou (approved):

gabizou:
Just maybe squash commits prior to merging.

PR #2213, opened 2019-03-10
unmarkRemoval when adding a te and avoid closing a Proxy twice

ImMorpheus:
Issue found while investigating https://github.com/SpongePowered/SpongeForge/issues/2639

This should fix furnace and other TE not being accessible

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/context/MultiBlockCaptureSupplier.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/context/MultiBlockCaptureSupplier.java
@@ -561,6 +561,9 @@ public final class MultiBlockCaptureSupplier implements ICaptureSupplier {
                 try (final SpongeProxyBlockAccess.Proxy transactionProxy = transaction.getProxy(mixinWorldServer)) {
                     transaction.process(eventTransaction, phaseState, phaseContext, currentDepth);
                 }
+                if (this.processingBlocks != null) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/event/tracking/context/MultiBlockCaptureSupplier.java
```diff
@@ -561,6 +561,9 @@ public boolean processTransactions(List<Transaction<BlockSnapshot>> transactions
                 try (final SpongeProxyBlockAccess.Proxy transactionProxy = transaction.getProxy(mixinWorldServer)) {
                     transaction.process(eventTransaction, phaseState, phaseContext, currentDepth);
                 }
+                if (this.processingBlocks != null) {
+                    this.processingBlocks.remove(mixinWorldServer);
```
gabizou, line 565: No.... It needs to retain the proxy to close the proxy itself when it's completed, otherwise, where is the proxy going to be told to close (since it's lazily created when logBlockChange etc. are called).

gabizou: Yeah, no... You really don't want to remove in the for loop. There's a reason why it's being closed and pruned in the finally block below. The proxy created for the `BlockTransaction` is auto-closed by the try with resources above.

Issue #2206, opened 2019-03-03
Server.copyWorld() gives same dimensionIDs if world copied multiple times


GlassSpirit:
**I am currently running**

- SpongeForge version: 1.12.2-2768-7.1.5
- Java version: 1.8
- Operating System: Windows 7

**Issue Description**
`Sponge.getServer().copyWorld()` gives same dimensionIDs for copied worlds if that world is copied multiple times.
For example if I copy default world with `dimensionId = 0`, first copied world would have `dimensionId = 2` because of nether and end. But if I copy default world more times, all copied worlds would have `dimensionId = 2`.
I think copyWorld doesn't properly update latest dimensionId value.

As I see from SpongeCommon code, dimensionId's are stored in `org.spongepowered.common.world.WorldManager.class` in `worldFolderByDimensionId` map and `dimensionBits` BitSet.
When worlds are loaded by game, `dimensionBits `and `dimensionTypeByDimensionId` are somehow populated, but when new worlds are created by `WorldManager.createWorldProperties()` or `WorldManager.copyWorld()`, dimensionId is get by `getNextFreeDimensionId()`, which gets first free id from dimensionBits, but doesn't change it anyhow (!).
So all next worlds created by that methods would have same dimensionId, which can break a lot of things...

Zidane:
Closed in PR that was merged.

PR #2215, opened 2019-03-12
Fix dimension ids duplication

GlassSpirit:
Details of main problem defined in issue #2206 and pr #2207, which was not fixed due to mods using negative dimensionIDs.
Since BitSet cannot work with negative dimensionIDs, it is neccessary to use IntSet.

All behavioral changes are taken from [Forge 1.12 DimensionManager.class](https://github.com/MinecraftForge/MinecraftForge/blob/1.12.x/src/main/java/net/minecraftforge/common/DimensionManager.java)
I don't know why forge marks "DimensionArray" compound as legacy data and uses "UsedIDs" compound.

### DIFF
--- a/src/main/java/org/spongepowered/common/world/WorldManager.java
+++ b/src/main/java/org/spongepowered/common/world/WorldManager.java
@@ -188,12 +188,7 @@ public final class WorldManager {
         return Optional.empty();
     }
 
-    /**
-     * Return the next free dimension ID. Note: you are not guaranteed a contiguous
-     * block of free ids. Always call for each individual ID you wish to get.
-     * @return the next free dimension ID
-     */
-    public static int getNextFreeDimensionId() {
+    public static Integer getNextFreeDimensionId() {
         int next = lastUsedDimensionId;
         while (usedDimensionIds.contains(next) || !checkAvailable(next)) {
             next++;
@@ -1228,12 +1223,12 @@ public final class WorldManager {
         if (compound == null) {
             dimensionTypeByDimensionId.keySet().stream().filter(dimensionId -> dimensionId >= 0).forEach(usedDimensionIds::add);
         } else {
-            for (int id : compound.getIntArray("UsedIDs")) {
+            for (int id : compound.getIntArray(NbtDataUtil.USED_DIMENSION_IDS)) {
                 usedDimensionIds.add(id);
             }
 
             // legacy data (load but don't save)
-            int[] intArray = compound.getIntArray("DimensionArray");
+            int[] intArray = compound.getIntArray(NbtDataUtil.LEGACY_DIMENSION_ARRAY);
             for (int i = 0; i < intArray.length; i++) {
                 int data = intArray[i];
                 if (data == 0) continue;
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/world/WorldManager.java
```diff
@@ -1205,30 +1222,31 @@ public static void sendDimensionRegistration(EntityPlayerMP playerMP, WorldProvi
...
-        dimensionBits.clear();
+        usedDimensionIds.clear();
+        lastUsedDimensionId = 0;
+
         if (compound == null) {
-            dimensionTypeByDimensionId.keySet().stream().filter(dimensionId -> dimensionId >= 0).forEach(dimensionBits::set);
+            dimensionTypeByDimensionId.keySet().stream().filter(dimensionId -> dimensionId >= 0).forEach(usedDimensionIds::add);
         } else {
-            final int[] intArray = compound.getIntArray("DimensionArray");
+            for (int id : compound.getIntArray("UsedIDs")) {
```
gabizou: If you can store these constants under `NbtDataUtil.Forge`, that'd be perfect (it's where we store a bunch of string constants for NbtData, so we have clearer migrations if we can provide ways to upgrade data from forge to sponge vanilla and vice versa).

src/main/java/org/spongepowered/common/world/WorldManager.java
```diff
@@ -186,8 +188,25 @@ public static void registerDimensionType(int dimensionTypeId, DimensionType type
         return Optional.empty();
     }
 
-    public static Integer getNextFreeDimensionId() {
-        return dimensionBits.nextClearBit(0);
+    /**
```
Zidane: Remove these docs. It is not intended for plugins to call this code.

Zidane:
Thanks a bunch @GlassSpirit for this PR. Appreciate it.

src/main/java/org/spongepowered/common/world/WorldManager.java
gabizou: Maybe not,  but it's good to document it for implementation (just like PhaseTracker is slowly being documented).

Zidane: These docs are focused on external code calling this method....

ImMorpheus:
@GlassSpirit Thank you for your contribution

This is missing a SpongeForge PR. 

We're shadowing the BitSet [here](https://github.com/SpongePowered/SpongeForge/blob/67c8b8fb928807b22684e0f6f2549cf14e5d62b1/src/main/java/org/spongepowered/mod/mixin/core/common/world/MixinWorldManager.java#L73).

You need to shadow the IntSet and update the method using that field.

GlassSpirit:
Does that mean that I need to make another PR to SpongeForge?
Or I must close this and open a PR directly in SpongeForge with all changes?
I think I don't know how to make a pr to SpongeForge without my changes in SpongeCommon...

ImMorpheus:
> Does that mean that I need to make another PR to SpongeForge?

yes

> I think I don't know how to make a pr to SpongeForge without my changes in SpongeCommon...

SpongeCommon is a submodule of SpongeForge.

1. Clone SpongeForge
2. Checkout `fix/dimension-ids-duplication` (your branch) on SpongeCommon
3. Create a new branch on SpongeForge and fix the issue with old shadowed field

PR #2220, opened 2019-03-16
Misc fixes for EntitySnapshot

ImMorpheus:
Fix `EntitySnapshot.builder().build().toContainer();` NPE
Fix `EntitySnapshot.builder().build(EntitySnapshot.builder().build().toContainer());` `Optional#empty()`
Fix tests
Fix EntitySnapshot hash-based collections (missing hashcode/equals)


gabizou:
EntitySnapshots are not supposed to be treated like EntityArchetypes in that they need AT LEAST valid information from an Entity to be created. Much like BlockSnapshots in a sense for TileEntity information. 

The issue with changing this is that you now make it acceptable to abuse snapshots to restore or create them as necessary, without really differing them from EntityArchetypes.

ImMorpheus:
> EntitySnapshots are not supposed to be treated like EntityArchetypes in that they need AT LEAST valid information from an Entity to be created

That's not the case at all. There's nothing stopping you from doing `EntitySnapshot.builder().build()`
This means the current system is broken. Working on it.

PR #2223, opened 2019-03-17
Refactor registry module registration

### LINKED ISSUE #1668: IllegalAccessException with @RegisterCatalog
An IllegalAccessException is thrown by the game registry when a **plugin provided** registry module is registered with the RegisterCatalog annotation. 

The ```final``` declaration of the "pseudo enum" fields is ignored by the [RegistryHelper](https://github.com/SpongePowered/SpongeCommon/blob/5d8ac978765dde886c022dda23dbfedd96f24107/src/main/java/org/spongepowered/common/registry/RegistryHelper.java#L75).

StackTrace:
```
[Sponge]: Error while mapping com.neolumia.quartz.arena.ArenaStates.DEAD
java.lang.IllegalAccessException: Can not set static final com.neolumia.quartz.arena.ArenaState field com.neolumia.quartz.arena.ArenaStates.DEAD to com.neolumia.quartz.core.arena.QuartzArenaState
        at sun.reflect.UnsafeFieldAccessorImpl.throwFinalFieldIllegalAccessException(Unknown Source) ~[?:1.8.0_131]
        at sun.reflect.UnsafeFieldAccessorImpl.throwFinalFieldIllegalAccessException(Unknown Source) ~[?:1.8.0_131]
        at sun.reflect.UnsafeQualifiedStaticObjectFieldAccessorImpl.set(Unknown Source) ~[?:1.8.0_131]
        at java.lang.reflect.Field.set(Unknown Source) ~[?:1.8.0_131]
        at org.spongepowered.common.registry.RegistryHelper.mapFields(RegistryHelper.java:75) [RegistryHelper.class:1.12.2-7.0.0-BETA-384]
        at org.spongepowered.common.registry.RegistryHelper.mapFields(RegistryHelper.java:43) [RegistryHelper.class:1.12.2-7.0.0-BETA-384]
        at org.spongepowered.common.registry.util.RegistryModuleLoader.tryModulePhaseRegistration(RegistryModuleLoader.java:69) [RegistryModuleLoader.class:1.12.2-7.0.0-BETA-384]
        at org.spongepowered.common.registry.SpongeGameRegistry.registerModulePhase(SpongeGameRegistry.java:529) [SpongeGameRegistry.class:1.12.2-7.0.0-BETA-384]
        at org.spongepowered.common.registry.SpongeGameRegistry.preInit(SpongeGameRegistry.java:553) [SpongeGameRegistry.class:1.12.2-7.0.0-BETA-384]
        at org.spongepowered.server.SpongeVanilla.preInitialize(SpongeVanilla.java:122) [SpongeVanilla.class:1.12.2-7.0.0-BETA-384]
        at net.minecraft.server.dedicated.DedicatedServer.handler$onServerLoad$zol000(SourceFile:1234) [nz.class:?]
        at net.minecraft.server.dedicated.DedicatedServer.func_71197_b(SourceFile:117) [nz.class:?]
        at net.minecraft.server.MinecraftServer.run(SourceFile:434) [MinecraftServer.class:?]
        at java.lang.Thread.run(Unknown Source) [?:1.8.0_131]
```

Code:
https://github.com/neolumia/quartz/tree/4f4e6e65fad5a51623889dc8107aa99291144b75

### LINKED ISSUE #2019: registerDefaults not called or called twice
**I am currently running**
- SpongeVanilla version: spongevanilla-1.12.2-7.1.0-BETA-93.jar

I'm writing some plugin and this plugin uses registry system.

Registry src: https://gist.github.com/XakepSDK/71bcbce4428e509639b7782976c2c49c
How i register registry: https://gist.github.com/XakepSDK/8e83d15ed0e01bcbe3bcb829ebdcae47

`registerDefaults()` isn't called, and if i annotate it with `AdditionalRegistration`, method called twice.

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeCommon/issues/2019
Fix https://github.com/SpongePowered/SpongeCommon/issues/1668
Add proper support for custom registry modules

### DIFF
--- a/src/main/java/org/spongepowered/common/registry/SpongeGameRegistry.java
+++ b/src/main/java/org/spongepowered/common/registry/SpongeGameRegistry.java
@@ -26,6 +26,7 @@ package org.spongepowered.common.registry;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.reflect.TypeToken;
@@ -134,7 +135,7 @@ import javax.annotation.Nullable;
 @Singleton
 public class SpongeGameRegistry implements GameRegistry {
 
-    public static final boolean PRINT_CATALOG_TYPES = Boolean.parseBoolean(System.getProperty("sponge.print_all_catalog_types"));
+    private static final boolean PRINT_CATALOG_TYPES = Boolean.parseBoolean(System.getProperty("sponge.print_all_catalog_types"));
 
     static {
         TypeSerializers.getDefaultSerializers().registerType(TypeToken.of(CatalogType.class), new CatalogTypeTypeSerializer());
@@ -149,9 +150,10 @@ public class SpongeGameRegistry implements GameRegistry {
     private RegistrationPhase phase = RegistrationPhase.PRE_REGISTRY; // Needed for module phase registrations
 
     protected final Map<Class<? extends CatalogType>, CatalogRegistryModule<?>> catalogRegistryMap = new IdentityHashMap<>();
-    private List<Class<? extends RegistryModule>> orderedModules = new ArrayList<>();
-    final Map<Class<? extends RegistryModule>, RegistryModule> classMap = new IdentityHashMap<>();
     private final Map<Class<?>, Supplier<?>> builderSupplierMap = new IdentityHashMap<>();
+
+    private final List<Class<? extends RegistryModule>> orderedModules = new ArrayList<>();
+    private final Map<Class<? extends RegistryModule>, RegistryModule> classMap = new IdentityHashMap<>();
     private final Set<RegistryModule> registryModules = new HashSet<>();
 
     @Inject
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/registry/SpongeGameRegistry.java
```diff
@@ -508,13 +513,11 @@ public VillagerRegistry getVillagerRegistry() {
         return SpongeVillagerRegistry.getInstance();
     }
 
-    @SuppressWarnings("deprecation")
     @Override
     public TextSerializerFactory getTextSerializerFactory() {
         return SpongeTextSerializerFactory.INSTANCE;
     }
 
-    @SuppressWarnings("deprecation")
```
gabizou, line 517: Why remove the warnings from the drepecation bits?

ImMorpheus: They are not used

dualspiral:
To note, I've been made aware by Morph that this will break Nucleus as I call `registerDefaults()` myself - I probably did it as a workaround. I can update this and release once I get time (though it will force all Nucleus users to update too if they update Sponge) - but my concern is that there might be other plugins like my own that will break and users (and developers, no doubt) will start complaining because they've worked around this not working previously.

If we want to merge this, we will need to potentially do this with care and ensuring that the developer community is alerted to this, in case they've worked around it.

ImMorpheus:
As I've said a month ago, I don't think we should worry about plugins implementing non-working features in sponge. If they choose to employ this kind of workaround it's on them.

However since this issue has not received any input, I did a bit of research.

There are around 40 plugins using custom registry modules. 
https://github.com/Atherys-Horizons/AtherysSkills and https://github.com/Sponge-RPG-dev/NT-RPG were breaking (fixed with the last commit).
The most popular workaround is not using `DummyObjectProvider` and delaying the registration to the INIT phase with `@DelayedRegistration`.

I've only found 2 plugins which would break with this PR:

https://github.com/Limeth/LimeFun
https://github.com/windy1/Reveries

Both are targeting unsupported sponge version (1.10.2/1.11.2) and (like Nucleus) are calling registerDefaults() themselves.

Reposting my message if someone missed the conversation on the matter a month ago:

> I'm not changing behaviour in the strict sense. Custom RegistryModules were not implemented in the first place. As a matter of fact, registering a plugin provided RegistryModule wouldn't call registerDefaults() and the fields wouldn't be remapped (see the testplugin). That's why I asked, my guess was that manually calling that method was a workaround to force the registration of the registry module. Also, keep in mind nucleus is not breaking because of binary incompatibility, it's because it's throwing exceptions when trying to register duplicate catalog, as opposed to Sponge which is failing silently https://github.com/SpongePowered/SpongeCommon/blob/7af4bde2a110f7dcdcb5917a7a3275ede7baad3b/src/main/java/org/spongepowered/common/registry/type/item/PotionTypeRegistryModule.java#L96. Different approach, sure, but my primary point is that using an unimplemented feature is not supported so it shouldn't be considered a breaking change. (Note1: test plugin I'm referring to https://github.com/SpongePowered/SpongeCommon/pull/2167) (To explain further: Nucleus is using reflection to map the catalog fields https://github.com/NucleusPowered/Nucleus/blob/bacf052297893440665af76d2c466169d9c6462f/src/main/java/io/github/nucleuspowered/nucleus/internal/registry/NucleusRegistryModule.java#L42, it has basically reimplemented the registry system by forcing the loading of registry modules and manually remapping the fields)

Issue #2228, opened 2019-03-29
Neighbor block updates broken


LXGaming:
**I am currently running**
- SpongeForge version: 1.12.2-2768-7.1.6-RC3634
- SpongeVanilla version: 1.12.2-7.1.6-RC157
- Java version: 1.8
- Operating System: Linux

- Plugins/Mods:
LuckPerms-Sponge-4.3.109

**Issue Description**
Neighboring blocks don't receive updates

Place something like redstone, torches or grow crops then break the block it's connected to.

Possibly related to https://github.com/SpongePowered/SpongeForge/issues/2662

![2019-03-30_12 41 14](https://user-images.githubusercontent.com/10443061/55267899-1d5d7600-52ea-11e9-8745-3044afbd5435.png)

LXGaming:
In a production environment IMixinBlock::hasNeighborChangedLogic returns false for blocks that do in fact have logic

https://github.com/SpongePowered/SpongeCommon/blob/85fafa0853b58e55d9ca214e1a17a671cd8b92ec/src/main/java/org/spongepowered/common/event/tracking/PhaseTracker.java#L662

This issue is not present when running in an IDE / Development environment

gabizou:
Retry with RC3636.

LXGaming:
Tested 3636 with LuckPerms, Nucleus and with and without GriefPrevention and it worked fine

However when testing with no plugins installed blocks such as sand and gravel simply disappear instead of falling and fluids don't drain.

gabizou:
> However when testing with no plugins installed blocks such as sand and gravel simply disappear instead of falling and fluids don't drain.

Yep, working on fixing the should fire flags. Issue with development was that shouldFire was always true, so this went unnoticed...

PR #2229, opened 2019-03-31
Fix Neighbor block updates in production environment

LXGaming:
Fixes https://github.com/SpongePowered/SpongeCommon/issues/2228


LXGaming:
Upon further testing this also fixes https://github.com/SpongePowered/SpongeForge/issues/2662

ImMorpheus:
Thank you for you contribution.

Good catch !
However, I'm still able to reproduce those issues with your PR.

PR #2231, opened 2019-04-03
Implement Schematic Format v2

gabizou:
[Specification](https://github.com/SpongePowered/Schematic-Specification/pull/13) | [SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1987) | **SpongeCommon**

Implements the new Schematic format version 2 as proposed in the linked PR above. 

This implements the new API with some artificial wrapping around the new generic `Palette` to still implement `BlockPalette` and `BlockPaletteType` as required by previous API contracts. This allows for non-binary breakage of the `Schematic` object and has proven to work without making any modifications to the CopyPasta test plugin.

Checklist:
- [x] Implement new Schematic API
- [x] Add EntityArchetype support 
- [x] Add Biome support
- [x] Add DataVersion for the intended Minecraft version
- [x] Add DataContentUpdater for the older `Schematic` format to the new format.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/persistence/SchematicTranslator.java
+++ b/src/main/java/org/spongepowered/common/data/persistence/SchematicTranslator.java
@@ -28,8 +28,10 @@ import com.flowpowered.math.vector.Vector3i;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.reflect.TypeToken;
-import net.minecraft.tileentity.TileEntity;
-import net.minecraft.util.ResourceLocation;
+import net.minecraft.nbt.NBTTagCompound;
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.util.datafix.DataFixer;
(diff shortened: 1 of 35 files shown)


src/main/java/org/spongepowered/common/data/persistence/SchematicTranslator.java
```diff
@@ -173,27 +214,87 @@ public Schematic translate(DataView view) throws InvalidDataException {
...
+                        break;
+                    }
+                    biomeJ++;
+                }
+                int y = biomeIndex / (width * length);
+                int z = (biomeIndex % (width * length)) / width;
+                int x = (biomeIndex % (width * length)) % width;
+                BiomeType type = biomePalette.get(bVal).get();
+                biomeBuffer.setBiome(x - offset[0], 0, z - offset[2], type);
+            }
```
wizjany, line 268: FYI this loop never increments `biomeIndex` so x/z(/y) will always be 0.

PR #2239, opened 2019-04-19
Fix MoveEntityEvent.Teleport isn't fired when a player uses the spectator-menu to teleport

### LINKED ISSUE #2099: MoveEntityEvent.teleport does not apply to spectator mode teleporting
This issue was originally submitted by @TheCrunchy in [SpongePowered/SpongeAPI#1908](https://github.com/SpongePowered/SpongeAPI/issues/1908).
___
While in spectator mode you can middle click and use the icons to teleport to other people, the MoveEntityEvent.teleport event doesnt listen for this.

karayuumac:
fix #2099


src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -421,6 +425,73 @@ public void processCreativeInventoryAction(CPacketCreativeInventoryAction packet
...
+                        break;
+                    }
+                }
+            }
+
+            if (entity != null) {
+                this.player.setSpectatingEntity(this.player);
+                this.player.dismountRidingEntity();
+                // Sponge start - handle MoveEntityEvent.Teleport
+                double x = entity.posX;
```
gabizou: If all that is being done is based on the player, and entity, you should be able to reduce this to a single cancellable injection:
```java
    @Inject(
        method = "handleSpectate",
        at = @At(
            value = "FIELD",
            target = "Lnet/minecraft/entity/Entity;world:Lnet/minecraft/world/World;",
            ordinal = 0
        ),
        locals = LocalCapture.CAPTURE_FAILHARD,
        cancellable = true
    )
    private void onSpectateTeleportForSponge(CPacketSpectate packetIn, CallbackInfo ci, Entity spectatingEntity) {
        double x = spectatingEntity.posX;
        double y = spectatingEntity.posY;
        double z = spectatingEntity.posZ;
        float yaw = spectatingEntity.rotationYaw;
        float pitch = spectatingEntity.rotationPitch;
        MoveEntityEvent.Teleport event = EntityUtil.handleDisplaceEntityTeleportEvent(this.player, x, y, z, yaw, pitch);
        if (event.isCancelled()) {
            ci.cancel();
        }

    }
```
Also, make sure your javadoc includes `@author` and the date, and the version of Minecraft targeted.

I've tested this injection in development and this is the result:
Screen Shot 2019-04-19 at 1 10 59 PM

gabizou (changes requested):

karayuumac:
Change to cancellable injection.
Also, change to use Transform because players can teleport to a different dimension.

src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java
```diff
@@ -421,6 +426,34 @@ public void processCreativeInventoryAction(CPacketCreativeInventoryAction packet
...
+                    ordinal = 0
+            ),
+            locals = LocalCapture.CAPTURE_FAILHARD,
+            cancellable = true
+    )
+    private void onHandleSpectateForSponge(CPacketSpectate packet, CallbackInfo ci, Entity entity) {
+        Transform<World> fromTransform = ((IMixinEntityPlayerMP) this.player).getTransform();
+        Transform<World> toTransform = ((IMixinEntity) entity).getTransform();
+
+        MoveEntityEvent.Teleport event = EntityUtil.handleDisplaceEntityTeleportEvent(this.player, fromTransform, toTransform);
```
JBYoshi: What happens if a plug-in changes the value of `event.getToTransform()`?

karayuumac: Oh sorry.
Currently, even though the transform is changed by a plug-in, the player's transform doesn't change.
The change doesn't apply to NetHandlerPlayServer.
I'll fix this.

karayuumac:
- fix: When the toTransform is changed by the plug-in, the player's transform is changed correctly.

I also have a question.
Plug-in can also set whether the player's velocity is keeping after teleportation by using 
MoveEntityEvent.Teleport.setKeepsVelocity(true).
I have no idea how to implement this.

ImMorpheus:
@karayuumac Thanks for your contribution.

You seem to have reverted your changes with this commit https://github.com/SpongePowered/SpongeCommon/pull/2239/commits/ce629ae77128c95f445b780a0889b1f5547a6aa0.
Please, change back to cancellable injection.

> Plug-in can also set whether the player's velocity is keeping after teleportation by using
> MoveEntityEvent.Teleport.setKeepsVelocity(true).
> I have no idea how to implement this.

As far as I can see you don't need to. It's already handled in `EntityUtil#handleDisplaceEntityPortalEvent`
https://github.com/SpongePowered/SpongeCommon/blob/4cd5f78d60ed71d034327e674f9527e03498ee49/src/main/java/org/spongepowered/common/entity/EntityUtil.java#L492-L496

karayuumac:
> As far as I can see you don't need to. It's already handled in `EntityUtil#handleDisplaceEntityPortalEvent`

Thank you for asking my question. I understood.

> You seem to have reverted your changes with this commit [ce629ae](https://github.com/SpongePowered/SpongeCommon/commit/ce629ae77128c95f445b780a0889b1f5547a6aa0).
> Please, change back to cancellable injection.

OK. I'll try to change to cancellable injection.

karayuumac:
I checked the code.
And, I think cancellable injection can't be used in this case.
The MoveEntityEvent.Teleport event in here
https://github.com/SpongePowered/SpongeCommon/blob/ce629ae77128c95f445b780a0889b1f5547a6aa0/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java#L459
produces ```Transform<World>``` and this is used in here
https://github.com/SpongePowered/SpongeCommon/blob/ce629ae77128c95f445b780a0889b1f5547a6aa0/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java#L473
and here.
https://github.com/SpongePowered/SpongeCommon/blob/ce629ae77128c95f445b780a0889b1f5547a6aa0/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java#L485-L486

karayuumac:
I checked my code and I think cancellable injection can't be used in this case.
MoveEntityEvent.Transport in here
https://github.com/SpongePowered/SpongeCommon/blob/a80813653c5d4de396d9db794d59559c527fd4a4/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java#L460-L461

is also used in here
https://github.com/SpongePowered/SpongeCommon/blob/a80813653c5d4de396d9db794d59559c527fd4a4/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java#L474-L476

and here.
https://github.com/SpongePowered/SpongeCommon/blob/a80813653c5d4de396d9db794d59559c527fd4a4/src/main/java/org/spongepowered/common/mixin/core/network/MixinNetHandlerPlayServer.java#L486-L489

Is it possible to implement this by injection?

parlough:
Thanks for the contribution! It is possible to implement it all without overwriting the method, it just gets a little messy.

Superseded by https://github.com/SpongePowered/SpongeCommon/commit/620cc8a9e6e12610377a53d82de6aa83a33e84b8

Issue #512, opened 2016-02-15
User lacking implementation of Data API integration


gabizou:
Currently, `SpongeUser` doesn't properly implement many of the Data API methods and thus throw exceptions and errors when trying to retrieve simple data from them. It'd be useful if `User` could construct the `DataManipulator`s appropriately without having to delegate to using the `DataProcessor` system. @simon816 would probably be interested in this.

gabizou:
Fixed with the merging of #2250.

PR #2250, opened 2019-04-26
Improve User Data API integration

### LINKED ISSUE #1880: Unable to properly set vanish on login
**I am currently running**

- SpongeForge version: 
- SpongeVanilla version: 1.12 from github (SpongeCommon to [this commit](https://github.com/SpongePowered/SpongeCommon/commit/36a82d2c6bfc2a916d0b9772b750ee7b9d87481c)) 

**Issue Description**
This was discovered while trying to fix a few vanish-related issues with Nucleus without using [hacky solution](https://github.com/NucleusPowered/Nucleus/commit/ae939f81f4dcfd2964cd8c7902e102cb1c466654).
This is the current situation when you try to login on a server:

- A ClientConnectionEvent.Login is fired here: https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java#L271 (You can't offer keys to the player during this event, as there's no "player") 
- The NetHandler is assigned to the player and the required packets are sent.
- Then [playerLoggedIn](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java#L343) is called, but it is injected [here](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java#L784) before it sends the `SPacketPlayerListItem` packet to all players to modify the tablist (so that we can support `Keys.VANISH` https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java#L789) 
- The [canSee](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/entity/Entity.java#L545) method should check `if this entity can see the provided entity`, however this check is always true as the only point where you can offer `Keys.VANISH` is during ClientConnectionEvent.Join which is fired [way after](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/server/MixinPlayerList.java#L430) the tab check.

Pinging @dualspiral, not sure if you found more to this while working on Nucleus

TLDR: a player is added to the tablist before we can offer any keys, so there's no way to properly vanish on login.

ImMorpheus:
Related https://github.com/SpongePowered/SpongeCommon/issues/512
Fix https://github.com/SpongePowered/SpongeCommon/issues/1880

VanishValueProcessor and InvulnerableValueProcessor are used as an example for future PRs.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/processor/value/entity/VanishValueProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/value/entity/VanishValueProcessor.java
@@ -24,16 +24,17 @@
  */
 package org.spongepowered.common.data.processor.value.entity;
 
-import net.minecraft.entity.Entity;
 import org.spongepowered.api.data.DataTransactionResult;
 import org.spongepowered.api.data.key.Keys;
 import org.spongepowered.api.data.value.ValueContainer;
 import org.spongepowered.api.data.value.immutable.ImmutableValue;
 import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.api.entity.Entity;
 import org.spongepowered.common.data.processor.common.AbstractSpongeValueProcessor;
 import org.spongepowered.common.entity.EntityUtil;
 import org.spongepowered.common.data.value.immutable.ImmutableSpongeValue;
 import org.spongepowered.common.data.value.mutable.SpongeValue;
+import org.spongepowered.common.entity.player.IUserAdapter;
 import org.spongepowered.common.interfaces.entity.IMixinEntity;
 
 import java.util.Optional;
@@ -51,16 +52,22 @@ public class VanishValueProcessor extends AbstractSpongeValueProcessor<Entity, B
 
     @Override
     protected boolean set(Entity container, Boolean value) {
-        if (!container.world.isRemote) {
-            EntityUtil.toMixin(container).setVanished(value);
-            return true;
+        if (container instanceof net.minecraft.entity.Entity) {
+            if (!((net.minecraft.entity.Entity) container).world.isRemote) {
+                EntityUtil.toMixin(container).setVanished(value);
+                return true;
+            }
         }
-        return false;
+        ((IUserAdapter) container).setVanished(value);
+        return true;
     }
 
     @Override
     protected Optional<Boolean> getVal(Entity container) {
-        return Optional.of(((IMixinEntity) container).isVanished());
+        if (container instanceof net.minecraft.entity.Entity) {
+            return Optional.of(((IMixinEntity) container).isVanished());
+        }
+        return Optional.of(((IUserAdapter) container).isVanished());
     }
 
     @Override
--- a/src/main/java/org/spongepowered/common/data/processor/value/entity/VanishValueProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/value/entity/VanishValueProcessor.java
@@ -29,20 +29,17 @@ import org.spongepowered.api.data.key.Keys;
 import org.spongepowered.api.data.value.ValueContainer;
 import org.spongepowered.api.data.value.immutable.ImmutableValue;
 import org.spongepowered.api.data.value.mutable.Value;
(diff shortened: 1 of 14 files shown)


src/main/java/org/spongepowered/common/data/util/NbtDataUtil.java
```diff
@@ -167,6 +167,7 @@ private NbtDataUtil() {
     public static final String PORTAL_AGENT_TYPE = "portalAgentType";
     public static final String WORLD_SERIALIZATION_BEHAVIOR = "serializationBehavior";
     public static final String HEALTH_SCALE = "HealthScale";
+    public static final String VANISH = "vanish";
```
ImMorpheus: "vanish" vs "isVanished" ?

src/main/java/org/spongepowered/common/entity/player/IUserAdapter.java
```diff
@@ -0,0 +1,14 @@
+package org.spongepowered.common.entity.player;
```
ImMorpheus: license

src/main/java/org/spongepowered/common/data/processor/value/entity/VanishValueProcessor.java
```diff
@@ -51,16 +52,22 @@ public VanishValueProcessor() {
...
     @Override
     protected boolean set(Entity container, Boolean value) {
-        if (!container.world.isRemote) {
-            EntityUtil.toMixin(container).setVanished(value);
-            return true;
+        if (container instanceof net.minecraft.entity.Entity) {
+            if (!((net.minecraft.entity.Entity) container).world.isRemote) {
+                EntityUtil.toMixin(container).setVanished(value);
+                return true;
+            }
```
XakepSDK: There is no `return false`, is this working as intended now?

ImMorpheus: No, it's not. Good catch !

src/main/java/org/spongepowered/common/data/util/NbtDataUtil.java
gabizou: Stick with proper CamelCasing.

src/main/java/org/spongepowered/common/entity/player/IUserAdapter.java
```diff
@@ -0,0 +1,14 @@
...
+
+import org.spongepowered.api.entity.living.Living;
+
+public interface IUserAdapter extends Living {
```
gabizou: Why is it a living as well?

src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
```diff
@@ -193,9 +371,12 @@ public void writeToNbt(NBTTagCompound compound) {
         compound.setInteger(NbtDataUtil.ENTITY_DIMENSION, this.dimension);
         compound.setTag(NbtDataUtil.ENTITY_ROTATION, NbtDataUtil.newFloatNBTList(this.rotationYaw, this.rotationPitch));
 
+        compound.setBoolean("Invulnerable", this.invulnerable);
```
gabizou: `NbtDataUtil` constants here.

Cybermaxke:
Rename `IUserAdapter` to something like `IUserOrEntity` and make `IMixinEntity` extend it. That should reduce some instance checks (and for in the future if you want to support more user data)

ImMorpheus:
Note: `Fix home testplugin` is not necessary, I've included that commit since that's the only testplugin with custom data and it wasn't working. There is no other way to test this commit (another testplugin would be pointless). Once this is approved I can rebase and drop that commit.

PR #2271, opened 2019-05-25
Remove useless semicolon

RDIL:


dualspiral:
Hi @RDIL,

Thanks for spotting this. While we appreciate contributions, [we ask that minor changes that make no functional difference are reported to the minor issues list for that repo](https://docs.spongepowered.org/stable/en/contributing/guidelines.html#general-steps), in this case, that is issue #23.  This ensure that our history isn't cluttered and that users can see what we have changed in the product more easily.

I also note that you targetted the `bleeding` branch. That branch is no longer active - our next API version will be on the `api-8` branch instead and so this commit would actually have been lost when we are able to complete the implementation of Sponge for 1.14.

If this exists in `stable-7`, please add this as a comment to #23. If you are looking to contribute to the project, [please have a read of our docs](https://docs.spongepowered.org/stable/en/contributing/index.html) and/or come and have a chat with us on [Discord](https://discord.gg/sponge).

Thanks!

PR #2274, opened 2019-06-01
Implement DisabledSlotsData for Armor Stands

tcmal:
[API PR](https://github.com/SpongePowered/SpongeAPI/pull/2010/)

Worth noting that the way this flag is implemented may change, as it currently doesn't apply to the stand's hands and has redundant parts ([MC-94252](https://bugs.mojang.com/browse/MC-94252))


gabizou:
I'm guessing this works now? With the problem of updating the armor stand that was being brought up over discord, I'm guessing that was resolved once the value processors were implemented?

I very much appreciate the implementation and API. Once I'm done cleaning up #2278, i'll get to merging this in.

gabizou:
So, I'll test this out and if it works, I'll be fine with merging the API and implementation. Any oppositions @Zidane ?

Issue #2094, opened 2018-09-03
ItemStack#builder#fromContainer not correctly deserializing HOCON containers


Zerthick:
**I am currently running**
- SpongeVanilla version: 1.12.2-7.1.0-BETA-100

**Issue Description**
As the title says, when attempting to recreate an itemstack with custom data manipulators attached from a HOCON string, those custom manipulators are lost. 

Serialization/Deserialization Code:
```java
public static String serializeSnapShot(ItemStackSnapshot snapshot) throws IOException {

    DataContainer container = snapshot.toContainer();
    StringWriter stringWriter = new StringWriter();

    DataFormats.HOCON.writeTo(stringWriter, container);

    return stringWriter.toString();
}

public static ItemStackSnapshot deserializeSnapShot(String serializedSnapshot) throws IOException {

    DataContainer container = DataFormats.HOCON.read(serializedSnapshot);

    return ItemStack.builder().fromContainer(container).build().createSnapshot();
}
```

The string I'm attempting to deserialize (which was created by the above code) is as follows:
```
ContentVersion=2
Count=1
Data=[
    {
        ContentVersion=2
        ManipulatorData {
            CommandBookCommands=[
                "$time set day"
            ]
            CommandBookUses=-1
            ContentVersion=1
        }
        ManipulatorId="commandbooks:command_book"
    }
]
ItemType="minecraft:enchanted_book"
UnsafeDamage=0
UnsafeData {
    display {
        Lore=[
            "Uses: ∞"
        ]
        Name="§b§o[Day]"
    }
}
```

Shiunin:
If it helps at all I am also experiencing this issue and can provide steps to reproduce if needed. I'm on this version:
![image](https://user-images.githubusercontent.com/44011334/55523472-4d847c00-563e-11e9-94a1-2b0cb5d8af7b.png)

randombyte-developer:
Still reproducible on SV `1.12.2-7.1.5`.

Some debugging from my side got me to this:

[This](https://github.com/SpongePowered/SpongeCommon/blob/c6724ece4e32d228c828a8190b6fae42b1adede9/src/main/java/org/spongepowered/common/data/builder/item/SpongeItemStackSnapshotBuilder.java#L69) shouldn't call `DataView#getViewList(...)`. The data view was built/translated from a config node by Sponge. The Data node is in the container. But the `getViewList()` doesn't return anything, because the Data node in there isn't a DataView.

[...]

PR #2275, opened 2019-06-01
Fix/configtranslator

### LINKED ISSUE #1598: ParticleOptions of a ParticleEffect not being deserialized from a config file
**I am currently running...**

- SpongeForge version: not tested
- SpongeVanilla version: spongevanilla-1.11.2-6.1.0-BETA-0
- Java version: 1.8.0_121
- Operating System:

 - [ ] Linux (Ubuntu, Debian, Fedora, etc)
 - [ ] macOS
 - [x] Windows

- Plugins/Mods: None

ParticleEffect deserialized from a config file contaning both the ParticleType and the various ParticleOptions contained in the config.

The ParticleEffect when deserialized loads only the ParticleType from the config, skipping entirely the ParticleOptions, even if actually in the config file.

**Further testing**
I made test plugin to save a ParticleEffect into a config file using Hocon and a ConfigurationNode.
I created the ConfigurationLoader and saved the ParticleEffect like this:
```java
ConfigurationLoader<CommentedConfigurationNode> config = LoaderHoconConfigurationLoader.builder().setPath(defaultConfigPath).build();
CommentedConfigurationNode rootNode = this.configLoader.load();

this.particle = ParticleEffect.builder()
                .type(ParticleTypes.CLOUD)
                .offset(new Vector3d(5d, 5d, 5d))
                .velocity(new Vector3d(0.5d, 0.5d, 0.5d))
                .build();

rootNode.setValue(TypeToken.of(ParticleEffect.class), this.particle);

this.configLoader.save(rootNode);

System.out.println("Created particle: " + particleEffectToString(this.particle));

private String particleEffectToString(ParticleEffect particleEffect) {
        StringBuilder string = new StringBuilder();

        string.append(particleEffect.getType());
        string.append(" options (" + particleEffect.getOptions().size() + "): ");
        particleEffect.getOptions().forEach((particleOption, o) -> {
            string.append(particleOption.getName());
            string.append(particleOption.getValueType());
        });

        return string.toString();
}
```
The println output is this
```
Created particle: SpongeParticleType{id=minecraft:cloud, name=cloud, internalType=CLOUD} options (2): offsetclass com.flowpowered.math.vector.Vector3dvelocityclass com.flowpowered.math.vector.Vector3d
```
The resulting config file is this:
```
Options=[
    {
        Option="minecraft:offset"
        Value {
            x=5.0
            y=5.0
            z=5.0
        }
    },
    {
        Option="minecraft:velocity"
        Value {
            x=0.5
            y=0.5
            z=0.5
        }
    }
]
Type="minecraft:cloud"
```
Now when I load the ParticleEffect at next execution from the config file like this:
```java
ParticleEffect particle = rootNode.getValue(TypeToken.of(ParticleEffect.class));

System.out.println("Loaded particle: " + particleEffectToString(this.particle));
```
The ParticleTypes is assigned but there are no ParticleOptions, as you can see from the println output:
```
Loaded particle: SpongeParticleType{id=minecraft:cloud, name=cloud, internalType=CLOUD} options (0): 
```
Gist containing both the logs at first and second execution and the test plugin main class [here](https://gist.github.com/anthonysan95/48a198b4cb2fad499976d07c810aab93).

If I did something wrong please tell me.. I tried in every possible way to deserialize it properly.
I even tried to go into the SpongeParticleEffectBuilder to see if the problem was there and this is what I found:
https://github.com/SpongePowered/SpongeCommon/blob/a5618c7a8bd32a10102fc443be8b0bc65002e511/src/main/java/org/spongepowered/common/effect/particle/SpongeParticleEffectBuilder.java#L64
I tested putting two println one before the line to see if the container contains the PARTICLE_OPTIONS and one inside the foreach printing the view, results: the first prints true while the other doesn't get called but I couldn't manage to get why.

### LINKED ISSUE #1963: `ConfigurateTranslator.translateMapOrList` throws NPE
- SpongeVanilla version: spongevanilla-1.12.2-7.1.0-BETA-74
  
`ConfigurateTranslator.translateMapOrList` throws NPE if node doesn't contain `key`.
Here is check if `key` is `null`: 
https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java#L123-L124
But inside this method there is `node.getKey().XXX` calls.
https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java#L83

So it throws NPE for me:
```
	at org.spongepowered.common.data.persistence.ConfigurateTranslator.translateMapOrList(ConfigurateTranslator.java:83) ~[ConfigurateTranslator.class:1.12.2-7.1.0-BETA-74]
	at org.spongepowered.common.data.persistence.ConfigurateTranslator.addTo(ConfigurateTranslator.java:124) ~[ConfigurateTranslator.class:1.12.2-7.1.0-BETA-74]
	at org.spongepowered.common.data.persistence.ConfigurateTranslator.translateFromNode(ConfigurateTranslator.java:71) ~[ConfigurateTranslator.class:1.12.2-7.1.0-BETA-74]
	at org.spongepowered.common.data.persistence.ConfigurateTranslator.translate(ConfigurateTranslator.java:115) ~[ConfigurateTranslator.class:1.12.2-7.1.0-BETA-74]
	at org.spongepowered.common.data.persistence.ConfigurateTranslator.translate(ConfigurateTranslator.java:45) ~[ConfigurateTranslator.class:1.12.2-7.1.0-BETA-74]
	at org.spongepowered.common.config.DataSerializableTypeSerializer.deserialize(DataSerializableTypeSerializer.java:50) ~[DataSerializableTypeSerializer.class:1.12.2-7.1.0-BETA-74]
	at org.spongepowered.common.config.DataSerializableTypeSerializer.deserialize(DataSerializableTypeSerializer.java:41) ~[DataSerializableTypeSerializer.class:1.12.2-7.1.0-BETA-74]
```

@gabizou

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeCommon/issues/1963
Fix https://github.com/SpongePowered/SpongeCommon/issues/1598
Fix x.0 being deserialized as int 

This does solve https://github.com/SpongePowered/SpongeAPI/issues/1543, however it still doesn't solve the type information loss mentioned in that issue. 

We need to keep in mind that `DataFormats.HOCON` and `DataFormats.JSON` SHOULD return the same DataView when serializing object.
Right now, HOCON is broken (see related issues mentioned above)

### DIFF
--- a/src/main/java/org/spongepowered/common/data/persistence/JsonDataFormat.java
+++ b/src/main/java/org/spongepowered/common/data/persistence/JsonDataFormat.java
@@ -144,18 +144,20 @@ public final class JsonDataFormat extends SpongeCatalogType implements StringDat
     private static Number readNumber(JsonReader reader) throws IOException {
         // Similar to https://github.com/zml2008/configurate/blob/master/configurate-gson/src/main/java/ninja/leaping/configurate/gson/GsonConfigurationLoader.java#L113
         // Not sure what's the best way to detect the type of number
-        double nextDouble = reader.nextDouble();
-        int nextInt = (int) nextDouble;
-        if (nextInt == nextDouble) {
-            return nextInt;
+        String str = reader.nextString();
+        try {
+            return Integer.parseInt(str);
+        } catch (NumberFormatException e) {
+            try {
+                return Double.parseDouble(str);
+            } catch (NumberFormatException ex) {
+                try {
+                    return Long.parseLong(str);
+                } catch (NumberFormatException exception) {
+                    throw new IOException(exception);
+                }
+            }
         }
-
-        long nextLong = (long) nextDouble;
-        if (nextLong == nextDouble) {
-            return nextLong;
-        }
-
-        return nextDouble;
     }
 
     private static List<?> readArray(JsonReader reader) throws IOException {
--- a/src/main/java/org/spongepowered/common/data/persistence/JsonDataFormat.java
+++ b/src/main/java/org/spongepowered/common/data/persistence/JsonDataFormat.java
@@ -145,16 +145,19 @@ public final class JsonDataFormat extends SpongeCatalogType implements StringDat
         // Similar to https://github.com/zml2008/configurate/blob/master/configurate-gson/src/main/java/ninja/leaping/configurate/gson/GsonConfigurationLoader.java#L113
         // Not sure what's the best way to detect the type of number
         String str = reader.nextString();
-        try {
-            return Integer.parseInt(str);
-        } catch (NumberFormatException e) {
-            try {
-                return Double.parseDouble(str);
-            } catch (NumberFormatException ex) {
+        if (str.contains(".")) {
+            return Double.parseDouble(str);
+        } else {
+            int limit = str.charAt(0) == '-' ? 11 : 10;
+            if (str.length() < limit) {
+                return Integer.parseInt(str);
+            } else if (str.length() > limit) {
+                return Long.parseLong(str);
+            } else {
                 try {
+                    return Integer.parseInt(str);
+                } catch (NumberFormatException e) {
                     return Long.parseLong(str);
-                } catch (NumberFormatException exception) {
-                    throw new IOException(exception);
                 }
             }
         }
--- a/src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
+++ b/src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
@@ -35,6 +35,7 @@ import org.spongepowered.api.data.DataView;
 import org.spongepowered.api.data.persistence.DataTranslator;
 import org.spongepowered.api.data.persistence.InvalidDataException;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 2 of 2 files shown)


src/main/java/org/spongepowered/common/data/persistence/JsonDataFormat.java
```diff
@@ -144,18 +144,20 @@ private static Object read(JsonReader reader) throws IOException {
...
-        int nextInt = (int) nextDouble;
-        if (nextInt == nextDouble) {
-            return nextInt;
-        }
-
-        long nextLong = (long) nextDouble;
-        if (nextLong == nextDouble) {
-            return nextLong;
+        String str = reader.nextString();
+        try {
```
dualspiral: Throwing exceptions (or allowing them to be thrown) is a rather expensive operation. Instead, maybe we could do a check to see if a decimal point exists in the string and parse a double based on that instead, though I don't know if that would be faster.

It would neaten this up slightly too.

randombyte-developer:
Does this fix #2094? Seems to be the same issue, but nothing linked to that issue as of now.

ImMorpheus:
> Does this fix #2094? Seems to be the same issue, but nothing linked to that issue as of now.

Likely. You can try use JSON and compare the DataContainers.

PR #2278, opened 2019-06-02
Overall Unit Testability with Schematics

gabizou:
**SpongeCommon** | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/2791)

After a few issues with Schematics and me hotfixing various little quirks, I decided to write some unit tests and integration tests.....

This PR is rather large due to the nature of Schematics and how they are processed.

Fixes:
- [x] Deserialized Schematics leak the entire DataContainer of the schematic's data through the metadata `DataView`
- [x] Always adding `Entities` list to file, even if it's empty.
- [x] Missing `RequiredMods` section for modded blocks/items/entities/tileentities
- [x] No equals implementation for any Schematic objects
- [x] TileEntity name/id resolution for TileEntityTypes doesn't work when registering custom mod ones.
- [x] Forceful addition of empty `SpongeData` on `TileEntity` datas
- [x] No fuzzy matching on `BlockState`s, can't match partial states with only specific properties
- [x] Palettes were not registered
- [x] Schematics don't store their own palette if the backing volume uses it's own palette

Cleanup:
- [x] Creating a Schematic from an `ArchetypeVolume` doesn't clean up references, retains duplicated collection objects


src/main/java/org/spongepowered/common/registry/type/BlockStateRegistryModule.java
```diff
@@ -44,9 +50,47 @@ public static BlockStateRegistryModule getInstance() {
...
+    @SuppressWarnings({"rawTypes", "unchecked"})
     @Override
     public Optional<BlockState> getById(String id) {
-        return Optional.ofNullable(this.blockStateMap.get(checkNotNull(id, "Id cannot be null!").toLowerCase(Locale.ENGLISH)));
+        final String state = checkNotNull(id, "Id cannot be null!").toLowerCase(Locale.ENGLISH);
+        final BlockState potential = this.blockStateMap.get(state);
+        if (potential != null) {
+            return Optional.of(potential);
+        }
+        if (state.contains("[")) {
```
gabizou, line 61: @Zidane I don't know if this can be improved, but how does it compare to your fuzzy maatching?

src/test/java/org/spongepowered/common/registry/CatalogTypeMethodTest.java
```diff
@@ -45,7 +45,7 @@
 import java.util.Optional;
 import java.util.Set;
 
-@RunWith(LaunchWrapperParameterized.class)
+//@RunWith(LaunchWrapperParameterized.class)
```
gabizou, line 48: Reminder to uncomment. Was adding too much time for gradle to figure out which tests to execute or not.

src/test/java/org/spongepowered/common/regression/SchematicUpgradeTest.java
```diff
@@ -0,0 +1,56 @@
...
+    public void testUpgradingv1Tov2() throws IOException {
+        File inputFile = new File(this.getClass().getClassLoader().getResource("loadv1.schematic").getFile());
+        DataContainer container = DataFormats.NBT.readFrom(new GZIPInputStream(new FileInputStream(inputFile)));
+        final Schematic v1Schem = DataTranslators.SCHEMATIC.translate(container);
+        inputFile = new File(this.getClass().getClassLoader().getResource("loadv2.schematic").getFile());
+        container = DataFormats.NBT.readFrom(new GZIPInputStream(new FileInputStream(inputFile)));
+        final Schematic v2Schem = DataTranslators.SCHEMATIC.translate(container);
+        assertEquals(v1Schem, v2Schem);
+    }
+
```
gabizou, line 55: Add another test for laoding a v1Schematic and saving it with the v2 format and compare to v2Schematic.

PR #2286, opened 2019-06-08
Clean up SpongePlayerDataHandler.saveFile().

JBYoshi:
- Shorten the try-with-resources block. This ensures the file is closed before we start moving it around.
- Remove a bunch of redundant method calls.
  - Files.createNewFile() was merged with Files.newOutputStream() using StandardOpenOption.CREATE.
  - Files.createDirectories() can always be used because it will not throw an error if the directory already exists.
  - The first Files.deleteIfExists() call can be ignored because we're using StandardCopyOption.REPLACE_EXISTING.
  - The second Files.deleteIfExists() call can be ignored because we're using Files.move() instead of Files.copy(), thus it will never exist.
- Use only one catch block for all operations. The stack trace will tell us what failed.


ImMorpheus (approved):
Working fine

PR #2293, opened 2019-06-14
Suppress exceptions caused by invalid UUIDs.

JBYoshi:
Occasionally, the user storage service will stumble across a player with an invalid UUID. This appears to be a [Mojang bug](https://bugs.mojang.com/browse/WEB-1312); their API is providing us with UUIDs that contain extra characters. This PR suppresses those errors until Mojang fixes that, by pretending the profile doesn't exist.

Fixes https://forums.spongepowered.org/t/29229.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/authlib/ProfileSearchResultsResponseSerializerMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/authlib/ProfileSearchResultsResponseSerializerMixin.java
@@ -1,0 +1,68 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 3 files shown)


dualspiral:
I think this workaround is in the wrong place. It should _at least_ be in the `GameProfileManager` [as it is possible to try to get a profile without using the USS (as I do here)](https://github.com/NucleusPowered/Nucleus/blob/sponge-api/7/src/main/java/io/github/nucleuspowered/nucleus/modules/core/commands/GetUserCommand.java#L51-L61).

You could be really adventurous and mixin to the Auth library before it even serialises, but that seems a bit overkill. We should protect against Mojang's bad responses equally thoughout Sponge though.

JBYoshi:
...Actually, since bulk lookups are possible, mixing in to the authentication library seems like the only right way to do this. Simply using a try/catch would cause all the lookups to fail if one was bad.

src/main/java/org/spongepowered/common/mixin/core/authlib/MixinProfileSearchResultsResponseSerializer.java
```diff
@@ -0,0 +1,61 @@
...
+            CallbackInfoReturnable<ProfileSearchResultsResponse> cir) {
+        // Occasionally we'll see a profile with an invalid UUID. The vanilla
+        // code, as well as most clients, will break when receiving these
+        // UUIDs. We'll suppress these errors for now; if one of them is a real
+        // user, Mojang will need to address the situation.
+        // See https://bugs.mojang.com/browse/WEB-1290.
+        for (Iterator<JsonElement> iterator = json.getAsJsonArray().iterator(); iterator.hasNext(); ) {
+            JsonObject element = iterator.next().getAsJsonObject();
+            String id = element.get("id").getAsString();
+            if (!id.matches("[0-9a-fA-F-]+")) {
```
ImMorpheus: Compile the regex and cache the result in a field

Zidane:
@JBYoshi Can you squash?

JBYoshi:
@Zidane Updated.

ImMorpheus:
Can you fix the conflicts and add a test plugin ?

ImMorpheus (approved):
Working as intended.

Please, fix the license for the testplugin and add name, version and description for the plugin annotation.

JBYoshi:
@ImMorpheus Added and updated to latest.

ImMorpheus (approved):

PR #2295, opened 2019-06-15
Change SchematicUpgradeTest to not use File

clienthax:
The current test code breaks on any paths with spaces in the name due to URL encoding,
Changes the test to use a stream directly from the resource, resolving the URL encoding step.

https://i.imgur.com/1DcZfGp.png

### DIFF
--- a/src/test/java/org/spongepowered/common/regression/SchematicUpgradeTest.java
+++ b/src/test/java/org/spongepowered/common/regression/SchematicUpgradeTest.java
@@ -43,7 +43,7 @@ public class SchematicUpgradeTest {
 
     @Test
     public void testUpgradingv1Tov2() throws IOException {
-        ClassLoader classLoader = this.getClass().getClassLoader();
+        final ClassLoader classLoader = this.getClass().getClassLoader();
         try (final InputStream v1InputStream = classLoader.getResource("loadv1.schematic").openStream();
              final InputStream v2InputStream = classLoader.getResource("loadv2.schematic").openStream()) {
                 final DataContainer v1Container = DataFormats.NBT.readFrom(new GZIPInputStream(v1InputStream));
--- a/src/test/java/org/spongepowered/common/regression/SchematicUpgradeTest.java
+++ b/src/test/java/org/spongepowered/common/regression/SchematicUpgradeTest.java
@@ -45,10 +45,12 @@ public class SchematicUpgradeTest {
     public void testUpgradingv1Tov2() throws IOException {
         final ClassLoader classLoader = this.getClass().getClassLoader();
         try (final InputStream v1InputStream = classLoader.getResource("loadv1.schematic").openStream();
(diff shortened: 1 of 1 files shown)


ImMorpheus (changes requested):
Wrap them in a try with resources

src/test/java/org/spongepowered/common/regression/SchematicUpgradeTest.java
```diff
@@ -34,23 +34,24 @@
...
-        DataContainer container = DataFormats.NBT.readFrom(new GZIPInputStream(new FileInputStream(inputFile)));
-        final Schematic v1Schem = DataTranslators.SCHEMATIC.translate(container);
-        inputFile = new File(this.getClass().getClassLoader().getResource("loadv2.schematic").getFile());
-        container = DataFormats.NBT.readFrom(new GZIPInputStream(new FileInputStream(inputFile)));
-        final Schematic v2Schem = DataTranslators.SCHEMATIC.translate(container);
-        assertEquals(v1Schem, v2Schem);
+        final ClassLoader classLoader = this.getClass().getClassLoader();
+        try (final InputStream v1InputStream = classLoader.getResource("loadv1.schematic").openStream();
+             final InputStream v2InputStream = classLoader.getResource("loadv2.schematic").openStream()) {
+                final DataContainer v1Container = DataFormats.NBT.readFrom(new GZIPInputStream(v1InputStream));
```
ImMorpheus: ~~that's better, but `GZIPInputStream` is still not closed~~
whoops, gabi is right. Actually, you don't need the try with resources at all. @clienthax can you update the PR ?

ImMorpheus (changes requested):

src/test/java/org/spongepowered/common/regression/SchematicUpgradeTest.java
clienthax: :'(
Should be fine now

gabizou: > that's better, but `GZIPInputStream` is still not closed

It gets closed by the [`NbtDataFormat.readFrom/writeTo`](https://github.com/SpongePowered/SpongeCommon/blob/bbb238a3292f5d988dbd0c4e9773ab78f00d778b/src/main/java/org/spongepowered/common/data/persistence/NbtDataFormat.java#L49-L79).

clienthax: @ImMorpheus not at home for 2 weeks, any chance you can just grab the older commit?

XakepSDK: Nothing wrong with closing already closed stream. It's ok by design.
Also you can just do something like this `new GZIPInputStream(classLoader.getResource("loadv2.schematic").openStream())`, close call shall close all inner streams

PR #2296, opened 2019-06-15
Privacy-security/timings

ImMorpheus:


src/main/java/org/spongepowered/common/relocate/co/aikar/timings/TimingsExport.java
```diff
@@ -293,14 +293,17 @@ public void run() {
...
+                } catch (IOException e) {
+                    SpongeImpl.getLogger().warn("Could not get own server hostname when uploading timings - falling back to 'localhost'", e);
+                }
             }
-            HttpURLConnection con = (HttpURLConnection) new URL("http://timings.aikar.co/post").openConnection();
+            HttpURLConnection con = (HttpURLConnection) new URL("https://timings.aikar.co/post").openConnection();
             con.setDoOutput(true);
-            con.setRequestProperty("User-Agent", "Sponge/" + getServerName() + "/" + hostname);
+            String name = TimingsManager.privacy ? "" : getServerName();
+            con.setRequestProperty("User-Agent", "Sponge/" + name + "/" + hostname);
```
ImMorpheus, line 306: @aikar Is this a valid user agent ?

simon816: The only user agent checking on the timings side is https://github.com/aikar/timings/blob/master/lib/Security.php#L24
So it doesn't really matter what it is.

ImMorpheus: That's just a template. It's not what he's using in production
https://github.com/aikar/timings/blob/c41b12ddacde256e3b54e667deca344a6587ff4f/lib/Security.php#L16

src/main/java/org/spongepowered/common/config/category/TimingsCategory.java
```diff
@@ -67,6 +67,10 @@ public void setEnabled(boolean enabled) {
         this.enabled = enabled;
     }
 
+    public boolean isServerNamePrivacy() {
```
Zidane: isServerNamePrivate

Zidane:
Not that this change is wrong but...the point? Is it to hide the plugins/mods a particular server is using?

phit:
yes for example, or maybe a server wanting to seek support and share stuff without wanting it to be attached to what server it is

ImMorpheus:
> Not that this change is wrong but...the point? Is it to hide the plugins/mods a particular server is using?

No, plugins/mods are always attached to a timings report.
https://github.com/SpongePowered/SpongeCommon/blob/bbb238a3292f5d988dbd0c4e9773ab78f00d778b/src/main/java/org/spongepowered/common/relocate/co/aikar/timings/TimingsExport.java#L195-L201

However the server name is stripped from the report if the privacy toggle is on.
https://github.com/SpongePowered/SpongeCommon/blob/bbb238a3292f5d988dbd0c4e9773ab78f00d778b/src/main/java/org/spongepowered/common/relocate/co/aikar/timings/TimingsExport.java#L124-L129

The point of this PR is stripping the server name from the user agent (and switching to https for the request).
There's no use in the privacy toggle if the server name is always sent in the user agent. 

EDIT: it turned out we were also ignoring the hiddenConfigEntries config. Thankfully, sponge sql category path is hardcoded and we were not leaking user credentials.
https://github.com/SpongePowered/SpongeCommon/blob/bbb238a3292f5d988dbd0c4e9773ab78f00d778b/src/main/java/org/spongepowered/common/relocate/co/aikar/timings/TimingsExport.java#L252

aikar:
I would of preferred if it stayed in the user agent incase a server is hitting me too hard, to give me a way to potentially identify and contact.
The user agent portion is only for my own logs, but the sponge/ prefix is the important part.

Zidane:
@ImMorpheus Please honor @aikar 's request on this.

PR #2297, opened 2019-06-16
Cleanup/config

### LINKED ISSUE #1626: remove unused config options
Review of the code revealed the configuration option:

https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/config/category/GeneralCategory.java#L33

Is not actually implemented anywhere in the rest of common or within SF.

ImMorpheus:
Tested on both SV and SF

Fix https://github.com/SpongePowered/SpongeCommon/issues/1626

### DIFF
--- a/src/main/java/org/spongepowered/common/config/category/EntityCollisionCategory.java
+++ b/src/main/java/org/spongepowered/common/config/category/EntityCollisionCategory.java
@@ -39,14 +39,10 @@ public class EntityCollisionCategory extends ConfigCategory {
                                                          + "performance when there are more than 8 entities on top of each other such as a 1x1 \n"
                                                          + "spawn pen. Set to 0 to disable.")
     private int maxEntitiesWithinAABB = 8;
-    @Setting(value = "defaults", comment = "Default maximum collisions used for all entities/blocks unless overridden.")
-    private Map<String, Integer> defaultMaxCollisions = new HashMap<>();
     @Setting(value = "mods", comment = "Per-mod overrides. Refer to the minecraft default mod for example.")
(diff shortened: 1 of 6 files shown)


src/main/java/org/spongepowered/common/config/category/EntityCollisionCategory.java
```diff
@@ -39,14 +39,10 @@
                                                          + "performance when there are more than 8 entities on top of each other such as a 1x1 \n"
                                                          + "spawn pen. Set to 0 to disable.")
     private int maxEntitiesWithinAABB = 8;
-    @Setting(value = "defaults", comment = "Default maximum collisions used for all entities/blocks unless overridden.")
```
phit, line 42: this is being used https://github.com/SpongePowered/SpongeCommon/blob/73bd5a690228331902862e20115ce9c841e68b70/src/main/java/org/spongepowered/common/mixin/entitycollisions/MixinBlock_Collisions.java
https://github.com/SpongePowered/SpongeCommon/blob/73bd5a690228331902862e20115ce9c841e68b70/src/main/java/org/spongepowered/common/mixin/entitycollisions/MixinEntity_Collisions.java

ImMorpheus: No, it's using https://github.com/SpongePowered/SpongeCommon/blob/bbb238a3292f5d988dbd0c4e9773ab78f00d778b/src/main/java/org/spongepowered/common/config/category/CollisionModCategory.java#L39

phit: my bad

PR #2302, opened 2019-06-21
Fix/user sync

ImMorpheus:
Address https://github.com/SpongePowered/SpongeCommon/pull/2124#issuecomment-500161476

Ensure Player is synced with its User object.

### DIFF
--- a/src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
+++ b/src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
@@ -159,8 +159,8 @@ public class SpongeUser implements ArmorEquipable, Tamer, DataSerializable, Carr
         }
 
         // Note: Uses the overworld's player data
-        SaveHandler saveHandler = (SaveHandler) worldServer.get().getSaveHandler();
-        File file = new File(saveHandler.playersDirectory, this.profile.getId().toString() + ".dat");
+        final SaveHandlerAccessor saveHandler = (SaveHandlerAccessor) worldServer.get().getSaveHandler();
+        final File file = new File(saveHandler.accessor$getPlayersDirectory(), this.profile.getId().toString() + ".dat");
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/common/entity/player/SpongeUser.java
```diff
@@ -534,23 +576,21 @@ public void markDirty() {
     }
 
     public void save() {
+        Preconditions.checkState(isInitialized(), "SpongeUser {} is not initialized", this.profile.getId());
```
Zidane: Just say User.

src/main/java/org/spongepowered/common/mixin/api/data/MixinDataHolder.java
```diff
@@ -65,16 +66,26 @@
     @Override
     public <T extends DataManipulator<?, ?>> Optional<T> get(Class<T> containerClass) {
         TimingsManager.DATA_GROUP_HANDLER.startTimingIfSync();
-        ;
         SpongeTimings.dataGetManipulator.startTimingIfSync();
+        boolean isUser = ((Object) this) instanceof SpongeUser;
```
Zidane: Maybe put this in a utility method to reduce the diff?

src/main/java/org/spongepowered/common/mixin/api/data/MixinDataHolder.java
```diff
@@ -65,16 +66,26 @@
...
+        DataHolder holder = this;
+        if (isUser) {
+            if (((User) this).isOnline()) {
+                holder = ((User) this).getPlayer().get();
```
Zidane: I kind of feel like this should be done in SpongeUser and not put in MixinDataHolder...but would involve duplicate code at that point. Just feels wrong, perhaps either way.

Thoughts @gabizou ?

PR #2304, opened 2019-06-22
Ensure the player is on the cause stack in tryHarvestBlock().

JBYoshi:
This is primarily intended for the mod Ore Excavator, which calls tryHarvestBlock() from a post-tick listener. See SpongePowered/SpongeForge#2789 and SpongePowered/SpongeForge#2832.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/server/management/PlayerInteractionManagerMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/server/management/PlayerInteractionManagerMixin.java
@@ -70,6 +70,8 @@ import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.common.SpongeImpl;
 import org.spongepowered.common.SpongeImplHooks;
 import org.spongepowered.common.bridge.entity.player.ServerPlayerEntityBridge;
 import org.spongepowered.common.bridge.inventory.ContainerBridge;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
```diff
@@ -523,4 +527,29 @@ public EnumActionResult processRightClick(EntityPlayer player, net.minecraft.wor
...
+    private final Deque<CauseStackManager.StackFrame> tryHarvestBlockStackFrames = new LinkedList<>();
+
+    /**
+     * Some mods (e.g. OreExcavator) will call tryHarvestBlock directly, and
+     * the cause stack may not have the player on it. This should fix that.
+     * See SpongeForge#2789.
+     */
+    @Inject(method = "tryHarvestBlock", at = @At("HEAD"))
+    private void onTryHarvestBlockHead(BlockPos pos, CallbackInfoReturnable<Boolean> cir) {
+        if (!Sponge.getCauseStackManager().getCurrentCause().contains(this.player)) {
```
Zidane: Considering this code came about due to tick logic, I would rather just throw it on the stack and pop it. Worst case, we have a dupe object. Best case we don't. Either case, we save a lookup.

JBYoshi: Done.

src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
```diff
@@ -523,4 +527,23 @@ public EnumActionResult processRightClick(EntityPlayer player, net.minecraft.wor
         }
     }
 
+    private final Deque<CauseStackManager.StackFrame> tryHarvestBlockStackFrames = new LinkedList<>();
```
gabizou: Move this field to the top with the other mixin-added fields please

src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
```diff
@@ -523,4 +527,23 @@ public EnumActionResult processRightClick(EntityPlayer player, net.minecraft.wor
...
+
+    /**
+     * Some mods (e.g. OreExcavator) will call tryHarvestBlock directly, and
+     * the cause stack may not have the player on it. This should fix that.
+     * See SpongeForge#2789.
+     */
+    @Inject(method = "tryHarvestBlock", at = @At("HEAD"))
+    private void onTryHarvestBlockHead(BlockPos pos, CallbackInfoReturnable<Boolean> cir) {
```
gabizou: with the ongoing cleanup/splitting of mixins, can you name this `impl$addPlayerToCauseStackForHarvesting`?

src/main/java/org/spongepowered/common/mixin/core/server/management/MixinPlayerInteractionManager.java
```diff
@@ -523,4 +527,23 @@ public EnumActionResult processRightClick(EntityPlayer player, net.minecraft.wor
...
+        this.tryHarvestBlockStackFrames.push(Sponge.getCauseStackManager().pushCauseFrame());
+        Sponge.getCauseStackManager().pushCause(this.player);
+    }
+
+    @Inject(method = "tryHarvestBlock", at = @At("RETURN"))
+    private void onTryHarvestBlockReturn(BlockPos pos, CallbackInfoReturnable<Boolean> cir) {
```
gabizou: And this `impl$PopPlayerFromStackForHarvesting`

gabizou (changes requested):
I'm debating on the whole whether we'd want to enter/exit phase states so that we'd avoid potentially forcing a stackframe creation (since stackframes are lazily initialized based on the phase states). PhaseState entry/exit is relatively cheap, and adding a new class here as generalized as it is, will allow us to potentially refine the entry to support whether using the `CapturePlayer` object, or whether we'd just write it as a new state providing the frame modifier.

Alternatively, can check on the phase state to default to adding the player as a modifier, or adding the player to the phase context as a captured player during harvesting sort of deal.

Zidane:
Can you update this PR @JBYoshi ?

JBYoshi:
@Zidane Updated.

gabizou:
Merged in to SF to use a mixin there so it's not always occurring in SpongeVanilla.

PR #2307, opened 2019-06-24
Finish merging TrackingPhase into IPhaseState.

JBYoshi:
**SpongeCommon** | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/2836)

#1567 removed almost all of the references to `TrackingPhase` in the Sponge codebase. This PR removes the last bit and the class itself.

The only remaining uses for `TrackingPhase` were for checking whether to enter a new phase for block decaying. They always followed a format similar to the following:
```java
// Seems like this would be better named isNotBlock, but it's always called isBlockAlready.
final boolean isBlockAlready = currentState.getPhase() != TrackingPhases.BLOCK; 
final boolean isWorldGen = currentState.isWorldGeneration();
try (PhaseContext context = isBlockAlready && !isWorldGen
        ? /* create BLOCK_DECAY context */
        : null) {
    if (context != null) context.buildAndSwitch();
    // run code
}
```

I replaced those checks with a method `includesDecays()`, which returns `true` when a new phase should *not* be started. (I originally thought about inverting the method and calling it `needsSeparateDecayState()`, but I felt this would fit better with the pattern for phase state methods focusing on the state itself.) From my testing, leaf decaying seems to be working fine with these changes.

Besides that, the only necessary changes were removing the `TrackingPhase` from log messages and inlining a few calls regarding `PacketPhase` as a registry of its phase states.


gabizou:
Gotta fix all the `toString()` now for the phase states to print a proper name based on their generic -> specific usage (it was my intention to have the names uniformed so that they'd all be named somewhat like `EntityPhase{DimensionChangeState}` or something thereof. I know the `toString()`s are all over the place, but if you're tackling this, might be worth to actually fix that too ;)

JBYoshi:
A few examples of the new `toString()` code:

```
BlockWorkerPhaseState.toString(): Plugin{BlockWorker}
BasicPluginState.toString(): Plugin{Basic}
ScheduledTaskPhaseState.toString(): Plugin{ScheduledTask}
BasicPluginState.toString(): Plugin{Basic}
BlockTickPhaseState.toString(): Tick{Block:BlockTickPhase}
BlockTickPhaseState.toString(): Tick{Block:RandomBlockTickPhase}
EntityTickPhaseState.toString(): Tick{Entity}
DimensionTickPhaseState.toString(): Tick{Dimension}
```
For the most part, it's automatically derived from the class name, but it's possible to add extra information, such as in the two `BlockTickPhaseState` entries.

PR #2311, opened 2019-06-25
Using URLDecoder to decode the possibly encoded strings

### LINKED ISSUE #2308: SQL service doesn't decode URL encoded usernames or passwords
So the core issue is that if you have special chars in your username or password when using a JDBC URL you are supposed to URL encode them. The way the current code is setup on the Sponge side makes that fairly irrelevant, until either the username contains `:` or `@` or the password contains `@`.

Then the regex parsing the username and password in the `SqlServiceImpl.ConnectionInfo` class just doesn't work anymore. As mentioned before the proper solution to the problem is providing the username and password URL encoded.  
But the `SqlServiceImpl.ConnectionInfo#fromUrl` method doesn't care about that.

So I propose to add `URLDecoder#decode` calls on these two lines: https://github.com/SpongePowered/SpongeCommon/blob/a57bec007c840e6eb4540d3695139d565711d87d/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java#L261-L262

Here's a little example:  
The URL `jdbc:mysql://test%40%3A%C3%A4%C3%B6%C3%BC%26%3F+%40test:pw%40%3A%C3%A4%C3%B6%C3%BC%26%3F+%40pw@localhost/db` should try to connect to `localhost` with the username `test@:äöü&? @test` and the password `pw@:äöü&? @pw` on the database `db`

BrainStone:
This PR should fix #2308 

It URL decodes the username and password in a JDBC connection URL.  
Also adds a unit test to make sure that functionality doesn't break.

Probably a good idea to add that change to API 8 and any other still maintained API version.

### DIFF
--- a/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
+++ b/src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
@@ -43,6 +43,9 @@ import org.spongepowered.common.config.SpongeConfigManager;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.sql.DriverManager;
@@ -175,6 +178,7 @@ public class SqlServiceImpl implements SqlService, Closeable {
     public static class ConnectionInfo {
 
         private static final Pattern URL_REGEX = Pattern.compile("(?:jdbc:)?([^:]+):(//)?(?:([^:]+)(?::([^@]+))?@)?(.*)");
+        private static final String UTF_8 = StandardCharsets.UTF_8.name();
         @Nullable private final String user;
         @Nullable private final String password;
         private final String driverClassName;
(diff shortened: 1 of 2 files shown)


BrainStone:
Travis failing is not my fault btw.

src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
```diff
@@ -269,6 +273,15 @@ public static ConnectionInfo fromUrl(@Nullable PluginContainer container, String
             final String driverClass = DriverManager.getDriver(unauthedUrl).getClass().getCanonicalName();
             return new ConnectionInfo(user, pass, driverClass, unauthedUrl, fullUrl);
         }
+
+        private static String urlDecode(String str) {
+            try {
+                return (str == null)? null : URLDecoder.decode(str, UTF_8);
```
randombyte-developer: OCD: missing space between `)` and `?`

src/main/java/org/spongepowered/common/service/sql/SqlServiceImpl.java
```diff
@@ -269,6 +273,15 @@ public static ConnectionInfo fromUrl(@Nullable PluginContainer container, String
...
+            } catch(UnsupportedEncodingException e) {
```
randombyte-developer: OCD: space between `catch` and `(`

PR #2318, opened 2019-07-09
Add Parallel GZIP Optimization.

clienthax:


clienthax:
Will likely change this to use https://engineering.linkedin.com/blog/2019/02/migz-for-compression-and-decompression

phit:
is this an actual issue? even the biggest data the game has is not above 8mb files usually, looking at those benchmarks from linkedin and the perf improvements on GB large files, this would give micro seconds improvements at best. depending on the initial overhead it may not be an improvement at all, do you have any data to support that this helps?

clienthax:
@phit none currently, it seems it could help during large pregens and retrogen operations but I need to test it properly

clienthax:
Closing this until i can do tests

Issue #2325, opened 2019-07-21
Keys.VANISH status stuck in true after logging out vanished


dualspiral:
**I am currently running**
- SpongeForge version: 1.12.2-2838-7.1.7-RC3854
- SpongeVanilla version: 1.12.2-7.1.7-RC244
- Java version: 1.8.0 u202
- Operating System: All, macOS, Linux

- Plugins/Mods:
  - Only https://gist.github.com/dualspiral/a2f13d1673dcaffa819a475b6c82d624

**Issue Description**

The issue here is two-fold:

* When a player respawns, the `Keys.VANISH` key is not persisted. If a player logged in vanished then unvanishes, then respawns, the vanish key is set to that upon login. So, if a player logged in false, but then vanishes then respawns, they will respawn unvanished.
* When a player logs out with `Keys.VANISH` set to `true`, when they log in, the key is set to `true` upon every subsequent login, even if an attempt is made to set it to `false`.

So - to reproduce (bold lines are the incorrect behaviour):

* Create a Sponge server, SF or SV, doesn't matter, with the above plugin only.
* Log in with your character
* Set the player vanished using `/vanishset [player] true`.
* Kill the player using `/kill [player]`
* **Note in the console that the player is no longer vanished.**
* Relog - the player is not vanished.
* Set the player to vanish again.
* Relog - the player is vanished.
* Unvanish then `/kill` your player
* **Note the player is vanished upon respawn**
* Unvanish and logout
* **Log back in, player is vanished.**

I imagine that other data may be affected by this, but I haven't had time to check or run this all through the debugger as of yet - hopefully someone has a good idea who has been touching this recently!

dualspiral:
Just as an update to ensure it's logged, Morph fixed the VANISH status remaining on login/logout in https://github.com/SpongePowered/SpongeCommon/commit/e67a112581b3cb975d69547875ae1ff3e7eaad43, but the issue of copying the data on respawn remains. From conversation, it looks like custom data is also affected, I suspect other data will be affected too.

ImMorpheus:
PR merged

PR #2329, opened 2019-07-28
Fix data lost on respawn

ImMorpheus:
Fix the last part of https://github.com/SpongePowered/SpongeCommon/issues/2325

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/server/PlayerListMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/server/PlayerListMixin.java
@@ -61,6 +61,7 @@ import net.minecraft.world.storage.WorldInfo;
 import org.apache.logging.log4j.Logger;
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.command.CommandSource;
+import org.spongepowered.api.data.manipulator.DataManipulator;
 import org.spongepowered.api.entity.Transform;
 import org.spongepowered.api.entity.living.player.Player;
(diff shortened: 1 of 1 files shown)


Zidane:
Do we want to run through the offer logic for this? We could take the old manipulator objects wholesale.

ImMorpheus:
> Do we want to run through the offer logic for this? We could take the old manipulator objects wholesale.

Yes, this also affects vanilla manipulators. We need the offer logic to call the correct data processor.

src/main/java/org/spongepowered/common/mixin/core/server/PlayerListMixin.java
```diff
@@ -368,6 +369,9 @@ public EntityPlayerMP recreatePlayerEntity(final EntityPlayerMP playerIn, int ta
         ((org.spongepowered.api.world.World) worldServer).spawnEntity(spongeEntity);
         this.playerEntityList.add(newPlayer);
         newPlayer.connection.sendPacket(new SPacketPlayerListItem(SPacketPlayerListItem.Action.UPDATE_GAME_MODE, newPlayer));
+        for (DataManipulator<?, ?> container : ((Player) playerIn).getContainers()) {
```
gabizou, line 372: Should be using the custom containers instead of all containers.

ImMorpheus: Wouldn't that be missing vanilla manipulators ?

Issue #2198, opened 2019-02-23
updateHealthPriorToArmor causing performance issues on large servers


bloodmc:
I've received numerous spark results from servers showing `updateHealthPriorToArmor` using 4-5% usage as its called every tick for each player.

Here are a few samples

https://sparkprofiler.github.io/?l09ftzunWm
https://sparkprofiler.github.io/?bNEgGLGEya

net.minecraft.entity.player.EntityPlayerMP.handler$updateHealthPriorToArmor$zhb000()4.05%
net.minecraft.entity.player.EntityPlayerMP.handler$updateHealthPriorToArmor$zhb000()2.46%

ImMorpheus:
PR merged

PR #2330, opened 2019-07-30
Fix scaled health

### LINKED ISSUE #2898: Health limit (or something similiar)
**I am currently running**

- SpongeForge version: spongeforge-1.12.2-2838-7.1.7-RC3854.
- Forge version: forge-1.12.2-14.23.5.2838-universal

- Plugins/Mods:
image

**Issue Description**

With increasing health through various modifiers, its indicators reach a certain limit, after which the amount of health ceases to increase. As an example, perks AstralSorcery, Shadow Armor ExtraBotany(Botania addon), Botaia Relic rings, and other modifier increase players HP, but on server with SF that's don't affect player health correctly. Demo:
image
image
In a singleplayer, everything works fine.

Similar problems have affected a lot of things, starting with damage modifiers and ending with the digging area.

ImMorpheus:
First off, let's define `HEALTH_SCALE`.

From the API https://github.com/SpongePowered/SpongeAPI/blob/c23f58d2a18b61b2ba457977e7aaf57dfbd749d0/src/main/java/org/spongepowered/api/data/key/Keys.java#L1029
```
     * Represents the {@link Key} for how much health a half-heart on a
     * {@link Player}'s GUI will stand for.
```

In vanilla max health value is 20 (or 10 hearts). Thus a half-heart is 1 health value.

Vanilla:

max health: 20
health scale (remember the definition): 1

In the implementation the default health scale is set to 20. Moreover `getMaxHealth()` and `getHealth()` take the health and apply the modifiers, so there's no need to run that logic twice (see `getInternalScaledHealth()`)

EDIT: the api is wrong, see gabi comment

I've played around with the ScaledHealth testplugin and botania's ring of odin.

This should fix https://github.com/SpongePowered/SpongeForge/issues/2898

### DIFF
--- a/src/main/java/org/spongepowered/common/bridge/entity/player/ServerPlayerEntityBridge.java
+++ b/src/main/java/org/spongepowered/common/bridge/entity/player/ServerPlayerEntityBridge.java
@@ -89,8 +89,6 @@ public interface ServerPlayerEntityBridge {
 
     boolean bridge$isHealthScaled();
 
-    void bridge$setHealthScaled(boolean scaled);
-
     void bridge$refreshScaledHealth();
 
     void bridge$injectScaledHealth(Collection<IAttributeInstance> set, boolean b);
--- a/src/main/java/org/spongepowered/common/bridge/entity/player/ServerPlayerEntityBridge.java
+++ b/src/main/java/org/spongepowered/common/bridge/entity/player/ServerPlayerEntityBridge.java
@@ -91,7 +91,7 @@ public interface ServerPlayerEntityBridge {
 
     void bridge$refreshScaledHealth();
 
-    void bridge$injectScaledHealth(Collection<IAttributeInstance> set, boolean b);
+    void bridge$injectScaledHealth(Collection<IAttributeInstance> set);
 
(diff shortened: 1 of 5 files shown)


Zidane:
I'll let @gabizou merge this one.

gabizou:
The API was ripped from the Bukkit standard, which basically had the scale to 20, so you'd have the artificial health scale of visible health on the client being 20 hp, or 10 hearts. It's a misdocumented javadoc in that sense to make it understood that the scale is for 1 heart to show for 1 hp. If the vanilla scale were used, 2 hp = 1 heart. But, again, because the original design was for the scaled maximum health to show on the client, it was implemented as such: What is the scaled maximum health to show on the client for that player? 20 hp by default. It got tricky when it came to attribute modifiers because mods would add attributes that would show half hearts or increment by half hearts and those modifiers need to show on the client, but *not scaled to what the internal scaled health would represent, but their actual values.*

ImMorpheus:
Last commit should fix https://github.com/SpongePowered/SpongeCommon/issues/2198
The injection was running every tick instead of when the `armorValue` changed

PR #2335, opened 2019-08-04
Add SoundEvent Implementation

clienthax:
API: https://github.com/SpongePowered/SpongeAPI/pull/2028

### DIFF
--- a/src/main/java/org/spongepowered/common/effect/record/SpongeRecordType.java
+++ b/src/main/java/org/spongepowered/common/effect/record/SpongeRecordType.java
@@ -27,10 +27,12 @@ package org.spongepowered.common.effect.record;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.flowpowered.math.vector.Vector3i;
+import net.minecraft.item.Item;
 import net.minecraft.network.play.server.SPacketEffect;
 import net.minecraft.util.math.BlockPos;
 import org.spongepowered.api.effect.sound.SoundType;
 import org.spongepowered.api.effect.sound.record.RecordType;
+import org.spongepowered.api.item.ItemType;
 import org.spongepowered.common.SpongeCatalogType;
 import org.spongepowered.common.text.translation.SpongeTranslation;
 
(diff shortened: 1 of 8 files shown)


clienthax:
https://github.com/SpongePowered/SpongeAPI/issues/1562

clienthax:
On a side note, if anyone knows of a SANE way which these magic ids match to any sponge representation, let me know.
1023 = ENTITY_WITHER_SPAWN
1028 = ENTITY_ENDERDRAGON_DEATH
1038 = BLOCK_END_PORTAL_SPAWN
RenderGlobal.broadcastSound

simon816:
Sounds are represented in the API as `SoundType` - https://github.com/SpongePowered/SpongeAPI/blob/stable-7/src/main/java/org/spongepowered/api/effect/sound/SoundTypes.java

clienthax:
@simon816 Those ids arnt soundtypes however by the looks of it they seem to be more "event keys"

clienthax:
Should be ready for review.

dualspiral:
[How will this work with the Forge PlaySoundAtEntityEvent](https://github.com/MinecraftForge/MinecraftForge/blob/1.12.x/src/main/java/net/minecraftforge/event/entity/PlaySoundAtEntityEvent.java#L52)?

I suspect this event impl is going to need some scaffolding in SpongeForge to either mix in to this event, or having some parings between Forge and Sponge events [such as this](https://github.com/SpongePowered/SpongeForge/blob/a480900d50fbc91a6621b5d25a2b00201a2f715f/src/main/java/org/spongepowered/mod/event/SpongeModEventManager.java#L174). A mod should have the oppotunity to cancel an event at the same time as its Sponge one and see that state, and you may want to change the Forge logic slightly to route to the Sponge event handlers.

A good example of where we do this is the `MessageChannelEvent.Chat` and `ServerChatEvent` pairing - they fire "concurrently" and sync status to each other as the events are sent to each member of their respective handler lists.

clienthax:
@dualspiral Im not too sure on how to link that one in without creating duplicate events currently, will leave it to someone more suited.

PR #2339, opened 2019-08-07
Increase valid Y position range for entity loading

clienthax:
Fixes disapearing Flat Transfer Nodes in ExtraUtils2 (Stores a master entity above the chunk at y512)
https://github.com/SpongePowered/SpongeForge/issues/2293

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/util/math/Vec3iMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/util/math/Vec3iMixin.java
@@ -43,7 +43,7 @@ public abstract class Vec3iMixin implements BlockPosBridge {
 
     @Override
     public boolean bridge$isValidPosition() {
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/core/util/math/Vec3iMixin.java
```diff
@@ -43,7 +43,7 @@
 
     @Override
     public boolean bridge$isValidPosition() {
-        return this.x >= -30000000 && this.z >= -30000000 && this.x < 30000000 && this.z < 30000000 && this.y >= 0 && this.y < 256;
+        return this.x >= -30000000 && this.z >= -30000000 && this.x < 30000000 && this.z < 30000000 && this.y >= 0 && this.y < 1024;
```
Zidane, line 46: I understand the below change but I don't think this change is correct. A valid position should be within Minecraft's typical bounds.

JBYoshi: How about leaving this method as-is and have the code using it call `isValidXZPosition()` instead?

PR #2342, opened 2019-08-09
Implement PlaySoundEvent.NoteBlock

clienthax:
API: https://github.com/SpongePowered/SpongeAPI/pull/2031

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/block/BlockNoteMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/block/BlockNoteMixin.java
@@ -1,0 +1,90 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/common/mixin/core/block/BlockNoteMixin.java
```diff
@@ -0,0 +1,87 @@
...
+            return;
+        }
+
+        TileEntity tileEntity = worldIn.getTileEntity(pos);
+        if(!(tileEntity instanceof Note))
+            return;
+
+        float pitch = (float) Math.pow(2.0D, (double) (param - 12) / 12.0D);
+
+        if(ShouldFire.PLAY_SOUND_EVENT_NOTE_BLOCK) {
```
ImMorpheus: You can place this at the beginning of the method.

ImMorpheus (changes requested):
just a small performance improvement.

(also, nitpick: `if(` -> `if (` )

ImMorpheus (approved):

PR #2353, opened 2019-08-18
Fix/translator again

### LINKED ISSUE #1142: NullPointerException loading configuration
I have ran into an issue loading configs (Initial discussion: https://forums.spongepowered.org/t/nullpointerexception-loading-configuration-after-initial-save/16648/7). 

I believe this may be an issue in Sponge.

### LINKED ISSUE #1963: `ConfigurateTranslator.translateMapOrList` throws NPE

### LINKED ISSUE #1973: Deserialization of MappedData is broken
Currently, any `DataManipulator` that extends `MappedData` (e.g. `StatisticData`) will not be deserialized properly if either its keys or values have special serialization behavior. `DataView#getMap` doesn't attempt to perform any sort of deserialization on the keys or values, which prevents a `MappedData` from being round-tripped through serialization properly.

### LINKED ISSUE #2352: The configurate translator failed on non-root nodes
**I am currently running**
- SpongeVanilla version: `1.12.2-7.1.7-RC268`
- Java version: `openjdk version "1.8.0_212"`
- Operating System: `#1 SMP Debian 4.19.37-6 (2019-07-18)`

The example plugin and part of the output:  
<https://gist.github.com/ustc-zzzz/dc6518488b62dbc19d6fdfa5df8a36f4>

The configurate translator successfully handled `node1` while failed to handle `node2` and threw an error.

### LINKED ISSUE #2439: Cannot deserialize FireworkEffect correctly.
Write to a config file with:
```java
node.getNode("FireworkEffects").setValue(TypeTokens.LIST_FIREWORK_TOKEN, fireworkEffects);
```
get:
```hocon
FireworkEffects: [
  {
    Colors=[
      {
        Blue=152
        ContentVersion=1
        Green=129
        Red=216
      }
    ]
    ContentVersion=1
    Fades=[]
    Flickers=false
    Trails=false
    Type="minecraft:ball"
  }
]
```
Read with:
```java
List<FireworkEffect> fireworkEffects = node.getNode("FireworkEffects").getChildrenList().stream().map(childNode -> {
    DataContainer dataContainer = DataTranslators.CONFIGURATION_NODE.translate(childNode);
    //For https://github.com/SpongePowered/Configurate/issues/93
    Consumer<DataQuery> fixEmptyList = (dataQuery) -> {
        if (!dataContainer.contains(dataQuery)) {
            dataContainer.set(dataQuery, Collections.emptyList());
        }
    };
    fixEmptyList.accept(DataQuery.of("Colors"));
    fixEmptyList.accept(DataQuery.of("Fades"));
    return Sponge.getDataManager().deserialize(FireworkEffect.class, dataContainer).get();
}).collect(Collectors.toList());
```
And the colors of read fireworkEffects is empty.

I find that colors' child is Map instead of DataView in the dataContainer.
So I edit the read part :
```java
List<FireworkEffect> fireworkEffects = node.getNode("FireworkEffects").getChildrenList().stream().map(childNode -> {
    DataContainer dataContainer = DataTranslators.CONFIGURATION_NODE.translate(childNode);
    //For https://github.com/SpongePowered/Configurate/issues/93
    Consumer<DataQuery> fixEmptyList = (dataQuery) -> {
        if (!dataContainer.contains(dataQuery)) {
            dataContainer.set(dataQuery, Collections.emptyList());
        }
    };
    fixEmptyList.accept(DataQuery.of("Colors"));
    fixEmptyList.accept(DataQuery.of("Fades"));
    
    Consumer<DataQuery> updateList = dataQuery -> {
        //noinspection OptionalGetWithoutIsPresent
        List<?> list = dataContainer.getList(dataQuery).get();
        dataContainer.set(dataQuery, list.stream().map(o -> {
            if (o instanceof Map) {
                DataContainer container = DataContainer.createNew();
                ((Map<?, ?>) o).forEach((k, v) -> container.set(DataQuery.of(k.toString()), v));
                return container;
            } else {
                return o;
            }
        }).collect(Collectors.toList()));
    };
    updateList.accept(DataQuery.of("Colors"));
    updateList.accept(DataQuery.of("Fades"));
    
    return Sponge.getDataManager().deserialize(FireworkEffect.class, dataContainer).get();
}).collect(Collectors.toList());
```
Things work correctly now. 
I wonder whether it is a bug.

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeCommon/issues/2352
Fix https://github.com/SpongePowered/SpongeCommon/issues/1142
Fix https://github.com/SpongePowered/SpongeCommon/issues/1963 
Fix https://github.com/SpongePowered/SpongeCommon/issues/1973 (need a test)
Fix https://github.com/SpongePowered/SpongeCommon/issues/2439 (need a test)

Also made sure this is not reintroducing https://github.com/SpongePowered/SpongeCommon/issues/1598 or https://github.com/SpongePowered/SpongeCommon/issues/1963 

Tests should cover the few regressions I could think of.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
+++ b/src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
@@ -65,13 +65,6 @@ public class ConfigurateTranslator implements DataTranslator<ConfigurationNode> 
         node.setValue(container.getMap(of()).get());
     }
 
-    private static DataContainer translateFromNode(ConfigurationNode node) {
-        checkNotNull(node, "node");
-        DataContainer dataContainer = DataContainer.createNew(DataView.SafetyMode.NO_DATA_CLONED);
-        ConfigurateTranslator.instance().addTo(node, dataContainer);
-        return dataContainer;
-    }
-
     @SuppressWarnings("unchecked")
     private static void translateMapOrList(ConfigurationNode node, DataView container) {
         Object value = node.getValue();
@@ -102,7 +95,10 @@ public class ConfigurateTranslator implements DataTranslator<ConfigurationNode> 
 
     @Override
     public DataContainer translate(ConfigurationNode obj) throws InvalidDataException {
-        return ConfigurateTranslator.translateFromNode(obj);
+        checkNotNull(obj, "node");
+        DataContainer dataContainer = DataContainer.createNew(DataView.SafetyMode.NO_DATA_CLONED);
+        ConfigurateTranslator.instance().addTo(obj, dataContainer);
+        return dataContainer;
     }
 
     @Override
--- a/src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
+++ b/src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
@@ -24,17 +24,22 @@
  */
 package org.spongepowered.common.data.persistence;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static org.spongepowered.api.data.DataQuery.of;
 
 import com.google.common.reflect.TypeToken;
 import ninja.leaping.configurate.ConfigurationNode;
 import ninja.leaping.configurate.SimpleConfigurationNode;
+import ninja.leaping.configurate.ValueType;
+import org.checkerframework.checker.nullness.qual.NonNull;
 import org.spongepowered.api.data.DataContainer;
+import org.spongepowered.api.data.DataQuery;
 import org.spongepowered.api.data.DataView;
 import org.spongepowered.api.data.persistence.DataTranslator;
 import org.spongepowered.api.data.persistence.InvalidDataException;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
@@ -65,18 +70,6 @@ public class ConfigurateTranslator implements DataTranslator<ConfigurationNode> 
         node.setValue(container.getMap(of()).get());
     }
 
-    @SuppressWarnings("unchecked")
-    private static void translateMapOrList(ConfigurationNode node, DataView container) {
-        Object value = node.getValue();
-        if (value instanceof Map) {
-            for (Map.Entry<Object, Object> entry : ((Map<Object, Object>) value).entrySet()) {
-                container.set(of('.', entry.getKey().toString()), entry.getValue());
-            }
-        } else if (value != null) {
-            container.set(of(node.getKey().toString()), value);
-        }
-    }
-
     public void translateContainerToData(ConfigurationNode node, DataView container) {
         ConfigurateTranslator.populateNode(node, container);
     }
@@ -103,14 +96,35 @@ public class ConfigurateTranslator implements DataTranslator<ConfigurationNode> 
 
     @Override
     public DataView addTo(ConfigurationNode node, DataView dataView) {
-        Object value = node.getValue();
-        Object key = node.getKey();
-        if (value != null) {
-            if (key == null || value instanceof Map || value instanceof List) {
-                translateMapOrList(node, dataView);
-            } else {
-                dataView.set(of('.', key.toString()), value);
+        if (node.hasMapChildren()) {
+            for (Map.Entry<?, ?> entry : ((Map<?, ?>) node.getValue()).entrySet()) {
+                final Object k = entry.getKey();
+                final ConfigurationNode n = node.getNode(k);
+                if (n.getValueType() == ValueType.MAP) {
+                    addTo(n, dataView.createView(of(k.toString())));
+                } else if (n.getValueType() == ValueType.LIST) {
+                    addTo(n, dataView);
+                } else {
+                    checkArgument(k != null, "map key");
+                    dataView.set(of(k.toString()), entry.getValue());
+                }
             }
+        } else if (node.hasListChildren()) {
+            final List<Object> l = new ArrayList<>();
+            for (ConfigurationNode o : node.getChildrenList()) {
+                if (o.getValueType().canHaveChildren()) {
+                    l.add(addTo(o, DataContainer.createNew()));
+                } else {
+                    l.add(o.getValue());
+                }
+            }
+            dataView.set(of(node.getKey().toString()), l);
+        } else {
+            final Object key = node.getKey();
+            final Object value = node.getValue();
+            checkArgument(key != null, "key");
+            checkArgument(value != null, "value");
+            dataView.set(DataQuery.of(key.toString()), node.getValue());
         }
         return dataView;
     }
--- a/src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
+++ b/src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
@@ -84,18 +84,8 @@ public class ConfigurateTranslator implements DataTranslator<ConfigurationNode> 
         }
     }
 
-    public ConfigurationNode translateData(DataView container) {
-        ConfigurationNode node = SimpleConfigurationNode.root();
-        translateContainerToData(node, container);
-        return node;
-    }
-
     public void translateContainerToData(ConfigurationNode node, DataView container) {
         ConfigurateTranslator.populateNode(node, container);
-    }
-
-    public DataContainer translateFrom(ConfigurationNode node) {
-        return ConfigurateTranslator.translateFromNode(node);
     }
 
     @Override
--- a/src/main/java/org/spongepowered/common/data/persistence/DataTranslatorTypeSerializer.java
+++ b/src/main/java/org/spongepowered/common/data/persistence/DataTranslatorTypeSerializer.java
@@ -44,7 +44,7 @@ public class DataTranslatorTypeSerializer<T> implements TypeSerializer<T> {
     @Override
     public T deserialize(TypeToken<?> type, ConfigurationNode value) throws ObjectMappingException {
         try {
-            return this.dataTranslator.translate(ConfigurateTranslator.instance().translateFrom(value));
+            return this.dataTranslator.translate(ConfigurateTranslator.instance().translate(value));
         } catch (InvalidDataException e) {
             // Since the value in the config node might be null, return null if an error occurs.
             return null;
--- a/src/main/java/org/spongepowered/common/data/persistence/HoconDataFormat.java
+++ b/src/main/java/org/spongepowered/common/data/persistence/HoconDataFormat.java
@@ -74,7 +74,7 @@ public class HoconDataFormat extends SpongeCatalogType implements StringDataForm
                 .setSource(source)
                 .build();
         CommentedConfigurationNode node = loader.load();
-        return ConfigurateTranslator.instance().translateFrom(node);
+        return ConfigurateTranslator.instance().translate(node);
     }
 
     @Override
@@ -98,7 +98,7 @@ public class HoconDataFormat extends SpongeCatalogType implements StringDataForm
         HoconConfigurationLoader loader = HoconConfigurationLoader.builder()
                 .setSink(sink)
                 .build();
-        ConfigurationNode node = ConfigurateTranslator.instance().translateData(data);
+        ConfigurationNode node = ConfigurateTranslator.instance().translate(data);
         loader.save(node);
     }
 
--- a/src/test/java/org/spongepowered/common/data/translator/ConfigurateDataViewTest.java
+++ b/src/test/java/org/spongepowered/common/data/translator/ConfigurateDataViewTest.java
@@ -24,19 +24,42 @@
  */
 package org.spongepowered.common.data.translator;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 
+import com.flowpowered.math.vector.Vector3d;
 import com.google.common.collect.Lists;
+import com.google.common.reflect.TypeToken;
 import ninja.leaping.configurate.ConfigurationNode;
 import ninja.leaping.configurate.SimpleConfigurationNode;
+import ninja.leaping.configurate.commented.SimpleCommentedConfigurationNode;
+import ninja.leaping.configurate.objectmapping.ObjectMappingException;
 import org.junit.Test;
 import org.spongepowered.api.data.DataContainer;
 import org.spongepowered.api.data.DataQuery;
 import org.spongepowered.api.data.DataView;
+import org.spongepowered.api.data.manipulator.mutable.entity.RespawnLocationData;
+import org.spongepowered.api.data.persistence.DataFormats;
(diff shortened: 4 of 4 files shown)


parlough (approved):

src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
```diff
@@ -75,64 +73,56 @@ private static DataContainer translateFromNode(ConfigurationNode node) {
         return dataContainer;
     }
 
-    private void translate(Map<Object, ? extends ConfigurationNode> node, DataView container, Map<?, ?> value) {
+    private void translate(Map<Object, ? extends ConfigurationNode> node, DataView container, Object key, Map<?, ?> value) {
```
WillBAnders: The `Object key` parameter is not used here, but is in the list variant.

In both cases, the key and value parameters are always the key and value of the node from what I see. Passing them as parameters is unnecessary.

src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
```diff
@@ -75,64 +73,56 @@ private static DataContainer translateFromNode(ConfigurationNode node) {
...
+                translate(n, sub);
+                container.set(of(String.valueOf(o.getKey())), sub);
+            }
         }
     }
 
-    private List<Object> translate(List<? extends ConfigurationNode> node, DataView container, List<?> value) {
+    private void translate(List<? extends ConfigurationNode> node, DataView container, Object key, List<?> value) {
         List<Object> list = new ArrayList<>(value.size());
         for (int i = 0; i < value.size(); i++) {
```
WillBAnders: This isn't affected by your change, but this for loop does not need an index - a simple foreach works.

src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
```diff
@@ -75,64 +73,56 @@ private static DataContainer translateFromNode(ConfigurationNode node) {
...
             }
         }
-        return list;
+        container.set(of(String.valueOf(key)), list);
     }
 
     private void translate(ConfigurationNode node, DataView container) {
         Object key = node.getKey();
         Object value = node.getValue();
         if (node.hasMapChildren()) {
```
WillBAnders: `if/else` for map, list, and normal value. Individual `if` conditions and `return` is complicating things here.

src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
```diff
@@ -75,64 +73,56 @@ private static DataContainer translateFromNode(ConfigurationNode node) {
...
-    public ConfigurationNode translateData(DataView container) {
-        ConfigurationNode node = SimpleConfigurationNode.root();
-        translateContainerToData(node, container);
-        return node;
+        if (node.hasListChildren()) {
+            translate(node.getChildrenList(), container, key, (List<?>) value);
+            return;
+        }
+
+        container.set(of('.', String.valueOf(key)), value);
```
WillBAnders: Configurate uses `.` as a separator itself, it will never be in the key of a node.

Additionally, I believe a test case for a root node holding only a value (such as `true`) will fail here as the key is null, which gets converted to a path consisting of the string `null`. This is the actual problem in the issue.

src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
```diff
@@ -75,64 +73,56 @@ private static DataContainer translateFromNode(ConfigurationNode node) {
...
         }
     }
 
-    private List<Object> translate(List<? extends ConfigurationNode> node, DataView container, List<?> value) {
+    private void translate(List<? extends ConfigurationNode> node, DataView container, Object key, List<?> value) {
         List<Object> list = new ArrayList<>(value.size());
         for (int i = 0; i < value.size(); i++) {
             ConfigurationNode n = node.get(i);
-            if (n.getValueType() == ValueType.SCALAR) {
+            if (!n.getValueType().canHaveChildren()) {
```
WillBAnders: This allows for null values, which can happen if they are explicitly set in the node. This will cause a NPE later on when set by the `DataContainer`. Given that nodes are likely coming from user input via configuration files, this case needs to be considered.

src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
```diff
@@ -75,64 +73,56 @@ private static DataContainer translateFromNode(ConfigurationNode node) {
...
+                list.add(sub);
             }
         }
-        return list;
+        container.set(of(String.valueOf(key)), list);
     }
 
     private void translate(ConfigurationNode node, DataView container) {
         Object key = node.getKey();
         Object value = node.getValue();
```
WillBAnders: Both key and value are used once, this should just be inlined. Using the dedicated methods like `getChildrenMap` will also remove the need for casting.

WillBAnders (changes requested):
I'm going to submit this as a request for changes as I don't believe this is ready yet. I've addressed a couple points and some formatting/style nitpicks, but I'll hit the main point below.

I do not think this is a proper solution to the issue. The main issue is that root notes have a null key, while others do not. For the purposes of translation, we need to treat *all* nodes as implicitly being 'root' so the path in the `DataContainer` is correct. This has a lot of changes to the structure, a fair bit of which is superfluous, and I would be worried about splash damage from this.

I believe the right change here is to add a `translateRoot` method as well as an `@Nullable Object key` parameter to `translate`. `translateRoot` would also use a null key so the node is seen as the root for the container, while inner calls can use the node's key. If the key is null (root), the value should be set directly on the container with an empty `DataQuery`.

src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
dualspiral: > Configurate uses . as a separator itself, it will never be in the key of a node.

That's wrong, it actually doesn't have a separator, which is why the path to a node is represented as `Object[]`. A `.` can actually be in the key of a node without being a separator. If you want an example of this, check Nucleus' `commands.conf` file.

> Additionally, I believe a test case for a root node holding only a value (such as true) will fail here as the key is null, which gets converted to a path consisting of the string null. This is the actual problem in the issue.

That is true ~, though I'm not sure the path that processes the root node will actually end up down here~. Quickly checking the code, it will, and this is a problem.

@ImMorpheus [The test case here](https://github.com/SpongePowered/SpongeCommon/pull/2353/files#diff-c5975971207a43e7feaecf528c5b4ba8R104) should be expanded to ensure that the end result is what we expect - that'll show that this is a problem.

Though, there iis an inherent problem here. What is the expected behaviour if I add a node that is a sub part of the tree (so the node has a key, say `a`) to a data view? Does that mean that the root view should have a single object `a`, or should it treat the node that you pass in as a root node always?

Javadocs might have the answer, I'm just being lazy here.

WillBAnders: Hmm, maybe this is outdated but I've always thought that calls to `getNode("this.that.other")` would split on the `.` automatically. Either way, I did forget about the case of quoted keys which could be used. However, I would say that taking the splitting on this is still not the right decision because the end result should match the structure of the node.

I was not able to find a javadoc specifying the behavior of child nodes for `DataView`, but I can try to explain the issues that would happen if the key were used. Because a `DataView` also has a path, we can link the structure of it to that to `ConfigurationNode`s and see what happens when serializing and deserializing.

When a value is set, that value is expected to be deserializeable along the given key. If I set a color to the node `my-color`, I expect something like `my-color { red=255, green=170, blue=0 }`. If I try to deserialize a color from this entire thing, it would fail - you need to get the value from the `my-color` node. If the key is included in the path during deserialization, I would get back a `DataContainer` containing `my-color { red=255, green=170, and blue=0}`. This `DataContainer` represents the color *value* (note that it's picked up an additional key that is user-specific), and is thus when set in a larger `DataView` under `my-color` the key becomes duplicated, resulting in `my-color { my-color { red=255, green=170, blue=0 } }`. Apologies if this was confusing but I tried xD

As additional example of this, [here is similar logic from Configurate](https://github.com/SpongePowered/Configurate/blob/master/configurate-core/src/main/java/ninja/leaping/configurate/SimpleConfigurationNode.java#L244) when setting a value to a `ConfigurationNode` - the key of the node is ignored and the value is set directly.

dualspiral: Okay, what I meant was what was the previous behaviour. That is, what is expected by plugins now? At this stage in the 7 cycle, it should be that, plus whatever is needed to fix the problems that were there.

WillBAnders: Ah, gotcha. My belief is the that the current behavior is inconsistent; it will deserialize to a different format than serialized from when using non-root nodes. The `populateNode` method, for example, is only ever called with a root. I will be able to write some tests later today to confirm this.

dualspiral:
Please note that I have backed out the previous commit that was meant to fix this due to data loss and serialisation issues. It will need to be reapplied to this tree for it to be mergable again (or just recreate the commit!)

You'll want to revert `41519ee5e2ab2771dee1db91ad6febb6e1215565`

ImMorpheus:
@WillBAnders I'm pretty sure the structure is fine, but if you find inconsistencies please write a reproducible test.

@dualspiral Expanding [this test](https://github.com/SpongePowered/SpongeCommon/pull/2353/files#diff-c5975971207a43e7feaecf528c5b4ba8R104) would be good, but I have no idea what's the expected behaviour.
The reason behind that test is to avoid a NPE (https://github.com/SpongePowered/SpongeCommon/issues/1963)
Excluding skipping the key-value (that would be difficult to debug), serializing the null key as "null" was the only thing that made sense to me.

dualspiral:
>  serializing the null key as "null" is the only thing that made sense to me.

If you serialise it as 'null', a null key will then be deserialised as that literal - you should be getting back what you put in, doing this means you don't. A `null` key is the root of a tree (except in lists, where each `ConfigurationNode` is also classed as the root of a tree - it's a list of trees).

Essentially - a `null` key -> `DataQuery.of()`, not `DataQuery.of("null")`.

ImMorpheus:
> If you serialise it as 'null', a null key will then be deserialised as that literal - you should be getting back what you put in, doing this means you don't.  

I can understand that

> Essentially - a null key -> DataQuery.of(), not DataQuery.of("null").

yeah, that's the issue. You can't do `DataQuery.of()`, not as a valid path inside a container at least.
`container.set(DataQuery.of(), value);` would throw an IllegalStateException.

If we want this level of consistency we should throw an exception too when translating a node with null keys

In any case, what's the expected result of `ConfigurateTranslator.instance().translate(SimpleConfigurationNode.root().setValue("test"));` ?

dualspiral:
Hmmm, that is a good point - how was it handled before?

ImMorpheus:
It wasn't. That line of code would throw a NPE.

src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
ImMorpheus: Added a final modifier, IMO this is clearer.

> Using the dedicated methods like getChildrenMap will also remove the need for casting.

I disagree, `getChildrenMap` and `getValue` are different

src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
ImMorpheus: Can you create a test to reproduce the issue ?

I'd like to address the lack of unit tests

dualspiral:
I just had a look at the NBT format and it’s the same as the DataContainers, you can’t have an object at the root. So, `DataQuery.of()` must only be a root map or list - and an exception should be thrown if a scale root node is passed. It’s a bit murkier for lists as it can be a list of scalars, but then why would you use a set of configuration nodes if you are storing doubles? So, all root nodes must be maps or lists.

So, to answer your question, I think it should be an exception, and documented as such. Perhaps not an NPE, probably an IAE in this case, along with an appropriate message.

src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
```diff
@@ -72,32 +75,52 @@ private static DataContainer translateFromNode(ConfigurationNode node) {
...
+                    DataContainer sub = DataContainer.createNew(DataView.SafetyMode.NO_DATA_CLONED);
+                    translate(n, sub);
+                    list.add(sub);
+                }
+            }
+            container.set(of(String.valueOf(key)), list);
+            return;
+        }
+
+        checkArgument(key != null, "Cannot use null key");
```
ImMorpheus: input wanted for an `appropriate message`

ImMorpheus:
Superseded by https://github.com/SpongePowered/SpongeCommon/pull/2603

PR #2358, opened 2019-08-21
Implement New Explosion Settings

Pulverizer:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2041) | **SpongeCommon**

Three new variables in ExplosionMixin and the code required for implementation.

All variables to be used in the newer versions of my plugin.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/ExplosionMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/ExplosionMixin.java
@@ -82,9 +82,9 @@ public abstract class ExplosionMixin implements ExplosionBridge {
     private boolean impl$shouldBreakBlocks;
     private boolean impl$shouldDamageEntities;
 //    private Cause createdCause;
-    private int impl$boundingBoxSize;
-    private boolean impl$reducedRandomness;
-    private double impl$entityKnockbackMultiplier;
+    private int impl$resolution;
+    private float impl$randomness;
+    private double impl$knockback;
 
     @Shadow @Final private List<BlockPos> affectedBlockPositions;
     @Shadow @Final private Map<EntityPlayer, Vec3d> playerKnockbackMap;
@@ -106,6 +106,9 @@ public abstract class ExplosionMixin implements ExplosionBridge {
         // Sponge-created explosions will explicitly set 'impl$shouldBreakBlocks' to its proper value
         this.impl$shouldBreakBlocks = this.damagesTerrain;
         this.impl$shouldDamageEntities = true;
+        this.impl$resolution = 16;
+        this.impl$randomness = 1.0F;
+        this.impl$knockback = 1.0;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/common/mixin/core/world/ExplosionMixin.java
```diff
@@ -116,22 +118,24 @@ private void onConstructed(final net.minecraft.world.World world, final Entity e
...
+                            double d0 = (double) ((float) j / (float)(impl$boundingBoxSize - 1) * 2.0F - 1.0F);
+                            double d1 = (double) ((float) k / (float)(impl$boundingBoxSize - 1) * 2.0F - 1.0F);
+                            double d2 = (double) ((float) l / (float)(impl$boundingBoxSize - 1) * 2.0F - 1.0F);
                             final double d3 = Math.sqrt(d0 * d0 + d1 * d1 + d2 * d2);
                             d0 = d0 / d3;
                             d1 = d1 / d3;
                             d2 = d2 / d3;
-                            float f = this.size * (0.7F + this.world.rand.nextFloat() * 0.6F);
+                            float f = this.size * (0.7F + impl$random * 0.6F);
+                            if (!impl$reducedRandomness)
```
dualspiral: Make sure any if statements (or any control blocks) are wrapped in braces please. We don't want Sponge to reprise [goto fail;](https://www.imperialviolet.org/2014/02/22/applebug.html)...

WillBAnders:
My primary concern here is that the behaviors modified here are arbitrary. There are plenty of ways explosions can be modified, and even the changes here can be done in different ways to obtain different results. Here are some of the weird edge cases that can happen with these values:

 - A resolution `<=0` disables the explosion
 - A resolution of `1` causes a `DivideByZeroException`
 - High randomness can cause `f` (which the wiki calls `intensity`) to go negative. This doesn't cause any direct problems that I can see though.
 - If knockback is `0`, knockback will not happen (which IMO is useful behavior to have)
 - Negative knockback values will effectively 'suck' the player towards the explosion. As cool as this sounds, don't think we should be doing this.

It was expressed on Discord that Bukkit used to have a plugin for customized explosions like this. Personally, this is the route I would recommend - all of these changes sound like very specific customization. I think it's best to handle this in a private project so it can be tailored to your needs without it being locked into Sponge.

XakepSDK:
I think we should leave knockback in API and move new settings to SpongeCommon explosion builder implementation.

Pulverizer:
@WillBAnders Thanks for pointing out the bugs. Guess I'll have to fix the first two before merging. However, a dev I spoke to on discord liked the idea of being able to create an implosion. 

In regards to being a separate project, are you suggesting I create a competing API or something because, to my understanding, it isn't possible to override a mixin with the `@Final` tag? As you said, bukkit could do it.

ryantheleach:
Yeah, implosions sound cool.

JBYoshi:
Tested and works as expected.

PR #2370, opened 2019-09-07
Add ability to change potion color

RedNesto:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2045) | **SpongeCommon**

### DIFF
--- a/src/main/java/org/spongepowered/common/data/DataRegistrar.java
+++ b/src/main/java/org/spongepowered/common/data/DataRegistrar.java
@@ -571,6 +571,9 @@ public final class DataRegistrar {
         DataUtil.registerDualProcessor(FluidItemData.class, SpongeFluidItemData.class, ImmutableFluidItemData.class,
                 ImmutableSpongeFluidItemData.class, new FluidItemDataProcessor());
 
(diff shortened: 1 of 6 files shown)


JBYoshi (approved):

ImMorpheus:
(probably related to https://github.com/SpongePowered/SpongeAPI/issues/1829#issuecomment-390369707)

Pinging @gabizou since I'm not sure if creating DataManipulators to edit NBT data is the route we want to take.

RedNesto:
@ImMorpheus I'm not sure if I understand you correctly, how would the implementation work without modifying the NBT data? Or maybe you mixed up DataManipulator and DataProcessor? 

Also I based this processor's implementation on [another potion data processor](https://github.com/SpongePowered/SpongeCommon/blob/8302b133ad54ce5fbbf20ef882e6977f4dfb4ba2/src/main/java/org/spongepowered/common/data/processor/data/item/ItemPotionEffectDataProcessor.java#L58-L72) so I guess this is fine to do so.

ImMorpheus:
We would modify the NBT data. What I'm saying is that I don't know if we plan to map nbt data to data processors. I was thiking of adding a key (to provide the query) without any data manipulator or data processor.
You can then rely on `DataHolder#setRawData` to change NBT data.

gabizou (approved):
Looks good.

PR #2373, opened 2019-09-08
Fix/move event

ImMorpheus:
MoveEntityEvent could be bypassed if `CPacketPlayer.PositionRotation` was sent since it has both `moving` and `rotating` field set to true and sponge was using 

`boolean positionOnly = packetIn.moving && !packetIn.rotating;`
`boolean rotationOnly = !packetIn.moving && packetIn.rotating;`

and it would fire an event only `if (rotationOnly || positionOnly)`

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
@@ -403,12 +403,16 @@ public abstract class NetHandlerPlayServerMixin implements NetHandlerPlayServerB
             if (this.impl$lastMoveLocation != null) {
                 fromPosition = this.impl$lastMoveLocation.getPosition();
             }
+
+            // if it's not a "moving packet", use the fromPosition so that the delta is 0
             final Vector3d toPosition = packetIn.moving ? Vector3d.from(packetIn.x, packetIn.y, packetIn.z) : fromPosition;
 
             final double deltaSquared = toPosition.distanceSquared(fromPosition);
 
 
             final Vector3d fromRotation = player.getRotation();
+
+            // if it's not a "rotation packet", use the fromRotation so that the delta is 0
             final Vector3d toRotation = packetIn.rotating ? Vector3d.from(packetIn.pitch, packetIn.yaw, 0) : fromRotation;
 
             final double deltaAngleSquared = fromRotation.distanceSquared(toRotation);
@@ -423,6 +427,8 @@ public abstract class NetHandlerPlayServerMixin implements NetHandlerPlayServerB
                 final Transform<World> fromTransform = player.getTransform().setPosition(fromPosition).setRotation(fromRotation);
                 Transform<World> toTransform = player.getTransform().setPosition(toPosition).setRotation(toRotation);
                 final Transform<World> originalToTransform = toTransform;
+
+                // RotateEntityEvent is fired when a player performs a rotation without moving their position.
                 if (rotating && ShouldFire.ROTATE_ENTITY_EVENT && !moving) {
                     final RotateEntityEvent event = SpongeEventFactory.createRotateEntityEvent(Sponge.getCauseStackManager().getCurrentCause(), fromTransform, toTransform, player);
                     if (SpongeImpl.postEvent(event)) {
@@ -431,6 +437,8 @@ public abstract class NetHandlerPlayServerMixin implements NetHandlerPlayServerB
                         toTransform = event.getToTransform();
                     }
                 }
+
+                // MoveEntityEvent.Position needs to be fired even if you move while rotating
                 if (moving && ShouldFire.MOVE_ENTITY_EVENT_POSITION) {
                     final MoveEntityEvent.Position event = SpongeEventFactory.createMoveEntityEventPosition(Sponge.getCauseStackManager().getCurrentCause(), fromTransform, toTransform, player);
                     if (SpongeImpl.postEvent(event)) {
--- a/src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
@@ -25,6 +25,7 @@
 package org.spongepowered.common.mixin.core.network;
 
 import com.flowpowered.math.vector.Vector3d;
+import com.flowpowered.math.vector.Vector3i;
 import io.netty.util.collection.LongObjectHashMap;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.entity.Entity;
@@ -399,13 +400,13 @@ public abstract class NetHandlerPlayServerMixin implements NetHandlerPlayServerB
             final Vector3d fromRotation = player.getRotation();
 
             // If Sponge used the player's current location, the delta might never be triggered which could be exploited
-            Location<World> fromLocation = player.getLocation();
+            Vector3d fromPosition = player.getPosition();
             if (this.impl$lastMoveLocation != null) {
-                fromLocation = this.impl$lastMoveLocation;
+                fromPosition = this.impl$lastMoveLocation.getPosition();
             }
 
-            Location<World> toLocation = new Location<>(player.getWorld(), packetIn.x, packetIn.y, packetIn.z);
-            Vector3d toRotation = new Vector3d(packetIn.pitch, packetIn.yaw, 0);
+            Vector3d toPosition = Vector3d.from(packetIn.x, packetIn.y, packetIn.z);
+            Vector3d toRotation = Vector3d.from(packetIn.pitch, packetIn.yaw, 0);
 
             // Minecraft does the same with rotation when it's only a positional update
             boolean positionOnly = packetIn.moving && !packetIn.rotating;
@@ -413,7 +414,7 @@ public abstract class NetHandlerPlayServerMixin implements NetHandlerPlayServerB
                 // Correct the new rotation to match the old rotation
                 toRotation = fromRotation;
 
-                positionOnly = !toLocation.getPosition().equals(fromLocation.getPosition()) && ShouldFire.MOVE_ENTITY_EVENT_POSITION;
+                positionOnly = !toPosition.equals(fromPosition) && ShouldFire.MOVE_ENTITY_EVENT_POSITION;
             }
 
             // Minecraft sends a 0, 0, 0 position when rotation only update occurs, this needs to be recognized and corrected
@@ -422,27 +423,29 @@ public abstract class NetHandlerPlayServerMixin implements NetHandlerPlayServerB
             if (rotationOnly) {
                 // Correct the to location so it's not misrepresented to plugins, only when player rotates without moving
                 // In this case it's only a rotation update, which isn't related to the to location
-                fromLocation = player.getLocation();
-                toLocation = fromLocation;
+                fromPosition = player.getLocation().getPosition();
+                toPosition = fromPosition;
 
                 rotationOnly = ShouldFire.ROTATE_ENTITY_EVENT;
             }
 
             if (packetIn.moving && packetIn.rotating) {
-                positionOnly = !toLocation.getPosition().equals(fromLocation.getPosition()) && ShouldFire.MOVE_ENTITY_EVENT_POSITION;
+                positionOnly = !toPosition.equals(fromPosition) && ShouldFire.MOVE_ENTITY_EVENT_POSITION;
                 rotationOnly = ShouldFire.ROTATE_ENTITY_EVENT;
             }
 
-            mixinPlayer.bridge$setVelocityOverride(toLocation.getPosition().sub(fromLocation.getPosition()));
+            mixinPlayer.bridge$setVelocityOverride(toPosition.sub(fromPosition));
 
-            final double deltaSquared = toLocation.getPosition().distanceSquared(fromLocation.getPosition());
+            final double deltaSquared = toPosition.distanceSquared(fromPosition);
             final double deltaAngleSquared = fromRotation.distanceSquared(toRotation);
+            final boolean blockChanged = !Vector3i.from((int) fromPosition.getZ(), (int) fromPosition.getY(), (int) fromPosition.getZ())
+                    .equals(Vector3i.from((int) packetIn.x, (int) packetIn.y, (int) packetIn.z));
 
             // These magic numbers are sad but help prevent excessive lag from this event.
             // eventually it would be nice to not have them
-            if (deltaSquared > ((1f / 16) * (1f / 16)) || deltaAngleSquared > (.15f * .15f)) {
-                final Transform<World> fromTransform = player.getTransform().setLocation(fromLocation).setRotation(fromRotation);
-                Transform<World> toTransform = player.getTransform().setLocation(toLocation).setRotation(toRotation);
+            if (deltaSquared > ((1f / 16) * (1f / 16)) || deltaAngleSquared > (.15f * .15f) || blockChanged) {
+                final Transform<World> fromTransform = player.getTransform().setPosition(fromPosition).setRotation(fromRotation);
+                Transform<World> toTransform = player.getTransform().setPosition(toPosition).setRotation(toRotation);
                 final Transform<World> originalToTransform = toTransform;
                 if (rotationOnly || positionOnly) {
                     final Event event;
@@ -453,7 +456,7 @@ public abstract class NetHandlerPlayServerMixin implements NetHandlerPlayServerB
                     }
                     if (SpongeImpl.postEvent(event)) {
                         ((EntityBridge) mixinPlayer).bridge$setLocationAndAngles(fromTransform);
-                        this.impl$lastMoveLocation = fromLocation;
+                        this.impl$lastMoveLocation = fromTransform.getLocation();
                         mixinPlayer.bridge$setVelocityOverride(null);
                         return true;
                     }
--- a/src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
@@ -423,7 +423,7 @@ public abstract class NetHandlerPlayServerMixin implements NetHandlerPlayServerB
                 final Transform<World> fromTransform = player.getTransform().setPosition(fromPosition).setRotation(fromRotation);
                 Transform<World> toTransform = player.getTransform().setPosition(toPosition).setRotation(toRotation);
                 final Transform<World> originalToTransform = toTransform;
-                if (rotating && ShouldFire.ROTATE_ENTITY_EVENT && !packetIn.moving) {
+                if (rotating && ShouldFire.ROTATE_ENTITY_EVENT && !moving) {
                     final RotateEntityEvent event = SpongeEventFactory.createRotateEntityEvent(Sponge.getCauseStackManager().getCurrentCause(), fromTransform, toTransform, player);
                     if (SpongeImpl.postEvent(event)) {
                         toTransform.setRotation(fromTransform.getRotation());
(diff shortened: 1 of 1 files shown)


JBYoshi (approved):
Looks good from what I can see.

src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
```diff
@@ -396,73 +397,49 @@ private boolean throwMoveEvent(final EntityPlayerMP playerMP, final CPacketPlaye
...
+            final Vector3d toRotation = packetIn.rotating ? Vector3d.from(packetIn.pitch, packetIn.yaw, 0) : fromRotation;
 
-            final double deltaSquared = toLocation.getPosition().distanceSquared(fromLocation.getPosition());
             final double deltaAngleSquared = fromRotation.distanceSquared(toRotation);
 
+            mixinPlayer.bridge$setVelocityOverride(toPosition.sub(fromPosition));
+
             // These magic numbers are sad but help prevent excessive lag from this event.
             // eventually it would be nice to not have them
-            if (deltaSquared > ((1f / 16) * (1f / 16)) || deltaAngleSquared > (.15f * .15f)) {
```
dualspiral, line 443: ~~16 -> 16f, this was otherwise returning zero. I'd be tempted to make these numbers named constants.~~

no, ignore me, one flat is enough, doesn't need to be the denominator.

dualspiral (changes requested):
Did you test it could be bypassed before doing this? In the current iteration, it resets the positionOnly/rotationOnly variables if a packet comes in containing both:

https://github.com/SpongePowered/SpongeCommon/blob/620cc8a9e6e12610377a53d82de6aa83a33e84b8/src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java#L439-L442

`positionOnly` and `rotationOnly` can both be true at this point. It therefore handled `CPacketPlayer.PositionRotation` fine.

I see what you've done in making rotate fire independently of the move event too. The previous behaviour was intentional, that the rotate event is only fired if the move event is not - there were discussions about that earlier this year. I think this was done for performance reasons. [Indeed, see the RotateEntityEvent javadocs](https://github.com/SpongePowered/SpongeAPI/blob/stable-7/src/main/java/org/spongepowered/api/event/entity/RotateEntityEvent.java#L33-L36) - it says:

```java
/**
 * Called when an {@link Entity} performs rotation of their body or, if
 * applicable, head without moving their position.
 */
```

So what you've done here is wrong - and there is no bug to fix with this. If you want to clean up the method, then sure, but it must act as it did before. ~~.. with one exception.~~

src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
```diff
@@ -396,73 +397,49 @@ private boolean throwMoveEvent(final EntityPlayerMP playerMP, final CPacketPlaye
...
+            final boolean rotating = deltaAngleSquared > (.15f * .15f);
+            if (moving || rotating) {
+                final Transform<World> fromTransform = player.getTransform().setPosition(fromPosition).setRotation(fromRotation);
+                Transform<World> toTransform = player.getTransform().setPosition(toPosition).setRotation(toRotation);
                 final Transform<World> originalToTransform = toTransform;
-                if (rotationOnly || positionOnly) {
-                    final Event event;
-                    if (rotationOnly) {
-                        event = SpongeEventFactory.createRotateEntityEvent(Sponge.getCauseStackManager().getCurrentCause(), fromTransform, toTransform, player);
+                if (rotating && ShouldFire.ROTATE_ENTITY_EVENT && !packetIn.moving) {
```
dualspiral: This introduces a bug. It should not fire if `moving` is `false` rather than this - you can get this packet but it not be in the tolerances. What this now does is not fire an event if `rotating` is `true` (rotation is large enough) but `moving` is `false` - if rotation is true but moving is false, then this should still fire.

dualspiral (changes requested):
I'm going to be honest, I'm not seeing the value in this PR. The previous behaviour was fine and while I get that you were attempting to fix what you saw as a bypass, it actually wasn't (as I've pointed out in a previous review) - the logic looks to me that it is absolutely fine and by changing it, we're losing comments explaining why we're doing what we are and it's just making things less clear.

I very much think this is a case of "if it ain't broke, don't fix it".

I'd like @Zidane, as the person who previously wrote this logic in (based on commit history), to review this.

ImMorpheus:
I'll explain this a bit more in detail since I probably took a few things for granted (and I did a poor job at explaining the issue).

Code to reproduce the issue:

```
    @SubscribeEvent //this is only fired on the client, which is why I'm using this event.
    public void onRightClickEmpty(PlayerInteractEvent.RightClickEmpty event) {
        EntityPlayer player = event.getEntityPlayer();
        CPacketPlayer.PositionRotation packet = new CPacketPlayer.PositionRotation(player.posX + 1, player.posY, player.posZ, 0.1f, 0.1f, true); 
        Minecraft.getMinecraft().getConnection().sendPacket(packet);
    }

    @Listener(beforeModifications = true)
    public void onMove(MoveEntityEvent.Position event) {
        event.setCancelled(true);
    }
```

You wouldn't expect a player to be able to move. 
note: this can also be bypassed by rotating while moving, but not in a reproducible way since some packets are handled just fine (moving true, rotating false).

Let's use the code snippet:

1. put a breakpoint here https://github.com/SpongePowered/SpongeCommon/blob/a00f52d6a59bfa89976f7257eae11e2565267b97/src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java#L447

2. *rightclick in air*

3. Both `rotationOnly` and `positionOnly` (the names don't help much, since the logic is wrong) are true, only RotateEventityEvent is fired. 

https://github.com/SpongePowered/SpongeCommon/blob/a00f52d6a59bfa89976f7257eae11e2565267b97/src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java#L447-L454

4. The player moves happily

So what's happening is:

- MoveEntityEvent is not fired even if the played moved 
- RotateEntityEvent is also fired (this is breaking the API contract)


Now, to the reviews:

> Did you test it could be bypassed before doing this?

yes :(

> It therefore handled CPacketPlayer.PositionRotation fine.

No, it didn't. See the explanation above, this is only happening with this type of `CPacketPlayer`.

> I see what you've done in making rotate fire independently of the move event too. The previous behaviour was intentional, that the rotate event is only fired if the move event is not - there were discussions about that earlier this year.

Thanks for linking the API doc, I wasn't aware of how RotateEntityEvent was documented.
However, you're contradicting yourself in stating the previous behaviour was intentional since (as I explained above) a RotateEntityEvent is fired when a player moves (with that specific packet). Either the previous behaviour was intentional (and the API doc is wrong) or the API doc (`rotation without moving their position`) is correct in stating a RotateEntityEvent is fired when there's no movement (and the previous behaviour was a bug).

> I'm going to be honest, I'm not seeing the value in this PR.

:(

> The previous behaviour was fine and while I get that you were attempting to fix what you saw as a bypass, it actually wasn't (as I've pointed out in a previous review) - the logic looks to me that it is absolutely fine

> I very much think this is a case of "if it ain't broke, don't fix it".

No, it wasn't (see above).

> by changing it, we're losing comments explaining why we're doing what we are and it's just making things less clear.

I'm interested in this, can you point out what part is unclear ? Adding comments would help future maintenance.

> I'd like Zidane, as the person who previously wrote this logic in (based on commit history), to review this.

I'm fine with other devs reviewing this (that's why I didn't push the commit directly), but the original author of the code is @modwizcode, thus I'm also pinging him to leave a comment.

> This introduces a bug. It should not fire if moving is false rather than this - you can get this packet but it not be in the tolerances. What this now does is not fire an event if rotating is true (rotation is large enough) but moving is false - if rotation is true but moving is false, then this should still fire.

Ah I see, will fix that

dualspiral:
Right, I see the bug now. The real issue was not that `positionOnly` and `rotationOnly` were mutally exclusive, they weren't, as I stated. The issue was that if both `positionOnly` and `rotationOnly` were `true`, the _rotation_ event took precidence, not the movement one - which is not what the original comment states.

In other words, to fix this,

https://github.com/SpongePowered/SpongeCommon/blob/a00f52d6a59bfa89976f7257eae11e2565267b97/src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java#L449-L453

needed to be rewritten to give position precidence (if (positionOnly) moveEvent else rotationEvent). 

I don't mind the rewrite, but the original comment is incorrect and this has caused the confusion.

dualspiral:
> I'm interested in this, can you point out what part is unclear ? Adding comments would help future maintenance.

I missed this part. Basically, any comment you removed from the code (such as https://github.com/SpongePowered/SpongeCommon/pull/2373/files#diff-5aa358b1827289e5a5d8e580204fc2bfL410). Some of them provide info about what Minecraft does and would be valuable to keep.

Lunaphied:
I'll defer to @Zidane on this as I'm not currently an active contributor, but for input as the one who wrote the original fix for the non-intuitive way minecraft sometimes sends packets, I think that all that really needs to be done here is make sure the movement event gets prioritized if both movement and rotation occur, as it seems I didn't test for that case. 

I'm not really seeing the point of a lot of the other modifications you've made here, and while you've added some comments, they don't really give the depth or any detail that the original comments documented about the internals of the way the game handles different packets. 

If you want to clean up this PR by reducing it to the minimal set of changes from the existing code, that would probably be the best way to get this fix in quickly.

ImMorpheus:
> I'll defer to Zidane on this as I'm not currently an active contributor, but for input as the one who wrote the original fix for the non-intuitive way minecraft sometimes sends packets, I think that all that really needs to be done here is make sure the movement event gets prioritized if both movement and rotation occur, as it seems I didn't test for that case.
> 
> I'm not really seeing the point of a lot of the other modifications you've made here, and while you've added some comments, they don't really give the depth or any detail that the original comments documented about the internals of the way the game handles different packets.


I'm assuming you are talking about `// Minecraft sends a 0, 0, 0 position when rotation only update occurs, this needs to be recognized and corrected` since that's the only comment which has effectively been removed (contrary to other comments being rephrased).

The reason is I didn't find anything wrong with how Minecraft sends moving/rotation packets. `CPacketPlayer`'s fields for positional updates (x, y, z and yaw, pitch) are primitive types (double and float). They can't be null, so they are using boolean flags (`moving` and `rotating`) to indicate a rotation/movement is taking place.
What I'm saying is that Minecraft does not send a `0, 0, 0 position when rotation only update occurs`, those are the values defaulted to the fields when they are initialized.
You shouldn't be looking at them at all if `moving` is false. 

The whole `positionOnly`/`rotationOnly` concept looks messy to me (it kinda gives the wrong impression `CPacketPlayer.PositionRotation` is not handled (even if I change priority)). Making the method clearer was also a (secondary) goal of this, but this may be a bias since I'm the author.

Lunaphied:
I seem to remember that there edge cases where minecraft's flags did not match the behavior that was actually occurring but i worked on this code a long time ago now so I could indeed be wrong. It did seem like a few other lines were superfluous changes that didn't really need to be changed and made the changelog more obscure. 

That comment explains that the code has to specifically handle 0,0,0 as an edge case where minecraft sometimes sends a position update but it's actually a rotation update (afaik this is one of the things I was fixing), and also why we can't just accept the location directly from the packet as would seem obvious. 

Like I said, maybe this behavior was changed. This is just my two cents.

ImMorpheus:
> I seem to remember that there edge cases where minecraft's flags did not match the behavior that was actually occurring 

I will investigate that, it sounds like a bug.

src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
```diff
@@ -396,73 +397,57 @@ private boolean throwMoveEvent(final EntityPlayerMP playerMP, final CPacketPlaye
...
+                    final MoveEntityEvent.Position event = SpongeEventFactory.createMoveEntityEventPosition(Sponge.getCauseStackManager().getCurrentCause(), fromTransform, toTransform, player);
                     if (SpongeImpl.postEvent(event)) {
-                        ((EntityBridge) mixinPlayer).bridge$setLocationAndAngles(fromTransform);
-                        this.impl$lastMoveLocation = fromLocation;
-                        mixinPlayer.bridge$setVelocityOverride(null);
-                        return true;
-                    }
-                    if (rotationOnly) {
-                        toTransform = ((RotateEntityEvent) event).getToTransform();
+                        toTransform.setPosition(fromTransform.getPosition());
```
dualspiral, line 445: This even might fire for a position and rotation update - so you need to set the rotation too - rather, it is better just to use:

```java
toTransform = fromTransform
```

That way, you're avoiding an object creation altogether.

(EDIT: In fact, as is, this is effectively a no-op. As I said in another comment, a `Transform` is immutable - so you should be setting the variable anyway - `setPosition` doesn't affect the `Transform` you are operating on. [See this javadoc](https://jd.spongepowered.org/7.1.0/org/spongepowered/api/entity/Transform.html#setPosition-com.flowpowered.math.vector.Vector3d-))

It also looks to me that you were saying that it should then fall into the if statement below (`(!toTransform.equals(originalToTransform))`) - the assumption here is that `fromTransform != originalToTransform`. While that is always true (at least, right now, the deltas take care of making sure `from != to`), that's not obvious at a glance - and is where a comment should be made explaining what is going on.

It took me some time to consider this situation and whether what you had done is valid.

src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
```diff
@@ -396,73 +397,57 @@ private boolean throwMoveEvent(final EntityPlayerMP playerMP, final CPacketPlaye
...
             final double deltaAngleSquared = fromRotation.distanceSquared(toRotation);
 
+            mixinPlayer.bridge$setVelocityOverride(toPosition.sub(fromPosition));
+
             // These magic numbers are sad but help prevent excessive lag from this event.
             // eventually it would be nice to not have them
-            if (deltaSquared > ((1f / 16) * (1f / 16)) || deltaAngleSquared > (.15f * .15f)) {
-                final Transform<World> fromTransform = player.getTransform().setLocation(fromLocation).setRotation(fromRotation);
-                Transform<World> toTransform = player.getTransform().setLocation(toLocation).setRotation(toRotation);
+            final boolean moving = deltaSquared > ((1f / 16) * (1f / 16));
```
dualspiral, line 424: Could we get away with not calculating the deltas if we know there is no movement or rotation?

src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
```diff
@@ -396,73 +397,57 @@ private boolean throwMoveEvent(final EntityPlayerMP playerMP, final CPacketPlaye
...
+            final boolean rotating = deltaAngleSquared > (.15f * .15f);
+            if (moving || rotating) {
+                final Transform<World> fromTransform = player.getTransform().setPosition(fromPosition).setRotation(fromRotation);
```
dualspiral, line 427: I know that this is in the original code, but as `Transform` is immutable, each step here creates a new `Transform`. Can't we just create a `new Transform<>(...)` instead?

src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
```diff
@@ -396,73 +397,57 @@ private boolean throwMoveEvent(final EntityPlayerMP playerMP, final CPacketPlaye
...
+                Transform<World> toTransform = player.getTransform().setPosition(toPosition).setRotation(toRotation);
```
dualspiral, line 428: As above.

dualspiral (changes requested):
I've taken another look through the code. If we are going to gut and rewrite part of the method to make it clear - we should also make it more effecient and make it clear as to what is going on. There is also a bug in here that would allow for rotation to occur even if the `MoveEntityEvent` is cancelled.

This is why I've been pushing against such big changes. I get your point about the `*Only` variables being confusing - we could just rename them. I get that you feel this is more logical. However, this has felt like a sledgehammer to crack a walnut - apart from the fact that the original issue was misdiagnosed, big changes like this introduce bugs. One has already been fixed, I've spotted another, in a comment below. 

However, the fact of the matter is that the only **logic** that was broken was that the rotation event was firing instead of the move event when the `PositionRotation` packet was recieved. [This is all that is really needed to fix the bug at hand](https://gist.github.com/dualspiral/07cdf856e3392186a2085374d8c0d1fb). I have tested it locally, ensured that position and rotation packets fire the move event and not the rotation event (with `sponge.shouldFireAll` set to `true`) and it works. I am considering commiting that patch to the tree so the fix is in and there is a new baseline to work against. 

I pass no judgement as to whether this should then follow - indeed, it might make it easier to update to 1.14. I just think we should do this in two steps, fix the original issue, then consider the refactor.

ImMorpheus:
> I've taken another look through the code. If we are going to gut and rewrite part of the method to make it clear - we should also make it more effecient and make it clear as to what is going on. 

Good, I opened a PR because I was looking for suggestions.

> There is also a bug in here that would allow for rotation to occur even if the MoveEntityEvent is cancelled.

I will look into it.

> I get your point about the *Only variables being confusing - we could just rename them. 

Ok ? They are being reassigned a couple of times and mixed with `ShouldFire`s flags, I don't think we can just rename them that easily (not without changing the flow of the method).

> This is why I've been pushing against such big changes. I get that you feel this is more logical. However, this has felt like a sledgehammer to crack a walnut - apart from the fact that the original issue was misdiagnosed, big changes like this introduce bugs. One has already been fixed, I've spotted another, in a comment below.
> 
> However, the fact of the matter is that the only logic that was broken was that the rotation event was firing instead of the move event when the PositionRotation packet was recieved. This is all that is really needed to fix the bug at hand. I have tested it locally, ensured that position and rotation packets fire the move event and not the rotation event (with sponge.shouldFireAll set to true) and it works.

That's a slippery slope and I find your last statement ironic - apart from the fact the you stated the previous behaviour was fine, your patch has also completely broken the contract with the API (**in some cases**). You can reproduce the behaviour with a listener for `RotateEntityEvent` (make sure there are no listeners for `MoveEntityEvent.Position` so that `ShouldFire#MOVE_ENTITY_EVENT_POSITION` is false), you will start seeing `RotateEntityEvent` being fired even when the position changed (contrary to what is stated in the api `Called when an Entity performs rotation of their body or, if applicable, head without moving their position.`)

> I am considering commiting that patch to the tree so the fix is in and there is a new baseline to work against.

I wouldn't mind. As a matter of fact I was thinking of closing this PR.

dualspiral:
> That's a slippery slope and I find your last statement ironic 

This entire PR has been a slippery slope to be honest. We have butted heads because of numerous problems:

a) The PR description did not actually describe the bug correctly. The original assumption was false, and I based my logic on them.
b) I didn't think there was a problem because I based my analysis on your description of what was going wrong. It took you actually explaining what you'd seen which made me realise what is wrong.
c) I _agree_ that I got it slightly wrong. Referring to the comment:

> apart from the fact the you stated the previous behaviour was fine

I admit that I was wrong, but I was also working against what was stated at the time and I should have looked into it in greater depth. See a).
d) Any big rewrite will likely incur issues. Bugs were introduced in this rewrite and by running it through my IDE I unconvered some significant ones. API contracts were broken or not fully understood - and it seems I have fallen into that trap too.
e) The reason I asked for @Zidane was because he _was_ the original author of the code in question. The rotation event was added by blood, [Zidane added a fix that resulted in the current behaviour](https://github.com/SpongePowered/SpongeCommon/commit/8d035f1a6842ef6729911aa2e3ee8e2894ac966a).

No-one is faultless here. I apologise for my part in this. I should have been more careful in my initial analysis.

> your patch has also completely broken the contract ith the API (in some cases)

Good catch. Easily enough solved.

Issue #2372, opened 2019-09-08
commands showing up in /help for players without permission to use


bhi-elliot:
**I am currently running**

- SpongeForge version: 1.12.2 2838 7.1.7 RC 3910
- Java version: 1.8.0_221-b11
- Operating System: Windows 10

Sponge plugins:

GriefPrevention 4.3.0.692
KaroglanSignEditor 3.4.1
LuckPerms Sponge 4.4.29
Nucleus 1.14.0 S7.1
project core 7.2.0 0.3.0
project inventories 7.2.0 0.12.2
project portals 7.2.0 0.15.2
project worlds 7.2.0 0.13.1
TimeWarp 2.3 r7.0.0
worldedit sponge 7.1.0–6.1.10 SNAPSHOT

Mods:

Aroma1997Core 1.12.2 2.0.0.2 b165
AromaBackup 1.12.2 3.0.0.0 b129
BetterFps 1.4.8
BiblioCraft 2.4.5
Biomes O’ Plenty 7.0.1.2441
BuildCraft 7.99.24.2
CodeChickenLib 3.2.3.357
foamfix 0.10.5
Forestry 5.8.2.394
IndustrialCraft 2 2.8.187 ex112
ironchest 7.0.59.842
JEI 4.15.0.289
NotEnoughItems 2.4.3.245

**Issue Description**
Players are able to see commands that they do not have permission to use.
This includes mod commands like
/forestry
And some plugin commands like
/timewarp
/luckperms
/brush
(from WorldEdit which produces the error to the player "You are not permitted to do that.")

- maybe these mods have these commands but do not link them to permissions.... WorldEdit is another one... none of my players should have WE permissions, but they are able to use "/sponge:help superpickaxe"  for example.... it doesn't give them a superpickaxe, but it doesn't autocomplete for them

/aroma1997 and /luckperms and /forestry
There are things like that that the players cannot actually use, but they are in the list of available commands regardless.

WillBAnders:
Dual and I have both confirmed the proper permission checks are being called on Sponge's side. I took a look at some of the plugins you've mentioned and it's pretty all over the place.

 - LuckPerms: Always [returns true](https://github.com/lucko/LuckPerms/blob/f49446011a63739522fe2d20815613a8322e1d53/sponge/src/main/java/me/lucko/luckperms/sponge/SpongeCommandExecutor.java#L76-L79) on permission checks, needs to be fixed there.
[...]

PR #2375, opened 2019-09-10
Fix wrapped command permission checks [#2372]

WillBAnders:
This is a draft PR awaiting testing and feedback to understand the scope of changes and consequences, as well as discussion towards a final solution.

The `MinecraftWrappedCommand` class contains commands from both Minecraft and Forge, which Sponge provides permissions for as shown on the [Sponge docs](https://docs.spongepowered.org/stable/en/server/spongineer/commands.html#forge). The current implementation uses the permission level system in the `testPermission` method, which effectively bypasses the permission system and includes commands with a permission level of `0`. This is how I believe these commands are being included in help commands and related.

I am concerned with this implementation for a number of reasons, the largest being the fact that these permissions presumably work for these commands, meaning the actual permission string is being checked elsewhere. I will have to examine this further in Vanilla/Forge to see what happens. Additionally, I'm not sure what happens if a permission service is not present and the permission level does come into play - this likely does not maintain the same behavior here.

Assistance with tracking this down would be greatly appreciated.

### DIFF
--- a/src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
+++ b/src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
@@ -185,8 +185,7 @@ public class MinecraftCommandWrapper implements CommandCallable {
 
     @Override
     public boolean testPermission(CommandSource source) {
-        ICommandSender sender = WrapperICommandSender.of(source);
-        return this.command.checkPermission(sender.getServer(), sender);
+        return source.hasPermission(getCommandPermission());
     }
 
     @Override


dualspiral:
> The current implementation uses the permission level system in the testPermission method, which effectively bypasses the permission system and includes commands with a permission level of 0

That's the key thing that had been missing. A permission level of zero means that non-ops can use the command - that is, it's intended by the mod that the command is available to all by default - so Sponge doing the same thing is correct. I suspect, without having looked at the implementation, that anything with level 1 or above gets a permission, anything at level 0 does not.

Of course, I'd be interested to see what happens with something like WorldEdit for Forge, but I think their WEPIF system will take care of the permission problem in those scenarios.

In this case, we shouldn't be removing what is already there, but if we wanted to slap Sponge's permission system on to it, we should `&&` it... though actually, we shouldn't do this blindly. 

This might be heading to a wontfix, unless we absolutely want to add a permission to every command, destined for ops or otherwise. The alternative is to add anything with perm level zero as default true in the permission system - at least then server owners could deny the command permission after.

WillBAnders:
Sponge does add permission to every command regardless of it's level; the main thing was why it wasn't working for the help command and tab completion and I've think I've figured that out.

The [`CommandPermissions`](https://github.com/SpongePowered/SpongeCommon/blob/a9423c7ec9f2b423b52c273f7e7a23818a8f3e41/src/main/java/org/spongepowered/common/command/CommandPermissions.java#L66) class has a custom [`testPermission`](https://github.com/SpongePowered/SpongeCommon/blob/a9423c7ec9f2b423b52c273f7e7a23818a8f3e41/src/main/java/org/spongepowered/common/command/CommandPermissions.java#L66) method which does an actual permission test if the command is a `MinecraftCommandWrapper` - bypassing the callable's `testPermission` because it's implemented differently. This is used in an injection to [`ICommandSender#canUseCommand`](https://github.com/SpongePowered/SpongeCommon/blob/da40458a2c60c8114fd567dc0dd35e5bb3799147/src/main/java/org/spongepowered/common/mixin/core/command/ICommandSenderMixin.java#L46), thus when the command actually runs the permission is properly checked.

`CommandPermissions` also has this [`populateMinecraftPermissions`](https://github.com/SpongePowered/SpongeCommon/blob/a9423c7ec9f2b423b52c273f7e7a23818a8f3e41/src/main/java/org/spongepowered/common/command/CommandPermissions.java#L95-L96) method which sets default permissions for commands (I will need to determine how exactly this works). This uses the internal permission level of the command, which is how I believe users get access to the level zero commands.

In summary:
 - Permissions are in place for all commands and checks work when executing them because of the specific handling in `CommandPermissions`.
 - Changing the implementation of `MinecraftCommandWrapper#testPermission` restores the callable interface and ensure the right permission checks are being made.
 - Users appear to be given permission for all level zero commands in `populateMinecraftPermissions`, though I need to verify exactly how this works as I've seen some weird things with it.

dualspiral:
> Users appear to be given permission for all level zero commands in `populateMinecraftPermissions`

If this is the case, then as long as you can revoke the permission and it disappears from the tab list, then this is a `wontfix` because all players should have access to level zero commands by default.

> as I've seen some weird things with it.

Like?

dualspiral:
[I created a test mod like so](https://gist.github.com/dualspiral/db53abdbbebd5af02c51e977e4934ac6) - three no-op commands:

* `/perms-true-test` - `checkPermission` always returns `true`
* `/perms-0-test` - `checkPermission` returns if a player has op level 0+
* `/perms-1-test` - `checkPermission` returns if a player has op level 1+

Without a permissions plugin, deopped I get `/perms-true-test` only (when I was actually expecting the op level 0 command too), opped, I get all three.

I installed LuckPerms and gave myself the `luckperms` permission only. I then enabled verbose logging and got the following on tab completing `/perms`:

```
[21:01:26] [pool-2-thread-5/INFO] [minecraft/DedicatedServer]: [LP] VB > dualspiral - icommand-test.command.perms-1-test - undefined
[21:01:26] [pool-2-thread-5/INFO] [minecraft/DedicatedServer]: [LP] VB > dualspiral - icommand-test.command.perms-0-test - undefined
[21:01:26] [pool-2-thread-5/INFO] [minecraft/DedicatedServer]: [LP] VB > dualspiral - icommand-test.command.perms-1-test - undefined
[21:01:26] [pool-2-thread-5/INFO] [minecraft/DedicatedServer]: [LP] VB > dualspiral - icommand-test.command.perms-0-test - undefined
```

There is no `icommand-test.command.perms-true-test` check - which should actually be unsurprising as the impl does this check in the `sender#canUseCommand` method.

I'm inclined to not worry about the op level 0 issue because in those cases, you'll be returning `true` anyway. In the case that you _do_ return `true`, no permission check is performed by the implementation at that time. I'm inclined to say that this is intentional by mods and so I don't really want to add a permission check that denies by default. If we wanted to put permissions on everything regardless, then the simple solution is to `&&` the permission check to the existing check - remember that mods might use `canUseCommand` in other ways too so we shouldn't outright blow that away.

WillBAnders:
To clarify, the test mod is the current behavior (aka not this PR), correct?

I think to start we need to define what the expected behavior is. My expectation, given that Sponge adds permissions to all Forge commands according to the documentation, is that when a permission plugin is present the player must have the permission for that command. If they are unable to use the command either by permission or op level, then it should not be shown in help or in tab complete. If this means that op level 0 commands have their permission given by default and must be denied rather than given, I'm okay with that provided the above behavior stands. Do you agree with this, or what do you think should happen?

> as I've seen some weird things with it.
> Like?

[`CommandPermissions#populateMinecraftPermissions`](https://github.com/SpongePowered/SpongeCommon/blob/a9423c7ec9f2b423b52c273f7e7a23818a8f3e41/src/main/java/org/spongepowered/common/command/CommandPermissions.java#L95-L96) goes through all wrapped commands and sets a permission for the command to `TRUE/FALSE` based on the wrapped `checkPermission` method.

Immediately after this, the [`AndPermissionLevelSubject` constructor](https://github.com/SpongePowered/SpongeCommon/blob/a9423c7ec9f2b423b52c273f7e7a23818a8f3e41/src/main/java/org/spongepowered/common/command/AndPermissionLevelSubject.java#L56-L58) appears to replace all true permissions with `UNDEFINED`. I tried tracking through this but I couldn't understand why this is done, so if you have any ideas that would be great.

dualspiral:
> To clarify, the test mod is the current behavior (aka not this PR), correct?

Yes.

> CommandPermissions#populateMinecraftPermissions goes through all wrapped commands and sets a permission for the command to TRUE/FALSE based on the wrapped checkPermission method.
>
> Immediately after this, the AndPermissionLevelSubject constructor appears to replace all true permissions with UNDEFINED. I tried tracking through this but I couldn't understand why this is done, so if you have any ideas that would be great.

That's to do with if you use op level permissions only. It collects the permissions for various oplevels and sets them in the sponge permission service groups - at least, that's how I understand it. If you install a permissions plugin when debugging, and you put a breakpoint in these methods, you'll see they never get called. Don't worry about those.

To go back to my original point adding bold for emphasis:

> I'm inclined to not worry about the op level 0 issue because in those cases, you'll be returning true anyway. In the case that you do return true, no permission check is performed by the implementation at that time. I'm inclined to say that this is intentional by mods and **so I don't really want to add a permission check that denies by default.**

And that's what happens with Minecraft commands. You can still use `/tell`, for example, with nothing defined. [You can see why this is here](https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/mixin/core/command/ServerCommandManagerMixin.java#L182-L217). When a command is registered, it checks for the oplevel of the command - which it can do if it is a `CommandBase`. If it is then oplevel `0`, [it registers the permission as true](https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/mixin/core/command/ServerCommandManagerMixin.java#L213-L214). Of course, my command that had an op level of 0 didn't hit this, because I implemented `ICommand` instead of `CommandBase` - like many Forge mods will do, and you can't specify the op level easily.

The rule of thumb is this - if you didn't require op to run a command before installing Sponge, this should be the case after installing Sponge. Small servers might very well not want to run a permission plugin, that's their perogative, but Sponge shouldn't inhibit their gameplay. 

Further, this PR, as it stands, blows away all mod checks. `checkPermission` may not solely rely on a permission check - we cannot (and must not) assume otherwise. At the very least, it must be an `&&` check, as I have stated previously.

There is clearly a problem with an op level 0 permission check from `ICommand` and so there needs to be some consideration there to _perhaps_ rectify that, but as I said before, it's unlikely a mod checks on a 0 op level, and wil likely just return `true` in that scenario.

WillBAnders:
> so I don't really want to add a permission check that denies by default.

Maybe there's a bit of clarification needed here because I don't believe that's what this issue is about.

The problem is that tab completion and Sponge's help command both use `CommandCallable#checkPermission` for determining whether to include them. For wrapped commands, the permissions that Sponge adds are only checked when executing the command and do not go through this method, which is inconsistent and breaks the contract of `CommandCallable`. These cases need to be handled consistently because right now there are two separate meaning of whether a user has access to a command.

Whether permission checks are denied by default or not doesn't matter with this; as of now they're not checked at all and that's what the issue is.

> At the very least, it must be an && check, as I have stated previously.

I'm interpreting this as being `senderPermissionLevel >= commandPermissionLevel && senderHasPermission`. If the permission service isn't in place then the has permission check doesn't do anything, and if it is then the permission level should be ignored anyways. I don't understand what's this is supposed to mean.

dualspiral:
> Maybe there's a bit of clarification needed here because I don't believe that's what this issue is about.

I don't think you understand how the system works, because it's _exactly_ what this issue is about.

> The problem is that tab completion and Sponge's help command both use CommandCallable#checkPermission for determining whether to include them.

Yes, everything I've said is linked to this.

>  For wrapped commands, the permissions that Sponge adds are only checked when executing the command and do not go through this method, which is inconsistent and breaks the contract of CommandCallable.

That's not quite right. 

`CommandCallable#checkPermission()` subsequently calls `testPermission`, in the case of mods, `MinecraftCommandWrapper#testPermission()`, which then delegates to the _mod_ as to whether it should execute via `ICommand#checkPermission`. As that's an interface, there is very little we can then do to alter the permission check from there. 

The one change we _can_ and _do_ make to that is [on `ICommandSender#canUseCommand`](https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/mixin/core/command/ICommandSenderMixin.java#L43-L48) - if the sender is then asked if they can execute the command, then we take over the check and do the permission check there. If this isn't called, we can't do the permission test.

[It is instructive to read the `testPermission` javadocs](
https://github.com/SpongePowered/SpongeAPI/blob/13da4949cc5ffd51a32144f331656887212bb5e8/src/main/java/org/spongepowered/api/command/CommandCallable.java#L74-L84), it says the following:

```
If implementations are unsure if the command can be executed by
the source, true should be returned. Return values of this method
may be used to determine whether this command is listed in command
listings.
```

That is what we are doing right now. We do not know whether permission would be granted by the mod, so we are being permissive. This statement alone means what is going on is not a bug, and crucially, _the contract is not being broken_ as you claim. We simply cannot be sure whether a mod is intended to be user space, or admin space. It says nothing about doing a permission check with a permission provider - it is about whether the conditions are correct for an invocation to take place.

As I said above:

> The rule of thumb is this - if you didn't require op to run a command before installing Sponge, this should be the case after installing Sponge. Small servers might very well not want to run a permission plugin, that's their perogative, but Sponge shouldn't inhibit their gameplay.

> Whether permission checks are denied by default or not doesn't matter with this

Of course it does, at least in the context of your proposed change. My point is that if a mod says a command can always be executed (that is, `ICommand#checkPermission` simply returns true, like WorldEdit does), then we should not then slap a permission on top of that and default to false. What I linked to explained why it worked for commands such as `/tell` and would be something I would want in place.

> as of now they're not checked at all and that's what the issue is.

The statement is wrong in some cases, as stated above.

> I'm interpreting this as being senderPermissionLevel >= commandPermissionLevel && senderHasPermission. If the permission service isn't in place then the has permission check doesn't do anything, and if it is then the permission level should be ignored anyways. I don't understand what's this is supposed to mean.

No, your `hasPermission` check would be `&&` with what is currently there.

WillBAnders:
I'm tracking things better now, previously I presumed that `canSenderUseCommand` would always be based on the op level, but what you're saying is that it can effectively do anything. I see what you mean by `&&` now and I agree with that.

Because Sponge adds these permissions, we know that when a permission plugin is present and that permission has been denied they cannot execute the command. It is this case that I want to be added to the `testPermission` method so tab completion and help works properly; not anything with defaults.

dualspiral:
> It is this case that I want to be added to the testPermission method so tab completion and help works properly; not anything with defaults

Except this is inherently tied to defaults - again, going back to a previous comment:

> The rule of thumb is this - if you didn't require op to run a command before installing Sponge, this should be the case after installing Sponge. Small servers might very well not want to run a permission plugin, that's their perogative, but Sponge shouldn't inhibit their gameplay.

I have no problem with the idea of this check being introduced, _as long as_ default permissions are set up such that a command a mod provides works the same way before and after Sponge is installed, with no additional changes. It's particularly important for sponge without any other permission plugin. As it stands, what you propose would make non-op commands effectively becomes op commands.

That's why defaults are important here.

WillBAnders:
Gotcha, I think we're on the same page again. It was my belief that the `CommandPermission#populateMinecraftPermissions` did this by setting the permission to whatever the wrapped command's `testPermission` returns. If this isn't the case, then we need to track down how to do this so it's handled properly when a permission service is not available.

dualspiral:
> If this isn't the case, then we need to track down how to do this so it's handled properly when a permission service is not available.

[I explained it here](https://github.com/SpongePowered/SpongeCommon/pull/2375#issuecomment-530502361)

dualspiral:
I thought about this some more and, while playing about trying to detect permissions, realised we have a config file we could use. So, I've created https://github.com/SpongePowered/SpongeCommon/pull/2379. I also include what I was considering doing to detect such permissions - there may be a discussion to have around that too.

I apologise for supersceding this, but as I had the IDE open I just went ahead and did it there and would rather just get it done rather than bat conversation back and forth. Comments are welcome over there.

Thanks for having a look though!

PR #2379, opened 2019-09-12
Add config entry for ensuring mod command permissions are checked.

dualspiral:
This is informed by, and supersedes, #2375 - there was a lot of discussion there regarding defaults that might be useful background reading. Also see #2372.

Some mod `ICommand`s, such as WorldEdit, have their `checkPermission` methods always return `true` - usually because they do their own permission handling during their invocation. This causes Sponge to not perform any permission checks before a command invocation (as we hook into `ICommandSender#canUseCommand`), and may not during, depending on what the command execute command contains. As a result, such commands appear in `/help` and tab completion, even though standard players cannot execute the commands themselves (because the permission check occurs later on).

A simple solution would be to **additionally** check for the Sponge permission during invocation before testing if the mod grants permission. However, this would break simple, op/no-op setups, effectively meaning that all non `CommandBase` derived `ICommands` would require op status, even if they wouldn't without Sponge.

I played about with trying to use a dummy `ICommandSender` on the `checkPermission` command to try to determine if a permission check occurs, and if not, defaulting the command permission to `true` using the `PermissionService`, as we do for regular Minecraft commands that provide an op level. However, I figured that the simplest solution would be to just put this behind a default off config. That way, if a user wants the additional Sponge permissions to take effect, this can be toggled on. I think this is a reasonable compromise, but it _does_ mean that users would have to be aware of the setting.

Obligatory screenshots to show it works ([using this test mod](https://gist.github.com/dualspiral/db53abdbbebd5af02c51e977e4934ac6)):

**Before**
before

**After (note that the `perms-true-test` command permission is now checked)**
after

[If we want to try to do this detection, this is what I had in mind](https://gist.github.com/dualspiral/7393dc8d7563d4806951f0a0d4630e92). I just don't think it's worth it.

### DIFF
--- a/src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
+++ b/src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java
@@ -185,8 +185,13 @@ public class MinecraftCommandWrapper implements CommandCallable {
 
     @Override
     public boolean testPermission(CommandSource source) {
-        ICommandSender sender = WrapperICommandSender.of(source);
-        return this.command.checkPermission(sender.getServer(), sender);
+        if (!SpongeImpl.getGlobalConfigAdapter().getConfig().getCommands().isEnforcePermissionChecksOnNonSpongeCommands()
+            || source.hasPermission(getCommandPermission())) {
+            ICommandSender sender = WrapperICommandSender.of(source);
+            return this.command.checkPermission(sender.getServer(), sender);
+        }
+
+        return false;
     }
 
     @Override
--- a/src/main/java/org/spongepowered/common/config/category/CommandsCategory.java
+++ b/src/main/java/org/spongepowered/common/config/category/CommandsCategory.java
@@ -37,10 +37,24 @@ public class CommandsCategory extends ConfigCategory {
                      + "Correct syntax is <unqualified command>=<plugin name> e.g. \"sethome=homeplugin\"")
     private Map<String, String> aliases = new HashMap<>();
 
+    @Setting(value = "enforce-permission-checks-on-non-sponge-commands",
+            comment = "Some mods may not trigger a permission check when running their command. Setting this to\n"
+                    + "true will enforce a check of the Sponge provided permission (\"<modid>.command.<commandname>\").\n"
+                    + "Note that setting this to true may cause some commands that are generally accessible to all to\n"
+                    + "require a permission to run.\n\n"
+                    + "Setting this to true will enable greater control over whether a command will appear in\n"
+                    + "tab completion and Sponge's help command.\n\n"
+                    + "If you are not using a permissions plugin, it is highly recommended that this is set to false\n"
+                    + "(as it is by default).")
+    private boolean enforcePermissionChecksOnNonSpongeCommands = false;
 
     @Setting(value = "multi-world-patches", comment = "Patches the specified commands to respect the world of the sender instead of applying the \n"
                                                     + "changes on the all worlds.")
     private Map<String, Boolean> multiWorldCommandPatches = new HashMap<>();
+
+    public boolean isEnforcePermissionChecksOnNonSpongeCommands() {
+        return this.enforcePermissionChecksOnNonSpongeCommands;
+    }
 
     public Map<String, String> getAliases() {
         return this.aliases;


WillBAnders:
I have one issue with this, which is that this still leaves Sponge's handling of permissions for mods in an inconsistent state. Regardless of whether the config is set, Sponge will still require the permission in order to execute the command. These situations should be unified, especially as the setting implies permission checks are not being made on mod commands when in fact they are. This might be why there is only permission check for the `perms-true-test` one; I would expect it to be the same as the others.

The first step of this can be fixed by modifying [`CommandPermissions#testPermission`](https://github.com/SpongePowered/SpongeCommon/blob/a9423c7ec9f2b423b52c273f7e7a23818a8f3e41/src/main/java/org/spongepowered/common/command/CommandPermissions.java#L65-L67) to just use the `CommandCallable` variant. The second piece has to do with the configuration - adding a setting in this manner will not be backwards compatible, it would have to default to true in order to maintain behavior when the command is executed. I would also contend that based on the Sponge documentation saying that it adds permissions for modded commands, this should be the default behavior regardless.

dualspiral:
I really advise you run the code and step through it because I'd hope how it all fits together will become a lot clearer. Right now, I don't think you are fully understanding what is going on - to me, it looks like what you think happens isn't actually happening. This PR isn't breaking anything, it's adding the option to enforce a permission check in the `testPermission` method, which is only an option as it changes mod behaviour when they suggest commands may always be used via their `checkPermission` method.

---

To take your points in turn:

> Sponge will still require the permission in order to execute the command. 

No it doesn't. Install LP and use it's verbose mode with this test mod. You'll see that the `/perms-true-test` checks no permissions with this set to false (and before this commit) and checks with this config set to true.

> These situations should be unified, especially as the setting implies permission checks are not being made on mod commands when in fact they are.

WorldEdit for Forge might look like it's checking a permission - but it will be checking it during their execution, not during `testPermission`. So no - if they don't during `checkPermission`, Sponge just doesn't know about it when tab completing etc.

> This might be why there is only permission check for the perms-true-test one; I would expect it to be the same as the others.

Once when my config option is set to true, zero when false when running it. There would be more for the other two commands. Double these for tab completion, as above (one of your other changes causes this doubling).

> The first step of this can be fixed by modifying CommandPermissions#testPermission to just use the CommandCallable variant. 

Then you'll end up with a stack overflow. This gets called from the `ICommand#checkPermission` method on the mod command when `ICommandSender#canUseCommand` method is called. 

So no, we cannot do that and I'm not interested in rewriting all these mixins and code to just make such a tweak.

> The second piece has to do with the configuration - adding a setting in this manner will not be backwards compatible.

Backwards compatible with _what_? This commit makes no changes to logic _by default_.

The reason I default it to `false` is for the reasons I gave in your previous PR about [if no permission plugin is present and op/no-op commands](https://github.com/SpongePowered/SpongeCommon/pull/2375#issuecomment-530543895) and I'm really not willing to go over it again.

> it would have to default to true in order to maintain behavior when the command is executed. 

What behaviour do you want to maintain? Or rather, what do you believe is happening now?

>  I would also contend that based on the Sponge documentation saying that it adds permissions for modded commands, this should be the default behavior regardless.

See reasons given above and in the previous PR. I'm really not interested in making those arguments yet again. I have explained how it works and why it's set to false by default for the reasons I've given, I'm not changing it to true by default.

I'm trying to provide an option to support those who want to ensure these Sponge provided permissions work with their mods, without inhibiting op/no-op servers. I'd rather not suddenly end up with issues occuring because commands suddenly don't work without op. That's why this is the method I've chosen.

WillBAnders:
Then we need to ask someone to help moderate this because I don't know how to convey the issues I'm seeing to you; I've done the best I can across multiple issues but that has not worked. I would not support merging this in the current state.

Zidane:
@WillBAnders Not seeing the issue here, I've read both your sides and this PR is very straightforward.

WillBAnders:
1. The permission check made [here](https://github.com/SpongePowered/SpongeCommon/blob/a9423c7ec9f2b423b52c273f7e7a23818a8f3e41/src/main/java/org/spongepowered/common/command/CommandPermissions.java#L65-L67) when executing commands is not affected by the config setting. The config setting therefore does not affect whether Sponge uses permissions when executing commands and is thus misleading as-is.
2. Because the setting was made false by default, this still leaves permission checks in an inconsistent state. The permission will be checked when executing but will not be checked when tab completing or using the help command. This is the issue described in #2372 that I set out to fix. Sponge adding permissions to commands is currently the expected behavior and it's stated as such in the docs.

dualspiral:
> The permission check made here when executing commands is not affected by the config 
setting. 

**Of course it is.** It **ensures** that a permission check takes place with the permission provider. I make no claim that this set to false turns a check on or off. All it does is add an additional check.

As it stands, for some mods, no permission check is made through the `ICommand#checkPermission` function. This is the equivalent of the `CommandCallable#testPermission`. We redirect any calls to `ICommandSource#canUseCommand` to our own permission checks, but if a command does not call this method in their `checkPermission` function, then we don't do a permission check before running the `ICommand#execute`. Mods might then do a permission check in there - it's the equivalent of doing your permission checks in `CommandCallable#process` or `CommandSpec#executor` instead of `CommandCallable#testPermission` or `CommandSpec#permission` - sure, that works, but we're not going to execute a command just to determine that tab completion shouldn't happen now, are we?

> The permission will be checked when executing but will not be checked when tab completing or using the help command.

> The config setting therefore does not affect whether Sponge uses permissions when executing commands and is thus misleading as-is.

**I've already explained this.** Copy/pasting from above:

> No it doesn't. Install LP and use it's verbose mode with this test mod. You'll see that the /perms-true-test checks no permissions with this set to false (and before this commit) and checks with this config set to true.

I should add that your PR did exactly what mine is doing, just without the config entry. Well, actually, no, worse, because it removed the ability for the mod to make that decision, but we did discuss why that was a bad idea so I take it a read that's what you intended.

> Sponge adding permissions to commands is currently the expected behavior and it's stated as such in the docs

Then the docs are wrong and I will go change them. That very well might be an ideal that we'd like to have by without having any sort of complex code analysis run on every mod you install we just _cannot_ make this guarantee without breaking the op/no-op scenario. I am trying to ensure we cater for everyone here. Sometimes we just have to admit that we were wrong and deal with it.

I have spent hours on this trying to find a better solution. I have worked through every line and I even came up with an alternative which I ultimately didn't think was worth it. I may come back and revisit this in the future. I **strongly** recommend that you take the time to setup a Sponge workspace, add a test mod like mine, throw in a breakpoint at `CommandPermissions#testPermission` and wherever else takes your fancy, and run commands on that mod. See where the checks happen - because I don't know how else to tell you that the assumptions that you make about permissions being checked differently on execution are incorrect.

dualspiral:
Unless there is a compelling reason otherwise, I'll merge this in the next day or two (unless someone else just wants to hit the trigger). As this is a simple "default status quo" thing - that is, it doesn't change behaviour by default, I see no reason to leave this sitting here much longer.

(plus, I've had a couple of people poke me about this - don't really want to incur the wrath of the server owner unnecessarily!)

gabizou (approved):
Looks good to me. Read the discussion, can't argue why this wouldn't be mergable with it being defaulting to false.

dualspiral:
2d60942e3fd6cd3e43461fedf524c4c32df925c5

Issue #1584, opened 2017-11-03
Decaying items not firing events


n00b42:
A decaying item is removed from the world without firing one of the expected events:
While ExpireEntityEvent seems logical, I was told at least an DestructEntityEvent should fire, but neither did that (at least not for decaying items)

For clarification: An decaying item is an item whose lifetime is over and is removed from the game. E.g. drops/sapling/... laying on the ground.

**Setup**
- Tried both:
  - spongeforge-1.12.2-2503-7.0.0-BETA-2716, Forge version: forge-1.12.2-14.23.0.2515
  - spongevanilla-1.12.2-7.0.0-BETA-349
- Java version: 1.8.0_144
- Operating System:  Windows

n00b42:
Maybe related #1583

ryantheleach:
@n00b42 In order to save developer time when it comes to recreating this issue for debugging, a small test plugin source uploaded to gist would be appreciated.

I've marked this as an accepted bug, due to it appearing on this list as implemented: https://github.com/SpongePowered/SpongeCommon/issues/41

DrZoddiak:
Just wanted to pop in and make note that this is still an issue.

https://gist.github.com/DrZoddiak/d6949c80be6bbda51c397e53a39aa8bf
Using this bit of code the event doesn't seem to fire when items are cleared naturally through the game.
Same with 
https://gist.github.com/DrZoddiak/a9f766b1913fea4af0044a4eeb708d34

Targeting: 7.2.0-SNAPSHOT

ImMorpheus:
Able to reproduce on SF (SC https://github.com/SpongePowered/SpongeCommon/commit/11a0398d9f3fe12ed7cf40ff758d6f7d36dbf154).

However it looks like `ExpireEntityEvent` is **not** being fired.
(`DestructEntityEvent` is working fine, see https://github.com/SpongePowered/SpongeCommon/pull/1157).

PR #2383, opened 2019-09-21
Fire event when Item Entity Expires

DrZoddiak:
Should resolve 1/2 of  #1584

Full disclosure I had no idea what I'm doing.
I just tinkered with it until it worked (with some assistance in discord)

It does fire an event every time an item entity expires. 
My test code: https://gist.github.com/DrZoddiak/6e542f070520e11193b8e2fd3c6f64f3


src/main/java/org/spongepowered/common/mixin/core/entity/item/EntityItemMixin.java
```diff
@@ -192,6 +192,22 @@
...
     }
+    
+    @Inject(
+        method = "onUpdate", 
+        at = @At (
+            value = "INVOKE", 
+            ordinal = 1, 
+            target = "Lnet/minecraft/entity/item/EntityItem;setDead()V")
+    )
+    private void fireExpireEntityEventTargetItem (final CallbackInfo ci) {
```
dualspiral: I hate to be the one to complain about formatting...

This method name needs to be prefixed with `spongeImpl$` to be consistent with the class (though, @gabizou, [I thought the policy was to prefix `impl$`, not `spongeImpl$`](https://github.com/SpongePowered/SpongeCommon/wiki/Implementation-Policy:-Bridges-and-Mixins#rules-3))? 

Also, very nitpicky, but while you're there, can you remove the space between the method name and parameters, and the space after the method name?

dualspiral (commented):
I've not tested this, but I do have a formatting comment.

A glance says it looks fine though, it's simply firing an event when an item decays. Nothing particularly special about that really...

src/main/java/org/spongepowered/common/mixin/core/entity/item/EntityItemMixin.java
```diff
@@ -192,6 +192,21 @@
...
     }
+    
+    @Inject(
+        method = "onUpdate", 
+        at = @At (
+            value = "INVOKE", 
+            ordinal = 1, 
+            target = "Lnet/minecraft/entity/item/EntityItem;setDead()V")
+    )
+    private void spongeImpl$fireExpireEntityEventTargetItem(final CallbackInfo ci) {
```
gabizou: Not to be too nit picky:
`impl$fireExpireEntityEvent` is what the method name should be. Prefix is `impl$` not `spongeImpl$`

src/main/java/org/spongepowered/common/mixin/core/entity/item/EntityItemMixin.java
```diff
@@ -192,6 +192,21 @@
             ci.cancel();
         }
     }
+    
+    @Inject(
+        method = "onUpdate", 
+        at = @At (
+            value = "INVOKE", 
+            ordinal = 1,
```
gabizou: Could this be redone with a slice instead of a single invoke that looks a little vague without looking at the target method?

src/main/java/org/spongepowered/common/mixin/core/entity/item/EntityItemMixin.java
```diff
@@ -192,6 +192,21 @@
...
+    
+    @Inject(
+        method = "onUpdate", 
+        at = @At (
+            value = "INVOKE", 
+            ordinal = 1, 
+            target = "Lnet/minecraft/entity/item/EntityItem;setDead()V")
+    )
+    private void spongeImpl$fireExpireEntityEventTargetItem(final CallbackInfo ci) {
+        try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
```
gabizou: Is it guaranteed that we're on the main thread in single player? I remember at times there's a lot that is written assuming this is run on the server only to blow up on the client.

gabizou (changes requested):
Quick look, address comments, need to test personally or another sponge dev to test.

src/main/java/org/spongepowered/common/mixin/core/entity/item/EntityItemMixin.java
DrZoddiak: So to be clear, I know nothing about mixins. My implementation was mostly based off of #1917 
and working around until it worked. That being said, I did give this a go trying to read a bit. 
(Honestly mixins aren't something I should be playing with)
I didn't want to upload it, as I have no idea if it is bad practice, but I did test the changes made here, and it still works.

https://gist.github.com/DrZoddiak/408a71583c974392beaf44d235c92c5d

src/main/java/org/spongepowered/common/mixin/core/entity/item/EntityItemMixin.java
DrZoddiak: I have no idea how to verify this without running the jar on my client, and I cannot seem to get my compiled jar to work on the client. As noted above I'm very inexperienced with this stuff.

DrZoddiak:
For ease of testing I've made a small test plugin someone can just toss into their plugins folder.
Code:
https://gist.github.com/DrZoddiak/bab23726553ab28ae4194c2cf3ac7f2d
Jar:
https://github.com/DrZoddiak/SpongeCommon/releases/tag/1.0

src/main/java/org/spongepowered/common/mixin/core/entity/item/EntityItemMixin.java
```diff
@@ -29,14 +29,17 @@
...
+import org.spongepowered.api.event.SpongeEventFactory;
+import org.spongepowered.api.event.entity.ExpireEntityEvent;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
-import org.spongepowered.asm.mixin.injection.At;
-import org.spongepowered.asm.mixin.injection.Constant;
-import org.spongepowered.asm.mixin.injection.Inject;
-import org.spongepowered.asm.mixin.injection.ModifyConstant;
-import org.spongepowered.asm.mixin.injection.Redirect;
+import org.spongepowered.asm.mixin.injection.*;
```
gabizou: No star imports. Sorry for nitpicking.

src/main/java/org/spongepowered/common/mixin/core/entity/item/EntityItemMixin.java
```diff
@@ -192,6 +195,29 @@
...
+        )
+    )
+    private void impl$fireExpireEntityEventTargetItem(final CallbackInfo ci) {
+        if (SpongeImplHooks.isMainThread()) {
+            try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
+                frame.pushCause(this);
+                ExpireEntityEvent.TargetItem event = SpongeEventFactory.createExpireEntityEventTargetItem(Sponge.getCauseStackManager().getCurrentCause(), (Item) this);
+                SpongeImpl.postEvent(event);
+            }
+        }
```
gabizou:
```suggestion
        if (!SpongeImplHooks.isMainThread() || this.getItem().isEmpty()) {
            // In the rare case the first if block is actually at the end of the method instruction list, we don't want to 
            // erroneously be calling this twice.
            return;
        }
        try (final CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
            frame.pushCause(this);
            final ExpireEntityEvent.TargetItem event = SpongeEventFactory.createExpireEntityEventTargetItem(frame.getCurrentCause(), (Item) this);
            SpongeImpl.postEvent(event);
        }
```

src/main/java/org/spongepowered/common/mixin/core/entity/item/EntityItemMixin.java
```diff
@@ -192,6 +195,29 @@
             ci.cancel();
         }
     }
+    
+    @Inject(
+        method = "onUpdate",
```
gabizou:
```suggestion
        method = "onUpdate",
        at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/item/EntityItem;setDead()V"),
        slice = @Slice(
            from = @At(value = "INVOKE", target = "Lnet/minecraft/entity/item/EntityItem;handleWaterMovement()Z"),
            to = @At("TAIL")
        )
```

gabizou (changes requested):
The injection point using ordinals is unclear, but what is clear is saying "target the method call `setDead()` after `handleWaterMovement()` is called" when you end up looking at the method. Because bytecode is weird and decompilation/recompilation is also weird, there's a corner case where the original if statement could be making a return at the end, where the `setDead()` is also after the instruction`handleWaterMovement()`. This ensures that the injection is going to consider that corner case where the event may be fired twice because the injection occurs twice.

ImMorpheus (approved):
Working as intended

PR #2385, opened 2019-09-24
Make hiding commands that subjects don't have permission for configurable

dualspiral:
~~[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2049)~~ | **SpongeCommon** | [Original Issue](https://github.com/SpongePowered/SpongeAPI/pull/2048)

---

See SpongePowered/SpongeAPI#2048.

Basically, after https://github.com/SpongePowered/SpongeAPI/commit/b38390e1751fff802d355155c45b2eb4913f5817, people setting up new servers found that commands that stated they had no permission before now state that they don't exist. The original intention was to remove commands from tab completion when a player doesn't have permission, and prevent players from running `/sponge:help <command>` to discover commands.

However, this has somewhat backfired. What this has done is effectively hide commands from players that don't have permissions for them - so, if you run, say `/tp` and you don't have permission, Sponge will now tell you the command is unknown, rather than you don't have permission.

I should have considered this being the case before pulling, oh well. So, I figured I'd go inbetween, add configuration to enable this command hiding behaviour with ~~four~~ two different options:

* The first is on execute - whether a command that you don't have permission for should either act as if it doesn't exist, or return a no permission message (default, 7.1.6 behaviour)
* The second is a combination of the following three, which I thought was the best idea after thinking on @ryantheleach's comment some more.
  * On Tab Completion - whether you can tab complete a command name that you don't have permissions for (default to no, not 7.1.6 behaviour)
  * On `/sponge which` - whether you can inspect commands you don't have access to (as an admin tool, this might not be important, but defaults to no anyway, not 7.1.6 behaviour)
  * On `/sponge:help <command>` - can you inspect commands you don't have access to (default no, not 7.1.6 behaviour)

Of course, what I thought would be a simple problem to solve actually turned out to be more complex - the API implementation of commands strikes again!

My original intention was to alter https://github.com/SpongePowered/SpongeAPI/commit/b38390e1751fff802d355155c45b2eb4913f5817#diff-89adf4090b06b371ecc8b77db21c4abfR305 and create a new method that accepts a `BiPredicate<CommandSource, CommandMapping>` that you would then do the filtering in. However... `SimpleDispatcher` is a `final` class in the API. So, that means:

* No access to the Sponge config - so we can't set filters based on the task that we need to do.
* Changing the impl to send `null` as the `CommandSource` parameter to this method is also a no-go, as it might have implications for those using the dispatcher with a different `Disambiguator`.
* We have to be mindful of other consumers of the API and I am now of mind, considering that the change is further widespread, to revert the commit involved.

The obvious solution to this, therefore, is to create our own dispatcher in Common. This is what I've suggested below, it's very rough as it's effectively starting with a copy of the `SimpleDispatcher`. The idea is that I basically add filters to the dispatcher where relavent, it becomes an implementation detail and we can play a little more fast and loose with how we deal with selecting commands, while the `SimpleDispatcher` can be left alone.

Of course, right now there is a lot of duplicate code. I thought about using Mixins to alter the API implementation to add our configuration, but again, that could change the fundamentals of the dispatcher if others are using it. So, I'd like to [revert this commit](https://github.com/SpongePowered/SpongeAPI/commit/b38390e1751fff802d355155c45b2eb4913f5817) to pull it into an implementation change, and then remove the `final` modifier from `SimpleDispatcher`. I don't think it's a breaking change to allow the `SimpleDispatcher` to be extended, but once we do it, we can't go back.

@gabizou @Zidane - particularly on that last point, I'd like some opinion of whether removing that modifier is okay. If it is not, then I think we'll just have to do what I'm doing right now, duplicating the class (as in the linked API PR).

---

As a resolution, no, we're just copying the dispatcher into common. That should give us more flexibility to fix things in the future should it be needed. On a further note, I won't revert the API commit, it actually works well for tab completion of child commands - if that was a problem I think we would have seen an issue by now.

### DIFF
--- a/src/main/java/org/spongepowered/common/SpongeImpl.java
+++ b/src/main/java/org/spongepowered/common/SpongeImpl.java
@@ -44,6 +44,7 @@ import org.spongepowered.api.event.SpongeEventFactory;
 import org.spongepowered.api.event.game.state.GameStateEvent;
 import org.spongepowered.api.plugin.PluginContainer;
 import org.spongepowered.api.util.Direction;
+import org.spongepowered.common.command.SpongeCommandManager;
 import org.spongepowered.common.config.SpongeConfig;
 import org.spongepowered.common.config.SpongeConfigSaveManager;
 import org.spongepowered.common.config.type.CustomDataConfig;
@@ -96,6 +97,7 @@ public final class SpongeImpl {
     @Inject @Nullable private static SpongeDataManager dataManager;
     @Inject @Nullable private static SpongePropertyRegistry propertyRegistry;
     @Inject @Nullable private static SpongeScheduler scheduler;
+    @Inject @Nullable private static SpongeCommandManager commandManager;
     @Inject @Nullable private static SpongeCauseStackManager causeStackManager;
 
     private static final List<PluginContainer> internalPlugins = new ArrayList<>();
@@ -154,6 +156,10 @@ public final class SpongeImpl {
 
     public static SpongeScheduler getScheduler() {
         return check(scheduler);
+    }
+
+    public static SpongeCommandManager getCommandManager() {
+        return check(commandManager);
     }
 
     public static SpongeCauseStackManager getCauseStackManager() {
--- a/src/main/java/org/spongepowered/common/command/SpongeCommandDisambiguator.java
+++ b/src/main/java/org/spongepowered/common/command/SpongeCommandDisambiguator.java
@@ -35,7 +35,6 @@ import org.spongepowered.api.util.annotation.NonnullByDefault;
 import org.spongepowered.api.command.CommandMapping;
 import org.spongepowered.api.command.CommandSource;
 import org.spongepowered.api.command.dispatcher.Disambiguator;
-import org.spongepowered.api.command.dispatcher.SimpleDispatcher;
 import org.spongepowered.common.SpongeImpl;
 
 import java.util.List;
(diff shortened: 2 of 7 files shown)


ryantheleach:
The idea is good, I'm just not a fan of the overly configurable config.

My choice would be to add a configuration option, "Cloak commands lacking permission"

This would cloak them from execute, complete, help and which.

with the ability to override this based on a permission.

dualspiral:
> This would cloak them from execute, complete, help and which.

This is the problem now, though. At the very least, we need to separate execute from the others - which is the whole premise of making the change.

We could reduce configurability. I considered having an option that would either:

* Hide commands without permissions in all scenarios (what we have now)
* Hide commands without permissions _except_ upon execution
* Show all commands, regardless of permission (mostly what we had before)

However, I'd rather keep it because I'm sure people will want different things. If the comments on the config are good, then what's the harm?

XakepSDK:
I hope this can be merged soon, admins ask for help because they think that plugin is not installed.

dualspiral:
Yep, it's been a real pain for me too and I've been a little tardy on this. I haven't properly tested it yet, which is really the only reason it hasn't been merged, but I won't get around to it until tomorrow. If someone would like to test if it works, whether hiding still works when the config entry is set to true, and that hiding is still enabled for `/help` and `/sponge which`, that would be great.

@Zidane More for information, annoyingly, the `SimpleDispatcher` in the API layer is used by the child command element - so that voids any notion of removing it from there because the child command element is API implemented itself. As a result, I've not touched the simple service manager either.

Zidane:
@dualspiral 

BAH

PR #2386, opened 2019-09-28
Use randomized SpawnPoint when respawning

### LINKED ISSUE #2983: Issue with spawnRadius and Sponge
**I am currently running**

- SpongeForge version: 1.12.2-2838-7.1.7-RC3924
- Forge version: 14.23.5.2838
- Java version: Java 8
- Operating System: Centos 7

- Plugins/Mods: None

**Issue Description**

When using spawnRadius with Sponge the first spawn will be correct, but when the worldSpawn gets set, respawning it doesn't take spawnRadius into account. It will always spawn you at worldSpawn instead of the set value of gamerule "spawnRadius". I have my spawnRadius set to 10,000, when I die and respawn I will spawn exactly at the worldSpawn.

LXGaming:
Fixes https://github.com/SpongePowered/SpongeForge/issues/2983


ImMorpheus (approved):
Working as intended

PR #2388, opened 2019-09-29
Fix chunk regeneration using old biome values.

### LINKED ISSUE #2343: Chunk regeneration problem, World.regenerateChunk( ... )
**I am currently running**
 - SpongeForge version: **RC3885**
 - Java version: **openjdk 1.8.0_222/64-bit**
 - Operating System: **Linux (4.15.0-54-generic/amd64)**
 - Plugins (10): Minecraft, Minecraft Coder Pack, SpongeAPI, Sponge, BungeeTabListPlus-SpongeBridge, LuckPerms, OEye, OServer, OAPI, OChat
 - Mods (7): Minecraft, Minecraft Coder Pack, Forge Mod Loader, Minecraft Forge, SpongeAPI, Sponge, SpongeForge

**Issue Description**
I'm trying to regenerate chunk using the `World.regenerateChunk( ... )` method, full code below:

```java
    @Override
    public CommandResult execute(CommandSource src, CommandContext args) throws CommandException {
        if (!(src instanceof Player)) {
            return CommandResult.empty();
        }

        Player player = (Player) src;

        BlockRay<World> ray = BlockRay.from(player)
                .stopFilter(BlockRay.continueAfterFilter(BlockRay.onlyAirFilter(), 1))
                .build();

        Optional<BlockRayHit<World>> end = ray.end();

        end.ifPresent(hit -> {
            Location<World> location = hit.getLocation();

            player.sendMessage(Text.of("End pos: ", hit.getPosition()));
            player.sendMessage(Text.of(" end type ", location.getBlockType().getName()));
            player.sendMessage(Text.of(" chunk at ", location.getChunkPosition()));

            player.getWorld()
                    .regenerateChunk(location.getChunkPosition())
                    .ifPresent(chunk -> {
                        Vector3i pos = chunk.getPosition();

                        player.sendMessage(Text.of(
                                TextColors.YELLOW, "Chunk ", TextColors.GOLD, "[",
                                TextColors.RED, pos.getX(), TextColors.YELLOW, ",",
                                TextColors.RED, pos.getZ(), TextColors.GOLD, "]",
                                TextColors.YELLOW, " regenerated"
                        ));
                    });
        });

    }
```

Maybe I'm doing something wrong?

With or without other mods, it doesn't matter. There is nothing interesting in the log files. Nothing about (un-)loading and/or (re-)generating chunks.

**Screenshots**

JBYoshi:
https://github.com/SpongePowered/SpongeCommon/blob/7271dde9b931e860bf1a60ca72d327a4d6371e30/src/main/java/org/spongepowered/common/mixin/core/world/gen/ChunkGeneratorOverworldMixin.java#L233-L237

During regeneration, `biomes` is never null, so it doesn't update the biome array until `getBiomes()` is called.

However, in SpongeVanilla, the `SpongeChunkGenerator` class will retrieve the biomes before this method is called. In this case, the old behavior is kept.

Tested on both SpongeForge and SpongeVanilla. Fixes #2343.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/gen/ChunkGeneratorOverworldMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/gen/ChunkGeneratorOverworldMixin.java
@@ -61,6 +61,7 @@ import org.spongepowered.common.bridge.world.gen.ChunkGeneratorOverworldBridge;
 import org.spongepowered.common.bridge.world.gen.PopulatorProviderBridge;
 import org.spongepowered.common.util.VecHelper;
 import org.spongepowered.common.util.gen.ObjectArrayMutableBiomeBuffer;
+import org.spongepowered.common.world.gen.SpongeChunkGenerator;
 import org.spongepowered.common.world.gen.WorldGenConstants;
 import org.spongepowered.common.world.gen.populators.AnimalPopulator;
 import org.spongepowered.common.world.gen.populators.FilteredPopulator;
@@ -215,6 +216,7 @@ public abstract class ChunkGeneratorOverworldMixin implements PopulatorProviderB
      *   This fixes the FuturePack mod as it extends the ChunkProviderOverworld generator.
      * @author gabizou - September 18th, 2018 - Updated modification to redirect for vanilla
      *   generators would re-apply the same modified math for the x and z values.
+     * @author JBYoshi - 
      *
      * Redirects this method call to just simply return the current biomes, as
      * necessitated by @Deamon's changes. This avoids an overwrite entirely.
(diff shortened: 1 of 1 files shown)


ImMorpheus (approved):
Working as intended.

The only affected call is https://github.com/SpongePowered/SpongeCommon/blob/2667e7729c4a854fac710107cb6486df2152e4be/src/main/java/org/spongepowered/common/mixin/api/mcp/world/gen/ChunkGeneratorOverworldMixin_API.java#L63 in which `biomesForGeneration` is [assigned](https://github.com/SpongePowered/SpongeCommon/blob/2667e7729c4a854fac710107cb6486df2152e4be/src/main/java/org/spongepowered/common/mixin/api/mcp/world/gen/ChunkGeneratorOverworldMixin_API.java#L61) two lines before and the chunkProvider is `ChunkProviderServer` (not an instance of `SpongeChunkGenerator`). This call is the same (`bridge$getBiomesForGeneration(x, z)`) so this is fine

PR #2389, opened 2019-09-29
Implement TeleporterMixin.bridge$placeEntity for SpongeVanilla.

JBYoshi:
Got an AbstractMethodError for this while doing some testing. I just copied over the code from the Forge mixin and removed all the code for "isVanilla == false". Not sure whether this is supposed to be implemented on vanilla (the method was designed for Forge) or whether I just hit a bug.

Optional companion PR: https://github.com/SpongePowered/SpongeForge/pull/3017

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/IPhaseState.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/IPhaseState.java
@@ -906,4 +906,13 @@ public interface IPhaseState<C extends PhaseContext<C>> {
     default boolean allowsGettingQueuedRemovedTiles() {
         return false;
     }
+
+    /**
+     * Allows phases to be notified when an entity successfully teleports
+     * between dimensions.
+     * 
+     * @param phaseContext The appropriate phase context
+     */
+    default void markTeleported(final C phaseContext) {
+    }
 }
--- a/src/main/java/org/spongepowered/common/mixin/core/world/TeleporterMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/TeleporterMixin.java
@@ -28,6 +28,7 @@ import com.google.common.base.MoreObjects;
 import it.unimi.dsi.fastutil.longs.Long2ObjectMap;
 import net.minecraft.block.state.pattern.BlockPattern;
 import net.minecraft.entity.Entity;
+import net.minecraft.entity.player.EntityPlayerMP;
 import net.minecraft.init.Blocks;
 import net.minecraft.util.EnumFacing;
 import net.minecraft.util.math.BlockPos;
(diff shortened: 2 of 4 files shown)


src/main/java/org/spongepowered/common/mixin/core/world/TeleporterMixin.java
```diff
@@ -82,6 +85,27 @@ public void placeInPortal(Entity entityIn, float rotationYaw) {
...
     }
 
+    @Override
+    public void bridge$placeEntity(net.minecraft.world.World world, Entity entity, float yaw) {
+        boolean didPort;
+        if (entity instanceof EntityPlayerMP) {
+            this.placeInPortal(entity, yaw);
+            didPort = true;
+        } else {
+            if (((WorldServerBridge) this.world).bridge$getDimensionId() == 1) {
```
gabizou: Is it guaranteed the world is a `WorldServer` instance?

src/main/java/org/spongepowered/common/mixin/core/world/TeleporterMixin.java
```diff
@@ -82,6 +85,27 @@ public void placeInPortal(Entity entityIn, float rotationYaw) {
...
+                didPort = true;
+            } else {
+                didPort = this.placeInExistingPortal(entity, yaw);
+            }
+        }
+        
+        if (PhaseTracker.getInstance().getCurrentContext() instanceof InvokingTeleporterContext) {
```
gabizou: either replace with a `isTeleporting()` or make it a state based method that's overridden in the state, much like `isCapturing` or `isTicking`.

gabizou (changes requested):
PhaseContext instanceof and casting, would be easier to make it on the state itself in my opinion.

src/main/java/org/spongepowered/common/mixin/core/world/TeleporterMixin.java
Zidane: That code comes from SpongeCommon but you don't have a bad point.

src/main/java/org/spongepowered/common/mixin/core/world/TeleporterMixin.java
Zidane: Yes

JBYoshi:
@gabizou Updated.

src/main/java/org/spongepowered/common/event/tracking/IPhaseState.java
```diff
@@ -906,4 +906,7 @@ default boolean isConvertingMaps() {
     default boolean allowsGettingQueuedRemovedTiles() {
         return false;
     }
+
```
gabizou: Can you document what this method is intending to do? I know that not all of the methods in `IPhaseState` are well documented, but it's best when adding new ones that they are documented.

src/main/java/org/spongepowered/common/mixin/core/world/TeleporterMixin.java
gabizou: Make the id a constant (Even though we know the end dimension id is `1` for all of 1.12's lifetime.

gabizou (changes requested):
Small documentation request and constant change. Otherwise, fine to merge.

PR #2390, opened 2019-10-01
Fix method name and avoid sending full chunk

LXGaming:
For information regarding this PR please read: https://github.com/SpongePowered/SpongeForge/issues/2764#issuecomment-537269223

This may fix the following issues:
 - https://github.com/SpongePowered/SpongeForge/issues/2764
 - https://github.com/SpongePowered/SpongeForge/issues/2984
 - https://github.com/SpongePowered/SpongeForge/issues/2989


Zidane:
@LXGaming Could you also move the query of the Chunk object to inside the if statement? No reason for us to be doing that lookup if we're not set to updating biomes.

LXGaming:
@Zidane Done

PR #2393, opened 2019-10-05
Refactor issue template

ImMorpheus:
- Remove `Java version/Operating System` (is there an issue where they were relevant ?)
- `Plugins/mods` moved under `**I am currently running**`
- Add a `log` section (for the bug report template)
- Add a minimal issue template for an implementation request (event not firing, datamanipulator not implemented, etc.)

This should hopefully improve the quality of bug reports and avoid common issues with them (people forgetting to post the crash log or ignoring the wall of text and making their own issue template).

I'd like to stress the fact that having a reproduction with minimal mods available would speed up development (currently, it is just recommended since that part was copied over from the previous template).


.github/ISSUE_TEMPLATE/BUG_REPORT.md
```diff
@@ -0,0 +1,31 @@
...
+<!-- Make sure you are using the latest version -->
+- SpongeForge version: 
+- SpongeVanilla version: 
+- Plugins/Mods:
+
+<!-- Assume we have no idea how mods work, be as detailed as possible 
+        and include a step by step reproduction. -->
+**Issue Description**
+
+
```
dualspiral, line 17: Add "Steps to Reproduce"?

.github/ISSUE_TEMPLATE/BUG_REPORT.md
```diff
@@ -0,0 +1,31 @@
...
+
+<!-- Any relevant log files and snippets should be uploaded to gist.github.com! -->
+**Log**
+
+
+
+<!-- Additional information -->
```
dualspiral, line 24: Put this information at the top - people are more likely to read what is immediately visible. The chance of people reading this before filling in the information is lower if it's after everything they need to fill.

dualspiral (commented):
> (is there an issue where they were relevant ?)

Yes - file system operations are one example where it does matter: e.g. https://github.com/SpongePowered/SpongeForge/issues/1991#issuecomment-364886453. While Java puports to be "write once, run everywhere", there are edge cases and systems where that _might_ matter.

Java versions may also matter - we required a patch version of Java because of problems with some collection type IIRC (we needed Java 8u20 - bugs in Java can affect things).

I'm not saying that it matters 99% of the time - but there are reasons why you _might_ request the information. However, this on itself is not a reason to just keep the requests, we could always request more information if we feel it is relavent.

Also, I don't think removing the comments, such as asking to include all mods/plugins, is a good idea. People don't follow this as is, much less so without description.

ImMorpheus:
> Yes - file system operations are one example where it does matter: e.g. SpongePowered/SpongeForge#1991 (comment)

Oh, that's a really good example. I don't mind adding the OS back, but I'll wait for other opinions on this.

> Java versions may also matter - we required a patch version of Java because of problems with some collection type IIRC (we needed Java 8u20 - bugs in Java can affect things).

I know, but that's (documented on the [docs](https://docs.spongepowered.org/stable/en/server/getting-started/jre.html#installing-java)  and) enforced at runtime https://github.com/SpongePowered/SpongeCommon/blob/4339372db7f0380258225f76e326f87a0e6a1181/src/java6/java/org/spongepowered/launch/JavaVersionCheckUtils.java#L40
You can't (literally) run an official version of Sponge on an older version of java.

> Also, I don't think removing the comments, such as asking to include all mods/plugins, is a good idea. People don't follow this as is, much less so without description.

`Plugins/Mods` was moved under `I am currently running`. That's self-explanatory to me, much like asking the Sponge version you are currently running when the issue happened.
I don't want to take the road of over-explaining things because some people don't follow the issue template. That would push reporters away from sponge.

The goal is to make it simpler and easier to report bugs.

dualspiral:
> You can't (literally) run an official version of Sponge on an older version of java.

I know that. My point is that there might be other bugs in Java we _don't_ know about that cause issues. But there is a chance it _might_ - which is why it was there in the first place. 

Again, 99% of the time it won't matter. I'm just pointing out why we might want to keep it.

> That's self-explanatory to me, much like asking the Sponge version you are currently running when the issue happened

It might be to you - but you're not the target audience! 

The main problem is that those who report bugs often take it as "this is what I think is relavant" and we _really_ need to consider the user and their behaviour here, not ourselves. We need to know what all the other mods and plugins are - that's why that's in the template.

> I don't want to take the road of over-explaining things because some people don't follow the issue template

So... why don't they follow it in the first place? Well, for one, Github Issues are rather terrible for any structure, to be honest - so we're always going to get people who think they know better and will delete the template. To those, if we spot it, maybe we should just close their report as invalid and get them to remake it with the template.

I would imagine that the language barrier is another - a lot of our users are not native English speakers.

> The goal is to make it simpler and easier to report bugs.
> This should hopefully improve the quality of bug reports

I'll be honest, I don't really think this does this. It looks like a few removed fields and the movement of how to fill it out placed in a less noticible location. If it's easier for the users but doesn't give us what we need to know, then it defeats the purpose.

I suggest that we should simplify the issue template not by removing fields, but by making each piece of information a short phrase that is easily translatable in Google Translate, or asks what we need to know without making people dart around, [perhaps something like this](https://gist.github.com/dualspiral/bdd8bb632bab77873270a0c8957bda23). Maybe go futher to make all the headers questions so it becomes a pseudo dialog? That way, it becomes obvious, as part of the template, rather than extra explanation.

ImMorpheus:
> It might be to you - but you're not the target audience!
> 
> The main problem is that those who report bugs often take it as "this is what I think is relavant" and we really need to consider the user and their behaviour here, not ourselves. We need to know what all the other mods and plugins are - that's why that's in the template.

That's why I asked for input (which you are providing) from other devs (and possibly the community)

> So... why don't they follow it in the first place? Well, for one, Github Issues are rather terrible for any structure, to be honest - so we're always going to get people who think they know better and will delete the template. 

I had people complaining the template was confusing which led to them ignoring it and improvising

> I would imagine that the language barrier is another - a lot of our users are not native English speakers.

Yeah, there's also that. Part of the changes are aimed to reduce the length of the fields description, to make it easier for non-english speaker.

> I'll be honest, I don't really think this does this. It looks like a few removed fields and the movement of how to fill it out placed in a less noticible location. If it's easier for the users but doesn't give us what we need to know, then it defeats the purpose.

> I suggest that we should simplify the issue template not by removing fields, but by making each piece of information a short phrase that is easily translatable in Google Translate, 

I'll be honest too, I think it does exactly that. 
You already provided a good example on why we should keep the OS field (which I agreed on). No examples of the java field being (statistically) relevant were provided and I still think it adds noise to the template.
The whole "removing fields" looks like cherry-picking to me, I even added a field for the log since a big part of the issues were missing that.

> or asks what we need to know without making people dart around, perhaps something like this. Maybe go futher to make all the headers questions so it becomes a pseudo dialog? That way, it becomes obvious, as part of the template, rather than extra explanation.

That's fine by me and by looking at your example it seems we're both moving toward a common direction (small and simple questions, no overly detailed descriptions, etc.).
Looking back at it, the "additional information" stuff is probably better suited for the docs (so that it could be localized).

> To those, if we spot it, maybe we should just close their report as invalid and get them to remake it with the template.

I have no problem with it, this is something I'd enforce right away.

dualspiral:
> Part of the changes are aimed to reduce the length of the fields description, to make it easier for non-english speaker.

As a note, to provide more explanation as to why I took the discursive route and why I think shortening phrases is not a good idea in itself - one of the few good things I learned on a course I recently took on learning how to teach, of all things, particularly when you have "english as an additional language" students. 

The reason why you'd want to increase the length of a two word field to a discursive phrase is so that non-English speakers can connect to simple verbs and actions and try to gather from context, rather than just a single word they won't understand at all. It also means that a phrase can get translated with context by a machine, giving a better chance that a user might understand what we're asking for if they need to use such assistance.

> The whole "removing fields" looks like cherry-picking to me, I even added a field for the log since a big part of the issues were missing that.

Sure, okay you did add an explicit field, but we did already ask for a log, and my point is what you did was simply move stuff about and reduce what we're asking of people. And that's why I didn't think what you did is going to help in itself - we have to think very carefully about whether reducing what we ask from them would help us - and I don't think it does. It might simplify what we ask of them, but could end up increasing our workload. 

Also, I must say once again, removing the java version field is not something I'm against, but I'm trying to point out why we _might_ need that information. As I said above:

> I'm not saying that it matters 99% of the time - but there are reasons why you might request the information. However, this on itself is not a reason to just keep the requests, we could always request more information if we feel it is relavent.

We just need to make sure that we don't cause more work for ourselves by over-simplifying. 

What I suggested did come to mind from reading your original response - simplification doesn't always mean fewer words in a field, or fewer fields. Simplification is also about making things easier to understand for those who might not otherwise understand. So, simplification in this case should mean better description of what we're asking for without going too far.

As long as we can still get the relavant information from our issues (and collectively decide on what that is), and increase our success rate in doing so, then that's the goal. I just think that we should do that by being clear in what we ask. But you're right in saying that blocks of text are useless - who wants to read that when their server is on fire and they want to report a severe bug?

Issue #2156, opened 2018-12-06
Fix needed for chunk savestate dupe glitch


tenten8401:
**I am currently running**
- Minecraft: 1.12.2
- SpongeAPI: 7.1.0-4f54ce0d
- Sponge: 1.12.2-7.1.5-SNAPSHOT
- SpongeForge: 1.12.2-2768-7.1.5-RC3506
- Minecraft Forge: 14.23.5.2768

**Issue Description**
There's apparently a dupe bug in vanilla minecraft that needs to be fixed, I don't know if there's an issue for it already but I couldn't seem to find one. It allows the player to prevent a chunk from saving, that way when they unload & load the chunks, it reverts back to the old state.
https://youtu.be/uw7vEGhKoH8?t=100

Paper Spigot has a patch for it already that basically just tries to check the book size and kicks the player off if it's too large.
https://github.com/PaperMC/Paper/blob/master/Spigot-Server-Patches/0403-Book-Size-Limits.patch

randombyte-developer:
Updated link: https://github.com/PaperMC/Paper/blob/master/Spigot-Server-Patches/0403-Book-Size-Limits.patch

gabizou:
Someone can write a PR for this.

Shybella:
I tested this dupe at one point and wasn't successful. Have you attempted this?

tenten8401:
@Shybella I had some players successfully do it on my Spongeforge server.

zand:
I don't think https://github.com/SpongePowered/SpongeCommon/commit/0f8df35e74e1e1cce7974220e8c5e1df68acd9fa fixes the problem describe in the video as they state its due to compressed chunks being limited to 1MB in size. They discuss different ways they thought of to trigger the bug starting with "shulker boxes" and only did it with books because it seemed the easiest.

A proper fix would be to get rid of that limit by doing something like saving over sized chunks to a separate file until it gets fixed in vanilla.

zand:
A video popped up of players abusing this bug without books https://youtu.be/-xDODejcM6c

Aaron1011:
@zand: A full fix can be found in the latest Forge build.

PR #2399, opened 2019-10-18
Migrate several patches to SV since they're now in Forge proper.

### LINKED ISSUE #3003: Just Installed SpongeForge and server wont start
I am currently running the newest 1.12.2 version of sponge forge through my hosted server company. And when I installed I recieved this crash. Heres the log:

[sponge crash.txt](https://github.com/SpongePowered/SpongeForge/files/3722190/sponge.crash.txt)

JBYoshi:
**SpongeCommon** | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/421)

All these patches were added to Forge just over a year ago in MinecraftForge/MinecraftForge#5160, so they no longer need to be applied to Forge.

Note that the configuration options are still available on Forge. Not sure whether they should be marked as SpongeVanilla-only.

Fixes https://github.com/SpongePowered/SpongeForge/issues/3003.


ImMorpheus:
`NetHandlerPlayServerMixin_LimitBookSize` is also useless. It was added by https://github.com/SpongePowered/SpongeCommon/commit/0f8df35e74e1e1cce7974220e8c5e1df68acd9fa to fix https://github.com/SpongePowered/SpongeCommon/issues/2156.

Forge fixed the core issue. (see https://github.com/MinecraftForge/MinecraftForge/blob/1.12.x/patches/minecraft/net/minecraft/world/chunk/storage/RegionFile.java.patch)

largeorbitalobject:
How do we migrate these patches into our Sponge?

JBYoshi:
@ImMorpheus Moved that patch and rebased on latest.

dualspiral:
> Not sure whether they should be marked as SpongeVanilla-only.

If you could just update the configuration comments to state that they only affect SpongeVanilla, then that should suffice for now, I would think. There is a wider question about config along with SF/SV specific options, but that's not really for solving here.

Otherwise, this is a straight move of code from SC to SV. Quick look suggests no issues - but I imagine you've throughly tested it!

PR #2400, opened 2019-10-19
Implement HealEntityEvent.

### LINKED ISSUE #2345: HealEntityEvent not working for players base health regeneration?
Dear Sponge-Team,

I am currently migrating my Spigot plugin to Sponge and figured out that the HealEntityEvent didn't get fired, when a player regenerates health. Is this event currently not working or isn't it triggering for the base player health regeneration?

**---EDIT 1---**
The event also get not triggered when a player consumes a health potion

SpongeForge Version: 1.12.2-2838-7.1.7-RC3884
SpongeApi Version: 7.1.0

Regards
Sascha

JBYoshi:
This is a basic implementation of `HealEntityEvent`. The original and final health works, but I'm not sure how to handle the `HealthModifier` list. It looks like it was copied from `DamageModifier`, but Minecraft has no built-in counterpart for healing. For now, I'm just going to leave that empty by default.

Test code (drop this in a plugin):
```java
    @Listener
    public void onHeal(HealEntityEvent e) {
        e.getCause().first(MessageReceiver.class).ifPresent(x -> {
            x.sendMessage(Text.of("Attempted to heal by " + e.getFinalHealAmount()));
        });
        e.setBaseHealAmount(5.0);
        e.getCause().first(MessageReceiver.class).ifPresent(x -> {
            x.sendMessage(Text.of("Will now heal by " + e.getFinalHealAmount()));
        });
    }
```

Fixes #2345.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/EntityLivingBaseMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/EntityLivingBaseMixin.java
@@ -65,6 +65,7 @@ import org.spongepowered.api.event.cause.entity.damage.DamageModifier;
 import org.spongepowered.api.event.cause.entity.damage.source.FallingBlockDamageSource;
 import org.spongepowered.api.event.entity.ChangeEntityEquipmentEvent;
 import org.spongepowered.api.event.entity.DamageEntityEvent;
+import org.spongepowered.api.event.entity.HealEntityEvent;
 import org.spongepowered.api.event.entity.MoveEntityEvent;
 import org.spongepowered.api.event.item.inventory.UseItemStackEvent;
 import org.spongepowered.api.item.inventory.Inventory;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/core/entity/EntityLivingBaseMixin.java
```diff
@@ -1169,5 +1172,16 @@ private void onResetActiveHand(final CallbackInfo ci) {
     }
 
     // End implementation of UseItemStackEvent
+    
+    @ModifyVariable(method = "heal", index = 1, at = @At("HEAD"))
+    private float sponge$fireHealEntityEvent(float original) {
```
gabizou, line 1177: I'd rather not start stubbing bare implementation events when we can properly have a real event that will actually work. None of the methods, objects, or types are being supported by simply throwing an event here, so it stands to confusion why the API says one thing, and the implementation is flat out implementing it "blindly" without the remaining objects to support the event.

Zidane:
Closing this as this isn't the direction I want to go to for data changes.

Issue #2396, opened 2019-10-10
Events are fired even when entity is dead


aromaa:
**I am currently running**

- SpongeForge version: 
- SpongeVanilla version: 1.12.2-7.1.7-RC277
- Java version: 1.8.0_211/64bit
- Operating System: Windows 10

**Issue Description**
Entity and player action related events can be fired even when the entity is dead. This seems to be due to checking isDead flag, which is only set after the player has been ticked once and declared dead due to health reaching zero. Because the network layer is being ticked first there can be packets that lead to players death and then after that also handle player move packets.

For example, if you fall to your death and two move packets are being handled in one tick where the first one is where you hit the ground and the server calls death event and the second one is regular move as you are not declared dead on the client until the servers says so.

Also some other player action related events are handled, for example interacts ones that seems to be lacking the isDead flag check.

This can lead to plugin state corruption and cause unintended consequences. I had a case where the player was revived unintentionally, leading to bad client & server state.

aromaa:
After looking at this more the [EntityPlayerMPMixin](https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/mixin/core/entity/player/EntityPlayerMPMixin.java#L223) is completely overwriting the onDeath() method.

EntityLivingBase seems to also have variable called "dead", which is set upon calling onDeath. This is seprate from the isDead boolean which is set after the entity has ticked.

aromaa:
I have come up with few solutions but all of them have drawback, I'm not sure what is acceptable from mods standpoint so I hope we can get some discussion going on how to do this.

1. Check entity health, this is easy one. If the entity's health is zero or below, do not fire events. However some entities have zero health from the beginning, like projectiles which would require some special handling. This seems most straightforward but could cause some issues.

2. For EntityLivingBase, check for the dead boolean. This is vanilla field and mods should be modifying this one in case of trying to bring the entity from death. However, should for all the other entity types use the isDead? Seems fine to me and is most likely more reliable one.

3. The worst thing we could do is introduce new boolean thats Sponge specific and has no meaning to other mods and plugins. This could cause a lot of issues when the player is indeed brought back from the dead intentionally

PR #2401, opened 2019-10-22
Don't process packets sent by dead players

LXGaming:
This is a potential fix for #2396 

The first fix is for the EntityPlayerMixin to ensure a consistent isDead state, When a player dies the isDead flag is set to true however when the player relogs without respawning this flag isn't updated on login resulting in a player whose health is 0 and isn't marked as dead.

The method has also been renamed from `recalculateXpOnLoad` -> `onReadEntityFromNBT` as it's no longer just recalculating xp.

The second fix is for the PacketPhaseUtil as this is where all packets end up getting processed on the Server Thread, The idea is that when a player dies the only thing they can realistic do is Respawn or Disconnect and as such all packets that aren't a Respawn packet are simply ignored.

The KeepAlive packet is processed on the Netty Thread so it will continue to function.

I have put the logic for this before the try-with-resources I'm not sure if this correct? My reasoning is that there is no point in Sponge attempting to track these packets if they are going to be ignored anyway.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketPhaseUtil.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketPhaseUtil.java
@@ -166,16 +166,10 @@ public final class PacketPhaseUtil {
     public static void onProcessPacket(final Packet packetIn, final INetHandler netHandler) {
         if (netHandler instanceof NetHandlerPlayServer) {
             EntityPlayerMP packetPlayer = ((NetHandlerPlayServer) netHandler).player;
+            // Don't process packets from a player when they are dead.
             if (!packetPlayer.isEntityAlive()) {
-                // Don't process packets from a player when they are dead.
-                if (!(packetIn instanceof CPacketClientStatus)) {
-                    return;
-                }
-
-                CPacketClientStatus clientStatusPacket = (CPacketClientStatus) packetIn;
-
                 // Allow the player to respawn.
-                if (clientStatusPacket.getStatus() != CPacketClientStatus.State.PERFORM_RESPAWN) {
+                if (!(packetIn instanceof CPacketClientStatus) || ((CPacketClientStatus) packetIn).getStatus() != CPacketClientStatus.State.PERFORM_RESPAWN) {
                     return;
                 }
             }
--- a/src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketPhaseUtil.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketPhaseUtil.java
@@ -46,6 +46,7 @@ import net.minecraft.network.Packet;
 import net.minecraft.network.datasync.DataParameter;
 import net.minecraft.network.play.client.CPacketClientSettings;
 import net.minecraft.network.play.client.CPacketClientStatus;
+import net.minecraft.network.play.client.CPacketCustomPayload;
 import net.minecraft.network.play.client.CPacketPlayer;
 import net.minecraft.network.play.server.SPacketSetSlot;
 import net.minecraft.util.EnumHand;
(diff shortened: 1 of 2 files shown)


aromaa:
To note this partially fixes #2396.

The second issue is that isDead is set to true after the player has ticked, meaning that isDead stays as false while the packets are processing. For EntityLivingBase it takes 20 ticks for the entity to be declared as isDead due to the death timer.

LXGaming:
@isokissa3 Thanks for pointing that out, didn't realize the isDead flag was delayed by the deathTime.

I've changed the check from `isDead` to `isEntityAlive()` which fixes the death time issue as the method checks isDead and also the players health. I've also tested with a plugin listening to the MoveEntityEvent and from my testing the event doesn't get fired after the entity has "died" (When it's health is at 0).

src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketPhaseUtil.java
```diff
@@ -165,8 +165,22 @@ public static boolean allTransactionsInvalid(final List<SlotTransaction> slotTra
...
     public static void onProcessPacket(final Packet packetIn, final INetHandler netHandler) {
         if (netHandler instanceof NetHandlerPlayServer) {
+            EntityPlayerMP packetPlayer = ((NetHandlerPlayServer) netHandler).player;
+            if (!packetPlayer.isEntityAlive()) {
+                // Don't process packets from a player when they are dead.
+                if (!(packetIn instanceof CPacketClientStatus)) {
+                    return;
+                }
+
+                CPacketClientStatus clientStatusPacket = (CPacketClientStatus) packetIn;
```
Zidane: You could likely collapse it down to one if, to lower the diff.

src/main/java/org/spongepowered/common/mixin/core/entity/player/EntityPlayerMixin.java
```diff
@@ -310,9 +310,14 @@ private void postEventAndUpdateExperience(final int finalExperience) {
     }
 
     @Inject(method = "readEntityFromNBT", at = @At("RETURN"))
-    private void recalculateXpOnLoad(final NBTTagCompound compound, final CallbackInfo ci) {
+    private void onReadEntityFromNBT(final NBTTagCompound compound, final CallbackInfo ci) {
```
Zidane: Change this to impl$readEntityFromNBT.

gabizou: I'm honestly of the opinion of naming it appropriately: `impl$recalculateExpOnReadNBT`.

src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketPhaseUtil.java
LXGaming: I've collapsed it slightly, I've left the `isEntityAlive` if statement as it easier to read if it's separate and it allows additional logic to be added in future if needed

src/main/java/org/spongepowered/common/mixin/core/entity/player/EntityPlayerMixin.java
LXGaming: Renamed to `impl$fixEntityProperties`

src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketPhaseUtil.java
```diff
@@ -165,8 +165,16 @@ public static boolean allTransactionsInvalid(final List<SlotTransaction> slotTra
     @SuppressWarnings({"rawtypes", "unchecked"})
     public static void onProcessPacket(final Packet packetIn, final INetHandler netHandler) {
         if (netHandler instanceof NetHandlerPlayServer) {
+            EntityPlayerMP packetPlayer = ((NetHandlerPlayServer) netHandler).player;
```
gabizou, line 169:
```suggestion
        if (!((NetHandlerPlayServer) netHandler).player.isEntityAlive() 
           || (!(packetIn instanceof CPacketClientStatus) 
               || ((CPacketClientStatus) packetIn).getStatus() != CPacketClientStatus.State.PERFORM_RESPAWN))) {
           return;        
        }
```

Pretty sure this'll make the whole lazy evaluation of the case where the player is dead, and the packet is a respawn packet, it'll continue. if the player is dead and the packet is not a respawn packet, it'll return.

LXGaming: The reason the packetPlayer variable is there is because I moved it out of the try-with-resources so it wasn't grabbing it twice. It's later passed to the StackFrame
https://github.com/SpongePowered/SpongeCommon/blob/f730075b39e8a22bad75391a1e948fef66333042/src/main/java/org/spongepowered/common/event/tracking/phase/packet/PacketPhaseUtil.java#L168-L178

LXGaming: I've collapsed the if statement as you've shown with the exception of the packetPlayer variable being removed as it's required later in the code.

ImMorpheus (approved):
Working as intended.

ImMorpheus:
@LXGaming Does this block mod provided packets sent after the player is dead ?

That's something we should probably avoid.

LXGaming:
Tested with JEI and the cheat item functionality still worked, Looks like Forge is injecting its NetworkDispatcher into the netty pipeline so its before the vanilla packet handler, so no this doesn't appear to block mod packets.

Might have to allow the CPacketCustomPayload just in case a mod calls `checkThreadAndEnqueue`.

Issue #2384, opened 2019-09-08
Shift-clicking in Crafting doesn't give custom item set in CraftedItemEvent.Preview


ratquaza:
**I am currently running**

- SpongeForge version: 7.1.0
- Forge version: 1.12.2-14.23.5.2838
- Java version: 1.8.0_221
- Operating System: Windows 10

- Plugins/Mods: SpongeForge and my own plugin

**During the `CraftItemEvent.Preview` event, when setting a custom ItemStack to be the preview stack and then shift-clicking to craft the item, the previewed item returns back to the original ItemStack. This only happens when shift-clicking, but not when clicking normally. [Here is the code](https://gist.github.com/BaitoY/d38fa32babf0980131c76c24e7151aea), and [here is a video showing what happens](https://drive.google.com/open?id=1g1OLET_3mMrRjaPUoDArfmQiBvMmsknu)**

PR #2414, opened 2019-11-09
Fix capturedCraftPreviewTransactions leaking when shift-crafting

ginyai:
An incorrect event will be fired by dragging items after shift-crafting.
And with codes like below, shift-crafting will not stop until [the crafting matrix is empty](            https://github.com/SpongePowered/SpongeCommon/blob/f7bde6a121d20b0f043b0d96ffd5edca711e103a/src/main/java/org/spongepowered/common/mixin/core/inventory/SlotCraftingMixin.java#L179).
Code
```java
    @Listener(order = Order.FIRST, beforeModifications = true)
    public void onCraft(CraftItemEvent.Preview event) {
        SlotTransaction craftTransaction = event.getPreview();
        craftTransaction.setCustom(craftTransaction.getFinal());
    }
```
Video
https://drive.google.com/file/d/1TWLuJrTkLygO1JWmqo4NdX4TcwDVFSgx/view?usp=sharing

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/inventory/ContainerMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/inventory/ContainerMixin.java
@@ -49,7 +49,6 @@ import org.spongepowered.api.item.inventory.Inventory;
 import org.spongepowered.api.item.inventory.InventoryArchetype;
 import org.spongepowered.api.item.inventory.ItemStackSnapshot;
 import org.spongepowered.api.item.inventory.crafting.CraftingInventory;
-import org.spongepowered.api.item.inventory.crafting.CraftingOutput;
 import org.spongepowered.api.item.inventory.query.QueryOperationTypes;
 import org.spongepowered.api.item.inventory.transaction.SlotTransaction;
 import org.spongepowered.api.item.inventory.type.CarriedInventory;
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/mixin/core/inventory/ContainerMixin.java
```diff
@@ -513,17 +512,6 @@ private ItemStack redirectTransferStackInSlot(final Container thisContainer, fin
                 result = ItemStack.EMPTY; // Return empty to stop shift-crafting
             }
         }
-
-        final Inventory craftInv = ((Inventory) thisContainer).query(QueryOperationTypes.INVENTORY_TYPE.of(CraftingInventory.class));
```
gabizou, line 517: Just want some clarity, how is deleting this changing behavior? Wouldn't technically the preview transactions no longer be populated in this case?

ginyai: Lines here was added in https://github.com/SpongePowered/SpongeCommon/commit/a3fac249b400056ca2544372dd0f01a9c28aa280 for https://github.com/SpongePowered/SpongeCommon/issues/2384.
But I find that the event is fired. but the SlotTransaction created at https://github.com/SpongePowered/SpongeCommon/blob/f7bde6a121d20b0f043b0d96ffd5edca711e103a/src/main/java/org/spongepowered/common/mixin/core/inventory/SlotCraftingMixin.java#L183 is not added to the SlotTransctionList. So the custom result will not be handled at https://github.com/SpongePowered/SpongeCommon/blob/f7bde6a121d20b0f043b0d96ffd5edca711e103a/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java#L1465

clienthax:
Can confirm this resolves the additional crafting output.

PR #2422, opened 2019-11-19
Use level_sponge.dat.old when unable to read level_sponge.dat

dualspiral:
This PR does two things:

* Try to read world data from `level_sponge.dat.old` if the `level_sponge.dat` file cannot be read (before, we just threw an exception if the file couldn't be read, meaning that when the `level_sponge.dat` file is saved, probably empty, the corrupted file is moved to `level_sponge.dat.old`, overwriting the potentially good backup)
* Rename any corrupted `level_sponge.dat` files with a timestamp and corrupted suffix, enabling us to analyse what might be going wrong with them later.

I'm opening a PR for this to get a quick review on it to ensure I'm not doing anything stupid. I have tested this, but I don't want to accidently overwrite good files.

I'll merge this later this week unless there is an issue.


dualspiral:
I'm preparing to merge this now. I tested this by intentionally corrupting my NBT file (destroying the GZIP header using a hex editor) and it does load the backup correctly, as well as keeping a copy of the corrupted file so anyone using a build containing this commit or beyond will be able to provide us with the file. That way, we can potentially get a better idea over what is going on.

PR #2424, opened 2019-11-23
Fix bug where parsing invalid block states would throw an exception.

nlipiarski:
This pull request fixes a problem where if a string block state had the opening and closing brackets, but either no key value pairs between them or an entry in the list did not include an equal sign an argument out of bounds exception would be thrown.  This change adds a check for an equal sign in the raw key value string.

### DIFF
--- a/src/main/java/org/spongepowered/common/registry/type/BlockStateRegistryModule.java
+++ b/src/main/java/org/spongepowered/common/registry/type/BlockStateRegistryModule.java
@@ -70,6 +70,10 @@ public final class BlockStateRegistryModule implements CatalogRegistryModule<Blo
             final String[] propertyValuePairs = properties.split(",");
             List<Tuple<BlockTrait<?>, ? extends Comparable<?>>> foundPairs = new ArrayList<>(propertyValuePairs.length);
             for (String propertyValuePair : propertyValuePairs) {
+                if (!propertyValuePair.contains("=")) {
+                    continue;
+                }
+
                 final String name = propertyValuePair.split("=")[0];
                 final String value = propertyValuePair.split("=")[1];
                 type.getTrait(name).ifPresent(trait -> {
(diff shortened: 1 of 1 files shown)


gabizou:
How is a block state's value not a property value pair? that block state would have to be invalid in the first place...

dualspiral:
This is a bad way to "fix" this, and to be honest, it's kind of correct as it is. As of now, if you don't use a well formed ID that is of one of the forms:

```
<pluginid>:<itemid>
<pluginid>:<itemid>[key=value,key=value,...]
```

you get either an exception or an empty optional. There is an argument you should just get an empty optional. However, with your change, the following (as an example) would be accepted:

```
minecraft:dirt[key=value]
minecraft:dirt[key=value,spam]
minecraft:dirt[spam,sausage,spam,spam,bacon,spam,tomato,and,spam,key=value,]
```

and, what's more, will resolve to the same ID.

We should not accept invalid IDs, of which `minecraft:dirt[]` is one of them. I understand what you're saying, in that `minecraft:dirt[]` is the same as saying no properties on the item, but if it's invalid in the spec, it should not be accepted in the element.

There might be a case for improved error handling here though.

nlipiarski:
I see what you mean with the parsing and accepting `minecraft:dirt[]` as valid.  The real problem I was having is that when passing in something like `minecraft:dirrt[]`, the parser will currently throw an `ArrayOutOfBoundsException` which did not seem like intended behavior.  Instead of what I did originally, I think changing it to return an empty value would be better.  I'll update the pull request to reflect this.

dualspiral:
Sorry, I meant to comment a while ago but things have been busy.

Quite simply, this isn't the way to fix it. Consider the following:

* An empty optional indicates that while the input is valid and could conceivably be a valid ID in the future, it doesn't map to anything now. For example, `minecraft:dirt[foo=bar]` follows the correct form and could be a valid block state in the future, or even a mod added state (though please, modders, no), but isn't now. Thus, the input is not erroneous, it's just maps to nothing - empty optional. 
* An exception should be thrown if the input is invalid, and always will be invalid, never returning a value. `minecraft:dirt[foo]` does not contain an valid key=value pair in the state, and so will never be a block state, and is therefore never valid. This is an error, the input is not a valid input and will never map to anything - this is an exception and so an exception should be thrown.

So, the bug isn't that an exception is thrown. The issue is that a **meaningful** exception isn't thrown - the current exception exposes internal state and is gibberish without that context. Returning an empty optional is the wrong thing to do here because the input itself is invalid. 

It should probably be an `IllegalArgumentException` that is thrown.

dualspiral:
https://github.com/SpongePowered/SpongeCommon/pull/2510 suprecedes this, after this has been quiet for a while. Thank you for bringing the issue to our attention.

PR #2430, opened 2019-12-04
Command implementation - part I

dualspiral:
After much deliberation, I've started gluing our system to Brigadier. I wasn't originally going to do so, just sending packets to emulate the client completion, but sitting and thinking about how we can ~~ab~~use it meant that I could come up with a couple of clever ways to do so.

~~We will create a Sponge specific Brigadier tree that will accept `CommandCause`.~~ We'll mix in `CommandCause` into `CommandSource`. Under the hood, this will be a `ICommandSource` and (crucially) `ISuggestionProvider` - which will allow us to send completions to the client with minimal fuss.

There are a couple of things we do to sit within Brig, the big one is where we add our command context to the String Reader. We do this to allow us to pass the context to parsers without mixins and further headaches with trying to make things painless for mods that use Brig directly.

The logic putting the nodes together hasn't been written yet. I expect a lot of fun when it comes to creating "optional" nodes.

This isn't complete and so won't work as is, but it's a good way to what we're expecting. ~Custom registrars and parsers~, along with suggestion providing for external frameworks, will come later.

This PR, then, is to get it checked over by @Zidane for compliance with the standards we wish to adopt. I think I got them all, but a good check would not go amiss before merging.


src/main/java/org/spongepowered/common/mixin/core/brigadier/tree/CommandNodeMixin.java
```diff
@@ -0,0 +1,75 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.mixin.core.brigadier.tree;
+
+import com.google.common.base.Preconditions;
+import com.mojang.brigadier.CommandDispatcher;
+import com.mojang.brigadier.tree.CommandNode;
+import jdk.internal.org.objectweb.asm.Opcodes;
```
i509VCB: Hmm... Wrong Opcodes.

dualspiral: Yes, because the right ones don't seem to be importable yet. I'll update them as soon as they become importable.

gabizou: Odd, thought I had fixed the asm lib dependency...

dualspiral: Probable it's just my workspace.

PR #2438, opened 2019-12-18
Small Fixes Part 1 - HumanEntity, SpongeUserInventory, SpongeTabList

ItsDoot:


src/main/java/org/spongepowered/common/entity/living/human/HumanEntity.java
```diff
@@ -340,10 +341,11 @@ public void removeFromTabListDelayed(@Nullable final ServerPlayerEntity player,
         if (delay == 0) {
             removeTask.run();
         } else {
-            SpongeImpl.getGame().getScheduler().createTaskBuilder()
+            SpongeImpl.getGame().getAsyncScheduler().submit(Task.builder()
```
simon816: This should be executed after an exact number of ticks, so can't use async here

PR #2443, opened 2019-12-24
Fix world unloading not correctly flushing data

Yeregorix:
While I was developing a plugin that needs to delete temporary worlds I found that the .mca files where still opened after calling `Server#unloadWorld`. This causes issues when trying to delete or to move the files. Even when doing it via API methods `renameWorld` or `deleteWorld`. This PR intends to fix this.

How to reproduce the issue (before this PR):
Create a world. Load it. Send a player in the world and generate few chunks. Remove the player. Unload the world. Try to delete or rename the world using the API.

Issues fixed by this PR:
- The world rename test plugin was never loading and then unloading the world before renaming it.
- `Server#renameWorld` was silently failing without telling why.
- `Server#unloadWorld` was flushing the world's data before stopping the async light executor.
- `Server#unloadWorld` wasn't flushing the world's data if the world had its serialization behavior set to `NONE`. (Files were still opened if the serialization behavior recently changed from `AUTOMATIC` or `MANUAL` to `NONE`)

Issue not fixed by this PR (should it be ?):
- Worlds are not flushed before the `GameStoppedServerEvent` and thus, in the same way, it's not possible for a plugin to delete or move worlds files.

### DIFF
--- a/src/main/java/org/spongepowered/common/world/WorldManager.java
+++ b/src/main/java/org/spongepowered/common/world/WorldManager.java
@@ -67,14 +67,13 @@ import org.spongepowered.api.world.SerializationBehaviors;
 import org.spongepowered.api.world.WorldArchetype;
 import org.spongepowered.api.world.storage.WorldProperties;
 import org.spongepowered.common.SpongeImpl;
-import org.spongepowered.common.bridge.entity.player.EntityPlayerMPBridge;
 import org.spongepowered.common.bridge.server.MinecraftServerBridge;
 import org.spongepowered.common.bridge.server.integrated.IntegratedServerBridge;
 import org.spongepowered.common.bridge.world.DimensionTypeBridge;
-import org.spongepowered.common.bridge.world.WorldServerBridge;
-import org.spongepowered.common.bridge.world.WorldServerBridge_AsyncLighting;
 import org.spongepowered.common.bridge.world.WorldBridge;
 import org.spongepowered.common.bridge.world.WorldInfoBridge;
+import org.spongepowered.common.bridge.world.WorldServerBridge;
+import org.spongepowered.common.bridge.world.WorldServerBridge_AsyncLighting;
 import org.spongepowered.common.bridge.world.WorldSettingsBridge;
 import org.spongepowered.common.bridge.world.chunk.ChunkProviderServerBridge;
 import org.spongepowered.common.config.SpongeConfig;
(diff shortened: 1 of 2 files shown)


dualspiral:
> The world rename test plugin was never loading and then unloading the world before renaming it.

Yup. Not sure we _really_ needed all the changes you made to it, but it's a test plugin so not critical...

> Server#renameWorld was silently failing without telling why.

Yup.

> Server#unloadWorld was flushing the world's data before stopping the async light executor.

Good catch, but is there a reason you removed the try block? It seems to me all you had to do was move the async lighting if block to above the try block (i.e. did you verify that `MinecraftException` cannot be thrown? I cannot see where it would be, but I'd like to be sure about that before I click the shiny green button.)

> Server#unloadWorld wasn't flushing the world's data if the world had its serialization behavior set to NONE. (Files were still opened if the serialization behavior recently changed from AUTOMATIC or MANUAL to NONE)

Yep, I've just dug through how this works - that's okay.

> Worlds are not flushed before the GameStoppedServerEvent and thus, in the same way, it's not possible for a plugin to delete or move worlds files.

Put it separately I think. I'd like to see this in ASAP.

src/main/java/org/spongepowered/common/world/WorldManager.java
```diff
@@ -540,37 +539,35 @@ public static boolean unloadWorld(final WorldServer worldServer, final boolean c
             final WorldServerBridge mixinWorldServer = (WorldServerBridge) worldServer;
             final int dimensionId = mixinWorldServer.bridge$getDimensionId();
 
-            try {
```
dualspiral, line 543: I was just having another think about this. This try/catch/finally should be reinstated. The change in order is okay, but if the save fails, we want to clean up gracefully - removing this means that it leaves us in an inconsistent state. The catch should change to just `Exception` or even `Throwable`, I'd think.

Similarly, just been talking to @Zidane about this, we should also `try/catch (Exception)` shutting down the async lighting thread too, for similar reasons.

dualspiral (changes requested):

Yeregorix:
> Not sure we really needed all the changes you made to it, but it's a test plugin so not critical…

Yes it wasn't really needed but between the commands that weren't working before and most commands not sending any message it was pretty difficult to test things correctly without doing these changes. Now it's much easier.

src/main/java/org/spongepowered/common/world/WorldManager.java
Yeregorix: Ok so I should just change `catch (MinecraftException e)` to `catch (Exception e)` and do the same around the async lighting shutdown ?

No need to keep a `finally` block if all exceptions are caught and no exception is rethrown. All lines after the try catch blocks will be executed in any case. 

Edit: Btw there is still a try catch around `saveWorld(worldServer, true);`

dualspiral: > No need to keep a finally block if all exceptions are caught and no exception is rethrown. All lines after the try catch blocks will be executed in any case.

If some other `Error` is thrown, it might - though they tend not to be recoverable, in the case it is we should still clean up.

Yeregorix: If we stick to `Exception` you're right. Do you want to catch `Exception` or `Throwable` ? `Throwable` seems maybe a bit overkill to me.

dualspiral: Exception is fine.

dualspiral (approved):
Will merge when I get a chance - thanks!

dualspiral:
f26e447

PR #2444, opened 2019-12-24
Lower Aquatic spawn tickrate

clienthax:
Resolves some broken behavior such as single stream squid farms
https://github.com/SpongePowered/SpongeForge/issues/3014


dualspiral:
...was that really it? That made me laugh.

I'll get around to merging this once I've dealt with #2443

PR #2449, opened 2019-12-29
Upgrade to Mixin 0.8

gabizou:
**SpongeCommon** | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/3045) | SpongeVanilla
- Several mixins that were once conflicted with a bug where
CallbackInjectors were being discovered artificially, something since
fixed since 0.8's previews.
- Pending migrations of some AT entries to true accessors, there will be
more commits that perform these migrations as they are tested.

Signed-off-by: Gabriel Harris-Rouquette <gabizou@me.com>


src/main/java/org/spongepowered/common/mixin/api/mcp/tileentity/TileEntityStructureAccessor.java
```diff
@@ -22,34 +22,40 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.common.bridge.tileentity;
+package org.spongepowered.common.mixin.api.mcp.tileentity;
```
gabizou: wrong package....

src/main/java/org/spongepowered/common/mixin/api/mcp/util/CooldownTrackerMixin_API.java
```diff
@@ -24,21 +24,21 @@
...
 import org.spongepowered.asm.mixin.Shadow;
-import org.spongepowered.common.bridge.util.CooldownTracker_CooldownBridge;
 import org.spongepowered.common.bridge.util.CooldownTrackerBridge;
+import org.spongepowered.common.mixin.core.util.CooldownTracker_CooldownAccessor;
 
 import java.util.Map;
 import java.util.OptionalDouble;
 import java.util.OptionalInt;
 
+import static com.google.common.base.Preconditions.checkNotNull;
```
gabizou: dammit IntelliJ, figure out the damn import ordering.

PR #2453, opened 2020-01-02
Rewrite entity collisions module.

### LINKED ISSUE #2452: With SpongeForge installed, zomibie pigman anger doesn't work properly
**I am currently running**
- SpongeForge version: 1.12.2-2838-7.1.7
- Forge version: 14.23.5.2838
- Java version: 1.8.0_232/64-bit (Private Build)
- Operating System: Linux (4.15.0-45-generic/amd64)

- Plugins/Mods:

Plugins : Minecraft, Minecraft Coder Pack, SpongeAPI, Sponge

Mods: Minecraft, Minecraft Coder Pack, Forge Mod Loader, Minecraft Forge, SpongeAPI, Sponge, Astral Sorcery, Baubles, Better Records, Blood Magic: Alchemical Wizardry, Chameleon, Corail Tombstone, Grimoire of Gaia 3, Guide-API, Iron Chest, JourneyMap, Just Enough Calculation, Just Enough Items, LagGoggles, Mantle, MrCrayfish's Furniture Mod, QuickStore, Realistic Terrain Generation, Shadowfacts' Forgelin, SlashBlade, SpongeForge, Tinker SlashBlade, Tinkers' Construct, Waila, What Are We Looking At

**Issue Description**
Attacking one of the pigmen within many of them should make all nearby pigmen aggressive. But in fact, it looks like with SpongeForge installed, the anger spreads much slower than it does in vanilla behavior. Also, each time attacking a pigman, it only makes part of nearby pigmen aggresive, not all of nearby pigmen. Some nearby pigmen even remain neutral for a long time after the attack.

This issue makes the [zombie pigman farming](https://minecraft.gamepedia.com/Tutorials/Zombie_pigman_farming) not working properly, which relies on anger of pigmen heavily.

As to how I confirmed this issue... With Sponge installed, I just spawned dozens of zombie pigmen with spawn eggs, then I switched to survival gamemode. I attacked one of the pigmen, then it happened like what I said above. Then I tried it **without** Sponge installed, and I did exactly what I had done before, the anger spread instantly. All nearby pigmen became aggresive moving to me.

The config of sponge is all default during test.

No crash-report and relevant server log generated.

JBYoshi:
Currently, the entity collisions module limits the return values of `World.getEntitiesWithinAABB()` and related methods to no more than 8 entries. However, this breaks many assumptions in both vanilla and mods - see #2452, SpongePowered/SpongeForge#1000, and SpongePowered/SpongeForge#1214.

This PR strips down the entity collisions module to do just what it was originally written for - entity collisions.

Fixes #2452.


dualspiral:
I'm wondering if this is the right way to do it. What is the performance impact of continuously calculating AABB dimensions and entity size dimensions?

Does pigman angering happen in its own phase on the phase tracker? If so, would we be better off just using that phase to set `ignoreCollisions` to `true` and bypass this altogether? It feels less hacky then saying "the size of the AABB is larger than the largest entity".

JBYoshi:
> What is the performance impact of continuously calculating AABB dimensions and entity size dimensions?

I added some profiling code, and the dimension calculation averages about 40-60 nanoseconds for each call to `isLimitExempt()` on my computer. It can peak to about 200 ns or so during the first calls, but drops quickly.

> Does pigman angering happen in its own phase on the phase tracker?

Unfortunately, no. I believe it's part of `AttackEntityPacketState`, but that also uses one of the `getEntitiesWithinAABB` methods for some attacks.

dualspiral:
> I believe it's part of `AttackEntityPacketState`

In that case, should that ignore collisions instead? I'm trying to get a sense of where checking for this entity collisions optimisation makes sense - does it make sense during _that_ state?

JBYoshi:
It's more than just attacks. Some other cases that cause problems:
- Selectors. Distance-based selectors use `getEntitiesWithinAABB` for optimization.
- Structure blocks. This will prevent certain entities from being saved.
- Blocks that change size or position (such as farmland turning back to dirt, or pistons).

I'm wondering if it would be better to implement a whitelist of places where this behavior should take effect. The original purpose of the optimization was to limit [entity collisions](https://github.com/SpongePowered/SpongeCommon/commit/fbdedf182772129f1d1d147221c30320d607fe42). So maybe we could just replace the code for entity collisions; I think that would deal with the biggest of the performance drops.

In addition, many calls to `getEntitiesWithinAABB` just look for the closest entity that matches a predefined filter. If there was a way we could optimize for this case, it might be possible to rewrite those calls.

dualspiral:
> The original purpose of the optimization was to limit entity collisions. So maybe we could just replace the code for entity collisions; I think that would deal with the biggest of the performance drops.

We could. 

Is the code/process that we think the original intention is for in its own `PhaseState`? If so, perhaps:

* Default `ignoreCollisions` to `true`
* Set it to `false` for that specific phase state
* See what happens

If it is not, then maybe we could do that by intoducing a phase state. 

I think it would be worth keeping the optimisation code as arms-length as possible from the original target (as I think it is now) rather than rewriting specific calls for specific reasons, because there may be multiple areas that _could_ benefit from it.

src/main/java/org/spongepowered/common/mixin/entitycollisions/ChunkMixin_Collisions.java
```diff
@@ -124,4 +128,18 @@
...
     }
+
+    /**
+     * Removes the limit when the bounding box is large. Fixes bugs like
+     * https://github.com/SpongePowered/SpongeCommon/issues/2452.
+     *
+     * @author JBYoshi
+     */
+    private boolean collisionsImpl$isLimitExempt(AxisAlignedBB aabb) {
+        if (!(this.world instanceof WorldServer)) {
```
gabizou: If we are keeping this logic, I'd much rather this be checking that `((WorldBridge) this.world).bridge$isFake()`

gabizou (changes requested):

JBYoshi:
The more I look at this, the more I think the collisions module needs a complete rewrite. It looks like the original code already replaces most of the `collideWithNearbyEntities` behavior, to the point that the collision code may not even be running on an entity that's colliding with more than 8 other entities.

It might take some time, but I'll start playing around with the code and see what I can come up with.

Pulverizer:
This limit is also an issue for plugins as I attempted to use it and was wondering why my plugin wasn't detecting all the target entities. There was nothing in the JavaDocs to warn me that there is a limit in the config.

I have since switched to using `getNearbyEntities()` but I have no idea if this is actually worse for performance.

JBYoshi:
I've started on the rewrite for this. It fixes support for `/gamerule maxEntityCramming`, which should help by clearing out tightly-packed groups of entities. It also may fix some cases where `CollideEntityEvent` could be called for all AABB searches, although I haven't done any testing with that yet.

src/main/java/org/spongepowered/common/mixin/core/entity/EntityLivingBaseMixin.java
```diff
@@ -1170,4 +1171,12 @@ private void onResetActiveHand(final CallbackInfo ci) {
 
     // End implementation of UseItemStackEvent
 
+    @Redirect(method = "onLivingUpdate", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/EntityLivingBase;collideWithNearbyEntities()V"))
+    private void collisions$runCollisions(EntityLivingBase self) {
+        try (PhaseContext<?> ignored = EntityPhase.State.COLLISION.createPhaseContext()
+                .source(this).buildAndSwitch()) {
```
gabizou: Almost guaranteed that this might still be called on the client thread.

src/main/java/org/spongepowered/common/mixin/core/entity/EntityLivingBaseMixin.java
```diff
@@ -179,6 +179,7 @@
     @Shadow private boolean checkTotemDeathProtection(final DamageSource p_190628_1_) {
         return false; // SHADOWED
     }
+    @Shadow protected abstract void collideWithNearbyEntities();
```
gabizou: prefix with `shadow$`.

src/main/java/org/spongepowered/common/mixin/entitycollisions/BlockMixin_Collisions.java
```diff
@@ -37,6 +37,9 @@
 import org.spongepowered.common.config.type.WorldConfig;
 import org.spongepowered.common.mixin.plugin.entitycollisions.interfaces.CollisionsCapability;
 
+// TODO this is currently inactive - No phase states for blocks override
+// isCollision, and CollideBlockEvent is handled with entities iterating
+// through blocks, not blocks iterating through entities
```
gabizou: Then remove it.

src/main/java/org/spongepowered/common/mixin/entitycollisions/ChunkMixin_Collisions.java
```diff
@@ -99,27 +95,26 @@
                 return true;
             }
 
+            CollisionsCapability capability;
```
gabizou: Should be `final`.

gabizou (changes requested):

JBYoshi:
I've finished most of the cleanup, so this PR should be ready for a full review.

ImMorpheus (approved):
Working as intended

src/main/java/org/spongepowered/common/mixin/core/entity/EntityLivingBaseMixin.java
```diff
@@ -1170,4 +1171,16 @@ private void onResetActiveHand(final CallbackInfo ci) {
 
     // End implementation of UseItemStackEvent
 
+    @Redirect(method = "onLivingUpdate", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/EntityLivingBase;collideWithNearbyEntities()V"))
+    private void collisions$runCollisions(EntityLivingBase self) {
```
gabizou: If this is supposed to be added when the collision plugin has enabled collisions, this should be in a separate mixin that gets applied.

ImMorpheus: This is necessary to enter the `EntityCollisionState` and it should always be applied. 
The name was probably copied from the old Mixin.

JBYoshi: I've renamed it to `impl$runCollisions`.

PR #2454, opened 2020-01-02
Implement SerializationBehaviors.NONE and SerializationBehaviors.METADATA_ONLY.

### LINKED ISSUE #2291: WorldArchetype SerializationBehaviors.NONE not obeyed
**I am currently running**

- SpongeForge version: 1.12.2-2825-7.1.7-RC3768
- Java version: 1.8.0_202/64-bit (Oracle Corporation)
- Operating System: Linux (4.15.0-43-generic/amd64)

- Plugins/Mods: Tekxit 3, Minecraft, Minecraft Coder Pack, SpongeAPI, Sponge, CatClearLag, EpicBanItem, GriefPrevention, LuckPerms, MMCReboot, MMCRules, Nucleus, ServerListPlus, Total Economy, spark, worldedit-sponge

Attempting to implement regeneration for worldedit sponge,
It seems that setting the SerializationBehavior to NONE still results in files being written to the worlds folder for the new world when it is loaded.

Code in question: https://github.com/EngineHub/WorldEdit/pull/486/files
(This is currently worked around by unloading and deleting the world after)

JBYoshi:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2084) | **SpongeCommon** | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/3048)

For test code, use the existing `MultiWorldTest` plugin - the world folders should not appear.

Fixes #2291.


dualspiral:
I'd like @Zidane to look at this when he's got chance - he's the worlds guy after all!

I am just a little concerned about not writing anything at all for two reasons: 

* Some mods might rely on the data files being there - this comment makes me think we may need to write something: https://github.com/SpongePowered/SpongeCommon/issues/2291#issuecomment-501449594. Some mods will save their own data on a per world basis and this could cause them to fall over. We need to test this thoughly with some of the bigger packs before pulling this.
* You might want to persist the _world_ upon creation, but leave it in a specific state. Something like a hunger games plugin that would reset the world between games would just want to unload/load, or some "mining" world that gets reset periodically through a unload/load. I get there is `MANUAL` behaviour, but that doesn't stop other plugins, such as backup plugins, causing it to update the data when they request a save manually. 

I wonder if we need to specify a separate behaviour for these scenarios - maybe `NO_CHUNKS` or `METADATA_ONLY`, to differentiate between having no chunks saved, and no data saved at all.

Zidane:
@dualspiral really hit the nail on the head with the second part. We need to be able to specify certain aspects. Most people care about chunks, few care about the other data.

src/main/java/org/spongepowered/common/mixin/core/world/storage/SaveHandlerMixin.java
```diff
@@ -70,9 +77,45 @@
...
+
     @Nullable private Exception impl$capturedException;
     // player join stuff
     @Nullable private Path impl$file;
+    private Set<File> impl$directoriesToCreate = new HashSet<>();
+
+    @Redirect(method = "<init>", at = @At(value = "INVOKE", target = "Ljava/io/File;mkdirs()Z", remap = false))
+    private boolean impl$mkdirs(File dir) {
+        if (WorldManager.NO_FILE_CREATION) {
+            impl$directoriesToCreate.add(dir);
```
Zidane: this.

Zidane:
@JBYoshi Does Player data files (that exist outside the root save container) still get saved if logging out in a world that does not save?

src/main/java/org/spongepowered/common/mixin/core/world/storage/SaveHandlerMixin.java
JBYoshi: That basically means "if the `SerializationBehavior` is ever switched from `NONE` to something else, create these directories before saving anything." I'll change the name to something more descriptive.

JBYoshi:
@dualspiral

> Some mods might rely on the data files being there

Good point. I'll add a `METADATA_ONLY` option; I just need to test it.

> You might want to persist the world upon creation, but leave it in a specific state.

I get the part about the metadata, but just to clarify: when you say "leave it in a specific state", you mean the initial state created by the world generator, correct?

@Zidane 
> Does Player data files (that exist outside the root save container) still get saved if logging out in a world that does not save?

Yes.

JBYoshi:
It's been a while, but `METADATA_ONLY` is now added. Also fixed a bug with region files being created with `SerializationBehaviors.NONE`.

src/main/java/org/spongepowered/common/mixin/core/world/storage/SaveHandlerMixin.java
dualspiral: I feeel silly for commenting on this much later on... but he means turn this line into `this.impl$directoriesToCreate.add(dir);`

src/main/java/org/spongepowered/common/mixin/core/world/chunk/storage/RegionFileCacheMixin.java
```diff
@@ -0,0 +1,58 @@
...
+    @Shadow public static RegionFile getRegionFileIfExists(File worldDir, int chunkX, int chunkZ) {return null;}
+
+    /**
+     * @author JBYoshi
+     * @reason Support for ChunkSerializationBehaviors that don't save chunks:
+     * don't create files if they aren't necessary. (The original method
+     * already returns null if the chunk doesn't exist.)
+     */
+    @Overwrite
+    public static DataInputStream getChunkInputStream(File worldDir, int chunkX, int chunkZ) {
```
dualspiral, line 50: I don't like that this is an overwrite, to be honest, but I see why you've done it.

@Zidane Should we prefer head injection and cancel in this case? I know you've done that in the past. Could also replace with a Redirect followed by a cancellable injection - but that then wanders into LVT land.

dualspiral (commented):
I think this is pretty much there and we should be able to consider pulling soon. Just one thing that I'd like @Zidane to rule on.

dualspiral:
I did miss something before, which I'll quickly address:

> I get the part about the metadata, but just to clarify: when you say "leave it in a specific state", you mean the initial state created by the world generator, correct?

Sort of. What a server might want to do is create a world, add some chests, features, buildings, etc. then turn off the chunk saving. So, not necesasrily the inital state by the world generator, but some snapshot in time. `METADATA_ONLY` will read from, but not write to, chunks, right?

To head this off now, I can also see server owners complaining that the world will "revert" when a chunk unloads and then reloads. I'd say that's fine and expected. If they are that bothered by that, they can just find a way to keep the chunks loaded!

This looks good to me. Let's see if we can get this in this weekend before I start asking testers to test.

src/main/java/org/spongepowered/common/mixin/core/advancements/AdvancementManagerMixin.java
```diff
@@ -24,19 +24,40 @@
...
+    }
+
     @Inject(method = "reload", at = @At("RETURN"))
     private void impl$reloadAdvancementProgressforPlayerList(final CallbackInfo ci) {
         ((PlayerListBridge) SpongeImpl.getServer().getPlayerList()).bridge$reloadAdvancementProgress();
     }
+
+    @Redirect(method = "loadCustomAdvancements", at = @At(value = "INVOKE",
+            target = "Lorg/apache/commons/io/FileUtils;listFiles(Ljava/io/File;[Ljava/lang/String;Z)Ljava/util/Collection;", remap = false))
+    private Collection<File> listFiles(File directory, String[] extensions, boolean recursive) {
```
gabizou: s/`listFiles`/`impl$listFilesIfDirectoryExists`

src/main/java/org/spongepowered/common/mixin/core/advancements/FunctionManagerMixin.java
```diff
@@ -0,0 +1,57 @@
...
+@Mixin(FunctionManager.class)
+public abstract class FunctionManagerMixin {
+    @Redirect(method = "loadFunctions", at = @At(value = "INVOKE", target = "Ljava/io/File;mkdirs()Z", remap = false))
+    private boolean impl$createDirectory(File dir) {
+        return WorldManager.mkdirsIfSaveable(dir);
+    }
+
+    @Redirect(method = "loadFunctions", at = @At(value = "INVOKE",
+            target = "Lorg/apache/commons/io/FileUtils;listFiles(Ljava/io/File;[Ljava/lang/String;Z)Ljava/util/Collection;", remap = false))
+    private Collection<File> listFiles(File directory, String[] extensions, boolean recursive) {
```
gabizou: Same here

src/main/java/org/spongepowered/common/mixin/core/world/chunk/storage/RegionFileCacheMixin.java
gabizou: If it's been done where we're only checking that the region file is `null` and if it is, returning `null` as a fail fast, then it should be a cancellable injection at `HEAD`.

src/main/java/org/spongepowered/common/mixin/core/world/gen/ChunkProviderServerMixin.java
```diff
@@ -92,6 +93,8 @@
     @Shadow protected abstract void saveChunkExtraData(Chunk chunkIn);
     @Shadow protected abstract void saveChunkData(Chunk chunkIn);
 
+    @Shadow public abstract boolean canSave();
```
gabizou: prefix with `shadow$` ( I get that this is in maintenance, and `api-8` branch is moving towards prefixing shadows always, but I'd like to keep it in line here as well).

src/main/java/org/spongepowered/common/mixin/core/world/gen/ChunkProviderServerMixin.java
```diff
@@ -250,7 +253,14 @@ public void queueUnload(final Chunk chunkIn)
     @Overwrite
     public boolean tick()
     {
-        if (!this.world.disableLevelSaving && !((WorldBridge) this.world).bridge$isFake())
+        // Sponge start
+        SerializationBehavior serializationBehavior = ((WorldProperties) this.world.getWorldInfo()).getSerializationBehavior();
```
gabizou: variable can be final.

src/main/java/org/spongepowered/common/mixin/core/world/gen/ChunkProviderServerMixin.java
```diff
@@ -325,4 +335,20 @@ public boolean tick()
         this.loadedChunks.remove(ChunkPos.asLong(chunk.x, chunk.z));
         ((ChunkBridge) chunk).bridge$setScheduledForUnload(-1);
     }
+
+    // This still returns true for METADATA_ONLY because other places (e.g. WorldServer) use it.
+    @Inject(method = "canSave", at = @At("HEAD"), cancellable = true)
+    public void onCanSave(CallbackInfoReturnable<Boolean> cir) {
```
gabizou: method name needs to reflect what it's actually doing. `impl$checkSerializationBehaviorForCanSave` or something

src/main/java/org/spongepowered/common/mixin/core/world/gen/ChunkProviderServerMixin.java
```diff
@@ -325,4 +335,20 @@ public boolean tick()
...
+        if (((WorldProperties) this.world.getWorldInfo()).getSerializationBehavior() == SerializationBehaviors.NONE) {
+            cir.setReturnValue(false);
+        }
+    }
+
+    @Inject(method = "saveChunks", at = @At("HEAD"), cancellable = true)
+    public void onSaveChunks(CallbackInfoReturnable<Boolean> cir) {
```
gabizou: Same here

src/main/java/org/spongepowered/common/mixin/core/world/storage/SaveHandlerMixin.java
```diff
@@ -70,9 +77,45 @@
 
     @Shadow @Final private File worldDirectory;
 
+    @Shadow protected abstract void setSessionLock();
```
gabizou: Prefix with `shadow$`

src/main/java/org/spongepowered/common/mixin/core/world/storage/SaveHandlerMixin.java
```diff
@@ -70,9 +77,45 @@
...
+
     @Nullable private Exception impl$capturedException;
     // player join stuff
     @Nullable private Path impl$file;
+    private Set<File> impl$directoriesToCreate = new HashSet<>();
+
+    @Redirect(method = "<init>", at = @At(value = "INVOKE", target = "Ljava/io/File;mkdirs()Z", remap = false))
+    private boolean impl$mkdirs(File dir) {
```
gabizou: `mkdirs` should be actually named what we're doing *on top of making directories*.

src/main/java/org/spongepowered/common/mixin/core/world/storage/SaveHandlerMixin.java
```diff
@@ -70,9 +77,45 @@
...
+    @Redirect(method = "<init>", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/storage/SaveHandler;setSessionLock()V"))
+    private void impl$onSetSessionLock(SaveHandler self) {
+        if (!WorldManager.NO_FILE_CREATION) {
+            this.setSessionLock();
+        }
+    }
+
+    @Redirect(method = "checkSessionLock",
+        at = @At(value = "NEW", target = "java/io/FileInputStream", remap = false))
+    private FileInputStream impl$onCheckSessionLock(File file) throws FileNotFoundException {
```
gabizou: change `onCheckSessionLock` to actually mean what this method is doing.

src/main/java/org/spongepowered/common/world/WorldManager.java
```diff
@@ -432,7 +436,14 @@ public static WorldProperties createWorldProperties(final String folderName, fin
             return optWorldProperties.get();
         }
 
-        final ISaveHandler saveHandler = new AnvilSaveHandler(WorldManager.getCurrentSavesDirectory().get().toFile(), folderName, true, ((MinecraftServerAccessor) SpongeImpl.getServer()).accessor$getDataFixer());
+        NO_FILE_CREATION = archetype.getSerializationBehavior() == SerializationBehaviors.NONE;
```
gabizou: I really don't like this sort of static flag flip and check. Is it possible to add this check through an `IPhaseState.allowsFileCreation()` method and switching in?

src/main/java/org/spongepowered/common/world/WorldManager.java
```diff
@@ -623,12 +636,19 @@ public static void saveWorld(final WorldServer worldServer, final boolean flush)
...
 
         final Path worldFolder = currentSavesDir.resolve(worldName);
-        if (!Files.isDirectory(worldFolder)) {
+        if (!Files.isDirectory(worldFolder) && properties.getSerializationBehavior() != SerializationBehaviors.NONE) {
             SpongeImpl.getLogger().error("Unable to load world [{}]. We cannot find its folder under [{}].", worldFolder, currentSavesDir);
             return Optional.empty();
         }
 
-        final ISaveHandler saveHandler = new AnvilSaveHandler(currentSavesDir.toFile(), worldName, true, ((MinecraftServerAccessor) SpongeImpl.getServer()).accessor$getDataFixer());
+        NO_FILE_CREATION = properties.getSerializationBehavior() == SerializationBehaviors.NONE;
```
gabizou: Same with here.. Especially since later you do a `try` block with finally. It screams that the state should be switching in.

src/main/java/org/spongepowered/common/world/WorldManager.java
```diff
@@ -893,6 +911,39 @@ private static WorldServer createWorldFromProperties(
         }
     }
 
+    public static boolean mkdirsIfSaveable(File dir) {
```
gabizou, line 911: Does this method have to be public?

src/main/java/org/spongepowered/common/world/WorldManager.java
```diff
@@ -893,6 +911,39 @@ private static WorldServer createWorldFromProperties(
...
+        if (NO_FILE_CREATION) {
```
gabizou: If you do end up making this a `PhaseTracker.getCurrentState().allowsFileCreation()` check, is there a case where this should be checked for threadedness?

src/main/java/org/spongepowered/common/world/WorldManager.java
```diff
@@ -1299,8 +1350,8 @@ public static NBTTagCompound saveDimensionDataMap() {
         if (optWorldServer.isPresent()) {
             return Optional.of(optWorldServer.get().getSaveHandler().getWorldDirectory().toPath());
         } else if (SpongeImpl.getGame().getState().ordinal() >= GameState.SERVER_ABOUT_TO_START.ordinal()) {
-            final SaveHandler saveHandler = (SaveHandler) SpongeImpl.getServer().getActiveAnvilConverter().getSaveLoader(SpongeImpl.getServer().getFolderName(), false);
-            return Optional.of(saveHandler.getWorldDirectory().toPath());
+            MinecraftServer server = SpongeImpl.getServer();
```
gabizou: Variable should be final

testplugins/src/main/java/org/spongepowered/test/MultiWorldTest.java
```diff
@@ -40,22 +40,38 @@
 
     @Override
     public void disable(CommandSource src) {
-        Sponge.getServer().getWorld("temp").ifPresent(world -> Sponge.getServer().unloadWorld(world));
+        Sponge.getServer().getWorld("no-save").ifPresent(world -> {
```
gabizou: Pull out the `Sponge.getServer()` as a local variable at this point with how many times it's being called.

testplugins/src/main/java/org/spongepowered/test/MultiWorldTest.java
```diff
@@ -40,22 +40,38 @@
...
+            Sponge.getServer().unloadWorld(world);
+            Sponge.getServer().deleteWorld(world.getProperties());
+        });
     }
 
     @Override
     public void enable(CommandSource src) {
         try {
-            final WorldArchetype archetype = Sponge.getRegistry().getType(WorldArchetype.class, "multi-world-test:overnether").orElse(
+            final WorldArchetype archetype1 = Sponge.getRegistry().getType(WorldArchetype.class, "multi-world-test:overnether").orElseGet(() ->
```
gabizou: Why the change?

testplugins/src/main/java/org/spongepowered/test/MultiWorldTest.java
```diff
@@ -40,22 +40,38 @@
...
                     .serializationBehavior(SerializationBehaviors.NONE)
                     .generator(GeneratorTypes.OVERWORLD)
                     .build("multi-world-test:overnether", "Overnether")
             );
-            final WorldProperties worldProperties = Sponge.getServer().createWorldProperties("temp", archetype);
+            final WorldProperties world1 = Sponge.getServer().createWorldProperties("no-save", archetype1);
+            Sponge.getServer().loadWorld(world1);
 
-            Sponge.getServer().loadWorld(worldProperties);
+            final WorldArchetype archetype2 = Sponge.getRegistry().getType(WorldArchetype.class, "multi-world-test:overend").orElseGet(() ->
```
gabizou: Should be explicit with the types here for variable names, if there's an "original" and "re-generated" or "nonSerializedArchetype" make it so, otherwise this looks like obfuscated smurf naming..

gabizou (changes requested):
Nitpicks, but I think the functionality can be sanity checked with a `PhaseTracker` introduction versus a global static variable check.

src/main/java/org/spongepowered/common/world/WorldManager.java
JBYoshi: It's used in several mixins in both SpongeCommon and SpongeForge, so making it public is the best option I can see.

src/main/java/org/spongepowered/common/world/WorldManager.java
JBYoshi: Potentially (if a mod decides to save a world from another thread), although it's very unlikely. I'll add in a check for now, just in case.

src/main/java/org/spongepowered/common/mixin/core/world/chunk/storage/RegionFileCacheMixin.java
JBYoshi: It would also need a redirect - the original method calls `createOrLoadRegionFile` instead of `getRegionFileIfExists`. Is that enough to merit an overwrite, or should I use the redirect and injection?

dualspiral:
I'd like to see this go in soon - does this just require a rebase, or is there much else to do?

JBYoshi:
I've squashed and rebased this; I think it's good to go if no one else has any objections.

src/main/java/org/spongepowered/common/event/tracking/phase/general/SaveHandlerCreationPhase.java
```diff
@@ -0,0 +1,42 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.event.tracking.phase.general;
+
+public class SaveHandlerCreationPhase extends GeneralState<SaveHandlerCreationContext> {
+
+    @Override
+    protected SaveHandlerCreationContext createNewContext() {
+        return new SaveHandlerCreationContext(this);
```
gabizou, line 33: Can this potentially be pooled?

src/main/java/org/spongepowered/common/mixin/core/advancements/AdvancementManagerMixin.java
```diff
@@ -24,19 +24,40 @@
...
 import org.spongepowered.common.bridge.server.management.PlayerListBridge;
+import org.spongepowered.common.world.WorldManager;
+
+import java.io.File;
+import java.util.Collection;
 
 @Mixin(AdvancementManager.class)
 public class AdvancementManagerMixin {
 
+    @Redirect(method = "loadCustomAdvancements", at = @At(value = "INVOKE", target = "Ljava/io/File;mkdirs()Z", remap = false))
```
gabizou:
```suggestion
    @Redirect(method = "loadCustomAdvancements",
          at = @At(value = "INVOKE", target = "Ljava/io/File;mkdirs()Z", remap = false)
    )
```

src/main/java/org/spongepowered/common/mixin/core/advancements/AdvancementManagerMixin.java
```diff
@@ -24,19 +24,40 @@
...
+        return WorldManager.mkdirsIfSaveable(dir);
+    }
+
     @Inject(method = "reload", at = @At("RETURN"))
     private void impl$reloadAdvancementProgressforPlayerList(final CallbackInfo ci) {
         ((PlayerListBridge) SpongeImpl.getServer().getPlayerList()).bridge$reloadAdvancementProgress();
     }
+
+    @Redirect(method = "loadCustomAdvancements", at = @At(value = "INVOKE",
+            target = "Lorg/apache/commons/io/FileUtils;listFiles(Ljava/io/File;[Ljava/lang/String;Z)Ljava/util/Collection;", remap = false))
```
gabizou:
```suggestion
    @Redirect(method = "loadCustomAdvancements", 
        at = @At(value = "INVOKE",
             target = "Lorg/apache/commons/io/FileUtils;listFiles(Ljava/io/File;[Ljava/lang/String;Z)Ljava/util/Collection;",
             remap = false
        )
    )
```

src/main/java/org/spongepowered/common/mixin/core/advancements/FunctionManagerMixin.java
```diff
@@ -0,0 +1,57 @@
...
+
+@Mixin(FunctionManager.class)
+public abstract class FunctionManagerMixin {
+    @Redirect(method = "loadFunctions", at = @At(value = "INVOKE", target = "Ljava/io/File;mkdirs()Z", remap = false))
+    private boolean impl$createDirectory(File dir) {
+        return WorldManager.mkdirsIfSaveable(dir);
+    }
+
+    @Redirect(method = "loadFunctions", at = @At(value = "INVOKE",
+            target = "Lorg/apache/commons/io/FileUtils;listFiles(Ljava/io/File;[Ljava/lang/String;Z)Ljava/util/Collection;", remap = false))
```
gabizou: indentation

src/main/java/org/spongepowered/common/mixin/core/world/chunk/storage/RegionFileCacheMixin.java
gabizou: I'd say the overwrite is fine.

src/main/java/org/spongepowered/common/mixin/core/world/chunk/storage/RegionFileCacheMixin.java
```diff
@@ -0,0 +1,62 @@
...
+import java.io.DataInputStream;
+import java.io.File;
+
+@Mixin(RegionFileCache.class)
+public abstract class RegionFileCacheMixin {
+
+    @Shadow public static RegionFile getRegionFileIfExists(File worldDir, int chunkX, int chunkZ) {return null;}
+
+    /**
+     * @author JBYoshi
```
gabizou:
```suggestion
     * @author JBYoshi - February xx, 2020 - Minecraft 1.12.2
```

src/main/java/org/spongepowered/common/mixin/core/world/gen/ChunkProviderServerMixin.java
```diff
@@ -92,6 +93,8 @@
     @Shadow protected abstract void saveChunkExtraData(Chunk chunkIn);
     @Shadow protected abstract void saveChunkData(Chunk chunkIn);
```
gabizou: empty line.

src/main/java/org/spongepowered/common/mixin/core/world/gen/ChunkProviderServerMixin.java
```diff
@@ -250,7 +253,14 @@ public void queueUnload(final Chunk chunkIn)
     @Overwrite
     public boolean tick()
     {
-        if (!this.world.disableLevelSaving && !((WorldBridge) this.world).bridge$isFake())
+        // Sponge start
+        final SerializationBehavior serializationBehavior = ((WorldProperties) this.world.getWorldInfo()).getSerializationBehavior();
+        if (serializationBehavior != SerializationBehaviors.AUTOMATIC) {
```
gabizou:
```suggestion
        final SerializationBehavior behavior = ((WorldProperties) this.world.getWorldInfo()).getSerializationBehavior();
        if (behavior != SerializationBehaviors.AUTOMATIC) {
```

src/main/java/org/spongepowered/common/mixin/core/world/gen/ChunkProviderServerMixin.java
```diff
@@ -325,4 +335,20 @@ public boolean tick()
...
+    @Inject(method = "canSave", at = @At("HEAD"), cancellable = true)
+    public void impl$checkSerializationBehaviorForCanSave(CallbackInfoReturnable<Boolean> cir) {
+        if (((WorldProperties) this.world.getWorldInfo()).getSerializationBehavior() == SerializationBehaviors.NONE) {
+            cir.setReturnValue(false);
+        }
+    }
+
+    @Inject(method = "saveChunks", at = @At("HEAD"), cancellable = true)
+    public void impl$checkSerializationBehaviorForSaveChunks(CallbackInfoReturnable<Boolean> cir) {
+        SerializationBehavior serializationBehavior = ((WorldProperties) this.world.getWorldInfo()).getSerializationBehavior();
```
gabizou:
```suggestion
        final SerializationBehavior serializationBehavior = ((WorldProperties) this.world.getWorldInfo()).getSerializationBehavior();
```

src/main/java/org/spongepowered/common/mixin/core/world/gen/ChunkProviderServerMixin.java
```diff
@@ -325,4 +335,20 @@ public boolean tick()
...
+        if (serializationBehavior == SerializationBehaviors.NONE || serializationBehavior == SerializationBehaviors.METADATA_ONLY) {
```
gabizou:
```suggestion
        final SerializationBehavior behavior = ((WorldProperties) this.world.getWorldInfo()).getSerializationBehavior();
        if (behavior == SerializationBehaviors.NONE || behavior == SerializationBehaviors.METADATA_ONLY) {
```

src/main/java/org/spongepowered/common/mixin/core/world/storage/SaveHandlerMixin.java
```diff
@@ -70,9 +79,49 @@
...
+
     @Nullable private Exception impl$capturedException;
     // player join stuff
     @Nullable private Path impl$file;
+    private Set<File> impl$directoriesToCreate = new HashSet<>();
+
+    @Redirect(method = "<init>", at = @At(value = "INVOKE", target = "Ljava/io/File;mkdirs()Z", remap = false))
+    @SuppressWarnings({"rawtypes", "unchecked"})
+    private boolean impl$createDirectoryIfSavingFiles(File dir) {
+        IPhaseState state = PhaseTracker.getInstance().getCurrentState();
```
gabizou:
```suggestion
        final IPhaseState state = PhaseTracker.getInstance().getCurrentState();
```

src/main/java/org/spongepowered/common/mixin/core/world/storage/SaveHandlerMixin.java
```diff
@@ -70,9 +79,49 @@
...
+            this.impl$directoriesToCreate.add(dir);
+            return false;
+        }
+        return dir.mkdirs();
+    }
+
+    @Redirect(method = "<init>", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/storage/SaveHandler;setSessionLock()V"))
+    @SuppressWarnings({"rawtypes", "unchecked"})
+    private void impl$setSessionLockIfCreatingFiles(SaveHandler self) {
+        IPhaseState state = PhaseTracker.getInstance().getCurrentState();
```
gabizou:
```suggestion
        final IPhaseState state = PhaseTracker.getInstance().getCurrentState();
```

src/main/java/org/spongepowered/common/mixin/core/world/storage/SaveHandlerMixin.java
```diff
@@ -70,9 +79,49 @@
...
+    private boolean impl$createDirectoryIfSavingFiles(File dir) {
+        IPhaseState state = PhaseTracker.getInstance().getCurrentState();
+        if (!state.shouldCreateWorldDirectories(PhaseTracker.getInstance().getCurrentContext())) {
+            this.impl$directoriesToCreate.add(dir);
+            return false;
+        }
+        return dir.mkdirs();
+    }
+
+    @Redirect(method = "<init>", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/storage/SaveHandler;setSessionLock()V"))
```
gabizou:
```suggestion
    @Redirect(method = "<init>", 
        at = @At(value = "INVOKE", 
            target = "Lnet/minecraft/world/storage/SaveHandler;setSessionLock()V"
        )
    )
```

src/main/java/org/spongepowered/common/mixin/core/world/storage/SaveHandlerMixin.java
```diff
@@ -70,9 +79,49 @@
...
+        if (state.shouldCreateWorldDirectories(PhaseTracker.getInstance().getCurrentContext())) {
+            this.shadow$setSessionLock();
+        }
+    }
+
+    @Redirect(method = "checkSessionLock",
+        at = @At(value = "NEW", target = "java/io/FileInputStream", remap = false))
+    private FileInputStream impl$createSessionLockAndCreateDirectories(File file) throws FileNotFoundException {
+        if (!file.exists()) {
+            WorldProperties props = Sponge.getServer().getWorldProperties(this.worldDirectory.getName()).get();
```
gabizou:
```suggestion
            final WorldProperties props = Sponge.getServer().getWorldProperties(this.worldDirectory.getName()).get();
```

src/main/java/org/spongepowered/common/mixin/core/world/storage/SaveHandlerMixin.java
```diff
@@ -70,9 +79,49 @@
...
+    private void impl$setSessionLockIfCreatingFiles(SaveHandler self) {
+        IPhaseState state = PhaseTracker.getInstance().getCurrentState();
+        if (state.shouldCreateWorldDirectories(PhaseTracker.getInstance().getCurrentContext())) {
+            this.shadow$setSessionLock();
+        }
+    }
+
+    @Redirect(method = "checkSessionLock",
+        at = @At(value = "NEW", target = "java/io/FileInputStream", remap = false))
+    private FileInputStream impl$createSessionLockAndCreateDirectories(File file) throws FileNotFoundException {
```
gabizou:
```suggestion
    private FileInputStream impl$createSessionLockAndCreateDirectories(final File file) throws FileNotFoundException {
```

src/main/java/org/spongepowered/common/mixin/core/world/storage/SaveHandlerMixin.java
```diff
@@ -70,9 +79,49 @@
...
+        if (!state.shouldCreateWorldDirectories(PhaseTracker.getInstance().getCurrentContext())) {
+            this.impl$directoriesToCreate.add(dir);
+            return false;
+        }
+        return dir.mkdirs();
+    }
+
+    @Redirect(method = "<init>", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/storage/SaveHandler;setSessionLock()V"))
+    @SuppressWarnings({"rawtypes", "unchecked"})
+    private void impl$setSessionLockIfCreatingFiles(SaveHandler self) {
```
gabizou:
```suggestion
    private void impl$setSessionLockIfCreatingFiles(final SaveHandler self) {
```

src/main/java/org/spongepowered/common/mixin/core/world/storage/SaveHandlerMixin.java
```diff
@@ -70,9 +79,49 @@
...
+    @Shadow protected abstract void shadow$setSessionLock();
+
     @Nullable private Exception impl$capturedException;
     // player join stuff
     @Nullable private Path impl$file;
+    private Set<File> impl$directoriesToCreate = new HashSet<>();
+
+    @Redirect(method = "<init>", at = @At(value = "INVOKE", target = "Ljava/io/File;mkdirs()Z", remap = false))
+    @SuppressWarnings({"rawtypes", "unchecked"})
+    private boolean impl$createDirectoryIfSavingFiles(File dir) {
```
gabizou:
```suggestion
    private boolean impl$createDirectoryIfSavingFiles(final File dir) {
```

gabizou (changes requested):
small changes so far. will continue reviewing another time.

src/main/java/org/spongepowered/common/event/tracking/phase/general/SaveHandlerCreationPhase.java
JBYoshi: I think so. The only state in there is whether files should be created or not, and that's reset every time. I've changed the phase to extend `PooledPhaseState`.

src/main/java/org/spongepowered/common/mixin/core/world/chunk/storage/RegionFileCacheMixin.java
```diff
@@ -0,0 +1,62 @@
...
+import java.io.DataInputStream;
+import java.io.File;
+
+@Mixin(RegionFileCache.class)
+public abstract class RegionFileCacheMixin {
+
+    @Shadow public static RegionFile getRegionFileIfExists(File worldDir, int chunkX, int chunkZ) {return null;}
+
+    /**
+     * @author JBYoshi - January 2, 2020
```
gabizou: Note the MC Version (in case this can be updated for 1.14 or gets lost in transition)

gabizou (approved):
LGTM

PR #2455, opened 2020-01-02
An attempt of some world gen stuff [1.14] [Draft]

i509VCB:
Currently this just implements the FeatureConfigs into their mixins. This only covers FeatureConfigs which actually exist (an api pr will come soon to remove the non-existent ones from API and add features that may not exist in API).

Don't expect a full World Gen API yet as that will require a considerable shifting of code around the place.

This will probably take a while as I add the actual FeatureConfig details to all the configs.


Zidane:
Closing as we're not touching WGEN now until post 1.16.x

PR #2460, opened 2020-01-04
Fix custom data not persisting across respawns

ImMorpheus:
This commit only affects custom data manipulators, vanilla manipulators (being on fire, etc.) are NOT custom data.
It doesn't affect Forge/mods.
It doesn't change how custom data serialization work: if a plugin wants its data to persist over a restart it needs to implement a DataTranslator.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/server/PlayerListMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/server/PlayerListMixin.java
@@ -88,6 +88,7 @@ import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 import org.spongepowered.common.SpongeImpl;
 import org.spongepowered.common.SpongeImplHooks;
 import org.spongepowered.common.bridge.advancements.PlayerAdvancementsBridge;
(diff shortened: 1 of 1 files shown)


dualspiral:
One thing we need to test before pulling is the following (this just came to mind, forgive me for not mentioning this earlier):

* Create data, put on player.
* Logout
* Login
* Remove data from player.
* Die
* Check data still isn't on the player.

We need to make sure the data we just removed hasn't been readded - if it has, this might get a bit hairier...

ImMorpheus:
@dualspiral Tested. The data is not there.

To explain further:

    Create data, put on player -> #data is added to the player (CustomDataHolderMixin)
    Logout                     -> #data is serialized to nbt
    Login                      -> #data is read from nbt and added to the player (CustomDataHolderMixin)
    Remove data from player.   -> #data is removed from the player (CustomDataHolderMixin)
    Die                        -> #data is copied from the old player to the new player (CustomDataHolderMixin#getCustomManipulators)

dualspiral (approved):
Looks good to me. I also checked my own query - works as intended.

PR #2461, opened 2020-01-04
Change impl$callSignChangeEvent to a Redirect to support SV upgrade to 0.8

dualspiral:
SpongeCommon | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/425)

This is required for SV due to a "Critical injection failure" in the LVT. I've gone for the path of least resistance and used a surrogate on the method in question. I'm not sure why it shows up now when updating mixin, but I'll let others answer that question. I did a build/clean and setupDecW, but if I've somehow missed something, happy to fix that.

This was tested in a production/obfuscated environment, letting SV download the minecraft server. Without this patch, this is the error: https://gist.github.com/dualspiral/f4d2649e7a44922e570b03836057eaa3

@gabizou just want a sanity check to make sure I've done this right.


dualspiral:
I've changed this to a redirect to avoid problems with the LVT - note that I return a zero length array and don't mark dirty - returning a zero length array means that the original function won't update the sign data itself, but it will then mark dirty and mark it for update, as the last two lines of the previous iteration would do.

PR #2462, opened 2020-01-04
Fix new te not been set on replacing te.

### LINKED ISSUE #3035: Empty result in Pit Kiln (TerraFirmaCraft)
**I am currently running**

- SpongeForge version: 1.12.2-2838-7.1.8-RC3963
- Forge version: 1.12.2-14.23.5.2847
- Java version: 8

- Plugins/Mods: TerraFirmaCraft-1.12.2-0.27.7.107

**Issue Description**
I do everything like on a wiki - and as a result I get nothing.
This bug reproduced only with SpongeForge. With Forge-server all good.
[Original issue in TFC git.](https://github.com/TerraFirmaCraft/TerraFirmaCraft/issues/664)
Help, please =/

ginyai:
The tileentity at this line should be null, but we process the removal later. 
fixes https://github.com/SpongePowered/SpongeForge/issues/3035

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/chunk/ChunkMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/chunk/ChunkMixin.java
@@ -545,7 +545,7 @@ public abstract class ChunkMixin implements ChunkBridge, CacheKeyBridge {
             // Sponge End
             TileEntity tileentity = this.getTileEntity(pos, net.minecraft.world.chunk.Chunk.EnumCreateEntityType.CHECK);
 
-            if (tileentity == null) {
+            if (tileentity == null || (transaction != null && transaction.queuedRemoval != null)) {
                 // Sponge Start - use SpongeImplHooks for forge compatibility
                 // tileentity = ((ITileEntityProvider)block).createNewTileEntity(this.worldObj, block.getMetaFromState(state)); // Sponge
                 tileentity = SpongeImplHooks.createTileEntity(newBlock, this.world, newState);
--- a/src/main/java/org/spongepowered/common/mixin/core/world/chunk/ChunkMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/chunk/ChunkMixin.java
@@ -545,8 +545,10 @@ public abstract class ChunkMixin implements ChunkBridge, CacheKeyBridge {
             // Sponge End
             TileEntity tileentity = this.getTileEntity(pos, net.minecraft.world.chunk.Chunk.EnumCreateEntityType.CHECK);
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/core/world/chunk/ChunkMixin.java
```diff
@@ -545,7 +545,7 @@ public IBlockState setBlockState(final BlockPos pos, final IBlockState state) {
             // Sponge End
             TileEntity tileentity = this.getTileEntity(pos, net.minecraft.world.chunk.Chunk.EnumCreateEntityType.CHECK);
 
-            if (tileentity == null) {
+            if (tileentity == null || (transaction != null && transaction.queuedRemoval != null)) {
```
gabizou, line 550: Wouldn't it just make more sense to check if `existing != null`? Or are you using the transaction as a check if we're capturing in the first place before checking if the capture had captured a tile entity to be removed?

ginyai: @gabizou The capture of the removal here [L434](https://github.com/ginyai/SpongeCommon/blob/8f87754fcf62686b167cdf96dd5c23c7883adbea/src/main/java/org/spongepowered/common/mixin/core/world/chunk/ChunkMixin.java#L434).
Affect the selection of this if expression

gabizou: Ok so it's bundling the combined check that required the tile entity being queued for removal in the first place. Can you comment that in the change then (since `tileentity == null` is the only vanilla check, by adding the transaction check it's now sponge modified). I try to keep this method commented about its changes compared to the original code to avoid confusion between updates and when reading the method to see where sponge is making a change compared to the original engine code.

ginyai: @gabizou like this?
```suggestion
            // Sponge Start - Additional check for the tile entity being queued for removal.
            // if (tileentity == null) { // Sponge
            if (tileentity == null || (transaction != null && transaction.queuedRemoval != null)) {
```

gabizou: ~~Edited suggestion~~

The suggestion sounds fine, but if you're doing a `//Sponge Start`, then it'd make sense to modify the later `// Sponge Start` since Sponge's modifications are starting earlier.

gabizou (commented):
So, going to test locally for some edge cases that I know have struggled with tile entity capturing, namely  chest breaking, furnace lighting, some other mod interactions etc. and if it all works out, fantastic! I'll keep you updated later today.

PR #2469, opened 2020-01-05
Invoke command on main thread if it is invoked from off-thread

dualspiral:
It's not obvious that a command cannot be invoked off thread, and there is no obvious exception to tell you this. We have had an instance in Discord where a command appeared to just hang - it took for me to be told that the command was being run off thread to realise what the problem was.

So, I figured it might be good to add a check in here to make it obvious as to what is going wrong.

### DIFF
--- a/src/main/java/org/spongepowered/common/command/SpongeCommandManager.java
+++ b/src/main/java/org/spongepowered/common/command/SpongeCommandManager.java
@@ -57,6 +57,8 @@ import org.spongepowered.api.text.serializer.TextSerializers;
 import org.spongepowered.api.util.TextMessageException;
 import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
+import org.spongepowered.common.SpongeImpl;
+import org.spongepowered.common.SpongeImplHooks;
 import org.spongepowered.common.bridge.inventory.TrackedInventoryBridge;
 import org.spongepowered.common.event.ShouldFire;
 import org.spongepowered.common.event.tracking.phase.general.CommandPhaseContext;
@@ -75,6 +77,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import java.util.function.BiPredicate;
 import java.util.function.Function;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/command/SpongeCommandManager.java
```diff
@@ -299,6 +299,9 @@ public boolean containsMapping(CommandMapping mapping) {
 
     @Override
     public CommandResult process(CommandSource source, String commandLine) {
+        if (!Sponge.getServer().isMainThread()) {
```
gabizou: Wouldn't it make more sense to re-submit the command on the main thread? Much like how packets are handled in `NetHandlerPlayServer`?

gabizou: Granted, it'd be a slight behavioral change, but the behavior change is more of "Hey, instead of crashing we re-submit the command to be processed on the main thread."

dualspiral: We could, but given this is non-void, what would we return in this case? Or would we just block the async thread - which admittedly, I'm okay with.

ImMorpheus: Wouldn't it be better to use SpongeImplHooks#isMainThread ?

ImMorpheus: Wouldn't it be better to use `SpongeImplHooks#isMainThread` ?

dualspiral: I don't really think it matters either way, but it doesn't hurt to be defensive.

gabizou: It does matter in the sense that the `SpongeImplHooks#isMainThread` has a stupid fast optimization check (much like the phase tracker)

gabizou: I'd be fine with blocking the async task if anything. Thoughts @Zidane ?

dualspiral: I'm actually curious about that statement, considering `SpongeImplHooks#isMainThread` is this check, with a `isServerAvailable` check prepended to it. I had a look at SF and SV to see if we overwrite that check, and we don't.

What is the optimization check? Checking two booleans is never going to be faster than checking one, is it?

Zidane: Block the async thread.

gabizou: It’s an overwrite (SpongeImplHooks is our “overwrite for optimizations and forge compat”) in ThreadChecks package

Zidane: They are both equal in performance. With that said, we do try (stress try here...) to not call API methods to keep the call stacks clean.

Biggest thing to consider is the isServerAvailable check. You want that check to happen here.

Zidane: Block the async task

dualspiral: > It’s an overwrite (SpongeImplHooks is our “overwrite for optimizations and forge compat”) in ThreadChecks package

Huh, Github didn't show that when I searched. Cool - learn something new every day. Neat.

> With that said, we do try (stress try here...) to not call API methods to keep the call stacks clean.

Fair point. I was going to change it anyway - but good to keep in mind.

gabizou: > Biggest thing to consider is the isServerAvailable check. You want that check to happen here.

That's what the `SpongeImplHooks` version will do, since it keeps itself updated whether the server is available, whether the server thread itself exists as a field, all in one method call.

PR #2474, opened 2020-01-09
[API-7] Implement methods for RailDirection and LogAxis

Lignium:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2078) | **SpongeCommon**

Implements the methods that introduced by https://github.com/SpongePowered/SpongeAPI/pull/2078.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/api/mcp/block/BlockLog_EnumAxisMixin_API.java
+++ b/src/main/java/org/spongepowered/common/mixin/api/mcp/block/BlockLog_EnumAxisMixin_API.java
@@ -26,11 +26,14 @@ package org.spongepowered.common.mixin.api.mcp.block;
 
 import net.minecraft.block.BlockLog;
 import org.spongepowered.api.data.type.LogAxis;
+import org.spongepowered.api.util.Axis;
 import org.spongepowered.asm.mixin.Implements;
 import org.spongepowered.asm.mixin.Interface;
 import org.spongepowered.asm.mixin.Intrinsic;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
+
+import java.util.Optional;
 
 @Mixin(BlockLog.EnumAxis.class)
 @Implements(@Interface(iface = LogAxis.class, prefix = "log$"))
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/mixin/api/mcp/block/BlockRailBase_EnumRailDirectionMixin_API.java
```diff
@@ -48,6 +53,78 @@
...
+            case 9:
+                return Direction.NORTH;
+            case 1:
+            case 3:
+                return Direction.EAST;
+            case 4:
+            case 6:
+            case 7:
+                return Direction.SOUTH;
+            case 2:
```
gabizou: All of these magic numbers need to be Constants in the `org.spongepowered.common.util.Constants`, whether it's in a new nested class for Rail Directions or whatever.

src/main/java/org/spongepowered/common/mixin/api/mcp/block/BlockLog_EnumAxisMixin_API.java
```diff
@@ -47,6 +50,20 @@
         return shadow$getName();
     }
 
+    @Override
+    public Optional<Axis> toAxis() {
+        switch ((BlockLog.EnumAxis) (Object) this) {
```
gabizou: Would it make better sense to just stick the `o.s.a.Axis` as a bridge method on `EnumAxis`? I don't like switch statements as is.

gabizou (changes requested):

src/main/java/org/spongepowered/common/mixin/api/mcp/block/BlockLog_EnumAxisMixin_API.java
Lignium: @gabizou Еxcuse me, I didn't catch your thought. What is the `o.s.a.`? How can I avoid using `switch` other than replacing them with `if`?

dualspiral: `org.spongepowered.api`

PR #2475, opened 2020-01-12
Fix and optimize DestructEntityEvent for non-living entities

Yeregorix:
Fixes an issue where `DestructEntityEvent` was not (always) called for non-living entities removed by API `Entity#remove`. Moreover, this PR only stores the cause of the event instead of the whole event.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/api/mcp/entity/EntityMixin_API.java
+++ b/src/main/java/org/spongepowered/common/mixin/api/mcp/entity/EntityMixin_API.java
@@ -447,7 +447,7 @@ public abstract class EntityMixin_API implements org.spongepowered.api.entity.En
 
     @Override
     public void remove() {
-        this.isDead = true;
+        setDead();
     }
 
     @Override
--- a/src/main/java/org/spongepowered/common/mixin/api/mcp/entity/EntityMixin_API.java
+++ b/src/main/java/org/spongepowered/common/mixin/api/mcp/entity/EntityMixin_API.java
@@ -447,7 +447,7 @@ public abstract class EntityMixin_API implements org.spongepowered.api.entity.En
 
     @Override
     public void remove() {
-        setDead();
+        this.setDead();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/mixin/api/mcp/entity/EntityMixin_API.java
```diff
@@ -447,7 +447,7 @@ public boolean isLoaded() {
 
     @Override
     public void remove() {
-        this.isDead = true;
+        setDead();
```
gabizou: Is this an API call or an implementation (native) method call?

Yeregorix: `remove` is from the API and `setDead` is from NMS. There is an injection in `setDead` to fire the `DestructEntityEvent`.

gabizou: Then rename the `setDead()` to `shadow$setDead()`. It's easier/clearer to see where the method comes from that way.

Yeregorix: 1. It will affect many other classes extending `EntityMixin_API` and calling `setDead`.
2. `setDead` is also shadowed in `EntityMixin` and `EntityMixin_Activation`.
Should I rename everywhere or just in `EntityMixin_API` and its subclasses ?

gabizou: > 1. It will affect many other classes extending `EntityMixin_API` and calling `setDead`.

Ah, never mind then. Thought it was an overall addition.

gabizou: Just remember to always qualify field and method calls, so this should be
```suggestion
        this.setDead();
```

gabizou (approved):
Will get to merging Monday.

PR #2476, opened 2020-01-12
PhaseTracker 1.14 Upgrade

gabizou:
This leaves a lot to be desired currently as it's a WIP.

### What's the update?

The `PhaseTracker` is essentially an adaptation of pairing server-focused changes occurring in a `World` to an `Event` factory with specific `Cause` population based on the processing of a `IPhaseState`. If that's enough to make you ask "***huh??***" then it might be recommended to read a little bit about [Events and their causes](https://docs.spongepowered.org/stable/en/plugin/event/causes.html). The `PhaseTracker` is a wedged implementation of logging `BlockState`, `BlockEntity`, `Entity`, and `ItemStack` changes within a `World` during specific operations, such as:
 - An `Entity` performing work during a Game tick
 - A `Player` interacting with a block
 - A `Block` "growing" into a different `Block`
 etc.

Because of how intrinsically "wired" the system is to the engine, it needs hooks in various areas of the Game logic to say "hey, we're gonna be ticking an `Entity` here!" Currently, for `stable-7` or Minecraft 1.12, it's doing this mostly well (sans some incompatibility with some mods and their design, no fault of theirs, just differences in how the system needs to be improved), but as it's shown, it's still very much tied to the game's core logic.

#### Ok, so update the hooks then?

Of course! How could I be so blind! Just update the tens of thousands of lines that make the system... Ok, it's not really *tens* of thousands, it's closer to maybe 20k, but short of the hooks, the actual logic processes need to be reviewed and re-analyzed with caution since 1.14's block logic has changed enough that some parts of the system are needing to be re-engineered.

#### PhaseTracker per Thread

With the `PhaseTracker` being so integral to many systems in the game logic, and to support client sided effects and events, along with asynchronous world generation events, the `PhaseTracker` is redesigned to have two primary instances (`CLIENT` and `SERVER`) while the purposed `getInstance()` will return a `Thread` appropriate variant. Because of this, we'll be able to appropriately associate individual `PhaseTracker`s to pooled `PhaseContext`s via `PooledPhaseState`. What this means is that the pools of recycled contexts will be per-thread, and furthermore be sanity checked to the `PhaseTracker` instance that created them. Uniquely, this will sanity check closures of `PhaseContext`s to avoid asynchronous closures without knowing which thread originally entered said state.

#### Splitting the `PhaseTracker` associated code from all Mixins into a separate package

This is mostly an organizational effort to verify "who does what and why". When looking at [WorldServerMixin](https://github.com/SpongePowered/SpongeCommon/blob/11a0398d9f3fe12ed7cf40ff758d6f7d36dbf154/src/main/java/org/spongepowered/common/mixin/core/world/WorldServerMixin.java#L962-L994), it's difficult to read as to what this method is doing and why, except that it's interacting with a "proxy" for some reason. This, on top of several scattered areas where we have `PhaseState` entries, it becomes difficult to see what sections of code are to support the `PhaseTracker`, using the `PhaseTracker` for context population, or simply using the `PhaseTracker` for throwing events. Lastly, because testing of Sponge code without the `PhaseTracker` involvement and possible bugs/side effects, there's no "switch" to flip, except by removing the main hooks and hoping nothing else breaks in the process of those missing hooks.

Likewise, because the injections/redirects/etc. would now be in a separate Mixin bundle, we can too prefix those with `tracker$` to further identify what methods are coming from where.

So here's a laundry list of things I need to have done for the update:

- [ ] Update `BlockTransaction` to keep track of neighbor "shape updates" being requested, etc.
 - [ ] Update `SpongeBlockProxy` to be a simplified `IWorld` proxy so fewer mixins have to exist 
  - This is up for debate, and only possible in 1.13+ because of the introduction of the super interface. Since we can implement `IWorld`, the contracts for various block methods only actually require an `IWorld` instance, allowing us to better track changes and `TileEntity` additions. 
 - [ ] Review whether there's enough "hooks" in some mixins to split out into another mixing package/plugin
 - This would theoretically allow us to entirely disable the `PhaseTracker` without disabling events being thrown
 - We'd be able to improve visibility of what mixins/injections are for what purpose, when some are stubs or proxy-related injections.
- [ ] Review `BlockChangeFlag` API exposure, we have a bit we can expand on for better plugin control


src/main/java/org/spongepowered/common/event/tracking/PooledPhaseState.java
```diff
@@ -25,45 +25,62 @@
...
+        this.contextPool = pool;
     }
 
     @Override
     public final C createPhaseContext() {
-        if (this.cached != null && !this.cached.isCompleted) {
-            final C cached = this.cached;
-            this.cached = null;
-            return cached;
+        this.lock.lock();
```
gabizou: @dualspiral @Zidane I'm introducing this due to the nature of now having possibly multiple threads accessing the same values/stacks, and avoiding the cache getting corrupted, I figured this would be optimal, but I'd ask for someone more knowledgeable than I about threading.

src/main/java/org/spongepowered/common/registry/builtin/supplier/EventContextKeySupplier.java
```diff
@@ -0,0 +1,73 @@
...
+import org.spongepowered.api.world.LocatableBlock;
+import org.spongepowered.api.world.World;
+import org.spongepowered.common.event.SpongeEventContextKey;
+
+import java.util.stream.Stream;
+
+@SuppressWarnings("UnstableApiUsage")
+public class EventContextKeySupplier {
+
+    public static Stream<EventContextKey<?>> stream() {
```
gabizou, line 30: @Zidane if you can, please tell me this is the correct way of doing it for Sponge provided types?

src/main/java/org/spongepowered/common/event/tracking/PooledPhaseState.java
dualspiral: Hooray, my numerical simulation knowledge comes into play here!

This will work, but I'm concerned about the performance penalty of this. Now, I'm not completely versed in the internals of the PhaseTracker, but I know enough about it to know that this code is going to be executed hundereds, if not thousands of times per second.

Consider this happening on multiple threads. A lock must always be treated as a bottleneck, because you're saying that you know multiple threads might access this so you need them to do it one at a time. That's going to happen a lot here - if we're not careful, the performance benefit of the async world generators will become a performance deficit because phases can only be switched one at a time - and this is happening both on build up and tear down of a phase.

I think that we may need to give each instance of the phase tracker a `UUID` that allows us to identify it, and/or let each phase tracker have its own pool of objects. I don't think there will be a massive cost memory wise _really_ and we'll not have to worry about dreadful performance because of many many multi thread contexts switches a second.

(Aside: I've done this before. I used MPI in my Monte-Carlo simulations but needed everything to sync up at one point. I was too judicious with my sync points and my simulation ended up being three times slower. I also had the risk of deadlock and this affected multiple nodes in a cluster hitting many other users waiting for their code to run! I don't think we have it here but that was hard to resolve...)

dualspiral: Looking a slight bit more, the cached context is linked to the state which I'm also guessing is pooled - is the state per phase-tracker instance? If it is, you wouldn't need to lock until after the first if block which would mitigate some of the hit with locking as the cached context would also be per phase-tracker instance.

src/main/java/org/spongepowered/common/event/tracking/PooledPhaseState.java
```diff
@@ -25,45 +25,62 @@
...
+                return;
+            }
+            if (this.cached == null) {
+                // We can cache this context to recycle it if it's requested later.
+                // If there's no requests and just pushing, then it can be pushed to the
+                // deque.
+                this.cached = context;
+                return;
+            }
+            this.contextPool.push(context);
```
dualspiral: Similarly, to my comment above, if this object is per phase tracker instance, you can just lock from here.

src/main/java/org/spongepowered/common/event/tracking/PooledPhaseState.java
gabizou: > Hooray, my numerical simulation knowledge comes into play here!

Yay! Primarily why I ask, since 

> Consider this happening on multiple threads. A lock must always be treated as a bottleneck, because you're saying that you know multiple threads might access this so you need them to do it one at a time. That's going to happen a lot here - if we're not careful, the performance benefit of the async world generators will become a performance deficit because phases can only be switched one at a time - and this is happening both on build up and tear down of a phase.

Right, suffice to say that this will happen for each `Entity`, `TileEntity`, `BlockState`, `FluidState`, and `BlockState` random ticks, along with every single `NeighborNotification` and possibly `NeighborShapeUpdate`, so it's going to be "hot" to say the least.

> I think that we may need to give each instance of the phase tracker a `UUID` that allows us to identify it, and/or let each phase tracker have its own pool of objects. I don't think there will be a massive cost memory wise _really_ and we'll not have to worry about dreadful performance because of many many multi thread contexts switches a second.

Currently each `PhaseTracker` gets it's own `Thread` and I've contemplated on generating some `UUID`s, possibly Type5 based on the "thread namespace", thereby guaranteeing the same uuid constant for the server thread and client thread, but world gen threads I've yet to find where those get spun off to hook a new PhaseTracker instance to them. I made this change to avoid the pool of objects existing on the `PhaseTracker` since it did nothing with the pools except shove them to a map for debugging purposes. If we are to move to per-`PhaseTracker` context creation, then we'll also need to supply the `PhaseTracker` instance being used (which is based on thread and expectations).

> (Aside: I've done this before. I used MPI in my Monte-Carlo simulations but needed everything to sync up at one point. I was too judicious with my sync points and my simulation ended up being three times slower. I also had the risk of deadlock and this affected multiple nodes in a cluster hitting many other users waiting for their code to run! I don't think we have it here but that was hard to resolve...)

I do understand the possibilities of deadlocking (There's a relatively easy deadlock you can do with a `Feature` that spawns entities and if any methods in the `Entity` call to the parent world to get a `Chunk`, boom, instant deadlock.

> Looking a slight bit more, the cached context is linked to the state which I'm also guessing is pooled - is the state per phase-tracker instance? If it is, you wouldn't need to lock until after the first if block which would mitigate some of the hit with locking as the cached context would also be per phase-tracker instance.

Currently every state instance is considered a singleton. We don't have pooled states since the `IPhaseState` is effectively a controller while the `PhaseContext` is the model. It's possible to shove the context pooling to the `PhaseTracker` instance since the context will know which is being used when it's being "switched" in.

src/main/java/org/spongepowered/common/data/type/SpongeSpawnType.java
```diff
@@ -30,7 +30,18 @@
 
 public final class SpongeSpawnType extends SpongeCatalogType implements SpawnType {
 
-    public SpongeSpawnType(CatalogKey key) {
+    private boolean isForced = false;
+
+    public SpongeSpawnType(CatalogKey key, String name) {
```
Zidane, line 35: Delete name parameter unless it is a NamedCatalogType.

src/main/java/org/spongepowered/common/event/SpongeEventContextKey.java
```diff
@@ -27,38 +27,34 @@
...
 
     SpongeEventContextKey(SpongeEventContextKeyBuilder<T> builder) {
-        this.id = builder.id;
+        this.id = builder.key;
         this.allowed = builder.typeClass;
-        this.name = builder.name;
     }
 
-    public SpongeEventContextKey(String id, String name, Class<T> allowed) {
+    public SpongeEventContextKey(CatalogKey id, TypeToken<T> allowed) {
```
Zidane, line 45: id -> key

src/main/java/org/spongepowered/common/event/tracking/BlockChangeFlagManager.java
```diff
@@ -0,0 +1,180 @@
...
+import org.spongepowered.common.world.SpongeBlockChangeFlag;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.StringJoiner;
+
+public final class BlockChangeFlagManager {
```
Zidane, line 41: These are not catalogs, correct?

src/main/java/org/spongepowered/common/event/tracking/PhaseTracker.java
```diff
@@ -342,363 +296,77 @@ void completePhase(final IPhaseState<?> prevState) {
...
+                        .map(TileEntity::getType)
+                        .orElse(null);
                 if (type != null) {
+                    ResourceLocation id = TileEntityType.getId(type);
+                    if (id == null) {
+                        id = new ResourceLocation(source.getClass().getCanonicalName());
+                    }
                     final Map<String, Boolean> autoFixedTiles = trackerConfig.getAutoFixedTiles();
-                    final boolean contained = autoFixedTiles.containsKey(type.getId());
+                    final boolean contained = autoFixedTiles.containsKey(type.toString());
```
Zidane, line 359: type.getKey().getFormatted()

src/main/java/org/spongepowered/common/event/tracking/PhaseTracker.java
```diff
@@ -791,22 +460,22 @@ public void notifyBlockOfStateChange(final ServerWorldBridge mixinWorld, final n
...
-                notifyState.neighborChanged(((ServerWorld) mixinWorld), notifyPos, sourceBlock, sourcePos);
+                notifyState.neighborChanged(((ServerWorld) mixinWorld), notifyPos, sourceBlock, sourcePos, isMoving);
             }
         } catch (final Throwable throwable) {
             final CrashReport crashreport = CrashReport.makeCrashReport(throwable, "Exception while updating neighbours");
             final CrashReportCategory crashreportcategory = crashreport.makeCategory("Block being updated");
             crashreportcategory.addDetail("Source block type", () -> {
                 try {
-                    return String.format("ID #%d (%s // %s)", Block.getIdFromBlock(sourceBlock),
+                    return String.format("ID #%d (%s // %s)", Registry.BLOCK.getId(sourceBlock),
```
Zidane, line 507: Do not pass back IDs to crash reports, pass their keys

src/main/java/org/spongepowered/common/event/tracking/PhaseTracker.java
```diff
@@ -815,51 +484,51 @@ public void notifyBlockOfStateChange(final ServerWorldBridge mixinWorld, final n
...
+     * with the additions of cause tracking
      *
      * @param pos The position of the block state to set
      * @param newState The new state
      * @param flag The notification flags
      * @return True if the block was successfully set (or captured)
      */
     @SuppressWarnings("rawtypes")
-    public boolean setBlockState(final ServerWorldBridge mixinWorld, final BlockPos pos, final net.minecraft.block.BlockState newState, final BlockChangeFlag flag) {
+    public boolean setBlockState(final ServerWorldBridge mixinWorld, final BlockPos pos,
```
Zidane: mixinWorld -> worldBridge

src/main/java/org/spongepowered/common/event/tracking/PhaseTracker.java
```diff
@@ -871,38 +540,36 @@ public boolean setBlockState(final ServerWorldBridge mixinWorld, final BlockPos
...
+            // this.profiler.startSection("queueCheckLight");
+            world.getProfiler().startSection("queueCheckLight");
+            // this.getChunkProvider().getLightManager().checkBlock(pos);
+            world.getChunkProvider().getLightManager().checkBlock(pos);
+            // this.profiler.endSection();
+            world.getProfiler().endSection();
         }
 
-        // Sponge Start - At this point, we can stop and check for captures.
+        // Sponge Start - At this point, we can stop and check for captures;
```
Zidane, line 597: .

src/main/java/org/spongepowered/common/event/tracking/PhaseTracker.java
```diff
@@ -1169,115 +857,21 @@ public boolean spawnEntityWithCause(final World world, final Entity entity) {
      * @return True if the entity spawn is on the main thread.
      */
     public static boolean isEntitySpawnInvalid(final Entity entity) {
-        if (Sponge.isServerAvailable() && (Sponge.getServer().isMainThread() || SpongeImpl.getServer().isServerStopped())) {
+        if (Sponge.isServerAvailable() && (Sponge.getServer().onMainThread() || SpongeImpl.getServer().isServerStopped())) {
```
Zidane, line 917: Use SpongeImpl method to collapse first two branches

src/main/java/org/spongepowered/common/event/tracking/PhaseTracker.java
```diff
@@ -84,25 +85,19 @@
 import org.spongepowered.common.event.tracking.phase.tick.NeighborNotificationContext;
 import org.spongepowered.common.event.tracking.phase.tick.TickPhase;
 import org.spongepowered.common.mixin.accessor.world.server.ServerWorldAccessor;
-import org.spongepowered.common.registry.type.event.SpawnTypeRegistryModule;
+import org.spongepowered.common.registry.builtin.supplier.SpawnTypeSupplier;
```
Zidane, line 91: Do not reference supplier classes in common code. Simply do SpawnTypes.BLAH.get()

src/main/java/org/spongepowered/common/event/tracking/PooledPhaseState.java
Zidane: Could you not grant each PhaseTracker instance it's own pool and avoid the lock altogether?

src/main/java/org/spongepowered/common/event/tracking/TrackingUtil.java
```diff
@@ -142,7 +143,36 @@
         });
     public static final int WIDTH = 40;
 
-    public static void tickEntity(final net.minecraft.entity.Entity entity) {
+    public static void tickEntity(final Consumer<net.minecraft.entity.Entity> consumer, final net.minecraft.entity.Entity entity) {
+        checkArgument(entity instanceof Entity, "Entity %s is not an instance of SpongeAPI's Entity!", entity);
+        checkNotNull(entity, "Cannot capture on a null ticking entity!");
+        final EntityBridge mixinEntity = (EntityBridge) entity;
```
Zidane, line 150: mixinEntity -> entityBridge

src/main/java/org/spongepowered/common/registry/builtin/supplier/EventContextKeySupplier.java
Zidane: This would go into the Stream package and be named like the other stream classes are.

src/main/java/org/spongepowered/common/registry/builtin/supplier/SpawnTypeSupplier.java
```diff
@@ -0,0 +1,39 @@
+package org.spongepowered.common.registry.builtin.supplier;
+
+import org.spongepowered.api.CatalogKey;
+import org.spongepowered.api.event.cause.entity.spawn.SpawnType;
+import org.spongepowered.common.data.type.SpongeSpawnType;
+
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+
+public class SpawnTypeSupplier {
```
Zidane, line 10: Same here, stream package and named like the stream classes.

In all honesty, the supplier classes are simply for supplying Vanilla registry entries.

src/main/java/org/spongepowered/common/registry/builtin/supplier/SpawnTypeSupplier.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public static SpawnType FORCED = new SpongeSpawnType(CatalogKey.sponge("forced"), "Forced").forced();
```
Zidane, line 12: Is there a reason this is here? Why isn't this in the API?

src/main/java/org/spongepowered/common/event/tracking/PooledPhaseState.java
dualspiral: That was my suggestion too: https://github.com/SpongePowered/SpongeCommon/pull/2476#discussion_r365579305

src/main/java/org/spongepowered/common/event/tracking/PhaseTracker.java
gabizou: Can't if the spawn type is an internal one. It's not exposed to the API for the reason that it's internal use only but still detectable for plugins to interpret for their own uses.

Issue #2482, opened 2020-01-20
Async tasks do not repeat after undefined time


LordRalex:
As of 7.1.7 (when scheduled tasks were touched to fix other async issues), other async tasks are getting stuck and failing to repeat.

It may not happen at the first iteration of it, however it eventually does stop executing. The test code below is a way to create and see that the task stops running.

Have not determined cause yet, but putting issue in for tracking.

```java
package test.threadtest;

import com.google.inject.Inject;
import org.slf4j.Logger;
import org.spongepowered.api.event.game.state.GameStartedServerEvent;
import org.spongepowered.api.event.Listener;
import org.spongepowered.api.plugin.Plugin;
import org.spongepowered.api.scheduler.Task;

import java.util.concurrent.*;

@Plugin(
        id = "threadtest",
        name = "Threadtest",
        description = "Test"
)
public class Threadtest {

    @Inject
    private Logger logger;

    @Listener
    public void onServerStart(GameStartedServerEvent event) {
        Task task = Task.builder().execute(() -> {
            logger.info("Time: " + System.currentTimeMillis());
        }).async().interval(3, TimeUnit.SECONDS).submit(this);
    }
}
```

dualspiral:
This is a horrible bug with threading, but I think I've solved it. I'll push up a branch shortly before merging this, but my solution has three parts - one of which has come about because I actually read the javadocs this time. I'll explain in detail what is going on as I think it would be useful and interesting for people who work on threading (*ahem* _async phase tracker_[<sup>1</sup>](#f1)).

## I: Locks, Conditions and Threading Madness

In our loop, we make use of a `ReentrantLock` and an associated `Condition`. In my previous fix, I made the assumption that if I put a lock on a thread, it stops all other threads aquiring the lock.

This is a true assumption... **except** if a thread subsequently awaits a `Condition`. [From the `Condition#await` javadocs](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/Condition.html#await()):

> **The lock associated with this Condition is atomically released** and the current thread becomes disabled for thread scheduling purposes and lies dormant until one of four things happens:
>
>  *  Some other thread invokes the signal() method for this Condition and the current thread happens to be chosen as the thread to be awakened; or
>  *  Some other thread invokes the signalAll() method for this Condition; or
>  *  Some other thread interrupts the current thread, and interruption of thread suspension is supported; or
>  *  A "spurious wakeup" occurs. 
>
[...]

PR #2483, opened 2020-01-21
Fix Async tasks stalling

dualspiral:
See #2482

* Add lock around post-task method - if the condition is awaiting it doesn't have the lock.
* Add a stateChanged boolean to signal to the loop that something's changed and that it should run the loop again at its earliest convenience if it isn't waiting for signalling, as a safety net.


dualspiral:
[spongeforge-1.12.2-2838-7.1.9-RC0.jar.zip](https://github.com/SpongePowered/SpongeCommon/files/4092887/spongeforge-1.12.2-2838-7.1.9-RC0.jar.zip)

Brend-Smits:
After testing the solution presented in this pull request, I can tell that my issue has been resolved. Thanks for the quick fix! I appreciate it a lot.

PR #2484, opened 2020-01-22
Don't handle health data changes on player spawn

### LINKED ISSUE #2524: Firing ChangeDataHolderEvent for health when a player logging in
**I am currently running**

- SpongeVanilla version: 1.12.2-7.1.10
- Java version: OpenJDK 1.8.0_242
- Operating System: Kubuntu 18.04 LTS
- Plugins/Mods: None

**Issue Description**

When a player enters the server, an entity spawns for him, however, this causes intermediate health change calls, and `ChangeDataHolderEvent.ValueChange` is called. These calls are in the `EntityLivingBase` constructor and in `EntityLivingBase#readEntityFromNBT()` method.

Temporary PR that solves this problem: https://github.com/SpongePowered/SpongeCommon/pull/2484

Lignium:
Fixes a bug, the essence of which is that when a player logs in, `ChangeDataHolderEvent.ValueChange` is called, which indicates a change of health. However, these calls are made when the player’s entity is constructing and deserializing, and should not be construed as a real data change.

Resolves my issue on the forum: https://forums.spongepowered.org/t/strange-firing-of-changedataholderevent-valuechange-when-player-entry/34701

Fixes: https://github.com/SpongePowered/SpongeCommon/issues/2524

### DIFF
--- a/src/main/java/org/spongepowered/common/bridge/network/datasync/EntityDataManagerBridge.java
+++ b/src/main/java/org/spongepowered/common/bridge/network/datasync/EntityDataManagerBridge.java
@@ -28,5 +28,6 @@ import net.minecraft.network.datasync.DataParameter;
 
 public interface EntityDataManagerBridge {
 
-    <T> void setSilently(DataParameter<T> key, T value);
+    <T> void bridge$setSilently(DataParameter<T> key, T value);
+
 }
--- a/src/main/java/org/spongepowered/common/bridge/network/datasync/EntityDataManagerBridge.java
+++ b/src/main/java/org/spongepowered/common/bridge/network/datasync/EntityDataManagerBridge.java
@@ -1,0 +1,32 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.bridge.network.datasync;
+
+import net.minecraft.network.datasync.DataParameter;
+
+public interface EntityDataManagerBridge {
+
+    <T> void setSilently(DataParameter<T> key, T value);
+}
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/mixin/core/entity/EntityLivingBaseMixin.java
```diff
@@ -588,6 +590,14 @@ public boolean attackEntityFrom(final DamageSource source, final float amount) {
         }
     }
 
+    @Redirect(method = {"<init>", "readEntityFromNBT"}, at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/EntityLivingBase;setHealth(F)V"))
+    private void onInitHealth(final EntityLivingBase entity, final float health) {
```
gabizou: method name should be `impl$initializeHealthSilently`

src/main/java/org/spongepowered/common/mixin/core/entity/EntityLivingBaseMixin.java
```diff
@@ -588,6 +590,14 @@ public boolean attackEntityFrom(final DamageSource source, final float amount) {
...
+        DataParameter<Float> healthParameter = EntityLivingBaseAccessor.accessor$getHealthParameter();
```
gabizou: `final` value please.

src/main/java/org/spongepowered/common/mixin/core/entity/EntityLivingBaseMixin.java
```diff
@@ -588,6 +590,14 @@ public boolean attackEntityFrom(final DamageSource source, final float amount) {
...
+        float clampedValue = MathHelper.clamp(health, 0.0F, this.getMaxHealth());
```
gabizou: And this too.

src/main/java/org/spongepowered/common/mixin/core/network/datasync/EntityDataManagerMixin.java
```diff
@@ -59,6 +62,18 @@
 
     @Shadow protected abstract <T> EntityDataManager.DataEntry<T> getEntry(DataParameter<T> key);
 
+    @Override
+    public <T> void setSilently(DataParameter<T> key, T value) {
+        EntityDataManager.DataEntry<T> dataentry = this.getEntry(key);
```
gabizou: `final`

src/main/java/org/spongepowered/common/bridge/network/datasync/EntityDataManagerBridge.java
```diff
@@ -0,0 +1,32 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.bridge.network.datasync;
+
+import net.minecraft.network.datasync.DataParameter;
+
+public interface EntityDataManagerBridge {
+
+    <T> void setSilently(DataParameter<T> key, T value);
```
gabizou: Method name should be `bridge$setSilently`

gabizou (changes requested):
Functionality wise looks good. Please abide by the naming conventions outlined on the [wiki regarding bridges and injections](https://github.com/SpongePowered/SpongeCommon/wiki/Implementation-Policy:-Bridges-and-Mixins).

src/main/java/org/spongepowered/common/mixin/core/entity/EntityLivingBaseMixin.java
```diff
@@ -588,6 +590,14 @@ public boolean attackEntityFrom(final DamageSource source, final float amount) {
         }
     }
 
+    @Redirect(method = {"<init>", "readEntityFromNBT"}, at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/EntityLivingBase;setHealth(F)V"))
```
gabizou, line 599: Javadoc why this is being done, link to the issue if need be. on the surface it doesn't look like it's changing anything except using a more direct approach to setting health (which honestly, might affect some mods).

gabizou:
I realize your submission of the fix is related to a side-effect you're encountering and reported on the forums, but as we'd like to keep track of issues with commits and mixins, I'd like to ask if you could open an issue on SpongeCommon to simply keep discussion and reference back. I'm tentative on the merge of these changes as the potential hooks not being called for mods (by redirecting `setHealth`, you could be missing a bunch of mod injected hooks to that method).

I'm not quite sure your approach is needed here, as we might have an alternative option for resolving your issue.

Lignium:
@gabizou I will definitely create an issue a little later. I hope you will help me in resolving this issue. :)

ImMorpheus:
This and other similar cases should be covered by https://github.com/SpongePowered/SpongeCommon/pull/2580

PR #2486, opened 2020-01-25
Register and implement TropicalFishShape

i509VCB:


src/main/java/org/spongepowered/common/data/type/SpongeTropicalFishShape.java
```diff
@@ -0,0 +1,43 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data.type;
+
+import org.spongepowered.api.CatalogKey;
+import org.spongepowered.api.data.type.TropicalFishShape;
+
+public class SpongeTropicalFishShape extends SpongeEntityMetadataType<Integer> implements TropicalFishShape {
+
+    private boolean large;
```
ImMorpheus: Can you add a final modifier here ?

i509VCB: Yea I didn't catch that one. I'll fix it when I am home.

src/main/java/org/spongepowered/common/registry/builtin/stream/TropicalFishShapeStreamGenerator.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.stream.Stream;
+
+public final class TropicalFishShapeStreamGenerator {
+    private TropicalFishShapeStreamGenerator() {
+    }
+
+    public static Stream<Tuple<TropicalFishShape, Integer>> stream() {
+        return Stream.of(
+            Tuple.of(new SpongeTropicalFishShape(CatalogKey.minecraft("kob"), false, 0), 0),
```
gabizou: I'm tempted to ask for these constants to be exposed in the `Constants` class so that Data Key Processors will be able to use them (if they're supported to be changed). Thoughts?

i509VCB: So like move the stream generation to Constants or did you mean something else.

i509VCB: Okay after sleeping on it I think I understand a bit more what you meant. There is actually a variety enum. But its private static so the third AT has come.

So I will change it so it corresponds to the enum. However since all the Tropical Fish data integer data value based, literally nothing prevents someone from using mixin to add to BODY_TEXTURES (Base Texture used to make the fish)
PATTERN_TEXTURES_A (Pattern textures used on smaller fish)
PATTERN_TEXTURES_B (Pattern textures used on larger fish) and creating a mod which adds fish that can exceed these values. Even vanilla's NBT (via /data merge) allows you to exceed these values.

It's also likely possible to this with just the integer values since translations do exist.

src/main/java/org/spongepowered/common/data/type/SpongeTropicalFishShape.java
```diff
@@ -0,0 +1,43 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.data.type;
+
+import org.spongepowered.api.CatalogKey;
+import org.spongepowered.api.data.type.TropicalFishShape;
+
+public class SpongeTropicalFishShape extends SpongeEntityMetadataType<Integer> implements TropicalFishShape {
```
Zidane: Finalize the class

src/main/java/org/spongepowered/common/registry/builtin/stream/TropicalFishShapeStreamGenerator.java
Zidane: I don't want these outside these basically util classes unless the need arises at some point.

i509VCB: I saw there was a TropicalFishEntity.Type enum (why didn't I see this earlier lol), so I changed to that, although it comes at the cost of one AT but you can use the easier method to just set the variation of the fish within data.

src/main/java/org/spongepowered/common/mixin/api/mcp/entity/passive/fish/TropicalFishEntityMixin_TypeMixin_API.java
```diff
@@ -0,0 +1,37 @@
+package org.spongepowered.common.mixin.api.mcp.entity.passive.fish;
```
Zidane, line 25: License headers

Issue #2487, opened 2020-01-20
Dupe some blocks


bigenergy:
**I am currently running**

- SpongeForge version: spongeforge-1.12.2-2838-7.1.9-RC3986
- Forge version: forge-1.12.2-14.23.5.2847
- Java version: 1.8.0_211
- Operating System: Ubuntu 19.04

- Plugins/Mods: no mods

**Issue Description**
1. Put a zombie skull (Any skull)
2. Break her down
3. Drops 2 instead of 1
Works with any skull

The same problem exists with the destruction of double blocks (an example of a telescope from astral sorcery) https://github.com/HellFirePvP/AstralSorcery/issues/1170

dualspiral:
Can you check 3983 for me - and if you can't reproduce on that, then 3984? Thanks in advance.

dualspiral:
For now, we have decided to revert all changes that led up to this issue and put the work in progress onto a branch. This will not be an issue in the next build.

Thank you for your help with this.

dualspiral:
https://github.com/SpongePowered/SpongeCommon/commit/65136e797b760d37c5846ad923649f44dba87f81

Typed in the wrong issue number on the commit... it's this one

bigenergy:
@dualspiral It is still sometimes reproduced on SF-1.12.2-2838-7.1.9

dualspiral:
Not due to this bug, it isn't - the commit fixed the serious dupe bug you reported which was consistent. 

File a new one with _exact_ reproduction steps - not just linking to this issue and saying "this happens sometimes".

PR #2488, opened 2020-01-25
Fix TE capturing with mods

dualspiral:
After having to revert this branch for now, I tried to see what was going on with these tile entity issues. It turns out, [mods such as TerraFirmaCraft inspect the TE after setting a block state which may replace another TE](https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/40d0ac3709e8ea0c24a2285060e8bf310b88c164/src/main/java/net/dries007/tfc/objects/te/TEPlacedItem.java#L38-L56).

One of the previous changes returned the outgoing TE when requested at all times if there was an outgoing TE - this actually caused duplication of a lot of TEs on drop. What mods were really looking for was the _incoming_ TE, that is, the pending change. Thankfully, it seems like that works.

I've tested furances, chests and skulls, they seem to work okay. Not tried piston flip flops yet. 

TFC pit kilns work fine to a point - I can fire pots and vessels, but ingot holders disappear. Debugging shows me that the TE is populated with them like other items, so that might be another bug altogether - but it is only apparent with Sponge.

I include a test version here, in case anyone wants to test this for me: [spongeforge-1.12.2-2838-7.1.9-RC0.zip](https://github.com/SpongePowered/SpongeCommon/files/4112574/spongeforge-1.12.2-2838-7.1.9-RC0.zip)

Related issues are:

#2487
https://github.com/SpongePowered/SpongeForge/issues/3058
https://github.com/SpongePowered/SpongeForge/issues/3055
https://github.com/SpongePowered/SpongeForge/issues/3001

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/WorldMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/WorldMixin.java
@@ -595,10 +595,15 @@ public abstract class WorldMixin implements WorldBridge {
             return;
         }
         if (this.isTileMarkedForRemoval(pos) && !this.bridge$isFake()) {
-            // if (PhaseTracker.getInstance().getCurrentState().allowsGettingQueuedRemovedTiles()) {
             // We must always return the queued TE for mods
             // See https://github.com/SpongePowered/SpongeForge/issues/3001
-            cir.setReturnValue(this.getQueuedRemovedTileFromProxy(pos));
+            if (PhaseTracker.getInstance().getCurrentState().allowsGettingQueuedRemovedTiles()) {
+                cir.setReturnValue(this.getQueuedRemovedTileFromProxy(pos));
+            } else {
+                // If we're not unwinding, then a mod wants what the tile entity
+                // will be
+                cir.setReturnValue(this.getProcessingTileFromProxy(pos));
+            }
         }
 
     }
--- a/src/main/java/org/spongepowered/common/mixin/core/world/WorldMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/WorldMixin.java
@@ -595,11 +595,10 @@ public abstract class WorldMixin implements WorldBridge {
             return;
         }
         if (this.isTileMarkedForRemoval(pos) && !this.bridge$isFake()) {
-            if (PhaseTracker.getInstance().getCurrentState().allowsGettingQueuedRemovedTiles()) {
-                cir.setReturnValue(this.getQueuedRemovedTileFromProxy(pos));
-                return;
-            }
-            cir.setReturnValue(null);
+            // if (PhaseTracker.getInstance().getCurrentState().allowsGettingQueuedRemovedTiles()) {
+            // We must always return the queued TE for mods
+            // See https://github.com/SpongePowered/SpongeForge/issues/3001
+            cir.setReturnValue(this.getQueuedRemovedTileFromProxy(pos));
         }
 
     }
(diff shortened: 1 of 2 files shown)


dualspiral:
https://github.com/SpongePowered/SpongeCommon/pull/2488/files#diff-622b739e0a75ba71805e24623ef5b9c2R602-R605 is the change I've made.

gabizou:
@dualspiral I feel I totally lost track of this, going through my notifications, is this something we should re-investigate again?

Zidane:
@dualspiral What is the status of this?

dualspiral:
I don't really remember all that well - only that I had to undo some changes blood made because they caused all sorts of dupes, this is the redo of the undone changes. I also made a change here to patch the dupes caused - basically mods were expecting an immediate change to the state of the game after some interactions, while Sponge's bulk capture held off on that.

I recall @gabizou saying we have to be super careful here with multi block captures, though I don't recall exactly what the conversation was, which is why it's kind of stood here while I've been doing other things.

We can try merging it as is after next RB, get people to play with it, see what happens. It should improve compatibility with some mods, such as TerraFirmaCraft.

src/main/java/org/spongepowered/common/mixin/core/world/WorldMixin.java
```diff
@@ -595,11 +595,15 @@ public IBlockState getBlockState(final BlockPos pos) {
...
         if (this.isTileMarkedForRemoval(pos) && !this.bridge$isFake()) {
+            // We must always return the queued TE for mods
+            // See https://github.com/SpongePowered/SpongeForge/issues/3001
             if (PhaseTracker.getInstance().getCurrentState().allowsGettingQueuedRemovedTiles()) {
                 cir.setReturnValue(this.getQueuedRemovedTileFromProxy(pos));
-                return;
+            } else {
+                // If we're not unwinding, then a mod wants what the tile entity
+                // will be
+                cir.setReturnValue(this.getProcessingTileFromProxy(pos));
```
gabizou, line 605: My only reservation is that this should already be failing through to the world's existing set of tile entities.. But it may be a catch 22 case that we *should* be passing back the tile entity from the proxy..

src/main/java/org/spongepowered/common/mixin/core/world/chunk/ChunkMixin.java
```diff
@@ -584,8 +576,22 @@ public IBlockState setBlockState(final BlockPos pos, final IBlockState state) {
...
+            // Sponge start - enqueue any changes to the transaction
+            if (transaction != null) {
+                // Go ahead and log the tile being replaced, the tile being removed will be at least already notified of removal
+                transaction.queueTileSet = tileentity;
+                if (tileentity != null) {
+                    ((TileEntityBridge) tileentity).bridge$setCaptured(true);
+                    tileentity.setWorld(this.world);
+                    tileentity.setPos(pos);
+                }
+                transaction.enqueueChanges(mixinWorld.bridge$getProxyAccess(), peek.getCapturedBlockSupplier());
```
gabizou, line 591: Reasoning why this was above was due to the expectations of what the "unchanged" original intentions would otherwise perform. I'm not 100% sure that the side effectual cases that are with the tile entity being potentially null here (since the if statement that this check/logic was otherwise taken from) already performs that check.

gabizou (commented):
I'd say I'd need to run through this step by step to verify what the cases look like and how the changes change those perspectives. 

I don't know whether this will affect some state specific cases like 1-tick flip flop piston machines, but I'll double check before merging either today or sometime this weekend..

src/main/java/org/spongepowered/common/mixin/core/world/WorldMixin.java
dualspiral: >  But it may be a catch 22 case that we should be passing back the tile entity from the proxy..

This was the original problem and why I suggested this as the fix. As far as a mod is concerned, they've done their action, so the previous (now) behaviour returns the current tile entity - the old TE that a mod doesn't expect. 

From memory, TFC does their action, changes the TE and then retrieves it again. In Forge, the TE is the new one, everything is okay for them. In Sponge, the TE is the old one, and thus not as they expect the world to be. This change gives them the new TE, and allows them to continue working as expected (in theory), with their future actions also being captured.

Zidane:
Closing this unless a member from the community picks it back up.

PR #2492, opened 2020-01-27
Fix ClassCastException in ValueProcessorDelegate#offerToStore() method

### LINKED ISSUE #2491: Trying to offer Keys.EXPERIENCE_FROM_START_OF_LEVEL raises an exception
**I am currently running**

- SpongeVanilla version: spongevanilla-1.12.2-7.1.9
- Java version: OpenJDK 1.8.0_232
- Operating System: Kubuntu 18.04 LTS
- Plugins/Mods: None

**Issue Description**

I accidentally discovered that my program crashes when it stumbles upon the `Keys.EXPERIENCE_FROM_START_OF_LEVEL` data type and tries to offer it to a player. Method `<E> CompositeValueStore#offer(BaseValue<E> value)`, when passing `ImmutableBoundedValue<Integer>`, throws a `ClassCastException` thrown in the class:

https://github.com/SpongePowered/SpongeCommon/blob/07d5c7207060c7a29de0740c62662c7ca217cc02/src/main/java/org/spongepowered/common/data/util/ValueProcessorDelegate.java#L115

Stack trace:

```
java.lang.ClassCastException: org.spongepowered.common.data.value.immutable.ImmutableSpongeBoundedValue cannot be cast to org.spongepowered.api.data.value.mutable.Value
	at org.spongepowered.common.data.util.ValueProcessorDelegate.offerToStore(ValueProcessorDelegate.java:115) ~[ValueProcessorDelegate.class:1.12.2-7.1.9-RC0]
	at net.minecraft.entity.Entity.offer(SourceFile:2789) ~[vg.class:?]
	at org.spongepowered.api.data.value.mutable.CompositeValueStore.offer(CompositeValueStore.java:214) ~[CompositeValueStore.class:1.12.2-7.1.9-RC0]
	...
	at java.lang.Thread.run(Thread.java:748) [?:1.8.0_232]
```

Please note that this is the only key in the API that is targeted specifically at an __immutable value__.

I get the data using `ValueContainer#getValues​​()`, I get about 40 values *​​(at the moment this function is broken, I will later offer a PR that allows to do this)*, save it as a set, and later try to offer them back to the player, all the values ​​are processed correctly, maybe some rejected, however, it is this immutable value that generate the unwanted exception.

Lignium:
Fixes: https://github.com/SpongePowered/SpongeCommon/issues/2491

### DIFF
--- a/src/main/java/org/spongepowered/common/data/util/ValueProcessorDelegate.java
+++ b/src/main/java/org/spongepowered/common/data/util/ValueProcessorDelegate.java
@@ -98,6 +98,7 @@ public final class ValueProcessorDelegate<E, V extends BaseValue<E>> implements 
         return false;
     }
 
+    @SuppressWarnings("unchecked")
     @Override
     public DataTransactionResult offerToStore(ValueContainer<?> container, E value) {
         for (ValueProcessor<E, V> processor : this.processors) {
@@ -113,14 +114,16 @@ public final class ValueProcessorDelegate<E, V extends BaseValue<E>> implements 
                 final Optional<V> currentValueOptional = processor.getApiValueFromContainer(container);
                 if (currentValueOptional.isPresent()) {
                     V currentValue = currentValueOptional.get();
-                    Value<E> rejectedValue;
+                    ImmutableValue<?> rejectedValue;
                     if (currentValue instanceof Value<?>) {
-                        rejectedValue = (Value<E>) currentValue;
+                        Value<E> mutableCurrentValue = (Value<E>) currentValue;
+                        mutableCurrentValue.set(value);
+                        rejectedValue = mutableCurrentValue.asImmutable();
                     } else {
-                        rejectedValue = ((ImmutableValue<E>) currentValue).asMutable();
+                        ImmutableValue<E> immutableCurrentValue = (ImmutableValue<E>) currentValue;
+                        rejectedValue = immutableCurrentValue.with(value);
                     }
-                    rejectedValue.set(value);
-                    return DataTransactionResult.failResult(rejectedValue.asImmutable());
+                    return DataTransactionResult.failResult(rejectedValue);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


gabizou:
Uhh, no apivaluefromcontainer should ever return an immutable value. That's a bug. the idea here is to recycle the value usage whereas the immutable values shouldn't be constructed in the first place.

JBYoshi:
Er... is this the problem? Everything else in the `Keys` class has a mutable type parameter.

```java
public static final Key<ImmutableBoundedValue<Integer>> EXPERIENCE_FROM_START_OF_LEVEL
```

https://github.com/SpongePowered/SpongeAPI/blob/4d86523fbb3b10dc60952465313f8ab9d2a1f64b/src/main/java/org/spongepowered/api/data/key/Keys.java#L770

Lignium:
Yes, this is a problem if you try to do this:

```java
Set<ImmutableValue<?>> dataValues = player.getValues();
```

And then later:

```java
Player player;
dataValues.forEach(player::offer);
```

The bottom line is to save all the player’s data, and then try to offer each of them to the player, even if some data cannot be changed. `Player#offer()` method should quietly return a failed transaction, but instead it throwing exception...

src/main/java/org/spongepowered/common/data/util/ValueProcessorDelegate.java
```diff
@@ -109,11 +110,17 @@ public DataTransactionResult offerToStore(ValueContainer<?> container, E value)
...
             if (processor.supports(container)) {
-                final Optional<V> optional = processor.getApiValueFromContainer(container);
-                if (optional.isPresent()) {
-                    V mutable = optional.get();
-                    ((Value<E>) mutable).set(value);
-                    return DataTransactionResult.failResult(((Value<E>) mutable).asImmutable());
+                final Optional<V> currentValueOptional = processor.getApiValueFromContainer(container);
+                if (currentValueOptional.isPresent()) {
+                    V currentValue = currentValueOptional.get();
+                    Value<E> rejectedValue;
```
dualspiral: Leaving aside the argument of what `getApiValueFromContainer` should be returning for the moment, if we go ahead with this way, this should be an `ImmutableValue`. If you have an `ImmutableValue`, it'll will be converted to a `Value` then back to an `ImmutableValue` again for no particular reason. Minimise conversions please.

Lignium: I hope I understand you correctly. I made some changes in the next commit.

gabizou (approved):
Looks good to me. Will get to merging this or if @dualspiral wants to merge this in.

dualspiral (approved):
It's fine by me if this wants to go in for the next RB, it's clearly not going to break anything. If I don't get to it first, feel free to merge (and whoever does, please also bump both SF and SV).

PR #2495, opened 2020-01-30
Mass BlockEntity fixup [1.14]

i509VCB:
Note this PR might not be ready yet but is mostly complete, of course leave comments if something seems odd or out of place.

This includes:
Implementation of new BlockEntities
Replaces dataManipulator methods with new system similar to one in use with entities right now
Handles renames of some keys such as REPRESENTED_PLAYER -> GAME_PROFILE

ComparatorType -> ComparatorMode
PlayerHead -> Skull (since it's generic now).

no flard


src/main/java/org/spongepowered/common/util/Constants.java
```diff
@@ -707,7 +707,7 @@ public static int blockPosToInt(final BlockPos pos) {
         public static final Supplier<DyeColor> DEFAULT_SHEEP_COLOR = DyeColors.WHITE;
         public static final Supplier<DyeColor> DEFAULT_SHULKER_COLOR = DyeColors.PURPLE;
         public static final Supplier<EntityType<Pig>> DEFAULT_SPAWNER_ENTITY = EntityTypes.PIG;
-        public static final Supplier<ComparatorType> DEFAULT_COMPARATOR_TYPE = ComparatorTypes.COMPARE;
+        public static final Supplier<ComparatorMode> DEFAULT_COMPARATOR_TYPE = ComparatorModes.COMPARE;
```
Zidane: DEFAULT_COMPARATOR_MODE

gabizou (approved):

PR #2504, opened 2020-02-14
Fix EntityHit distance calculation

Yeregorix:
This PR fixes an issue when working with `World#getIntersectingEntities`:
Currently, `EntityHit#getDistance` returns the distance between the entity and the intersection of the ray with the entity's chunk, instead of the distance between the entity and the start of the ray.


dualspiral:
I was confused as to why you were also getting rid of `chunkStart` and had to spend a bit of time going through the impl, but I've got it!

To make it clearer, what's going on here is that the ray that is passed to the `ChunkMixin#bridge$getIntersectingEntities` method is no longer clamped at the chunk boundaries. This allowed for a change to the distance parameter that is passed in, as the implementation is now checking to see how far from the start of the full ray the intersection is, not how far from the chunk boundary.

I think this looks okay, but I'd like a second opinion @gabizou, in case there is something funky that this was doing that I didn't catch in my review by clamping the ray to chunk boundaries, or if you'd like the original impl to stay and just pass the extra information down instead. I don't think there is anything wrong with this though.

gabizou (approved):
Looks good. Tested locally, makes sense. The original implementation of this was down the git history where I didn't even really write it, but I've authored it by commit due to it being moved around a fair bit with mixins changing between renaming and project splitting.

PR #2506, opened 2020-02-15
Replace craft preview transaction list with a single entry.

### LINKED ISSUE #2505: Issue with the `CraftItemEvent.Preview`
**I am currently running**

- SpongeForge version: spongeforge-1.12.2-2838-7.1.10-RC3994
- SpongeVanilla version: 
- Java version: 1.8.0_231
- Operating System: windows 10

- Plugins/Mods: only a test plugin with a listener
```java
    @Listener
    public void onCraft(CraftItemEvent.Preview event) {
        event.getTransactions().forEach(slotTransaction -> slotTransaction.setCustom(slotTransaction.getFinal()));
    }
```

**Issue Description**
When drag in crafting slots then release the left button, the output slot will show the craft as if there is only the first crafting slot is set.
images:
N4M`9Y%XGNZZ49AW{683PST
S})A$Q10UA6`IHL9_WOPTUJ
QTMSWXWFUBSS9KB8VL{~S)J

The event is fired form DragInventoryStopState
https://github.com/SpongePowered/SpongeCommon/blob/72a1d111b0aeceacf74f7595c26454f7e16a2737/src/main/java/org/spongepowered/common/event/tracking/phase/packet/drag/DragInventoryStopState.java#L61-L76

I add a breakpoint and find that the captured `previewTransactions` is more one.
In the first image, it is like:
1. air -> iron_nugget*9
2. iron_nugget*9 -> air
3. air -> air
4. air -> air

But we choose the first transaction as craft result.

JBYoshi:
Basically everything that uses the crafting transaction list relies on it having only a single entry. However, when the player drags in the crafting table, `slotChangedCraftingGrid` is called multiple times, putting multiple entries into the list. This PR merges these calls into a single transaction to prevent issues like this.

Fixes #2505.

### DIFF
--- a/src/main/java/org/spongepowered/common/config/category/LoggingCategory.java
+++ b/src/main/java/org/spongepowered/common/config/category/LoggingCategory.java
@@ -69,8 +69,9 @@ public class LoggingCategory extends ConfigCategory {
     private boolean logEntityCollisionChecks = false;
     @Setting(value = "entity-speed-removal", comment = "Whether to log entity removals due to speed")
     private boolean logEntitySpeedRemoval = false;
-    @Setting(value = "transaction-merge-fail", comment = "Log when two conflicting changes are merged into one. (This number specifies the maximum "
-            + "number of messages to log. Set to 0 to show all messages.)")
+    @Setting(value = "transaction-merge-fail",
+            comment = "Log when two conflicting changes are merged into one. (This number specifies the maximum number of"
+                    + "\nmessages to log. Set to 0 to show all messages.)")
     private int logTransactionMergeFailure = 25;
     @Setting(value = "world-auto-save", comment = ""
             + "Log when a world auto-saves its chunk data.\n"
--- a/src/main/java/org/spongepowered/common/config/category/LoggingCategory.java
+++ b/src/main/java/org/spongepowered/common/config/category/LoggingCategory.java
@@ -69,8 +69,9 @@ public class LoggingCategory extends ConfigCategory {
     private boolean logEntityCollisionChecks = false;
     @Setting(value = "entity-speed-removal", comment = "Whether to log entity removals due to speed")
     private boolean logEntitySpeedRemoval = false;
-    @Setting(value = "transaction-merge-fail", comment = "Log when two conflicting changes are merged into one")
-    private boolean logTransactionMergeFailure = true;
+    @Setting(value = "transaction-merge-fail", comment = "Log when two conflicting changes are merged into one. (This number specifies the maximum "
+            + "number of messages to log. Set to 0 to show all messages.)")
+    private int logTransactionMergeFailure = 25;
     @Setting(value = "world-auto-save", comment = ""
             + "Log when a world auto-saves its chunk data.\n"
             + "Note: This may be spammy depending on the auto-save-interval configured for world.")
@@ -188,7 +189,7 @@ public class LoggingCategory extends ConfigCategory {
         this.logEntitySpeedRemoval = flag;
     }
 
-    public boolean logTransactionMergeFailure() {
+    public int logTransactionMergeFailure() {
         return this.logTransactionMergeFailure;
     }
 
--- a/src/main/java/org/spongepowered/common/config/category/LoggingCategory.java
+++ b/src/main/java/org/spongepowered/common/config/category/LoggingCategory.java
@@ -69,6 +69,8 @@ public class LoggingCategory extends ConfigCategory {
     private boolean logEntityCollisionChecks = false;
     @Setting(value = "entity-speed-removal", comment = "Whether to log entity removals due to speed")
     private boolean logEntitySpeedRemoval = false;
+    @Setting(value = "transaction-merge-fail", comment = "Log when two conflicting changes are merged into one")
+    private boolean logTransactionMergeFailure = true;
     @Setting(value = "world-auto-save", comment = ""
             + "Log when a world auto-saves its chunk data.\n"
             + "Note: This may be spammy depending on the auto-save-interval configured for world.")
@@ -184,6 +186,10 @@ public class LoggingCategory extends ConfigCategory {
 
     public void setLogEntitySpeedRemoval(boolean flag) {
         this.logEntitySpeedRemoval = flag;
+    }
+
+    public boolean logTransactionMergeFailure() {
+        return this.logTransactionMergeFailure;
     }
 
     public boolean logWorldAutomaticSaving() {
--- a/src/main/java/org/spongepowered/common/event/tracking/phase/packet/drag/DragInventoryStopState.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/phase/packet/drag/DragInventoryStopState.java
@@ -65,7 +65,7 @@ public abstract class DragInventoryStopState extends NamedInventoryState {
 
         Inventory craftInv = ((Inventory) player.openContainer).query(QueryOperationTypes.INVENTORY_TYPE.of(CraftingInventory.class));
         if (craftInv instanceof CraftingInventory) {
-            SlotTransaction previewTransaction = ((ContainerBridge) player.openContainer).bridge$getPreviewTransaction();
+            final SlotTransaction previewTransaction = ((ContainerBridge) player.openContainer).bridge$getPreviewTransaction();
             if (previewTransaction != null) {
                 CraftingRecipe recipe = SpongeCraftingRecipeRegistry
                         .getInstance().findMatchingRecipe(((CraftingInventory) craftInv).getCraftingGrid(), ((World) player.world)).orElse(null);
--- a/src/main/java/org/spongepowered/common/event/tracking/phase/packet/drag/DragInventoryStopState.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/phase/packet/drag/DragInventoryStopState.java
@@ -24,6 +24,7 @@
  */
 package org.spongepowered.common.event.tracking.phase.packet.drag;
 
+import com.google.common.collect.ImmutableList;
 import net.minecraft.entity.player.EntityPlayerMP;
 import net.minecraft.network.Packet;
 import org.spongepowered.api.item.inventory.Inventory;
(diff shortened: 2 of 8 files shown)


src/main/java/org/spongepowered/common/event/tracking/phase/packet/drag/DragInventoryStopState.java
```diff
@@ -64,13 +65,13 @@ public static void unwindCraftPreview(InventoryPacketContext context) {
 
         Inventory craftInv = ((Inventory) player.openContainer).query(QueryOperationTypes.INVENTORY_TYPE.of(CraftingInventory.class));
         if (craftInv instanceof CraftingInventory) {
-            List<SlotTransaction> previewTransactions = ((ContainerBridge) player.openContainer).bridge$getPreviewTransactions();
-            if (!previewTransactions.isEmpty()) {
+            SlotTransaction previewTransaction = ((ContainerBridge) player.openContainer).bridge$getPreviewTransaction();
```
gabizou: (nitpick) `final` this value please.

src/main/java/org/spongepowered/common/event/tracking/phase/packet/inventory/PlaceRecipePacketState.java
```diff
@@ -79,15 +80,14 @@ public void unwind(final InventoryPacketContext context) {
             return;
         }
 
-        final List<SlotTransaction> previewTransactions = ((ContainerBridge) player.openContainer).bridge$getPreviewTransactions();
-        if (previewTransactions.isEmpty()) {
+        SlotTransaction previewTransaction = ((ContainerBridge) player.openContainer).bridge$getPreviewTransaction();
```
gabizou: (nitpick) `final` this variable please

src/main/java/org/spongepowered/common/mixin/core/inventory/ContainerMixin.java
```diff
@@ -431,26 +424,35 @@ private void beforeSlotChangedCraftingGrid(final InventoryCraftResult output, fi
...
         final SlotAdapter slot = this.impl$getAdapters().get(index);
-        this.impl$capturedCraftPreviewTransactions.add(new SlotTransaction(slot, orig, repl));
+        if (this.impl$capturedCraftPreviewTransaction == null) {
+            this.impl$capturedCraftPreviewTransaction = new SlotTransaction(slot, orig, repl);
+        } else if (this.impl$capturedCraftPreviewTransaction.getSlot().equals(slot)
+                && this.impl$capturedCraftPreviewTransaction.getFinal().equals(orig)) {
+            this.impl$capturedCraftPreviewTransaction = new SlotTransaction(slot, this.impl$capturedCraftPreviewTransaction.getOriginal(), repl);
+        } else {
+            SlotTransaction replace = new SlotTransaction(slot, orig, repl);
+            SpongeImpl.getLogger().warn("Could not merge craft preview transactions - some events may break (original {}, replace {})",
```
gabizou: Could this be configurable?

src/main/java/org/spongepowered/common/mixin/core/inventory/SlotCraftingMixin.java
```diff
@@ -174,23 +175,21 @@ private void afterTake(final EntityPlayer thePlayer, final ItemStack stack, fina
         ((ContainerBridge) container).bridge$setFirePreview(true);
         this.impl$craftedStack = null;
 
-        final List<SlotTransaction> previewTransactions = ((ContainerBridge) container).bridge$getPreviewTransactions();
+        SlotTransaction previewTransaction = ((ContainerBridge) container).bridge$getPreviewTransaction();
```
gabizou: (nitpick) `final` this variable please

gabizou (approved):
Nitpicks aside, looks good to me

JBYoshi:
@gabizou Updated.

src/main/java/org/spongepowered/common/config/category/LoggingCategory.java
```diff
@@ -179,6 +181,10 @@ public void setLogEntitySpeedRemoval(boolean flag) {
         this.logEntitySpeedRemoval = flag;
     }
 
+    public boolean logTransactionMergeFailure() {
```
dualspiral: I know @gabizou said "could this be configurable?" and attached it to the log line, but I'm not sure this is what he meant - I think we should _always_ be reporting a failure to the log.

Was the meaning that this entire behaviour on failure should be configurable - as in, original vs replacement?

gabizou: Well this and whether silencing after X number of failed/merged transactions happens.

JBYoshi: I can understand the configuration to "silence after X messages", but why would the behavior of the failure need to be configurable? I can't imagine a server owner changing that.

src/main/java/org/spongepowered/common/config/category/LoggingCategory.java
```diff
@@ -64,6 +64,9 @@
     private boolean logEntityCollisionChecks = false;
     @Setting(value = "entity-speed-removal", comment = "Whether to log entity removals due to speed")
     private boolean logEntitySpeedRemoval = false;
+    @Setting(value = "transaction-merge-fail", comment = "Log when two conflicting changes are merged into one. (This number specifies the maximum "
+            + "number of messages to log. Set to 0 to show all messages.)")
```
gabizou: Should this comment be multi-lined to roughly 100 characters?

src/main/java/org/spongepowered/common/mixin/core/inventory/ContainerMixin.java
```diff
@@ -431,26 +425,43 @@ private void beforeSlotChangedCraftingGrid(final InventoryCraftResult output, fi
         final ItemStackSnapshot repl = ItemStackUtil.snapshotOf(output.getStackInSlot(index));
 
         final SlotAdapter slot = this.impl$getAdapters().get(index);
-        this.impl$capturedCraftPreviewTransactions.add(new SlotTransaction(slot, orig, repl));
+        if (this.impl$capturedCraftPreviewTransaction == null) {
+            this.impl$capturedCraftPreviewTransaction = new SlotTransaction(slot, orig, repl);
+        } else if (this.impl$capturedCraftPreviewTransaction.getSlot().equals(slot)
+                && this.impl$capturedCraftPreviewTransaction.getFinal().equals(orig)) {
```
gabizou, line 435: Are you doing the correct equality check here? Or are you aiming to just allow instance equality checking the `ItemStackSnapshot`?

src/main/java/org/spongepowered/common/mixin/core/inventory/ContainerMixin.java
```diff
@@ -431,26 +425,43 @@ private void beforeSlotChangedCraftingGrid(final InventoryCraftResult output, fi
         final ItemStackSnapshot repl = ItemStackUtil.snapshotOf(output.getStackInSlot(index));
 
         final SlotAdapter slot = this.impl$getAdapters().get(index);
-        this.impl$capturedCraftPreviewTransactions.add(new SlotTransaction(slot, orig, repl));
+        if (this.impl$capturedCraftPreviewTransaction == null) {
+            this.impl$capturedCraftPreviewTransaction = new SlotTransaction(slot, orig, repl);
+        } else if (this.impl$capturedCraftPreviewTransaction.getSlot().equals(slot)
```
gabizou, line 434: Off the top of my head, do `SlotAdapter`s implement `equals`?

src/main/java/org/spongepowered/common/mixin/core/inventory/ContainerMixin.java
```diff
@@ -431,26 +425,43 @@ private void beforeSlotChangedCraftingGrid(final InventoryCraftResult output, fi
...
+                && this.impl$capturedCraftPreviewTransaction.getFinal().equals(orig)) {
+            this.impl$capturedCraftPreviewTransaction = new SlotTransaction(slot, this.impl$capturedCraftPreviewTransaction.getOriginal(), repl);
+        } else {
+            SlotTransaction replace = new SlotTransaction(slot, orig, repl);
```
gabizou:
```suggestion
            final SlotTransaction replace = new SlotTransaction(slot, orig, repl);
```

src/main/java/org/spongepowered/common/mixin/core/inventory/ContainerMixin.java
```diff
@@ -431,26 +425,43 @@ private void beforeSlotChangedCraftingGrid(final InventoryCraftResult output, fi
...
+            ContainerMixin.impl$numTransactionErrorsLogged++;
+            int maxLogs = SpongeImpl.getGlobalConfigAdapter().getConfig().getLogging().logTransactionMergeFailure();
```
gabizou:
```suggestion
            final int maxLogs = SpongeImpl.getGlobalConfigAdapter().getConfig().getLogging().logTransactionMergeFailure();
```

src/main/java/org/spongepowered/common/mixin/core/inventory/ContainerMixin.java
```diff
@@ -119,7 +120,8 @@
     //private boolean postPreCraftEvent = true; // used to prevent multiple craft events to fire when setting multiple slots simultaneously
     private List<SlotTransaction> impl$capturedSlotTransactions = new ArrayList<>();
     private List<SlotTransaction> impl$capturedCraftShiftTransactions = new ArrayList<>();
-    private List<SlotTransaction> impl$capturedCraftPreviewTransactions = new ArrayList<>();
+    @Nullable private SlotTransaction impl$capturedCraftPreviewTransaction;
+    private static int impl$numTransactionErrorsLogged = 0;
```
gabizou: Static fields should be in their own body imo. Make it well aware they are static.

gabizou (commented):
Some nitpicks

Zidane:
@JBYoshi Get this updated please.

src/main/java/org/spongepowered/common/mixin/core/inventory/ContainerMixin.java
JBYoshi: It has to be `equals()` - `orig` is created fresh each time the method is called.

src/main/java/org/spongepowered/common/mixin/core/inventory/ContainerMixin.java
JBYoshi: Not currently. It seems to be an easy implementation - it just requires checking the underlying inventory and the index - so I'll do that.

dualspiral:
@JBYoshi Is this good to go? Anything we need to be aware of if we merge it?

JBYoshi:
I think it should be good to merge, if there aren't any other objections.

dualspiral:
I'm sure someone will complain if it doesn't work.

PR #2514, opened 2020-02-24
More core plugin updates

i509VCB:


src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -897,19 +900,20 @@ public static boolean handleCollideImpactEvent(final net.minecraft.entity.Entity
             final Optional<User> owner = PhaseTracker.getInstance().getCurrentContext().getOwner();
             owner.ifPresent(user -> frame.addContext(EventContextKeys.OWNER, user));
 
-            final Location<World> impactPoint = new Location<>((World) projectile.world, VecHelper.toVector3d(movingObjectPosition.hitResult));
+            final Location<World> impactPoint = new Location<>((World) projectile.world, VecHelper.toVector3d(movingObjectPosition.getHitVec()));
             boolean cancelled = false;
 
             if (movingObjectType == RayTraceResult.Type.BLOCK) {
-                final BlockPos blockPos = movingObjectPosition.getBlockPos();
+                BlockRayTraceResult blockMovingObjectPosition = (BlockRayTraceResult) movingObjectPosition;
```
gabizou:
```suggestion
                final BlockRayTraceResult blockMovingObjectPosition = (BlockRayTraceResult) movingObjectPosition;
```

src/main/java/org/spongepowered/common/mixin/core/entity/projectile/ThrowableEntityMixin.java
```diff
@@ -47,8 +47,7 @@
     @Shadow protected LivingEntity owner;
     @Shadow protected abstract void onImpact(RayTraceResult movingObjectPosition);
 
-    @Nullable
-    public ProjectileSource projectileSource;
+    @Nullable public ProjectileSource impl$projectileSource;
```
gabizou: Shouldn't be public at all, at the most `protected` if it's being used in sub mixins.

src/main/java/org/spongepowered/common/mixin/core/item/EnderEyeItemMixin.java
```diff
@@ -113,14 +114,14 @@ private void implThrowForPreEvent(final World worldIn, final PlayerEntity player
      */
     @SuppressWarnings("Duplicates")
     @Surrogate
-    private void implThrowForPreEvent(final World worldIn, final PlayerEntity playerIn, final Hand handIn,
-        final CallbackInfoReturnable<ActionResult<ItemStack>> cir, final ItemStack used, @Nullable final BlockPos targetPos) {
+    private void impl$ThrowForPreEvent(final World worldIn, final PlayerEntity playerIn, final Hand handIn,
+                                       final CallbackInfoReturnable<ActionResult<ItemStack>> cir, final ItemStack used, @Nullable final BlockPos targetPos) {
```
gabizou: Extra spaces == bad

gabizou (changes requested):
small nitpicks, but looks good.

Some of these changes, are you absolutely sure they are still necessary? I'm not sure that the `ClassInheritanceMultiMap` changes are still necessary, but I haven't done that investigation

i509VCB:
I gave the class inheritance multimap another look and the condition it allows for is quite broad as one of the halves only requires just an interface:
```java
private boolean impl$isAssignableFromOrInterface(final Class<?> baseClass, final Class<?> clazz) {
    return clazz.isInterface() || baseClass.isAssignableFrom(clazz);
}
```
I threw a todo comment on it for now but I can still make changes if needed

src/main/java/org/spongepowered/common/mixin/core/util/ClassInheritanceMultiMapMixin.java
```diff
@@ -34,9 +34,13 @@
 
     /**
      * @author Minecrell
+     * @author i509VCB - February 23rd, 2020 - 1.14.4
+     *
+     * @todo Is this even necessary anymore?
```
gabizou: Yes, this is still necessary. Since the logic is currently looking at the passed in class, we want to be able to ask if the class is either an interface OR an extended class of the base class.

gabizou (changes requested):
Remove unnecessary todo comment on the ClassInheritanceMultiMapMixin.

src/main/java/org/spongepowered/common/mixin/core/util/ClassInheritanceMultiMapMixin.java
i509VCB: Will do once I pack since I have to go to Dallas tomorrow

src/main/java/org/spongepowered/common/mixin/core/entity/LivingEntityMixin.java
```diff
@@ -170,7 +170,6 @@
     @Shadow private boolean checkTotemDeathProtection(final DamageSource p_190628_1_) {
         return false; // SHADOWED
     }
-
```
gabizou: Undo this change, whiteline noise.

gabizou (changes requested):

gabizou (approved):

PR #2516, opened 2020-02-25
[Discussion] Block Explosion Resistance Event

Pulverizer:
No code changes have yet been made to the fork. I am opening this PR to discuss the best method of implementing this feature.

I planned a feature for my plugin that changes the explosion resistance of a block based on the size of the craft containing it. e.g. a 2000 block craft would increase the blast resistance of it blocks by 4%.

I have been unable to find a better way to do this.


ImMorpheus:
That's not what PRs are for.

If you want to suggest a feature (a new event in this case) you need to open an **issue** in the **SpongeAPI** repository.

PR #2523, opened 2020-03-01
Add support of offering ConnectedDirectionsData for some block states

Lignium:
This changes allow to offer `ConnectedDirectionsData` for the following blocks:

- Fences
- Panes
- Redstone Wire
- Trip Wire
- **Vine**
- Wall

The most important thing here is vine. It was these blocks that I could not correctly placed in a world with all flags turned off. But for completeness, I implemented wherever possible.
I also noticed that there are many more keys that cannot be offered to blocks. Of course, it probably makes no sense to offer, for example, to a redstone wire `Keys.POWER`, but it seems to me that this should be implemented anyway. I can do this, but **I need your feedback on this**.


Lignium:
I looked at all the mixin classes for blocks, and found the following data keys that cannot be "written" to block states:

- ~~ATTACHED~~
- **CONNECTED_DIRECTIONS** (was implemented by this PR)
- **CONNECTED_EAST** (was implemented by this PR)
- **CONNECTED_NORTH** (was implemented by this PR)
- **CONNECTED_SOUTH** (was implemented by this PR)
- **CONNECTED_WEST** (was implemented by this PR)
- EXTENDED
- IN_WALL
- OCCUPIED
- POWER
- SNOWED
- _SUSPENDED_ (never implemented at all)
- WIRE_ATTACHMENT_EAST
- WIRE_ATTACHMENT_NORTH
- WIRE_ATTACHMENTS
- WIRE_ATTACHMENT_SOUTH
- WIRE_ATTACHMENT_WEST

As a result, the keys remain candidates for implementation:

- EXTENDED
- IN_WALL
- OCCUPIED
- POWER
- SNOWED
- WIRE_ATTACHMENT_EAST
- WIRE_ATTACHMENT_NORTH
- WIRE_ATTACHMENTS
- WIRE_ATTACHMENT_SOUTH
- WIRE_ATTACHMENT_WEST

The question is why this data was not implemented before?

gabizou:
Have you actually tried to offer these block states back to the world? From how I've understood these to be *represented by the server* is that they're non-existent because things like "Connected directions" is heavily reliant on knowing their neighbors, but at the end of the day, they're serialized as being just wire with x power. That's the reason why I've never implemented or asked for the states that are considered "virtual" to be implemented before.

Lignium:
In general, I directly faced the problem of placing vines in the world, I need to hang the vines on certain surfaces in certain positions. However, by virtue of the implementation, Sponge ignores these calls and the vine at placing has some default value. After applying this PR, I can now properly place vines. I also thought about other blocks that may also have such pitfalls.
UPD: I also use this key to transform blocks in space (rotation and reflection). And of course, vines will break.

src/main/java/org/spongepowered/common/mixin/core/block/BlockFenceMixin.java
```diff
@@ -37,14 +39,27 @@
...
 
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
 @Mixin(BlockFence.class)
 public abstract class BlockFenceMixin extends BlockMixin {
 
+    private static final Map<Direction, PropertyBool> DIRECTION_TO_PROPERTY_MAPPING;
```
gabizou: I'd rather not have a static field at all. Also, the field needs to be prefixed according to the naming standards for mixin-added-by-sponge names, i.e. `impl$DIRECTION_TO_PROPERTY_MAPPING`.

src/main/java/org/spongepowered/common/mixin/core/block/BlockFenceMixin.java
```diff
@@ -37,14 +39,27 @@
...
+
+    static {
+        ImmutableMap.Builder<Direction, PropertyBool> directionToPropertyMappingBuilder = ImmutableMap.builder();
```
gabizou: This really should not be statically initialized, I'd rather a constructor injection to initialize the field.

src/main/java/org/spongepowered/common/mixin/core/block/BlockFenceMixin.java
```diff
@@ -59,16 +74,33 @@
     @Override
     public Optional<BlockState> bridge$getStateWithData(final IBlockState blockState, final ImmutableDataManipulator<?, ?> manipulator) {
         if (manipulator instanceof ImmutableConnectedDirectionData) {
-            return Optional.of((BlockState) blockState);
+            ImmutableConnectedDirectionData connectedDirectionData = (ImmutableConnectedDirectionData) manipulator;
+            return Optional.of((BlockState) DirectionalBlockUtils.applyConnectedDirections(blockState,
+                                                                                           DIRECTION_TO_PROPERTY_MAPPING,
```
gabizou: Don't align multi-line like this, just indent on continuation.

src/main/java/org/spongepowered/common/mixin/core/block/BlockFenceMixin.java
```diff
@@ -59,16 +74,33 @@
     @Override
     public Optional<BlockState> bridge$getStateWithData(final IBlockState blockState, final ImmutableDataManipulator<?, ?> manipulator) {
         if (manipulator instanceof ImmutableConnectedDirectionData) {
-            return Optional.of((BlockState) blockState);
+            ImmutableConnectedDirectionData connectedDirectionData = (ImmutableConnectedDirectionData) manipulator;
```
gabizou:
```suggestion
            final ImmutableConnectedDirectionData connectedDirectionData = (ImmutableConnectedDirectionData) manipulator;
```

src/main/java/org/spongepowered/common/mixin/core/block/BlockFenceMixin.java
```diff
@@ -59,16 +74,33 @@
...
 
+    @SuppressWarnings("unchecked")
     @Override
     public <E> Optional<BlockState> bridge$getStateWithValue(final IBlockState blockState, final Key<? extends BaseValue<E>> key, final E value) {
-        if (key.equals(Keys.CONNECTED_DIRECTIONS) || key.equals(Keys.CONNECTED_EAST) || key.equals(Keys.CONNECTED_NORTH)
-                || key.equals(Keys.CONNECTED_SOUTH) || key.equals(Keys.CONNECTED_WEST)) {
-            return Optional.of((BlockState) blockState);
+        if (key.equals(Keys.CONNECTED_DIRECTIONS)) {
+            return Optional.of((BlockState) DirectionalBlockUtils.applyConnectedDirections(blockState,
+                                                                                           DIRECTION_TO_PROPERTY_MAPPING,
```
gabizou: Same with this. Don't indent to align.

src/main/java/org/spongepowered/common/util/DirectionalBlockUtils.java
```diff
@@ -0,0 +1,63 @@
...
+import net.minecraft.block.properties.IProperty;
+import net.minecraft.block.state.IBlockState;
+import org.spongepowered.api.util.Direction;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.BiFunction;
+
+public final class DirectionalBlockUtils {
```
gabizou: Don't make new utils. Put it in an existing utils class.

gabizou (changes requested):

src/main/java/org/spongepowered/common/mixin/core/block/BlockFenceMixin.java
Lignium: So you suggest using a non-static field? But this field is not an object property in any way in terms of OOP, purely supporting information. Or did I misunderstand you?

src/main/java/org/spongepowered/common/util/DirectionalBlockUtils.java
Lignium: Where, for example? I tried to find something suitable, but got confused. I'm not good at codebase yet.
UPD: Constants.DirectionFunctions would be a good choice?

src/main/java/org/spongepowered/common/mixin/core/block/BlockFenceMixin.java
gabizou: If it's not static to the Block itself, and it's something as a translation, then there's a [TypeProvider](https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/registry/TypeProvider.java) bit of the registry implementation, and there's also [Constants that can contain these type mappings](https://github.com/SpongePowered/SpongeCommon/blob/a865595b36ab95a876c8ebd46b464b23a73b09aa/src/main/java/org/spongepowered/common/util/Constants.java#L1097-L1105) where you can make some better static constants like this rather than in the mixin itself.

src/main/java/org/spongepowered/common/util/Constants.java
```diff
@@ -1358,5 +1364,25 @@ public static Axis convertAxisToSponge(final EnumFacing.Axis axis) {
                     return Axis.X;
             }
         }
+
```
gabizou, line 1365: Pretty much all the mappings for the specific block types can exist in here, or as a type provider. That's what I mean by reducing the amount of static usage in the mixins themselves.

Zidane:
@Lignium Could you update this for us please?

Lignium:
@Zidane Rebase? Done

gabizou (approved):
LGTM, thanks for the changes!

Issue #2526, opened 2020-03-06
InteractItemEvent.Primary delayed while jumping


SocraticPhoenix:
**Tested on:**
- Windows 10, Java 1.8.0_241

- Sponge Vanilla 1.12.2-7,1,10 (recommended/stable build as of today)
- Sponge Vanilla 1.12.2-7.1.11-RC346

**Issue:**
`InteractItemEvent.Primary` seems to fire late, sometimes, if the player is jumping.

Test plugin: https://gist.github.com/SocraticPhoenix/f1942d4e39b6a1e101669a1c6f56b997
Video showing behavior: https://youtu.be/1aSCkg8HWEI
(delay most visible at 0:10)

Yeregorix:
I've found a similar issue few days ago with `InteractBlockEvent.Primary`. The event can be delayed indefinitely if the player is in adventure mode and if he doesn't move. It seems to be related to this method in the implementation: https://github.com/SpongePowered/SpongeCommon/blob/4a9742b465d095769fa4e6ca5fbf22865e3de87b/src/main/java/org/spongepowered/common/mixin/core/server/MinecraftServerMixin.java#L321

If we read this method we can also quickly understand another big issue: if multiple players click in the same tick, there will be a event only for the last player.

JBYoshi:
Looks like the PR was closed a while ago, so I'll close this. Reopen if it still happens.

PR #2528, opened 2020-03-14
Fix interaction events not being called correctly when left clicking in the air.

Yeregorix:
This PR fixes several issues related to `InteractItemEvent.Primary` and `InteractBlockEvent.Primary` when a player is clicking in the air:
- If multiple players click in the air during the same tick, only the last one fires the events.
- The distance of the ray tracing used to determinate if the player is interacting a non-air block was incorrect. It was 6, now it's 5. It should be 4.5 for non-creative players but I didn't dare to touch `SpongeImplHooks#getBlockReachDistance` because it's used in many places and I don't want to break something.
- Black magic based on static variables and delay between packets.

These 3 things were causing problems like #2526.


ImMorpheus (approved):
Working as intended

PR #2532, opened 2020-03-25
Some more core and impl fixes [api8]

i509VCB:


i509VCB:
Lot of changes here have been done before I could rebase lol. Well next time

PR #2550, opened 2020-04-03
Implement CatalogKey.Builder

i509VCB:


src/main/java/org/spongepowered/common/registry/SpongeCatalogKeyBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+import net.minecraft.util.ResourceLocation;
+import net.minecraft.util.ResourceLocationException;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.spongepowered.api.CatalogKey;
+import org.spongepowered.api.plugin.PluginContainer;
+
+public class SpongeCatalogKeyBuilder implements CatalogKey.Builder {
```
Zidane: I like final things.

i509VCB: finally finialized

PR #2551, opened 2020-04-04
Implement SaveChunkEvent.

Zidane:
For discussion, @dualspiral 

Also note I fixed the fact that we were calling LoadChunk on the client thread (or attempting to) as well as we were not respecting ShouldFire for unload.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/chunk/storage/AnvilChunkLoaderMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/chunk/storage/AnvilChunkLoaderMixin.java
@@ -42,6 +42,8 @@ import org.spongepowered.api.entity.EntityType;
 import org.spongepowered.api.entity.Transform;
 import org.spongepowered.api.event.CauseStackManager.StackFrame;
 import org.spongepowered.api.event.SpongeEventFactory;
+import org.spongepowered.api.event.cause.Cause;
+import org.spongepowered.api.event.cause.EventContext;
 import org.spongepowered.api.event.cause.EventContextKeys;
 import org.spongepowered.api.event.cause.entity.spawn.SpawnTypes;
 import org.spongepowered.api.event.entity.ConstructEntityEvent;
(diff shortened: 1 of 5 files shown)


dualspiral:
I mean, it's bog standard throwing events and fixing bugs! I see nothing contentious here. If someone does, I'd be mightly surprised. When the API is good to go, this will be good to go.

src/main/java/org/spongepowered/common/mixin/core/world/chunk/storage/AnvilChunkLoaderMixin.java
```diff
@@ -300,4 +305,11 @@ public boolean writeNextIO() {
         return this.chunkSaveLocation.toPath();
     }
 
-}
+    @Inject(method = "writeChunkData", at = @At("RETURN"))
+    private void impl$callSaveChunkEventPost(ChunkPos pos, NBTTagCompound compound, CallbackInfo ci) {
+        if (ShouldFire.SAVE_CHUNK_EVENT || ShouldFire.SAVE_CHUNK_EVENT_POST) {
```
ImMorpheus:
```suggestion
        if (ShouldFire.SAVE_CHUNK_EVENT_POST) {
```

src/main/java/org/spongepowered/common/mixin/core/world/gen/ChunkProviderServerMixin.java
```diff
@@ -325,4 +326,15 @@ public boolean tick()
         this.loadedChunks.remove(ChunkPos.asLong(chunk.x, chunk.z));
         ((ChunkBridge) chunk).bridge$setScheduledForUnload(-1);
     }
+
+    @Inject(method = "saveChunkData", at = @At("HEAD"), cancellable = true)
+    private void impl$callSaveChunkEventPre(Chunk chunkIn, CallbackInfo ci) {
+        if (ShouldFire.SAVE_CHUNK_EVENT || ShouldFire.SAVE_CHUNK_EVENT_PRE) {
```
ImMorpheus:
```suggestion
        if (ShouldFire.SAVE_CHUNK_EVENT_PRE) {
```

ImMorpheus (changes requested):
Checking the flag for the generic type will also make the event fire when there are listeners for other sub events but not the event itself.

ImMorpheus (approved):

gabizou (approved):
LGTM. Some reservations about changing `world.isRemote` to a `((WorldBridge) world).bridge$isFake()` but that isn't likely going to be an issue.

PR #2558, opened 2020-04-08
Fix small compile errors in API mixins

TheGlitch76:

PR #2561, opened 2020-04-10
Don't attempt to load worlds when changing player location

### LINKED ISSUE #2559: Incorrect implementation of User.getLocation
**stable-7**
User.setLocation should only work on already loaded worlds, and not attempt to load one if it is now found. (see [here](https://github.com/SpongePowered/SpongeCommon/pull/2558/commits/1a4242a42677580ea024a7eda4514202fd7a47bb#r405228485))
[This method](https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/mixin/api/mcp/entity/player/EntityPlayerMPMixin_API.java#L653) should be corrected to attempt to get the world directly through the list of loaded worlds (unsure of the api7 specifics on that) and orElseThrow if it is not found.

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeCommon/issues/2559

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/api/mcp/entity/player/EntityPlayerMPMixin_API.java
+++ b/src/main/java/org/spongepowered/common/mixin/api/mcp/entity/player/EntityPlayerMPMixin_API.java
@@ -651,8 +651,7 @@ public abstract class EntityPlayerMPMixin_API extends EntityPlayerMixin_API impl
 
     @Override
     public boolean setLocation(final Vector3d position, final UUID world) {
-        final WorldProperties prop = Sponge.getServer().getWorldProperties(world).orElseThrow(() -> new IllegalArgumentException("Invalid World: No world found for UUID"));
-        final World loaded = Sponge.getServer().loadWorld(prop).orElseThrow(() -> new IllegalArgumentException("Invalid World: Could not load world for UUID"));
+        final World loaded = Sponge.getServer().getWorld(world).orElseThrow(() -> new IllegalArgumentException("Invalid World: No loaded world found for UUID"));
         return this.setLocation(new Location<>(loaded, position));
     }
 }


dualspiral:
@Zidane Are we changing the behaviour of `stable-7` to do this - or should we be looking at allowing this without setting the world - (i.e., set the location based on the existence of `WorldProperties` alone)?

The thing that bothers me is that this method clearly exists to set the location of a user to a world that hasn't loaded, but is enabled. If we only wanted to allow setting a player's location to a world that is loaded, then `setLocation(Location<World>)` would suffice, for offline and online players. I agree we shouldn't load the world, but perhaps we should set the location based on the `WorldProperties` existing instead.

I'd argue that this should be the case of 8 too, though I know that world management happens differently there so may need a better workaround (or just the removal of that API method, whatever you prefer).

ImMorpheus:
A more "backward compatible" way of doing it would be to return false instead of throwing an exception (Just like when the teleport event is canceled).

AFAIK that method exists to be used by a SpongeUser (offline player) for offline location here: https://github.com/SpongePowered/SpongeCommon/blob/4db9bcd110abf7a94062f5f5a6e1a08f949238ca/src/main/java/org/spongepowered/common/mixin/api/common/entity/player/SpongeUserMixin_API.java#L117-L134
The fact Player is extending that method is a side effect. It shouldn't be used for online players just like there's no similar method for the Entity class.

For API 8 there's an opened PR changing the behaviour of the method: https://github.com/SpongePowered/SpongeCommon/pull/2558/files#diff-dbdb1fdeab655e15e5e318cbe807bf1bR451

dualspiral:
Good point. I'd somehow missed the fact this was on Player, not User. Duh - I'm an idiot.

Well, the contract is that is passed through to `setLocation(Location)`, and `Location` must have an online world in this case. However, it might be that a user is obtained by the USS and so you're just operating on `User` objects, not `Player` objects - and so you might just use this method across all entries you pull out. It shows a bit of a discrepency in the abilities of `User#setLocation` and `Player#setLocation`, though I can see why that is...

I think we should return false in this case at the very least, but:

> The fact Player is extending that method is a side effect. It shouldn't be used for online players just like there's no similar method for the Entity class.

But it is there, and so we have to assume people might use it. We really cannot fall into the trap of "well, this method exists but shouldn't be used". It's there, it's potentially in use.

Kind of makes me feel like what the method is currently doing is right, to be honest. I see your point, I see @Zidane's point, but I think that it's the wrong point to make. We can use this method to send an offline player to an offline world, but not an online player to an offline world? Doesn't seem right to me.

PR #2569, opened 2020-04-15
Use Identity/Reference based collection where possible

ImMorpheus:
[SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/3132) | **SpongeCommon**

### DIFF
--- a/src/main/java/org/spongepowered/common/data/property/SpongePropertyRegistry.java
+++ b/src/main/java/org/spongepowered/common/data/property/SpongePropertyRegistry.java
@@ -31,7 +31,6 @@ import static com.google.common.base.Preconditions.checkState;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
 import com.google.inject.Singleton;
 import it.unimi.dsi.fastutil.objects.ReferenceOpenHashSet;
 import org.spongepowered.api.data.Property;
--- a/src/main/java/org/spongepowered/common/data/property/SpongePropertyRegistry.java
+++ b/src/main/java/org/spongepowered/common/data/property/SpongePropertyRegistry.java
@@ -31,6 +31,7 @@ import static com.google.common.base.Preconditions.checkState;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 import com.google.inject.Singleton;
 import org.spongepowered.api.data.Property;
 import org.spongepowered.api.data.property.PropertyHolder;
@@ -104,7 +105,7 @@ public class SpongePropertyRegistry implements PropertyRegistry {
     public Collection<Property<?, ?>> getPropertiesFor(final PropertyHolder holder) {
         final ImmutableList.Builder<Property<?, ?>> builder = ImmutableList.builder();
         if (this.tempRegistry != null) { // Still doing registrations
-            final Set<Class<? extends Property<?, ?>>> used = new HashSet<>();
+            final Set<Class<? extends Property<?, ?>>> used = Sets.newIdentityHashSet();
             for (final Map.Entry<Class<? extends Property<?, ?>>, PropertyStoreDelegate<?>> entry : this.tempRegistry.delegateMap.entrySet()) {
                 used.add(entry.getKey());
                 entry.getValue().getFor(holder).ifPresent(builder::add);
--- a/src/main/java/org/spongepowered/common/data/property/SpongePropertyRegistry.java
+++ b/src/main/java/org/spongepowered/common/data/property/SpongePropertyRegistry.java
@@ -33,6 +33,7 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.inject.Singleton;
+import it.unimi.dsi.fastutil.objects.ReferenceOpenHashSet;
 import org.spongepowered.api.data.Property;
 import org.spongepowered.api.data.property.PropertyHolder;
 import org.spongepowered.api.data.property.PropertyRegistry;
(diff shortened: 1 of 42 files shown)


dualspiral:
I'm curious, what are the performance gains from doing this?

ImMorpheus:
https://docs.oracle.com/javase/8/docs/api/java/util/IdentityHashMap.html

> For many JRE implementations and operation mixes, this class will yield better performance than HashMap (which uses chaining rather than linear-probing). 

We're already largely taking advantage of the performance gain of `IdentityHashMap` in the codebase, this is also a matter of consistency. Some of the collections were not using it even if they could. 
The performance gain is hard to measure because this PR is not targeting hotspots:
Ex.
`ChunkSaveHelper#writeChunks` is only used by the command `/sponge chunks`
the `VirtualPortalAgent` class is not used at all

the `builders` field in `SpongeDataManager` on the other hand is used when serializing/deserializing `MemoryDataContainer`s

Zidane:
Speaking of inconsistency...

Can we stop using Guava to create basic maps like this? I see `Sets.` and then `new Identity`.

ImMorpheus:
> Speaking of inconsistency...
> 
> Can we stop using Guava to create basic maps like this? I see `Sets.` and then `new Identity`.

While I hate the abuse of Guava, there's nothing I can do in this case.
Java doesn't have an `IdentityHashSet`.

EDIT: as a matter of fact `new` is used with Identity**HashMap**s
EDIT2: k, moved to fastutil. Now there's "new" everywhere.

Zidane:
Ha @ImMorpheus I completely forgot about that...

Ignore me.

Zidane:
@ImMorpheus Remove the bad Sets imports, I think they are left over before you replaced Guava usages.

Otherwise, great PR. Thanks as always.

src/main/java/org/spongepowered/common/event/tracking/phase/packet/inventory/BasicInventoryPacketState.java
```diff
@@ -24,6 +24,8 @@
  */
 package org.spongepowered.common.event.tracking.phase.packet.inventory;
 
+import com.google.common.collect.Sets;
```
gabizou:
```suggestion
```

src/main/java/org/spongepowered/common/item/inventory/archetype/SpongeInventoryArchetypeBuilder.java
```diff
@@ -24,6 +24,8 @@
  */
 package org.spongepowered.common.item.inventory.archetype;
 
+import com.google.common.collect.Sets;
```
gabizou:
```suggestion
```

src/main/java/org/spongepowered/common/item/inventory/lens/impl/ReusableLens.java
```diff
@@ -65,7 +66,7 @@ private ReusableLens(SlotProvider slots, Function<SlotProvider, T> lensProvider)
     public static <T extends Lens> ReusableLens<T> getLens(Class<T> lensType, InventoryAdapter adapter,
             Supplier<SlotProvider> slots, Function<SlotProvider, T> lens) {
         Map<Class<? extends Lens>, Int2ObjectMap<ReusableLens>>
```
gabizou, line 67: Isn't this going to be changing to a different type of map?

ImMorpheus:
yep, I'll remove the old imports with the next commits. 

I've also found out that `Key`s and `EventContextKey`s are treated as if they don't have an equals contract (compared with `==` and used in `IdentityHashMap`s) so there are a few spots that can be updated.

ImMorpheus:
Ok, this is ready now.
Turned out `EvenContextKey`s are following the `equals` contract, so unless we decide that it's perfectly fine to compare them with `==` or use them in Identity maps then they should stay as they are now.
`Key`s are already being used that way (see https://github.com/SpongePowered/SpongeCommon/blob/a4f4c117ad89266a57762eaf27cd01d30251a7bb/src/main/java/org/spongepowered/common/data/SpongeManipulatorRegistry.java#L94 for an example)

The map passed to `NotifyNeighbourEvent` is now an EnumMap, that should help a bit given how often that's called.

src/main/java/org/spongepowered/common/data/property/SpongePropertyRegistry.java
```diff
@@ -104,7 +105,7 @@ public void completeRegistration() {
     public Collection<Property<?, ?>> getPropertiesFor(final PropertyHolder holder) {
         final ImmutableList.Builder<Property<?, ?>> builder = ImmutableList.builder();
         if (this.tempRegistry != null) { // Still doing registrations
-            final Set<Class<? extends Property<?, ?>>> used = new HashSet<>();
+            final Set<Class<? extends Property<?, ?>>> used = new ReferenceOpenHashSet<>();
```
gabizou: this is a moot set to even bother using. It's a temp registry that gets dropped as soon as the game has finished loading.

src/main/java/org/spongepowered/common/item/inventory/lens/impl/ReusableLens.java
ImMorpheus: `IdentityHashMap` implements `Map`. The declaration is fine, I'm changing the value here
https://github.com/SpongePowered/SpongeCommon/pull/2569/files#diff-79c49477a2574946c4d8b9f22e05e208R68

src/main/java/org/spongepowered/common/data/util/DataUtil.java
```diff
@@ -580,10 +581,9 @@ public static void readCustomData(final NBTTagCompound compound, final DataHolde
                 // Re-attempt to deserialize custom data
                 final SerializedDataTransaction transaction = deserializeManipulatorList(builder.build());
                 final List<DataManipulator<?, ?>> manipulators = transaction.deserializedManipulators;
-                final List<Class<? extends DataManipulator<?, ?>>> classesLoaded = new ArrayList<>();
+                final Set<Class<? extends DataManipulator<?, ?>>> classesLoaded = new ReferenceOpenHashSet<>();
                 for (final DataManipulator<?, ?> manipulator : manipulators) {
-                    if (!classesLoaded.contains(manipulator.getClass())) {
-                        classesLoaded.add((Class<? extends DataManipulator<?, ?>>) manipulator.getClass());
+                    if (classesLoaded.add((Class<? extends DataManipulator<?, ?>>) manipulator.getClass())) {
```
gabizou, line 586: Damn, this must've been here for ages.....

gabizou (commented):
looks good. Would be good to get some profiling results if you can, Might be worth spinning up some stupid Redstone machine that uses hoppers and enable `-Dsponge.shouldfireAll=true` to ping the event listeners and phase states and see what the comparison is. That being said, I'd like to take that approach to do some base testing with some profiling results as an example (acknowledging some limited testing, and of course, letting the JVM "warm up").

src/main/java/org/spongepowered/common/data/property/SpongePropertyRegistry.java
ImMorpheus: Reverted with other temporary maps.

PR #2570, opened 2020-04-15
Update JoinData only after the player has joined

ImMorpheus:
`JoinData` was updated before firing `ClientConnectionEvent.Login`. 
The event can be canceled because of the whitelist, a ban or by a plugin listening to said event.

This PR moves the code after the event to update `JoinData` only after the player has joined the server


dualspiral:
I was wondering whether we should put this after the `ClientConnectionEvent.Join` event instead, but the player has really joined at that point, just awaiting a spawn. So this seems reasonable and makes sense.

PR #2573, opened 2020-04-17
[API7] Map API

tyhdefu:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2112) | SpongeCommon | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/429) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/3149)

MapItemData/ImmutableMapItemData, a bunch of keys, an event, see SpongeAPI pull request for more details.


src/main/java/org/spongepowered/common/bridge/optimization/OptimizedMapDataBridge.java
```diff
@@ -42,4 +42,8 @@
     void mapOptimizationBridge$updateItemFrameDecoration(EntityItemFrame frame);
 
     void mapOptimizationBridge$removeItemFrame(EntityItemFrame frame);
+
+    int getWorldId();
```
Zidane: This is not the appropriate section to put this bridge method.

It needs to go into org.spongepowered.common.bridge.item.MapItemBridge (make the interface if it doesn't exist)

You need to then implement that with a mixin in org.spongepowered.common.mixin.core.item.MapItemMixin

Lastly, this should be get/setDimensionId

**EDIT**

Lastly, this should be bridge$get/setDimensionId

src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeMapItemData.java
```diff
@@ -0,0 +1,128 @@
...
+public class ImmutableSpongeMapItemData extends AbstractImmutableData<ImmutableMapItemData, MapItemData> implements ImmutableMapItemData {
+
+    private final Vector2i center;
+    private final World world;
+    private final boolean trackingPosition;
+    private final boolean unlimitedTracking;
+    private final byte scale;
+
+    public ImmutableSpongeMapItemData() {
+        this(Vector2i.ZERO, Sponge.getServer().getWorld(Sponge.getServer().getDefaultWorldName()).get(),
```
Zidane: Should call the WorldManager directly here and not the API, minimizes that call stack.

src/main/java/org/spongepowered/common/data/manipulator/immutable/item/ImmutableSpongeMapItemData.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Override
+    public ImmutableValue<Vector2i> location() {
+        return new ImmutableSpongeValue<>(Keys.MAP_LOCATION, Vector2i.ZERO, this.center);
+    }
+
+    @Override
+    public ImmutableValue<World> world() {
+        return new ImmutableSpongeValue<>(Keys.MAP_WORLD, Sponge.getServer().getWorld(Sponge.getServer().getDefaultWorldName()).get(), this.world);
```
Zidane: Same as above.

Zidane: Also, why do you keep the world field but not make use of it?

src/main/java/org/spongepowered/common/data/manipulator/mutable/item/SpongeMapItemData.java
```diff
@@ -0,0 +1,153 @@
...
+        this.unlimitedTracking = unlimitedTracking;
+        this.scale = scale;
+        registerGettersAndSetters();
+    }
+
+    public SpongeMapItemData(MapData mapData) {
+        this(
+                new Vector2i(mapData.xCenter, mapData.zCenter),
+                (org.spongepowered.api.world.World)WorldManager
+                        .getWorldByDimensionId(((OptimizedMapDataBridge) mapData).getWorldId()).get(),
```
Zidane: Change this to use the bridge changes I mentioned above.

src/main/java/org/spongepowered/common/data/processor/multi/item/MapItemDataProcessor.java
```diff
@@ -0,0 +1,131 @@
...
+        MapData mapData = (MapData)
+                ((net.minecraft.world.storage.MapStorage)Sponge.getServer().getMapStorage().get())
+                        .getOrLoadData(MapData.class, Constants.ItemStack.MAP_PREFIX + metaData);
+        if (mapData == null)
+            return false;
+        Vector2i location = (Vector2i)keyValues.get(Keys.MAP_LOCATION);
+        mapData.xCenter = location.getX();
+        mapData.zCenter = location.getY();
+        int id = ((WorldServerBridge)keyValues.get(Keys.MAP_WORLD)).bridge$getDimensionId();
+        ((OptimizedMapDataBridge)mapData).setWorldId(id);
```
Zidane: Change this to use the bridge changes I mentioned above.

src/main/java/org/spongepowered/common/data/processor/multi/item/MapItemDataProcessor.java
```diff
@@ -0,0 +1,131 @@
...
+        int metaData = itemStack.getMetadata();
+        MapData mapData = (MapData)
+                ((net.minecraft.world.storage.MapStorage)Sponge.getServer().getMapStorage().get())
+                        .getOrLoadData(MapData.class, Constants.ItemStack.MAP_PREFIX + metaData);
+        if (mapData == null)
+            return false;
+        Vector2i location = (Vector2i)keyValues.get(Keys.MAP_LOCATION);
+        mapData.xCenter = location.getX();
+        mapData.zCenter = location.getY();
+        int id = ((WorldServerBridge)keyValues.get(Keys.MAP_WORLD)).bridge$getDimensionId();
```
Zidane: dimensionId

src/main/java/org/spongepowered/common/data/processor/value/item/ItemMapWorldValueProcessor.java
```diff
@@ -0,0 +1,77 @@
...
+    @Override
+    protected Value<World> constructValue(World actualValue) {
+
+        return new SpongeValue<>(Keys.MAP_WORLD,
+                Sponge.getServer().getWorld(Sponge.getServer().getDefaultWorldName()).get(), actualValue);
+    }
+
+    @Override
+    protected boolean set(MapData container, World value) {
+        container.dimension = (byte)((WorldServerBridge)value).bridge$getDimensionId();
```
Zidane: This is where things are so much fun.

Mods very much have dimension ids of worlds well above 256. Since they can set the number to whatever they like! Forge patches this to be an int for a good reason..

@gabizou How do you want to deal with this?

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -101,24 +90,10 @@
...
 import org.spongepowered.api.event.cause.entity.spawn.SpawnTypes;
-import org.spongepowered.api.event.entity.ChangeEntityEquipmentEvent;
-import org.spongepowered.api.event.entity.CollideEntityEvent;
-import org.spongepowered.api.event.entity.ConstructEntityEvent;
-import org.spongepowered.api.event.entity.DestructEntityEvent;
-import org.spongepowered.api.event.entity.InteractEntityEvent;
-import org.spongepowered.api.event.entity.MoveEntityEvent;
-import org.spongepowered.api.event.entity.RotateEntityEvent;
-import org.spongepowered.api.event.entity.SpawnEntityEvent;
+import org.spongepowered.api.event.entity.*;
```
Zidane: No star imports please.

src/main/java/org/spongepowered/common/mixin/core/world/storage/MapStorageMixin.java
```diff
@@ -35,12 +41,50 @@
...
+import org.spongepowered.common.data.manipulator.mutable.item.SpongeMapItemData;
+import org.spongepowered.common.world.WorldManager;
+
+import javax.annotation.Nullable;
+import java.util.Map;
+import java.util.Optional;
 
 @Mixin(MapStorage.class)
-public abstract class MapStorageMixin {
+public abstract class MapStorageMixin implements org.spongepowered.api.world.map.MapStorage {
```
Zidane: If this implements just the API, this mixin needs to go to org.spongepowered.common.mixin.api

If this has injections in it, you need to separate out the API implementation into the above namespace. If the API needs something set via an injection, the API calls a bridge method added via a bridge.

https://github.com/SpongePowered/SpongeCommon/wiki/Implementation-Policy:-Bridges-and-Mixins

src/main/java/org/spongepowered/common/bridge/optimization/OptimizedMapDataBridge.java
tyhdefu: i dont think a mixin is actually needed, i just put it there because it had
`private static Field mapOptimizationImpl$dimensionField;`
which you can then use to access the dimension field with reflection so spongeforge's patch can work.
Where should I put the code in this case?

src/main/java/org/spongepowered/common/data/processor/value/item/ItemMapWorldValueProcessor.java
tyhdefu: this should be using the .setDimensionId() to access this field, and since bridge$getDimensionId() shouldn't return a value that the field can't hold - eg. in SpongeVanilla it wont return more than a byte's max value and in SpongeForge the field is able to take it.

tyhdefu:
All Implementation is completed, and a test plugin called MapTest has been added to testplugins.

src/main/java/org/spongepowered/common/bridge/world/storage/MapStorageBridge.java
```diff
@@ -0,0 +1,47 @@
...
+public interface MapStorageBridge {
+    Optional<MapData> bridge$getMinecraftMapData(int id);
+
+    /**
+     * Only call if you know what you are doing, if used
+     * incorrectly, will cause maps to be overwritten.
+     * Next map created will use this id + 1
+     * @param id int To set highest map number to.
+     */
+    void bridge$setHighestMapId(short id);
```
dualspiral, line 40: Should this be an int to support Forge?

src/main/java/org/spongepowered/common/data/processor/value/mapinfo/MapInfoCanvasValueProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    protected MapInfoData createManipulator() {
+        return new SpongeMapInfoData();
+    }
+
+    @Override
+    protected boolean set(MapInfo mapInfo, MapCanvas value) {
+        MapData mapData = (MapData)mapInfo;
```
dualspiral: Style note: there should be a space between the cast and the variable.

src/main/java/org/spongepowered/common/data/processor/value/mapinfo/MapInfoLocationValueProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public MapInfoLocationValueProcessor() {
+        super(Keys.MAP_LOCATION, MapInfo.class);
+    }
+
+    @Override
+    protected boolean set(MapInfo mapInfo, Vector2i value) {
+        MapData mapData = (MapData)mapInfo;
+        // This also sets xCenter and zCenter.
+        mapData.calculateMapCenter(value.getX(), value.getX(), mapData.scale);
```
dualspiral: Should the second parameter be `value.getY()`?

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -1185,7 +1185,7 @@ public static boolean callInteractInventoryOpenEvent(final EntityPlayerMP player
     }
 
     public static InteractInventoryEvent.Close callInteractInventoryCloseEvent(final Container container, final EntityPlayerMP player,
-            final ItemStackSnapshot lastCursor, final ItemStackSnapshot newCursor, final boolean clientSource) {
+                                                                               final ItemStackSnapshot lastCursor, final ItemStackSnapshot newCursor, final boolean clientSource) {
```
dualspiral, line 1219: Please leave code alone if you're not otherwise touching it - this should have been left alone.

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -1778,4 +1778,9 @@ public static ItemStack throwDropItemAndConstructEvent(final net.minecraft.entit
         return event;
     }
 
+    public static CreateMapEvent callCreateMapEvent(final Cause cause, MapInfo mapInfo) {
```
dualspiral: Just make these calls inline, no need to add to the common event factory.

src/main/java/org/spongepowered/common/map/canvas/SpongeEmptyCanvas.java
```diff
@@ -0,0 +1,86 @@
...
+
+import net.minecraft.world.storage.MapData;
+import org.spongepowered.api.data.DataContainer;
+import org.spongepowered.api.data.DataQuery;
+import org.spongepowered.api.map.MapCanvas;
+import org.spongepowered.api.map.color.MapColor;
+import org.spongepowered.api.map.color.MapColorTypes;
+import org.spongepowered.common.util.Constants;
+
+import java.awt.*;
```
dualspiral: No star imports

src/main/java/org/spongepowered/common/map/canvas/SpongeMapCanvasBuilder.java
```diff
@@ -0,0 +1,138 @@
...
+import org.spongepowered.api.data.DataView;
+import org.spongepowered.api.data.persistence.InvalidDataException;
+import org.spongepowered.api.map.MapCanvas;
+import org.spongepowered.api.map.color.MapColor;
+import org.spongepowered.api.map.color.MapColorType;
+import org.spongepowered.common.map.color.SpongeMapColor;
+import org.spongepowered.common.util.Constants;
+
+import javax.annotation.Nullable;
+import java.awt.*;
```
dualspiral: No star imports

src/main/java/org/spongepowered/common/map/color/SpongeMapColor.java
```diff
@@ -0,0 +1,82 @@
...
+        return Color.ofRgb(r,g,b);
+    }
+
+    private int addShade(int color) {
+        return Math.floorDiv(color * Constants.Map.SHADE_MULTIPLIER[shade], Constants.Map.SHADE_DIVIDER);
+    }
+
+    @Override
+    public int getContentVersion() {
+        return 0;
```
dualspiral: 1

src/main/java/org/spongepowered/common/map/decoration/SpongeMapDecorationType.java
```diff
@@ -0,0 +1,54 @@
...
+import net.minecraft.world.storage.MapDecoration;
+import org.spongepowered.api.map.decoration.MapDecorationType;
+
+public class SpongeMapDecorationType implements MapDecorationType {
+    private MapDecoration.Type type;
+    private String id;
+    private String name;
+
+    public SpongeMapDecorationType(MapDecoration.Type mapDecorationType, String id, String name) {
+        type = mapDecorationType;
```
dualspiral: `this.type`

src/main/java/org/spongepowered/common/mixin/api/mcp/map/MapDecorationMixin_API.java
```diff
@@ -0,0 +1,186 @@
...
+
+@Mixin(MapDecoration.class)
+public class MapDecorationMixin_API implements SpongeMapDecoration {
+    @Shadow @Final private MapDecoration.Type type;
+    @Shadow private byte x;
+    @Shadow private byte y;
+
+    @Shadow private byte rotation;
+    // If should save to disk
+    private boolean isPersistent;
```
dualspiral: In mixins, if you're adding your own method or field in the common `api` package, prefix it with `api$`

Here, `api$isPersistent`

src/main/java/org/spongepowered/common/mixin/api/mcp/map/MapDecorationMixin_API.java
```diff
@@ -0,0 +1,186 @@
...
+    //
+    private String key = Constants.Map.DECORATION_KEY_PREFIX + UUID.randomUUID().toString();
```
dualspiral: `api$key`

src/main/java/org/spongepowered/common/mixin/api/mcp/map/MapDecorationMixin_API.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    @Override
+    public int getY() {
+        return this.y;
+    }
+
+    @Override
+    public void setPosition(Vector2i position) {
+        checkState(MapUtil.isInMapDecorationBounds(position.getX()), "x position out of bounds");
```
dualspiral: Don't statically import here please (I know we have done it elsewhere, but we are shifting away from that)

src/main/java/org/spongepowered/common/mixin/api/mcp/map/MapDecorationMixin_API.java
```diff
@@ -0,0 +1,186 @@
...
+    @Shadow private byte y;
+
+    @Shadow private byte rotation;
+    // If should save to disk
+    private boolean isPersistent;
+    //
+    private String key = Constants.Map.DECORATION_KEY_PREFIX + UUID.randomUUID().toString();
+
+    @Inject(method = "<init>", at = @At("RETURN"))
+    public void initialiser(MapDecoration.Type typeIn, byte xIn, byte yIn, byte rotationIn, CallbackInfo ci) {
```
dualspiral: `api$initialiser`

src/main/java/org/spongepowered/common/mixin/api/mcp/map/MapInfoMixin_API.java
```diff
@@ -0,0 +1,223 @@
...
+     *  - 'frame-' + item frame entity id
+     *
+     * No decorations are saved to disk. Player/frames are recalculated and treasure maps have nbt on their
+     * ItemStacks, therefore, due to our adding of the ability to make maps and add decorations, it could
+     * get out of sync quite easily. Therefore we opt to save the ones that would be saved on ItemStacks
+     * that aren't ones that are depend on the world surroundings (Players, item frames)
+     * (but also leave them there)
+     */
+    @Shadow public Map<String, MapDecoration> mapDecorations;
+    private int mapId;
```
dualspiral: `api$mapId`

src/main/java/org/spongepowered/common/mixin/api/mcp/map/MapInfoMixin_API.java
```diff
@@ -0,0 +1,223 @@
...
+    // Turns out Minecraft 1.14 actually has exactly what we want
+    // Locked = stops terrain updates
+    // So we're just adding it early, and it should be directly compatible
+    // https://minecraft.gamepedia.com/Map#Locking
+    private boolean locked = Constants.Map.DEFAULT_MAP_LOCKED;
```
dualspiral: `api$locked`

src/main/java/org/spongepowered/common/mixin/api/mcp/map/MapInfoMixin_API.java
```diff
@@ -0,0 +1,223 @@
...
+
+    @Inject(method = "<init>", at = @At("RETURN"))
+    public void setMapId(String mapname, CallbackInfo ci) {
```
dualspiral: `api$setMapId`...

...though I wonder if some of these should be in the core mixins package instead, then using a bridge interface if it's needed to be accessed here, this is non-api facing logic and I think it should probably be moved out of here.

src/main/java/org/spongepowered/common/mixin/api/mcp/map/MapInfoMixin_API.java
```diff
@@ -0,0 +1,223 @@
...
+            this.mapId = Integer.parseInt(id);
+        } catch (NumberFormatException e) {
+            SpongeImpl.getLogger().error("Map id could not be got from map name, (" + mapname + ")", e);
+        }
+    }
+
+    // Save to disk if we don't want to auto explore the map
+    // No point saving additional data if its default.
+    @Inject(method = "writeToNBT", at = @At("RETURN"))
+    public void writeNBT(NBTTagCompound compound, CallbackInfoReturnable<NBTTagCompound> cir) {
```
dualspiral: As above.

src/main/java/org/spongepowered/common/mixin/api/mcp/map/MapInfoMixin_API.java
```diff
@@ -0,0 +1,223 @@
...
+            }
+            NBTTagCompound nbt = mapDecoration.getMCNBT();
+            nbt.setString("id", entry.getKey());
+            nbtList.appendTag(nbt);
+        }
+        compound.setTag(Constants.Map.DECORATIONS_KEY, nbtList);
+    }
+
+    @Inject(method = "readFromNBT", at = @At("RETURN"))
+    public void readNBT(NBTTagCompound nbt, CallbackInfo ci) {
```
dualspiral: And again.

src/main/java/org/spongepowered/common/mixin/api/mcp/map/MapInfoMixin_API.java
```diff
@@ -0,0 +1,223 @@
...
+            }
+        }
+    }
+
+    // We could technically Inject after, but that would mean
+    // searching the map for the value we just had, seems like a waste.
+    @Nullable
+    @SuppressWarnings("unchecked")
+    @Redirect(method = "updateDecorations", at = @At(value = "INVOKE", target = "Ljava/util/Map;put(Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;"))
+    public Object redirectPut(Map map, Object key, Object value) {
```
dualspiral: and again.

src/main/java/org/spongepowered/common/mixin/api/mcp/map/MapInfoMixin_API.java
```diff
@@ -0,0 +1,223 @@
...
+        Optional<DataView> manipulator = container.getView(Constants.Map.MAP_DATA);
+        manipulator.flatMap(dataView -> new SpongeMapInfoData().from(dataView.copy()))
+            .map(this::offer)
+            .orElseThrow(() -> new InvalidDataException("Manipulator data given to MapInfo was invalid!"));
+        this.mapId = container.getInt(Constants.Map.MAP_ID).orElseThrow(() -> new InvalidDataException("Could not get the MapId from the container given to MapInfo"));
+    }
+
+    @Override
+    public int getContentVersion() {
+        return 0;
```
dualspiral: 1

src/main/java/org/spongepowered/common/mixin/core/item/ITemMapMixin.java
```diff
@@ -39,6 +39,7 @@
 public class ITemMapMixin extends ItemMapBase {
 
 
+
```
dualspiral, line 52: Didn't need to be touched.

src/main/java/org/spongepowered/common/map/MapUtil.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * Returns MapInfo of newly created map, if the event was not cancelled.
+     * @param mapInfoData MapInfoData for the data in the MapData
+     * @param cause Cause of the event
+     * @return MapInfo if event was not cancelled
+     */
+    public static Optional<MapInfo> fireCreateMapEvent(MapInfoData mapInfoData, Cause cause) {
+        int id = Sponge.getServer().getMapStorage()
+                .map(mapStorage -> (MapStorageBridge)mapStorage)
+                .flatMap(MapStorageBridge::bridge$getHighestMapId)
```
dualspiral, line 71: Is there a good way to perhaps find the _lowest_ ID that is available? For example, if IDs 1 to 4 are used, but then 3 is deleted (can you delete a map via this API?), it may make sense to grab ID 3 rather than the highest.

Leaving it as highest ID might be fine, but we should really reuse any map ID that is then unused otherwise we're limiting the server to 32,767 maps over its lifetime (I don't think the negative IDs are used?).

src/main/java/org/spongepowered/common/map/MapUtil.java
```diff
@@ -0,0 +1,98 @@
...
+                .orElse(-1);
+        id++;
```
dualspiral: Rather than do this and have some magic statement, just fold this into your map statements and return `orElse(0)`.

dualspiral (changes requested):
There's quite a lot to take in here but there's a fair few things I noticed that do need looking at.

* This needs rebasing on top of the latest `stable-7`.
* Style is inconsistent and really should be followed - this will help us pull this into 1.14.4 too!
  * Use `this.` on instance variables and methods calls.
  * Prefix static calls with the name of the class (like `Class.method()` or `Class.variable`), even if it's the same class.
  * Casts must have a space between the cast and the call (`(Object) blah`, not `(Object)blah`)
* I realise that you've touched some classes when working out how to get things working, but any class that you don't end up modifying should not be in the diff - don't alter a class you don't need to alter!
* Mixins:
  * Make sure that you prefix variables and methods in mixins with the appropriate prefix, `api$` for API mixins, `impl$` for core mixins.
  * If a redirect or injection does not affect the implementation of the API, it might need to be in a core mixin.
* Content version should be 1, not 0.
* We need to consider map data that may need to be discarded pretty quickly, such as if a QR code is displayed to link a player to an account somewhere - obviously this doesn't want to be around for long. Can map data be obliterated, and then the ID reused if necessary later? Should we have a way for a map to never reach the disk?
* Once we reach the highest map ID, what happens then? Can we then reuse a map ID no longer in use? How would we do that?

src/main/java/org/spongepowered/common/bridge/world/storage/MapStorageBridge.java
tyhdefu: Only the dimension is changed from a short to an int, the map id (the amount of possible maps), is unchanged. Though I don't think it would be a terrible idea to try and change this to a int.

src/main/java/org/spongepowered/common/data/processor/value/mapinfo/MapInfoLocationValueProcessor.java
tyhdefu: Yes, good catch, I was only testing setting location with the same X and Y cords.

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
tyhdefu: Is there any git magic I can do to undo this?

src/main/java/org/spongepowered/common/bridge/world/storage/MapStorageBridge.java
dualspiral: Right, okay.

So why is `bridge$getHighestMapId` returning an optional integer?

src/main/java/org/spongepowered/common/mixin/api/mcp/map/MapInfoMixin_API.java
tyhdefu: I would move it, but I think it has to stay here since it needs to be able to access api$mapId.

src/main/java/org/spongepowered/common/bridge/world/storage/MapStorageBridge.java
tyhdefu: getHighestMapId, gets the highest map id, that exists, so when 1 map exists, its 0, etc. It doesn't make much sense to return -1 or something, because if you actually tried to get that map id, it wouldn't exist.
I will update the javadoc to make this more clear.

dualspiral:
@tyhdefu I've spent a little bit of time making sure things are within our style and fixed some of it up:

* Many mixins didn't have the right prefix `api$` or `impl$`, that's fixed
* A couple of mixins were in the API package when they didn't implement any API, they're in core
* Many things have had `this.` added, or made `final`.
* An `INVOKE` mixin used an `ordinal`, but no `target` - _always always always_ make sure you have a target.
* The test plugin has been migrated to use our "loadable module" framework.

This has been committed to your branch. However, it's not ready yet. I was testing the make map blue on create listener and while the event fired, the map came back as a map of the terrain - so it seems that offering map data to a map as it's created doesn't work. Can you look into this?

For reference:

* Pull the code down
* Run the server in the IDE, run `/testplugins e maptest`
* Create a map via right clicking an empty map

I expected it to be blue, and it wasn't.

src/main/java/org/spongepowered/common/map/canvas/SpongeMapCanvasBuilder.java
```diff
@@ -0,0 +1,146 @@
...
+        );
+        return Optional.of(mapCanvas);
+    }
+
+    @Override
+    public MapCanvas build() {
+        if (this.canvas == null) {
+            return SpongeEmptyCanvas.INSTANCE;
+        }
+        return this.canvas;
```
dualspiral: This is problematic.

Imagine that I create a canvas using this builder, then build it. You return the backing canvas here. Now I change something using this builder - that's going to change the already built canvas.

**The builder and provided object should have no relationship after the object is built**. The canvas currently has various paint methods, they really should be here on this builder, then a deep copy of the final data should be placed onto the canvas. The provided canvas should be immutable.

Please separate the canvas and the builder, only creating an immutable cavas when you build.

src/main/java/org/spongepowered/common/map/canvas/SpongeMapCanvasBuilder.java
```diff
@@ -0,0 +1,146 @@
...
+import java.awt.Graphics2D;
+import java.awt.Image;
+import java.awt.image.BufferedImage;
+import java.awt.image.DataBufferInt;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+public class SpongeMapCanvasBuilder extends AbstractDataBuilder<MapCanvas> implements MapCanvas.Builder {
```
dualspiral: Looking at the methods on this class - I see no reason for the `AbstractDataBuilder` and the `MapCanvas.Builder` to be the same class. They do not interact with each other in any way - I think they should be separated.

dualspiral (changes requested):
There may be more comments as I go through, but this needs to be resolved.

dualspiral:
Sorry again, I've not been the quickest with this and that is somewhat unfair to you! I'm gonna try to go through it again over the next few days, lots have been capturing my time so I will do what I can.

How much of this can be ported to API 8 - both API and Common? We have to think about the future API, and I don't want to add anything to a potential 7.4 that would not really be portable to API 8. I would think the vast majority can be moved across but it's not something I've looked at in great detail.

I will admit that it would be easier for us to look at taking this on for 1.15.2+ instead, rather than 1.12.2, as that's where almost all eyes are for Sponge. However, I'm also painfully aware of how much work has gone into this and I do want to avoid just deleting it. If you're up for having a look and seeing if this work will apply somewhat well to 1.15.2, that would be useful - if it's close to 1:1 (ignoring class location changes) that would be a big boon for us.

@Zidane @gabizou If you get a chance, please have a look at the API with an API 8 hat on - would the way this is built transfer to our API 8 concepts cleanly? Especially - would we want maps to be data holders? They are NBT compounds after all.

PR #2581, opened 2020-04-24
Improve vanish

ImMorpheus:
A few improvement for Vanish related data, specifically:

- Uncollideable now means Uncollideable. (this wasn't the case, as you could for example collide with `EntityItem`s and pick them up).
- Vanished players don't affect mob spawning (just like spectactor).

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/player/EntityPlayerMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/player/EntityPlayerMixin.java
@@ -391,7 +391,7 @@ public abstract class EntityPlayerMixin extends EntityLivingBaseMixin implements
 
     @Override
     public boolean bridge$affectsSpawning() {
-        return this.affectsSpawning && !this.isSpectator();
+        return this.affectsSpawning && !this.isSpectator() && !this.bridge$isVanished();
     }
 
     @Override
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/player/EntityPlayerMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/player/EntityPlayerMixin.java
@@ -401,7 +401,7 @@ public abstract class EntityPlayerMixin extends EntityLivingBaseMixin implements
 
     @Override
     public boolean bridge$affectsSpawning() {
-        return this.affectsSpawning && !this.isSpectator() && !this.bridge$isVanished();
+        return this.affectsSpawning && !this.isSpectator() && !this.bridge$isUntargetable();
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/core/entity/player/EntityPlayerMixin.java
```diff
@@ -361,6 +363,14 @@ public void onDeath(final DamageSource cause) {
         this.setFlag(0, false);
     }
 
+    @Redirect(method = "onLivingUpdate", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/World;getEntitiesWithinAABBExcludingEntity(Lnet/minecraft/entity/Entity;Lnet/minecraft/util/math/AxisAlignedBB;)Ljava/util/List;"))
+    private List<Entity> onSpongeGetEntitiesWithinAABBExcludingEntity(final World world, Entity entityIn, AxisAlignedBB bb) {
+        if (this.bridge$isUncollideable()) {
```
dualspiral, line 368: [Should this depend on `VANISH_IGNORES_COLLISION`](https://github.com/SpongePowered/SpongeAPI/blob/stable-7/src/main/java/org/spongepowered/api/data/key/Keys.java#L2248-L2257)?

src/main/java/org/spongepowered/common/mixin/core/entity/player/EntityPlayerMixin.java
```diff
@@ -391,7 +401,7 @@ private void spongePlaySound(final World world, final EntityPlayer player, final
 
     @Override
     public boolean bridge$affectsSpawning() {
-        return this.affectsSpawning && !this.isSpectator();
+        return this.affectsSpawning && !this.isSpectator() && !this.bridge$isVanished();
```
dualspiral: [If a player is targetable becuase `VANISH_PREVENTS_TARGETING` is `false`](https://github.com/SpongePowered/SpongeAPI/blob/stable-7/src/main/java/org/spongepowered/api/data/key/Keys.java#L2271), should spawning also be unaffected?

src/main/java/org/spongepowered/common/mixin/core/entity/player/EntityPlayerMixin.java
ImMorpheus: `VANISH_IGNORES_COLLISION` is calling `bridge$isUncollideable` and `bridge$setUncollideable`

https://github.com/SpongePowered/SpongeCommon/blob/9be33cf2c38e66fe8ac774503ef88f0b33d51faa/src/main/java/org/spongepowered/common/data/processor/multi/entity/InvisibilityDataProcessor.java#L66

https://github.com/SpongePowered/SpongeCommon/blob/9be33cf2c38e66fe8ac774503ef88f0b33d51faa/src/main/java/org/spongepowered/common/data/processor/multi/entity/InvisibilityDataProcessor.java#L84

src/main/java/org/spongepowered/common/mixin/core/entity/player/EntityPlayerMixin.java
ImMorpheus: good point, I'll change that

PR #2582, opened 2020-04-26
Improve config documentation

ST-DDT:
*SpongeCommon* | [**SpongeDocs**](https://github.com/SpongePowered/SpongeDocs/pull/832)

- Added missing comments
- Improved formatting of existing comments

**TODO**

- [x] Add missing comments
- [x] Improve lines marked with `TODO`

Requesting help for adding the missing comments.
The docs team takes care of formatting and wording.

Please tell me, if the formatting for the docs affects the java source readability negatively.


src/main/java/org/spongepowered/common/config/category/OptimizationCategory.java
```diff
@@ -33,103 +33,106 @@
...
-                                                   + "will allow for hoppers to save performing server -> client updates\n"
-                                                   + "when transferring items. Because hoppers can transfer items multiple\n"
-                                                   + "times per tick, these updates can get costly on the server, with\n"
-                                                   + "little to no benefit to the client. Because of the nature of the\n"
-                                                   + "change, the default will be 'false' due to the inability to pre-emptively\n"
-                                                   + "foretell whether mod compatibility will fail with these changes or not.\n"
-                                                   + "Refer to: https://github.com/PaperMC/Paper/blob/8175ec916f31dcd130fe0884fe46bdc187d829aa/Spigot-Server-Patches/0269-Optimize-Hoppers.patch\n"
-                                                   + "for more details.")
+    @Setting(value = "optimize-hoppers", comment = ""
+            + "Based on Aikar's optimizationo of Hoppers, setting this to 'true'\n"
```
phit: typo

ST-DDT: Fixed

src/main/java/org/spongepowered/common/config/category/EntityActivationModCategory.java
```diff
@@ -33,11 +33,12 @@
...
 
-    @Setting(value = "enabled", comment = "If 'false', entity activation rules for this mod will be ignored and always tick.")
+    @Setting(value = "enabled",
+            comment = "If 'false', entity activation rules for this mod will be ignored and always tick.")
     private boolean isEnabled = true;
-    @Setting(value = "defaults")
+    @Setting(value = "defaults", comment = "Default activation ranges used for mod entities unless overridden.")
     private Map<String, Integer> defaultRanges = new HashMap<>();
-    @Setting(value = "entities")
+    @Setting(value = "entities", comment = "???")
```
phit: this is the list of entities with the name as key and range as value

src/main/java/org/spongepowered/common/config/category/ModuleCategory.java
```diff
@@ -30,32 +30,33 @@
 @ConfigSerializable
 public class ModuleCategory extends ConfigCategory {
 
-    @Setting(value = "bungeecord")
+    @Setting(value = "bungeecord", comment = "???")
     private boolean pluginBungeeCord = false;
```
phit, line 36: BungeeCord Module, allows Sponge to be used with a Proxy like Bungeecord or Velocity, usually you want to also enable `bungeecord.ip-forwarding` with this

src/main/java/org/spongepowered/common/config/category/ModuleCategory.java
```diff
@@ -30,32 +30,33 @@
...
 
-    @Setting(value = "entity-activation-range")
+    @Setting(value = "entity-activation-range", comment = "???")
```
phit: Entity activation module, allows you to tweak entity activation ranges, similarly to "Entity Distance" in the latest 1.16 snapshots but per mob. Essentially it changes how far an entity can be away from a player, until it stops being updated (e.g. moving).

src/main/java/org/spongepowered/common/config/category/ModuleCategory.java
```diff
@@ -30,32 +30,33 @@
...
 
-    @Setting(value = "tileentity-activation", comment = "Controls block range and tick rate of tileentities. \n"
-                                                      + "Use with caution as this can break intended functionality.")
+    @Setting(value = "tileentity-activation", comment = ""
+            + "Controls block range and tick rate of tileentities.\n"
+            + "Use with caution as this can break intended functionality.")
     private boolean pluginTileEntityActivation = false;
 
-    @Setting(value = "entity-collisions")
+    @Setting(value = "entity-collisions", comment = "???")
```
phit: Module that allows you to configure the maximum amount of entities a specific entity or block can collide with.

src/main/java/org/spongepowered/common/config/category/ModuleCategory.java
```diff
@@ -30,32 +30,33 @@
...
     private boolean pluginEntityCollisions = true;
 
-    @Setting("timings")
+    @Setting(value = "timings", comment = "???")
```
phit: Module to run Aikar's Timings, profile your server performance and get a sharable web result https://github.com/aikar/timings#aikars-minecraft-timings-viewer-v2

src/main/java/org/spongepowered/common/config/category/ModuleCategory.java
```diff
@@ -30,32 +30,33 @@
...
+            + "before disabling all exploits with this toggle.")
     private boolean enableExploitPatches = true;
 
-    @Setting("optimizations")
+    @Setting(value = "optimizations")
     private boolean enableOptimizationPatches = true;
 
-    @Setting("tracking")
+    @Setting(value = "tracking", comment = "???")
     private boolean tracking = true;
```
phit, line 68: this used to disable some part of the tracking, i dont think its used anymore? https://github.com/SpongePowered/SpongeCommon/blob/4a9742b465d095769fa4e6ca5fbf22865e3de87b/src/main/java/org/spongepowered/common/mixin/plugin/TrackingPlugin.java#L48

src/main/java/org/spongepowered/common/config/category/TileEntityActivationModCategory.java
```diff
@@ -33,15 +33,16 @@
 @ConfigSerializable
 public class TileEntityActivationModCategory extends ConfigCategory {
 
-    @Setting(value = "enabled", comment = "If 'false', tileentity activation rules for this mod will be ignored and always tick.")
+    @Setting(value = "enabled",
+            comment = "If 'false', tileentity activation rules for this mod will be ignored and always tick.")
     private boolean isEnabled = true;
-    @Setting(value = "default-block-range")
+    @Setting(value = "default-block-range", comment = "???")
```
phit: my config file has this for it, not sure where that comment went? `Default activation block range used for all tileentities unless overridden.`

src/main/java/org/spongepowered/common/config/category/TileEntityActivationModCategory.java
```diff
@@ -33,15 +33,16 @@
...
     private Integer defaultBlockRange;
-    @Setting(value = "default-tick-rate")
+    @Setting(value = "default-tick-rate", comment = "???")
```
phit: same as previous `Default tick rate used for all tileentities unless overridden.`

src/main/java/org/spongepowered/common/config/category/TileEntityActivationModCategory.java
```diff
@@ -33,15 +33,16 @@
...
     private Integer defaultTickRate;
-    @Setting(value = "block-range")
+    @Setting(value = "block-range", comment = "???")
```
phit: list of tile/block entities with block name as key and range as value

src/main/java/org/spongepowered/common/config/category/TileEntityActivationModCategory.java
```diff
@@ -33,15 +33,16 @@
...
     private Map<String, Integer> tileEntityRangeList = new HashMap<>();
-    @Setting(value = "tick-rate")
+    @Setting(value = "tick-rate", comment = "???")
```
phit: list of tile/block entities with block name as key and rate as value

src/main/java/org/spongepowered/common/config/category/TileEntityActivationModCategory.java
```diff
@@ -33,15 +33,16 @@
 @ConfigSerializable
 public class TileEntityActivationModCategory extends ConfigCategory {
```
phit, line 35: I think there needs to be some general explanation in the config to what this does, maybe a copy of the module description? same for all categories really

src/main/java/org/spongepowered/common/config/category/TimingsCategory.java
```diff
@@ -33,22 +33,22 @@
 @ConfigSerializable
 public class TimingsCategory extends ConfigCategory {
 
-    @Setting
+    @Setting(comment = "???")
```
phit: ??? no clue i looked briefly and it seems to end in https://github.com/SpongePowered/SpongeCommon/blob/6077a35adcb7fd95d4886ffa5aa408a63ac1b4f2/src/main/java/org/spongepowered/common/relocate/co/aikar/timings/TimingHandler.java#L72

not sure what the point is

src/main/java/org/spongepowered/common/config/category/TimingsCategory.java
```diff
@@ -33,22 +33,22 @@
...
     private boolean verbose = false;
 
-    @Setting
+    @Setting(comment = "???")
```
phit: disable timings, might as well disable the module at that point?

src/main/java/org/spongepowered/common/config/category/TimingsCategory.java
```diff
@@ -33,22 +33,22 @@
...
     private boolean enabled = true;
 
-    @Setting("server-name-privacy")
+    @Setting(value = "server-name-privacy", comment = "???")
```
phit: Hides the server name in the Airkar webviewer to no leak the l33t server you are running

src/main/java/org/spongepowered/common/config/category/TimingsCategory.java
```diff
@@ -33,22 +33,22 @@
...
     private boolean serverNamePrivacy = false;
 
-    @Setting("hidden-config-entries")
+    @Setting(value = "hidden-config-entries", comment = "???")
```
phit: allows you to hide specific config entries from showing up on the webviewer, e.g. not leak your sql creds

src/main/java/org/spongepowered/common/config/category/TimingsCategory.java
```diff
@@ -33,22 +33,22 @@
...
     private List<String> hiddenConfigEntries = Lists.newArrayList("sponge.sql");
 
-    @Setting("history-interval")
+    @Setting(value = "history-interval", comment = "???")
     private int historyInterval = 300;
 
-    @Setting("history-length")
+    @Setting(value = "history-length", comment = "???")
```
phit: how long the list of said history can get

src/main/java/org/spongepowered/common/config/category/TimingsCategory.java
```diff
@@ -33,22 +33,22 @@
...
+    @Setting(value = "server-name-privacy", comment = "???")
     private boolean serverNamePrivacy = false;
 
-    @Setting("hidden-config-entries")
+    @Setting(value = "hidden-config-entries", comment = "???")
     private List<String> hiddenConfigEntries = Lists.newArrayList("sponge.sql");
 
-    @Setting("history-interval")
+    @Setting(value = "history-interval", comment = "???")
     private int historyInterval = 300;
```
phit, line 55: ticks % history-interval
modulo how often a tick should be pushed to the timings history, not sure exactly what the history is

src/main/java/org/spongepowered/common/config/type/GeneralConfigBase.java
```diff
@@ -42,29 +42,45 @@
...
+            + "Note: The radius uses the lower value of mob spawn range and server's view distance.")
     private SpawnerCategory spawner = new SpawnerCategory();
-    @Setting(value = "tileentity-activation")
+
+    @Setting(value = "tileentity-activation",
+            comment = "Configuration options related to activation ranges of tile entities.")
     private TileEntityActivationCategory tileEntityActivationCategory = new TileEntityActivationCategory();
-    @Setting
+
+    @Setting(comment = "???")
```
phit: > Module to run Aikar's Timings, profile your server performance and get a sharable web result https://github.com/aikar/timings#aikars-minecraft-timings-viewer-v2

src/main/java/org/spongepowered/common/config/type/GlobalConfig.java
```diff
@@ -58,46 +58,51 @@
...
-    @Setting
+    @Setting(comment = "Configuration options related to permissions and permission handling")
     private PermissionCategory permission = new PermissionCategory();
 
-    @Setting(value = "modules")
+    @Setting(value = "modules", comment = "Options related to optional modules, that can be enabled or disabled.")
     private ModuleCategory mixins = new ModuleCategory();
 
-    @Setting("ip-sets")
+    @Setting(value = "ip-sets", comment = "???")
```
phit: https://github.com/SpongePowered/SpongeForge/issues/239

phit (commented):
left some comments on all the ??? to give some context

generally would be nice to give all categories a general description of what it actually does for people not familiar with the games inner workings, stuff like tile entity activation means nothing to most users

src/main/java/org/spongepowered/common/config/category/ModuleCategory.java
ST-DDT: For now I just add a description. If this is no longer used it can be removed in a different PR.

src/main/java/org/spongepowered/common/config/category/StructureSaveCategory.java
```diff
@@ -33,19 +33,22 @@
...
 public class StructureSaveCategory extends ConfigCategory {
 
-    @Setting(value = "enabled", comment = "If 'false', disables the modification to prevent certain structures\n"
-                                          + "from saving to the world's data folder. If you wish to prevent certain\n"
-                                          + "structures from saving, leave this \"enabled=true\". When \'true\', the\n"
-                                          + "modification allows for specific \'named\' structures to NOT be saved to\n"
-                                          + "disk. Examples of some structures that are costly and somewhat irrelivent\n"
-                                          + "is 'mineshaft's, as they build several structures and save, even after\n"
-                                          + "finished generating.")
+    @Setting(value = "enabled", comment = "" // TODO: Unclear
```
ImMorpheus: This is used by `OptimizationCategory#useStructureSave`
https://github.com/SpongePowered/SpongeCommon/blob/bafbfa0a6849bee99d6fd1c0b8f563e5fb1bee35/src/main/java/org/spongepowered/common/config/category/OptimizationCategory.java#L137-L139

which is used by the OptimizationPlugin to decide if the Mixin should be applied:
https://github.com/SpongePowered/SpongeCommon/blob/bafbfa0a6849bee99d6fd1c0b8f563e5fb1bee35/src/main/java/org/spongepowered/common/mixin/plugin/OptimizationPlugin.java#L118-L119

The mixin is then used to allows servers to opt-out of saving specific structures 
https://github.com/SpongePowered/SpongeCommon/blob/bafbfa0a6849bee99d6fd1c0b8f563e5fb1bee35/src/main/java/org/spongepowered/common/mixin/optimization/world/gen/structure/MapGenStructureMixin_Structure_Saving.java#L53

This is basically a global switch for the `StructureSaveCategory`.

src/main/java/org/spongepowered/common/config/type/GlobalConfig.java
```diff
@@ -58,46 +58,53 @@
...
+            comment = "Options related to minecraft's movement checks, that can be enabled or disabled.")
     private MovementChecksCategory movementChecks = new MovementChecksCategory();
 
     @Setting(value = "broken-mods", comment = "Stopgap measures for dealing with broken mods")
     private BrokenModCategory brokenMods = new BrokenModCategory();
 
-    @Setting(value = "metrics")
+    @Setting(value = "metrics", comment = ""
+            + "Configuration options related to metric collection.\n"
+            + "Metric collection is disabled by default.")
```
ImMorpheus: The collection is set to `UNDEFINED` (which happens to be treated as disabled). It's not `disabled` by default.
Wouldn't this be better suited inside the `MetricsCategory` ?

src/main/java/org/spongepowered/common/config/category/TimingsCategory.java
```diff
@@ -33,22 +33,24 @@
...
+    @Setting(comment = "Enables the verbose mode for the timings module.")
     private boolean verbose = false;
 
-    @Setting
+    @Setting(comment = "???")
     private boolean enabled = true;
 
-    @Setting("server-name-privacy")
+    @Setting(value = "server-name-privacy", 
+            comment = "Hides the server name in the Airkar webviewer to no leak the l33t server you are running")
```
ImMorpheus: `Airkar` -> `Aikar`

ST-DDT: Fixed

src/main/java/org/spongepowered/common/config/type/GlobalConfig.java
ST-DDT: Fixed

src/main/java/org/spongepowered/common/config/category/StructureSaveCategory.java
ST-DDT: Well, this is a though one. I rephrased the setting's comments a bit and hope that it is now easier to understand.

ST-DDT:
AFAICT these are the last entries that are still missing:

- `TimingsCategory`
  - `enabled`
  - `hidden-config-entries`

src/main/java/org/spongepowered/common/config/category/TimingsCategory.java
ImMorpheus: This is used by the TimingsFactory.
Timings is enabled only if the `Timing` module and this option are both set to true

https://github.com/SpongePowered/SpongeCommon/blob/a865595b36ab95a876c8ebd46b464b23a73b09aa/src/main/java/org/spongepowered/common/relocate/co/aikar/timings/SpongeTimingsFactory.java#L63

src/main/java/org/spongepowered/common/config/category/TimingsCategory.java
ImMorpheus: Phit is correct.
Just a small note: these config entries are skipped and are never sent to the webviewer (aka they don't leave your server).
Additionally the sponge.sql section is **always** ignored and is never sent to the webviewer (regardless of this option).

ImMorpheus:
>     TimingsCategory
>         enabled
>         hidden-config-entries

Added a few comments

ST-DDT:
Thanks for your input! This should be ready for review and merge now.

IMO there seems to be a few config properties which seems to be redundant or unclear in at least their naming or function. Maybe that should be addressed for API 8 / the respective SpongeCommon release.

src/main/java/org/spongepowered/common/config/category/CollisionModCategory.java
```diff
@@ -34,17 +34,17 @@
 public class CollisionModCategory extends ConfigCategory {
 
     @Setting(value = "enabled", comment = "If 'false', entity collision rules for this mod will be ignored.")
-    private boolean isEnabled = true;
+    private final boolean isEnabled = true;
```
dualspiral: Undo these final modifiers on `@Setting` fields, Configurate's object mapping will not work with these changes.

ST-DDT: Fighting the auto formatter ... Done.

src/main/java/org/spongepowered/common/config/category/CustomDataRegistrationCategory.java
```diff
@@ -30,72 +30,68 @@
...
+     * Again, another auto-populated list of FAILED id's. This is never read from
+     * file, except after startup, to specify which id's can be "saved".
      */
-    @Setting(value = "failed-data-list", comment = "An auto generated list, by Sponge, to discover and list \n"
-                                                 + "all failed custom data deserializations at runtime due \n"
-                                                 + "to a lack of the registrations being made by a plugin. \n"
-                                                 + "Not to be confused by failed deserialization due to bad data. \n"
-                                                 + "Modifying the list will result in no effect as Sponge auto \n"
-                                                 + "generates this list. This is merely for user configuration.")
+    @Setting(value = "failed-data-list", comment = ""
```
dualspiral, line 67: This might seem like a nitpick, but there should not be an empty string like this at the end of a line. I would much rather just put `comment = ` on the next line (though, what was wrong with what was there before?)

src/main/java/org/spongepowered/common/config/type/GlobalConfig.java
```diff
@@ -58,46 +58,51 @@
     @Setting(comment = "Configuration options related to the Sql service, including connection aliases etc")
     private SqlCategory sql = new SqlCategory();
 
-    @Setting
+    @Setting(comment = "Configuration options related to commands, including command aliases and hidden commands.")
```
dualspiral, line 61: We're being very inconsistent here between "Configuration options" and just "Options" - choose one or the other please (though do we need to specify they are options? We _are_ in the configuration file! Though now I think about it more, if it natrually comes up in the description, so be it).

src/main/java/org/spongepowered/common/config/type/GlobalConfig.java
```diff
@@ -58,46 +58,51 @@
...
     private CommandsCategory commands = new CommandsCategory();
 
-    @Setting
+    @Setting(comment = "Configuration options related to permissions and permission handling")
     private PermissionCategory permission = new PermissionCategory();
 
-    @Setting(value = "modules")
+    @Setting(value = "modules", comment = "Options related to optional modules, that can be enabled or disabled.")
```
dualspiral: I think it's worth explaining what modules are a bit more. For example:

```
Sponge provides a number of modules that allow for enabling or disabling\n
certain specific features. These may be enabled or disabled below.\n\n
Any changes here require a server restart as modules are applied at startup.
```

src/main/java/org/spongepowered/common/config/type/GlobalConfig.java
```diff
@@ -58,46 +58,51 @@
...
     private ModuleCategory mixins = new ModuleCategory();
 
-    @Setting("ip-sets")
+    @Setting(value = "ip-sets", comment = ""
+            + "Automatically assigns (permission) contexts to users that use any of the given ips.\n"
+            + "This can be used to restrict/grand permissions, based on the player's source or target ip.")
```
dualspiral: s/grand/grant

src/main/java/org/spongepowered/common/config/type/GlobalConfig.java
```diff
@@ -58,46 +58,51 @@
...
     private Map<String, List<IpSet>> ipSets = new HashMap<>();
 
-    @Setting(value = "bungeecord")
+    @Setting(value = "bungeecord", comment = "Configuration options related for server proxies, such as bungeecord")
```
dualspiral: `Controls how Sponge interacts with server proxies, such as BungeeCord and Velocity. Requires that the 'bungeecord' module is enabled.`

src/main/java/org/spongepowered/common/config/category/CustomDataRegistrationCategory.java
ST-DDT: I changed it to this variant to put more characters in each line. So there are roughly 100 characters in each line both in the yaml and docs.
If I move comment to a new line then I only have 80 chars.
Also in most cases, the 100 char limit seem to match the sentence part length, so the lines end with a comma or dot.

src/main/java/org/spongepowered/common/config/type/GlobalConfig.java
ST-DDT: Fixed

src/main/java/org/spongepowered/common/config/type/GlobalConfig.java
ST-DDT: Changed

src/main/java/org/spongepowered/common/config/type/GlobalConfig.java
ST-DDT: Fixed

src/main/java/org/spongepowered/common/config/type/GlobalConfig.java
ST-DDT: Changed

ImMorpheus (approved):

gabizou (approved):
Short of `""` being abused for text alignment, documentation looks fine

PR #2584, opened 2020-04-29
Don't call closeChannel on the Main Thread

LXGaming:
Reported by Skyline/Skyine918 on Discord and GitHub
https://github.com/SpongePowered/SpongeForge/issues/3145

The `NetworkManager::closeChannel` method will block the thread until the channel has closed.

The fix replicates Vanilla logic for closing connections as seen in `NetHandlerPlayServer::disconnect` and `NetworkSystem::networkTick` which is too have a Netty Thread handle closing the connection and to also disable auto read.


dualspiral:
Thanks for spotting this - I'll pull this in now.

PR #2585, opened 2020-04-29
Fix target of StateImplementationMixin_API

Lignium:
Fixes warning:

`@Mixin target net/minecraft/block/state/BlockStateContainer$StateImplementation is public in mixins.common.api.json:mcp.block.state.StateImplementationMixin_API and should be specified in value`

Already my eyes hurt to see this yellow string...


ImMorpheus:
This should be added to the OCD list to be merged at a later time in bulk.

PR #2588, opened 2020-05-01
Implement a basic PlaceholderService

dualspiral:
[Sponge API](https://github.com/SpongePowered/SpongeAPI/pull/2116) | **SpongeCommon (that's here!)** |  [SpongeDocs](https://github.com/SpongePowered/SpongeDocs/pull/834)

Simple implementation of the placeholder API. See API PR for more details.

* Also, moved economy registry modules in Common into a `service` package to make it clear they are from services.

### DIFF
--- a/src/main/java/org/spongepowered/common/registry/CommonModuleRegistry.java
+++ b/src/main/java/org/spongepowered/common/registry/CommonModuleRegistry.java
@@ -103,12 +103,15 @@ import org.spongepowered.api.scoreboard.objective.Objective;
 import org.spongepowered.api.scoreboard.objective.displaymode.ObjectiveDisplayMode;
 import org.spongepowered.api.service.economy.account.AccountDeletionResultType;
 import org.spongepowered.api.service.economy.transaction.TransactionType;
+import org.spongepowered.api.text.placeholder.PlaceholderContext;
+import org.spongepowered.api.text.placeholder.PlaceholderParser;
 import org.spongepowered.api.statistic.Statistic;
 import org.spongepowered.api.statistic.StatisticType;
 import org.spongepowered.api.text.chat.ChatType;
 import org.spongepowered.api.text.chat.ChatVisibility;
 import org.spongepowered.api.text.format.TextColor;
 import org.spongepowered.api.text.format.TextStyle;
+import org.spongepowered.api.text.placeholder.PlaceholderText;
 import org.spongepowered.api.text.selector.SelectorType;
 import org.spongepowered.api.text.serializer.TextSerializer;
 import org.spongepowered.api.util.ban.Ban;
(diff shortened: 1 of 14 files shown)


src/main/java/org/spongepowered/common/registry/type/service/placeholder/PlaceholderParserRegistryModule.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+        this.register(extraCatalog);
+    }
+
+    @Override
+    public void registerDefaults() {
+        register(new PlaceholderParser() {
+            @Override
+            public Text parse(PlaceholderText placeholderText) {
+                return placeholderText.getAssociatedReceiver().<Text>map(x -> Text.of(((CommandSource) x).getName())).orElse(Text.EMPTY);
```
ST-DDT: You violate the api contract here if the receiver is not a command source.

src/main/java/org/spongepowered/common/registry/type/service/placeholder/PlaceholderParserRegistryModule.java
```diff
@@ -0,0 +1,98 @@
...
+            public String getName() {
+                return "Name";
+            }
+        });
+        register(new PlaceholderParser() {
+            @Override
+            public Text parse(PlaceholderText placeholderText) {
+                return Text.of(
+                        placeholderText.getAssociatedReceiver()
+                            .map(x -> ((Locatable) x).getWorld().getName())
```
ST-DDT: Same here

src/main/java/org/spongepowered/common/registry/type/service/placeholder/PlaceholderParserRegistryModule.java
dualspiral: Yep, I'm meant to filter first.

src/main/java/org/spongepowered/common/registry/type/service/placeholder/PlaceholderParserRegistryModule.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+        this.register(extraCatalog);
+    }
+
+    @Override
+    public void registerDefaults() {
+        register(new PlaceholderParser() {
+            @Override
+            public Text parse(PlaceholderText placeholderText) {
+                return placeholderText.getAssociatedObject()
```
gabizou: Nitpick: Could this be reduced down to a buildable method instead of two anonymous classes?

src/main/java/org/spongepowered/common/service/placeholder/SpongePlaceholderService.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private Supplier<Object> createMessageReceiverSupplier(final Object receiver) {
+        if (receiver instanceof ConsoleSource) {
+            return () -> (ConsoleSource) SpongeImpl.getServer();
+        } else if (receiver instanceof Player) {
+            final UUID uuid = ((Player) receiver).getUniqueId();
+            return () -> SpongeImpl.getGame().getServer().getPlayer(uuid).orElse(null);
+        }
+
+        return () -> receiver;
```
gabizou: Should this be a `WeakReference` maybe? Just in case?

gabizou (approved):
Small nitpicks, but otherwise looks good. Only worry is that weak reference mention. Better to throw an NPE because someone provided an `Entity` that could leak several more objects and the entity is now so far removed.

src/main/java/org/spongepowered/common/registry/type/service/placeholder/PlaceholderParserRegistryModule.java
dualspiral: It could and I don't really see the harm in it, but I don't necessarily think such a builder will get used by other plugins, hence why I did this. I think it's far more likely they'll implement the methods on their own.

I could do something such as this:

```
PlaceholderParser.builder().plugin(Object).id(String).name(String).executor(Function<PlaceholderText, Text>).build();
```

where the ID is the unnamespaced version?

src/main/java/org/spongepowered/common/service/placeholder/SpongePlaceholderService.java
dualspiral: Yes, that's a good idea.

dualspiral: Just to clarify, I've moved this method into the builder itself. To also make it clear that you can use suppliers to make a `PlaceholderText`, I've added overloads onto the service itself, such that you can either provide an object, or a `Supplier<Object>` for the text.

src/main/java/org/spongepowered/common/registry/type/service/placeholder/PlaceholderParserRegistryModule.java
dualspiral: I've done as I suggested above.

gabizou (approved):
LGTM

PR #2591, opened 2020-05-02
Fix tileentitysign memory leak

ImMorpheus:
`TileEntitySign` has a player field used by `NetHandlerPlayServer#processUpdateSign`
The player field is set in `EntityPlayerMP#openEditSign` but it's never set to null.
```
if (!tileentitysign.getIsEditable() || tileentitysign.getPlayer() != this.player) {
    this.server.logWarning("Player " + this.player.getName() + " just tried to change non-editable sign");
    return;
}
```

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
@@ -271,11 +271,9 @@ public abstract class NetHandlerPlayServerMixin implements NetHandlerPlayServerB
         return ZERO_LENGTH_ARRAY; // will bypass the for loop after this method.
     }
 
-    @Inject(method = "processUpdateSign", locals = LocalCapture.CAPTURE_FAILHARD,
-            at = @At(value = "INVOKE", target = "Lnet/minecraft/world/WorldServer;notifyBlockUpdate(Lnet/minecraft/util/math/BlockPos;Lnet/minecraft/block/state/IBlockState;Lnet/minecraft/block/state/IBlockState;I)V"))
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
```diff
@@ -270,6 +271,13 @@
         return ZERO_LENGTH_ARRAY; // will bypass the for loop after this method.
     }
 
+    @Inject(method = "processUpdateSign", locals = LocalCapture.CAPTURE_FAILHARD,
+            at = @At(value = "INVOKE", target = "Lnet/minecraft/world/WorldServer;notifyBlockUpdate(Lnet/minecraft/util/math/BlockPos;Lnet/minecraft/block/state/IBlockState;Lnet/minecraft/block/state/IBlockState;I)V"))
+    private void impl$setPlayer(CPacketUpdateSign packetIn, final CallbackInfo ci, WorldServer worldserver,
```
dualspiral: Why not just redirect `tileEntitySign.markDirty()` and call `tileEntitySign.setPlayer(null)` there? Better than relying on local capture.

ImMorpheus: Done

XakepSDK:
What if i place sign without saving text and kill client? Will it be cleared?

dualspiral:
Maybe eventually when the chunk unloads or someone else goes to edit the sign, it will happen.

Do note that this is a best effort guess here. We could go all out and make a weak reference field to store this in instead, but quite frankly, this is going to fix the majority of the leak involved and we're just not going to catch everything.

dualspiral (dismissed):
~~Looks good.~~

dualspiral:
Actually, wait, you forgot to call `markDirty` within the redirect - it's still important to do that!

PR #2593, opened 2020-05-02
Fix map optimization dimension check for vanilla servers Fixes #2592

### LINKED ISSUE #2592: Map-optimization means that players aren't shown on maps
**I am currently running**

- SpongeForge version: Not yet tested but i believe the problem code is in spongecommon
- SpongeVanilla version: spongevanilla-1.12.2-7.2.0
- Java version: 8

- Plugins/Mods:
None.

**Issue Description**
When map-optimization in global.conf is set to true, a player decoration (the pointer thing) showing where you are, doesn't appear. It does appear if you disable map-optimization.

1. Create a map by right clicking an empty map,
2. Observe that there is no player decoration on the map

clienthax:
Byte + int equality check was failing before on spongevanilla.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/EntityLivingBaseMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/EntityLivingBaseMixin.java
@@ -928,12 +928,12 @@ public abstract class EntityLivingBaseMixin extends EntityMixin implements Livin
 
                 slotAdapter = slotLens.getAdapter(((InventoryAdapter) inventory).bridge$getFabric(), (Inventory) inventory);
             } else {
-                if (this.slotLens.isEmpty()) {
-                    for (final EntityEquipmentSlot slot : EntityEquipmentSlot.values()) {
-                        this.slotLens.put(slot, new SlotLensImpl(slot.getSlotIndex()));
-                    }
+                SlotLens lens = this.slotLens.get(entityEquipmentSlot);
+                if (lens == null) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/optimization/world/storage/MapDataMixin_MapOptimization.java
```diff
@@ -178,7 +178,7 @@ public void updateVisiblePlayers(final EntityPlayer player, final ItemStack mapS
                 if (!mixinMapInfo.mapOptimizationBridge$isValid()) {
                     this.mapDecorations.remove(player.getName());
                 } else {
-                    if (this.trackingPosition && mapOptimizationImpl$dimensionField.get(this).equals(player.dimension)) {
+                    if (this.trackingPosition && (((Number)mapOptimizationImpl$dimensionField.get(this)).intValue() == player.dimension)) {
```
dualspiral: Couldn't you do this instead, rather than boxing and casting to `Number`?

```suggestion
                    if (this.trackingPosition && mapOptimizationImpl$dimensionField.getInt(this) == player.dimension) {
```

[I say this because of what the javadocs have to say about `Field#getInt(Object)`](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Field.html#getInt-java.lang.Object-), `byte` to `int` is a widening conversion so this should work fine in both cases, and would allow for this to succeed.

(Also, style nitpick - space between the cast operator and the target!)

clienthax:
Ugh.. just kill me with git

PR #2595, opened 2020-05-02
Fix/invisibility processor

ImMorpheus:
https://github.com/SpongePowered/SpongeAPI/blob/f12f8e7743844b8761b890ce3e440911e97c6f38/src/main/java/org/spongepowered/api/data/key/Keys.java#L2266

https://github.com/SpongePowered/SpongeAPI/blob/f12f8e7743844b8761b890ce3e440911e97c6f38/src/main/java/org/spongepowered/api/data/key/Keys.java#L2252

when turning of vanish with something like this:

```
id.set(Keys.VANISH, false);
id.set(Keys.VANISH_IGNORES_COLLISION, false);
id.set(Keys.VANISH_PREVENTS_TARGETING, false);
```

the invisibilitydataprocessor would just ignore `collision` and `untargetable` (keeping the old values) and set `vanish` to false. 

The redirect was also breaking the api contract (uncollideable is ignored if you're not in vanish).

### DIFF
--- a/src/main/java/org/spongepowered/common/data/processor/multi/entity/InvisibilityDataProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/multi/entity/InvisibilityDataProcessor.java
@@ -67,13 +67,9 @@ public class InvisibilityDataProcessor
         final boolean untargetable = (Boolean) keyValues.get(Keys.VANISH_PREVENTS_TARGETING);
         final boolean vanish = (Boolean) keyValues.get(Keys.VANISH);
         dataHolder.bridge$setInvisible(invis);
-        if (vanish) {
-            dataHolder.bridge$setVanished(true);
-            dataHolder.bridge$setUncollideable(collision);
-            dataHolder.bridge$setUntargetable(untargetable);
-        } else {
-            dataHolder.bridge$setVanished(false);
-        }
+        dataHolder.bridge$setVanished(vanish);
+        dataHolder.bridge$setUncollideable(collision);
+        dataHolder.bridge$setUntargetable(untargetable);
         return true;
     }
 
--- a/src/main/java/org/spongepowered/common/data/processor/multi/entity/InvisibilityDataProcessor.java
+++ b/src/main/java/org/spongepowered/common/data/processor/multi/entity/InvisibilityDataProcessor.java
@@ -68,8 +68,8 @@ public class InvisibilityDataProcessor
         final boolean vanish = (Boolean) keyValues.get(Keys.VANISH);
         dataHolder.bridge$setInvisible(invis);
         dataHolder.bridge$setVanished(vanish);
-        dataHolder.bridge$setUncollideable(collision);
-        dataHolder.bridge$setUntargetable(untargetable);
+        dataHolder.bridge$setUncollideable(vanish && collision);
+        dataHolder.bridge$setUntargetable(vanish && untargetable);
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/data/processor/multi/entity/InvisibilityDataProcessor.java
```diff
@@ -67,13 +67,9 @@ protected boolean set(final VanishableBridge dataHolder, final Map<Key<?>, Objec
...
         dataHolder.bridge$setInvisible(invis);
-        if (vanish) {
-            dataHolder.bridge$setVanished(true);
-            dataHolder.bridge$setUncollideable(collision);
-            dataHolder.bridge$setUntargetable(untargetable);
-        } else {
-            dataHolder.bridge$setVanished(false);
-        }
+        dataHolder.bridge$setVanished(vanish);
+        dataHolder.bridge$setUncollideable(collision);
```
dualspiral: So... is this right? When vanish is false, the entity in question should unquestionably be collidable. I'm a little worried that we're getting into a spiderweb here where we have a bridge method named one thing and the key named the other.

One school of thought is setting `VANISH_IGNORES_COLLISION` and `VANISH_PREVENTS_TARGETING` to `true` should not be cleared by setting `VANISH` to `false`. That feels largely expected and I err on the side of agreeing with that. I'm thus not against this change myself, but we need to make other changes to insulate ourselves against other issues.

However, I know this is in impl, but `bridge$setUncollideable` and `bridge$setUntargetable` should really be named better to indicate that they are on vanish only. Their representive `is` methods should then be renamed similarly AND include a vanish check within their bodies. I'd rather check one method for collisions, rather than tacking on a vanish check everywhere too because it's not obvious that's what needs to happen.

If we want/need separate non-vanish uncollidable/targetting states then we need to think about that more, because we now need to think about how this will work with the vanish based states.

dualspiral (commented):
I'm going to ask @gabizou to look at this, because we've previously had a little back and forth about what should/shouldn't be set when vanish is off. However, I do have a reservation about this, I think we need to step back and think about what's going on and if we can avoid another bug in future because something wasn't obvious...

src/main/java/org/spongepowered/common/data/processor/multi/entity/InvisibilityDataProcessor.java
ImMorpheus: From what I gathered from the implementation and the API  (`This state will be ignored if the {@link Entity} is not also vanished as per {@link #VANISH}.`) the uncollideable state should be ignored in the implementation, not in the data processor.

> One school of thought is setting VANISH_IGNORES_COLLISION and VANISH_PREVENTS_TARGETING to true should not be cleared by setting VANISH to false. That feels largely expected and I err on the side of agreeing with that. I'm thus not against this change myself, but we need to make other changes to insulate ourselves against other issues.

Agreed, but not only that. As it stands now, setting `VANISH_IGNORES_COLLISION` explicitily to `false` is ignored if you're also setting `VANISH` to `false`. That doesn't feel right to me.

> However, I know this is in impl, but bridge$setUncollideable and bridge$setUntargetable should really be named better to indicate that they are on vanish only. Their representive is methods should then be renamed similarly AND include a vanish check within their bodies. I'd rather check one method for collisions, rather than tacking on a vanish check everywhere too because it's not obvious that's what needs to happen.

EDIT: I'm not sure anymore about this. Discussions welcomed

src/main/java/org/spongepowered/common/data/processor/multi/entity/InvisibilityDataProcessor.java
```diff
@@ -67,13 +67,9 @@ protected boolean set(final VanishableBridge dataHolder, final Map<Key<?>, Objec
...
+        dataHolder.bridge$setUntargetable(untargetable);
```
dualspiral: Okay, so I was looking about and thinking about previous discussions, and if vanish is `false`, the other two keys are removed from the user compound anyway ([see here](https://github.com/SpongePowered/SpongeCommon/blob/7eeea95c4b167d8bd728950e349e0bb9480090ba/src/main/java/org/spongepowered/common/entity/player/SpongeUser.java#L448-L451) and [here](https://github.com/SpongePowered/SpongeCommon/blob/7eeea95c4b167d8bd728950e349e0bb9480090ba/src/main/java/org/spongepowered/common/entity/player/SpongeUser.java#L469-L473)). So in the interests of getting fix in, I think we should change the calls as follows:

```suggestion
        dataHolder.bridge$setUncollideable(vanish && collision);
        dataHolder.bridge$setUntargetable(vanish && untargetable);
```

because that's clearly the intention behind what's going on - even though I'm not 100% convinced it's what we should be doing in the wider notion, as I have talked about above.

I would much rather have the `isUncolliable` methods renamed and updated to also consider vanish state, but I am more interested in having this fixed without having this stuck for ages because it's otherwise not good for our users.

PR #1769, opened 2018-02-04
Refactoring and implementation of the LaunchProjectileEvent

RedNesto:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1766) | **SpongeCommon**

This PR implement its SpongeAPI counterpart.

### When is the event fired ?

Here are each cases I tested:
- [x] A non-player entity shoots a projectile (skeletons shooting arrows, llamas spitting, ghasts shooting fireballs, snowmans shooting snowballs, blaze shooting small fireballs)
- [x] A player shoots arrows (with a bow heh)
- [x] A player throws projectiles (snowballs, eggs, potions - splash and lingering -, enderpearls, eyes of ender, using fishing rod, experience bottle)
- [x] A dispenser dispenses projectiles (arrows, snowballs, eggs, potions - splash and lingering -, experience bottle)
- [x] A plugin shoots a projectile
- [x] A player uses a firework.
- [x] A dispenser dispenses a firework.
- [x] A Wither attacks.
- [x] The dragon shoots a fireball

Tell me if you see a case of projectile I missed.

[And there is the content of a test plugin](https://gist.github.com/RedNesto/a1c45a36bf79e190dae512ee867daeac)

~Note: It seems that there are two projectile-related `EventContextKeys`, `PROJECTILE_SOURCE` and `THROWER`, from what I saw, only one would be necessary, I may address this issue later if it is considered as one.~
Fixed in 5d0184de3b1eeba951c5873d9c20c919c7a8d307

PR #2598, opened 2020-05-06
Implement LaunchProjectileEvent, Take 2

RedNesto:
This is an adapted version of #1769 for API 7

One big issue with the current event is that it is limited to one projectile, where mods (and crossbows with the Multishot enchantment) can shoot multiple at once . I'd like to fix that later for API 8 in another PR.

**What works**

- [x] A non-player entity shoots a projectile (skeletons shooting arrows, llamas spitting, ghasts shooting fireballs, snowmans shooting snowballs, blaze shooting small fireballs, withers, ender dragon)
  - [x] Cancelling, at least for vanilla entities, that shoot one projectile and not do not consume stacks from an inventory
- [x] A player shoots arrows
  - [x] Cancelling, if only one projectile is launched
  - [ ] ~Cancelling when several projectiles are shot (possible with mods) and only one is cancelled: this may restore all consumed stacks~ - I guess we'll have to forget this one for API 7, let's consider one cancelled event as the cancellation of all the other events
- [x] A player throws projectiles (snowballs, eggs, potions - splash and lingering -, enderpearls, eyes of ender, using fishing rod, experience bottle)
  - [x] Cancelling (assuming it only launches one at a time)
- [x] A dispenser launches a projectile (arrows, snowballs, eggs, potions - splash and lingering -, experience bottle)
  - [x] Cancelling (assuming it only launches one at a time)
- [x] A plugin shoots a projectile
- [x] A player uses a firework - I had to add a condition operand, otherwise fireworks are always restored to the player's inventory. I need help on this one
  - [x] Cancelling
- [x] A dispenser dispenses a firework
  - [x] Cancelling

There can be issues in cases where multiple projectiles are launched at the same time (but in vanilla it shouldn't be an issue). I think the best way to handle it is to cancel all events if one is cancelled.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
@@ -88,6 +88,7 @@ import org.spongepowered.api.entity.living.Human;
 import org.spongepowered.api.entity.living.Living;
 import org.spongepowered.api.entity.living.player.Player;
 import org.spongepowered.api.entity.living.player.User;
+import org.spongepowered.api.entity.projectile.Projectile;
 import org.spongepowered.api.entity.projectile.source.ProjectileSource;
 import org.spongepowered.api.event.CauseStackManager;
 import org.spongepowered.api.event.Event;
@@ -110,6 +111,7 @@ import org.spongepowered.api.event.entity.RotateEntityEvent;
 import org.spongepowered.api.event.entity.SpawnEntityEvent;
 import org.spongepowered.api.event.entity.ai.SetAITargetEvent;
 import org.spongepowered.api.event.entity.explosive.DetonateExplosiveEvent;
+import org.spongepowered.api.event.entity.projectile.LaunchProjectileEvent;
 import org.spongepowered.api.event.item.inventory.ChangeInventoryEvent;
 import org.spongepowered.api.event.item.inventory.ClickInventoryEvent;
 import org.spongepowered.api.event.item.inventory.CraftItemEvent;
@@ -1786,4 +1788,39 @@ public class SpongeCommonEventFactory {
         return event;
     }
 
+    /**
+     * Common method to fire a LaunchProjectileEvent and populate its context, then calls SpawnEntityEvent.
+     *
+     * This must be called in the main thread.
+     *  @param frame a frame to populate the context and use for the event.
+     * @param player the player who launched the projectiles.
+     * @param projectile the projectiles to fire.
+     * @param context the context in which this method is called
+     */
+    public static void callProjectileLaunchEvent(CauseStackManager.StackFrame frame, EntityPlayerMP player, Projectile projectile,
+            PhaseContext<?> context) {
+        if (!Sponge.getServer().isMainThread()) {
+            SpongeImpl.getLogger().error("callProjectileLaunchEvent called outside of the main thread");
+            return;
+        }
+
+        frame.addContext(EventContextKeys.SPAWN_TYPE, SpawnTypes.PROJECTILE);
+        frame.addContext(EventContextKeys.PROJECTILE_SOURCE, (ProjectileSource) player);
+
+        LaunchProjectileEvent launchProjectileEvent =
+                SpongeEventFactory.createLaunchProjectileEvent(Sponge.getCauseStackManager().getCurrentCause(), projectile);
+        if (SpongeImpl.postEvent(launchProjectileEvent)) {
+            TrackedInventoryBridge playerMixinContainer = (TrackedInventoryBridge) player.openContainer;
+            PacketPhaseUtil.handleSlotRestore(player, player.openContainer, playerMixinContainer.bridge$getCapturedSlotTransactions(), true);
+        } else {
+            if (ShouldFire.SPAWN_ENTITY_EVENT) {
+                List<Entity> entities = new ArrayList<>();
+                entities.add(projectile);
+                callSpawnEntity(entities, context);
+            } else {
+                Optional<User> creator = Optional.of((User) player);
+                EntityUtil.processEntitySpawn(projectile, () -> creator);
+            }
+        }
+    }
 }
--- a/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
@@ -1794,10 +1794,10 @@ public class SpongeCommonEventFactory {
      * This must be called in the main thread.
      *  @param frame a frame to populate the context and use for the event.
      * @param player the player who launched the projectiles.
-     * @param projectile the projectiles to fire.
-     * @param context the context in which this method is called
+     * @param projectiles the launched projectiles.
+     * @param context the context in which this method is called.
      */
-    public static void callProjectileLaunchEvent(CauseStackManager.StackFrame frame, EntityPlayerMP player, Projectile projectile,
+    public static void callProjectileLaunchEvent(CauseStackManager.StackFrame frame, EntityPlayerMP player, List<Projectile> projectiles,
             PhaseContext<?> context) {
         if (!Sponge.getServer().isMainThread()) {
             SpongeImpl.getLogger().error("callProjectileLaunchEvent called outside of the main thread");
@@ -1807,18 +1807,22 @@ public class SpongeCommonEventFactory {
         frame.addContext(EventContextKeys.SPAWN_TYPE, SpawnTypes.PROJECTILE);
         frame.addContext(EventContextKeys.PROJECTILE_SOURCE, (ProjectileSource) player);
 
-        LaunchProjectileEvent launchProjectileEvent =
-                SpongeEventFactory.createLaunchProjectileEvent(Sponge.getCauseStackManager().getCurrentCause(), projectile);
-        if (SpongeImpl.postEvent(launchProjectileEvent)) {
-            TrackedInventoryBridge playerMixinContainer = (TrackedInventoryBridge) player.openContainer;
-            PacketPhaseUtil.handleSlotRestore(player, player.openContainer, playerMixinContainer.bridge$getCapturedSlotTransactions(), true);
+        for (Projectile projectile : projectiles) {
+            LaunchProjectileEvent event = SpongeEventFactory.createLaunchProjectileEvent(frame.getCurrentCause(), projectile);
+            if (SpongeImpl.postEvent(event)) {
+                List<SlotTransaction> transactions = ((TrackedInventoryBridge) player.openContainer).bridge$getCapturedSlotTransactions();
+                PacketPhaseUtil.handleSlotRestore(player, player.openContainer, transactions, true);
+                // Since the event can't handle several plugins at once we interpret one cancel as a cancel of the others
+                return;
+            }
+        }
+
+        if (ShouldFire.SPAWN_ENTITY_EVENT) {
+            //noinspection unchecked,rawtypes
+            callSpawnEntity((List) projectiles, context);
         } else {
-            if (ShouldFire.SPAWN_ENTITY_EVENT) {
-                List<Entity> entities = new ArrayList<>();
-                entities.add(projectile);
-                callSpawnEntity(entities, context);
-            } else {
-                Optional<User> creator = Optional.of((User) player);
+            Optional<User> creator = Optional.of((User) player);
+            for (Projectile projectile : projectiles) {
                 EntityUtil.processEntitySpawn(projectile, () -> creator);
             }
         }
--- a/src/main/java/org/spongepowered/common/event/tracking/phase/block/DispensePhaseState.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/phase/block/DispensePhaseState.java
@@ -25,15 +25,26 @@
 package org.spongepowered.common.event.tracking.phase.block;
 
 import org.spongepowered.api.Sponge;
+import org.spongepowered.api.block.BlockSnapshot;
+import org.spongepowered.api.block.tileentity.TileEntity;
+import org.spongepowered.api.data.LocatableSnapshot;
 import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.entity.projectile.Projectile;
+import org.spongepowered.api.entity.projectile.source.ProjectileSource;
+import org.spongepowered.api.event.CauseStackManager;
+import org.spongepowered.api.event.SpongeEventFactory;
 import org.spongepowered.api.event.cause.EventContextKeys;
 import org.spongepowered.api.event.cause.entity.spawn.SpawnTypes;
+import org.spongepowered.api.world.Location;
+import org.spongepowered.common.SpongeImpl;
+import org.spongepowered.common.bridge.block.DispenserBridge;
 import org.spongepowered.common.event.SpongeCommonEventFactory;
 import org.spongepowered.common.event.tracking.TrackingUtil;
 import org.spongepowered.common.event.tracking.context.GeneralizedContext;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 
 final class DispensePhaseState extends BlockPhaseState {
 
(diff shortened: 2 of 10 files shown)


src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -1786,4 +1788,43 @@ public static ItemStack throwDropItemAndConstructEvent(final net.minecraft.entit
...
+    /**
+     * Common method to fire a LaunchProjectileEvent and populate its context, then calls SpawnEntityEvent.
+     *
+     * This must be called in the main thread.
+     *  @param frame a frame to populate the context and use for the event.
+     * @param player the player who launched the projectiles.
+     * @param projectiles the launched projectiles.
+     * @param context the context in which this method is called.
+     */
+    public static void callProjectileLaunchEvent(CauseStackManager.StackFrame frame, EntityPlayerMP player, List<Projectile> projectiles,
```
gabizou, line 1800:
```suggestion
    public static void callProjectileLaunchEvent(final CauseStackManager.StackFrame frame, final EntityPlayerMP player, final List<Projectile> projectiles,
```

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -1786,4 +1788,43 @@ public static ItemStack throwDropItemAndConstructEvent(final net.minecraft.entit
...
+            PhaseContext<?> context) {
```
gabizou, line 1801:
```suggestion
            final PhaseContext<?> context) {
```

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -1786,4 +1788,43 @@ public static ItemStack throwDropItemAndConstructEvent(final net.minecraft.entit
...
+        if (!Sponge.getServer().isMainThread()) {
+            SpongeImpl.getLogger().error("callProjectileLaunchEvent called outside of the main thread");
+            return;
+        }
+
+        frame.addContext(EventContextKeys.SPAWN_TYPE, SpawnTypes.PROJECTILE);
+        frame.addContext(EventContextKeys.PROJECTILE_SOURCE, (ProjectileSource) player);
+
+        for (Projectile projectile : projectiles) {
+            LaunchProjectileEvent event = SpongeEventFactory.createLaunchProjectileEvent(frame.getCurrentCause(), projectile);
```
gabizou, line 1811:
```suggestion
            final LaunchProjectileEvent event = SpongeEventFactory.createLaunchProjectileEvent(frame.getCurrentCause(), projectile);
```

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -1786,4 +1788,43 @@ public static ItemStack throwDropItemAndConstructEvent(final net.minecraft.entit
...
 
+    /**
+     * Common method to fire a LaunchProjectileEvent and populate its context, then calls SpawnEntityEvent.
+     *
+     * This must be called in the main thread.
+     *  @param frame a frame to populate the context and use for the event.
+     * @param player the player who launched the projectiles.
+     * @param projectiles the launched projectiles.
+     * @param context the context in which this method is called.
+     */
```
gabizou, line 1799: Gonna need a `@SuppressWarnings({"rawTypes", "uncheckedCast"})` or something for the raw `List` casting going on.

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -1786,4 +1788,43 @@ public static ItemStack throwDropItemAndConstructEvent(final net.minecraft.entit
...
+                // Since the event can't handle several plugins at once we interpret one cancel as a cancel of the others
+                return;
+            }
+        }
+
+        if (ShouldFire.SPAWN_ENTITY_EVENT) {
+            //noinspection unchecked,rawtypes
+            callSpawnEntity((List) projectiles, context);
+        } else {
+            Optional<User> creator = Optional.of((User) player);
```
gabizou, line 1824:
```suggestion
            final Optional<User> creator = Optional.of((User) player);
```

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -1786,4 +1788,43 @@ public static ItemStack throwDropItemAndConstructEvent(final net.minecraft.entit
...
+            for (Projectile projectile : projectiles) {
+                EntityUtil.processEntitySpawn(projectile, () -> creator);
```
gabizou, line 1826: Was almost sure there was a list based one that would work?

src/main/java/org/spongepowered/common/event/tracking/phase/block/DispensePhaseState.java
```diff
@@ -58,8 +69,44 @@ public void unwind(final GeneralizedContext phaseContext) {
             });
         phaseContext.getCapturedEntitySupplier()
             .acceptAndClearIfNotEmpty(entities -> {
-                Sponge.getCauseStackManager().addContext(EventContextKeys.SPAWN_TYPE, SpawnTypes.DISPENSE);
-                SpongeCommonEventFactory.callSpawnEntity(entities, phaseContext);
+                List<Entity> projectiles = new ArrayList<>();
+                List<Entity> rest = new ArrayList<>();
```
gabizou, line 73:
```suggestion
                final List<Entity> projectiles = new ArrayList<>();
                final List<Entity> rest = new ArrayList<>();
```

src/main/java/org/spongepowered/common/event/tracking/phase/block/DispensePhaseState.java
```diff
@@ -58,8 +69,44 @@ public void unwind(final GeneralizedContext phaseContext) {
...
+                for (Entity entity : entities) {
+                    if (entity instanceof Projectile) {
+                        projectiles.add(entity);
+                    } else {
+                        rest.add(entity);
+                    }
+                }
+
+                if (!rest.isEmpty()) {
+                    Sponge.getCauseStackManager().addContext(EventContextKeys.SPAWN_TYPE, SpawnTypes.DISPENSE);
```
gabizou, line 83: Should not be adding to the context outside a frame, or without popping the context afterwards.

src/main/java/org/spongepowered/common/event/tracking/phase/block/DispensePhaseState.java
```diff
@@ -58,8 +69,44 @@ public void unwind(final GeneralizedContext phaseContext) {
...
+                    SpongeCommonEventFactory.callSpawnEntity(rest, phaseContext);
+                }
+
+                if (!projectiles.isEmpty()) {
+                    try (CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
+                        frame.addContext(EventContextKeys.SPAWN_TYPE, SpawnTypes.PROJECTILE);
+                        TileEntity tileEntity = phaseContext.getSource(BlockSnapshot.class)
```
gabizou, line 90: Dispensing uses a `BlockSnapshot`? If you want to retain a reference to the original `TileEntity` and making this performant (we always do), make a specialized `PhaseContext` for `DispensePhaseState` that will take a `launcher(TileEntity)` and provide the nullable `TileEntity` that might be launching the projectile.

src/main/java/org/spongepowered/common/event/tracking/phase/block/DispensePhaseState.java
```diff
@@ -58,8 +69,44 @@ public void unwind(final GeneralizedContext phaseContext) {
...
+                                .flatMap(LocatableSnapshot::getLocation)
+                                .flatMap(Location::getTileEntity)
+                                .orElse(null);
+                        if (tileEntity instanceof ProjectileSource) {
+                            frame.addContext(EventContextKeys.PROJECTILE_SOURCE, ((ProjectileSource) tileEntity));
+                        }
+
+                        int projectilesCount = projectiles.size();
```
gabizou, line 98:
```suggestion
                        final int projectilesCount = projectiles.size();
```

src/main/java/org/spongepowered/common/event/tracking/phase/packet/player/InteractionPacketState.java
```diff
@@ -84,6 +84,7 @@ public boolean isInteraction() {
 
     @Override
     public void populateContext(final EntityPlayerMP playerMP, final Packet<?> packet, final InteractionPacketContext context) {
+        ((TrackedInventoryBridge) playerMP.openContainer).bridge$setCaptureInventory(true);
```
gabizou, line 87: I don't feel comfortable setting the inventory to capture blindly like this, are we for certain that this doesn't negatively interact with other inventory related captures that might take place when interacting with other bits (like right clicking a water bucket to a cauldron)

src/main/java/org/spongepowered/common/event/tracking/phase/packet/player/PlaceBlockPacketState.java
```diff
@@ -119,19 +128,37 @@ public void unwind(final BasicPacketContext context) {
         final EntityPlayerMP player = context.getPacketPlayer();
         final ItemStack itemStack = context.getItemUsed();
         final SpongeItemStackSnapshot snapshot = context.getItemUsedSnapshot();
+
+        boolean entityCaptured = !context.getCapturedEntities().isEmpty();
```
gabizou, line 132:
```suggestion
        final boolean entityCaptured = !context.getCapturedEntities().isEmpty();
```

src/main/java/org/spongepowered/common/event/tracking/phase/packet/player/PlaceBlockPacketState.java
```diff
@@ -119,19 +128,37 @@ public void unwind(final BasicPacketContext context) {
...
         context.getCapturedEntitySupplier()
             .acceptAndClearIfNotEmpty(entities -> {
                 try (final CauseStackManager.StackFrame frame = Sponge.getCauseStackManager().pushCauseFrame()) {
-                    frame.addContext(EventContextKeys.SPAWN_TYPE, SpawnTypes.SPAWN_EGG);
-                    SpongeCommonEventFactory.callSpawnEntity(entities, context);
+                    List<Projectile> projectiles = new ArrayList<>();
+                    List<Entity> rest = new ArrayList<>();
```
gabizou, line 137:
```suggestion
                    final List<Projectile> projectiles = new ArrayList<>();
                    final List<Entity> rest = new ArrayList<>();
```

src/main/java/org/spongepowered/common/event/tracking/phase/tick/EntityTickPhaseState.java
```diff
@@ -148,9 +152,23 @@ protected void processCaptures(final Entity tickingEntity, final EntityTickConte
...
-                        SpongeCommonEventFactory.callSpawnEntity(projectile, phaseContext);
-                        frame.removeContext(EventContextKeys.SPAWN_TYPE);
+                        if (tickingEntity instanceof ProjectileSource) {
+                            frame.addContext(EventContextKeys.PROJECTILE_SOURCE, (ProjectileSource) tickingEntity);
+                        }
+                        final Cause cause = Sponge.getCauseStackManager().getCurrentCause();
+                        boolean launchEventCancelled = false;
+                        for (Entity proj : projectile) {
+                            if (SpongeImpl.postEvent(SpongeEventFactory.createLaunchProjectileEvent(cause, (Projectile) proj))) {
+                                launchEventCancelled = true;
```
gabizou, line 162: So if any of the launch events are cancelled, then how would you tell plugins that the events prior were cancelled?

src/main/java/org/spongepowered/common/mixin/core/block/BlockDispenserMixin.java
```diff
@@ -144,6 +145,7 @@
     )
     private void impl$InjectToStoreOriginalItem(
         final World worldIn, final BlockPos pos, final CallbackInfo ci, final BlockSourceImpl source, final TileEntityDispenser dispenser, final int slotIndex, final ItemStack dispensedItem, final IBehaviorDispenseItem behavior) {
+        ((DispenserBridge) dispenser).bridge$setDispensedItem(dispensedItem);
```
gabizou, line 148: reset anywhere?

src/main/java/org/spongepowered/common/mixin/core/tileentity/TileEntityDispenserMixin.java
```diff
@@ -24,17 +24,23 @@
...
-import org.spongepowered.common.item.inventory.lens.impl.minecraft.SingleGridLens;
 import org.spongepowered.common.item.inventory.lens.impl.collections.SlotCollection;
+import org.spongepowered.common.item.inventory.lens.impl.minecraft.SingleGridLens;
 
 @Mixin(TileEntityDispenser.class)
-public abstract class TileEntityDispenserMixin extends TileEntityLockableLootMixin {
+public abstract class TileEntityDispenserMixin extends TileEntityLockableLootMixin implements DispenserBridge {
+
+    private @Nullable ItemStack originalStack;
+    private int originalStackSize;
```
gabizou, line 43:
```suggestion
    private @Nullable ItemStack impl$originalStack;
    private int impl$originalStackSize;
```

gabizou: You might as well make a snapshot instead.

src/main/java/org/spongepowered/common/mixin/core/tileentity/TileEntityDispenserMixin.java
```diff
@@ -50,4 +56,16 @@
         return new SingleGridLens(0, 3, 3, (Class) TileEntityDispenser.class, slots);
     }
 
+    @Override
+    public void bridge$setDispensedItem(ItemStack dispensed) {
+        this.originalStack = dispensed;
```
gabizou, line 61: You do know this is 100% mutable, or are you intending on copying it outside of this?

src/main/java/org/spongepowered/common/mixin/core/tileentity/TileEntityDispenserMixin.java
```diff
@@ -50,4 +56,16 @@
...
+        this.originalStackSize = dispensed.getCount();
+    }
+
+    @Override
+    public void bridge$restoreDispensedItem(int count) {
+        if (originalStack != null) {
+            this.originalStack.setCount(this.originalStackSize + count);
```
gabizou, line 68: What is this restoring exactly? The field seems to exist solely to be stored and updated, but not actually read anywhere that I can see.

gabizou (changes requested):
Context leakage that should not exist outside a Stackframe, other comments are more clarification needed or some explanation fo what's going on (can even be some comments in code so people looking back at this will understand in the future ;) )

dualspiral:
It's also worth noting [LaunchProjectileEvent has actually been deprecated](https://github.com/SpongePowered/SpongeAPI/commit/79fb019bd6245e8415016dcfc71a75242f298f55) and a [barebones firing of the event after spawning in the entity](https://github.com/SpongePowered/SpongeCommon/commit/8e4ee8e93e2acfc4f6b8c2349a0242950b8cd4df) was added for projectiles - the rationale being that when launching a projectile you're spawning an entity in anyway and the `Cause` should tell you why.

RedNesto:
Yeah I noticed it. I'm not sure if this PR is worth it, what is essentially adds is cancellation handling and context now.

dualspiral:
I think we'll leave it as is as we have a barebones event in 7 and it's completely gone in 8. Regardless, thanks for looking into this.

RedNesto:
Agree, we'll need to make sure the context is correctly set for projectiles though (I'm not sure if it is properly filled right now).

dualspiral:
I'd be happy for a PR if that's the case!

Issue #2599, opened 2020-05-02
Sponge has problems with executing Minecraft Functions properly


CedeLovesKat:
**I am currently running**

- SpongeForge version: spongeforge-1.12.2-2838-7.2.1
- Forge version: 3238
- Java version: 8
- Operating System: x64 Windows 10

- Plugins/Mods:
[19:57:48] [Server thread/INFO] [minecraft/DedicatedServer]: Mods (24): Minecraft, Minecraft Coder Pack, Forge Mod Loader, Minecraft Forge, SpongeAPI, Sponge, Armourer's Workshop, Chisels & Bits, CreativeCore, CreativeCoreDummy, CustomNPCs, Dynmap, Fluids++, LegendaryPlus, LittleTiles, LittleTilesCore, MrCrayfish's Furniture Mod, Pixelmon, Pixelmon Trainer Commands, SpongeForge, Thut Core, Thut's Elevators, ThutCore Compat, Wondertrade
> sponge plugins
[19:57:55] [Server thread/INFO] [nucleus]: Server ran the command: /sponge plugins
[19:57:55] [Server thread/INFO] [minecraft/DedicatedServer]: Plugins (17): Minecraft, Minecraft Coder Pack, SpongeAPI, Sponge, CarrotShop, CommandSigns Reloaded, GriefPrevention, LuckPerms, Nucleus, PixelHunt, PixelMoney, PixelmonBroadcasts, PixelmonEconomyBridge, SimpleDexRewards, Total Economy, VoxelSniper, WorldEdit

**Issue Description**
I wrote some minecraft:functions for my "radio" system I implemented with comandblocks. They basically play music according to the carrot_on_a_stick in the left hand and the location the player is right now. However they dont work in sponge. I tested them on mohist, vanilla, forge and singleplayer and they work totally fine. The command even says that the corresponding function commands are executed but nothing happens at all. This needs to be fixed desperately.
The pastebin of my function: https://pastebin.com/4zNefxz7
As I mentioned, it works on my singeplayer/ map / Mohist / Vanilla and Forge but not on sponge...!

I hope I could explain it as good as possible! Im down for further questions : )

Thanks in advance!

CedeLovesKat:
Hey thanks for the answer. It makes me really happy that it seems like you guys can fix this problem. I was very sad when it diddnt work like it should on my server and Im looking forward to use this to the fullest potential. However, I have problems with getting something similiar to work as well. 

[...]

PR #2600, opened 2020-05-06
Ensure default permissions are used with some wrapped subjects

dualspiral:
See #2599

The reason I make this a PR at first is to:

a) make sure I'm not accidently causing a big security hole; and
b) talk about the notion of a default permission in the API.

As you can see in the diff, I have the following:

```
if (target instanceof SubjectBridge) {
    return ((SubjectBridge) target).bridge$permDefault(permission).asBoolean();
}
```

Would it be better to graduate `bridge$permDefault(permission)` to the API `Subject` interface (as `defaultPermission`) so that we don't need to check if the object is of type `SubjectBridge` and allow third-party `Subjects` (e.g. `ProxySource`s) to have their own defaults, and make `hasPermission(...)` check `defaultPermision(...)` if the value is undefined?

Such an API change shouldn't affect permission plugins.

(Note, I'm willing to pull this whenever, even if we decide to put defaults into the API, it's just a thought I had)

@clienthax please test this - it worked for me but I'd like to make sure it works for you too.

### DIFF
--- a/src/main/java/org/spongepowered/common/command/AndPermissionLevelSubject.java
+++ b/src/main/java/org/spongepowered/common/command/AndPermissionLevelSubject.java
@@ -27,13 +27,16 @@ package org.spongepowered.common.command;
 import com.google.common.collect.ImmutableSet;
 import net.minecraft.command.ICommandSender;
 import org.spongepowered.api.command.CommandSource;
+import org.spongepowered.api.command.source.ProxySource;
 import org.spongepowered.api.service.context.Context;
 import org.spongepowered.api.service.permission.MemorySubjectData;
 import org.spongepowered.api.service.permission.PermissionService;
+import org.spongepowered.api.service.permission.Subject;
 import org.spongepowered.api.service.permission.SubjectCollection;
 import org.spongepowered.api.service.permission.SubjectData;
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/command/AndPermissionLevelSubject.java
```diff
@@ -93,4 +96,24 @@ public SubjectData getSubjectData() {
...
+    @Override
+    public boolean hasPermission(Set<Context> contexts, String permission) {
+        final Tristate value = getPermissionValue(contexts, permission);
+        if (value == Tristate.UNDEFINED) {
+            Subject target = this.delegate;
+            while (target instanceof ProxySource) {
+                if (target instanceof SpongeProxySource) {
+                    target = ((SpongeProxySource) target).getSubjectDelegate();
+                } else {
+                    target = ((ProxySource) target).getOriginalSource();
```
dualspiral: Point of discussion: I feel like maybe I don't need to, or worse, _shouldn't_ check for this. A `ProxySource` might define its own `hasPermission` et. al. checks - we just know `SpongeProxySource` doesn't. Does it make sense to remove this?

PR #2603, opened 2020-05-10
Updates for Configurate 3.7

### LINKED ISSUE #1142: NullPointerException loading configuration

### LINKED ISSUE #1598: ParticleOptions of a ParticleEffect not being deserialized from a config file

zml2008:
Alright, this update is about ready for some actual review

These changes bring Sponge up to date with changes in configurate, modifies the Configurate DataTranslator in order to resolve many of its known issues, and extends test coverage a little bit. I've tested this PR + Configurate 3.7's snapshots on a local server with Nucleus and some of my plugins and things seem fine, but with the number of changes I'd like to have more extensive testing.

### Outstanding action items + questions
- [x] Figure out how to bump the configurate dependency version in common and not API
- [ ] Publish 3.7 snapshots to the Sponge maven repo
- [ ] Verify that #1973 is unrelated
- [ ] Do we want to handle a root node with a list value in any way? right now both configurate and json's translators require the root data to be a Map.

### Longer-term changes
- API 8: Centralize type serializer registration in preparation for immutable collections in Configurate 4.0
- Look at extracting sponge's hierarchical configuration into some part of Configurate for usage in plugins (I know at least @bloodmc's GriefDefender uses a similar system)
- Implement single server-wide `WatchServiceListener` and expose auto-reloading `ConfigurationReference`s for plugins.


ImMorpheus:
See https://github.com/SpongePowered/SpongeCommon/pull/2353/files#diff-c5975971207a43e7feaecf528c5b4ba8R96

You may want to include those test since they are taken from real issue which has been reported in the past

build.gradle
```diff
@@ -50,7 +50,6 @@ version = minecraftVersion + '-' + ext.apiSuffix + '.' + common.recommendedVersi
 
 ext.apiVersion = isSnapshot ? ext.fixedApiVersionWithDecreasedMinor : ext.apiReleaseVersion
 
-
```
gabizou: random line

gradle/minecraft.gradle
```diff
@@ -9,6 +9,18 @@ evaluationDependsOn api.path
...
 
+repositories {
+    mavenLocal()
+}
+
+configurations.all {
+    resolutionStrategy {
+        force 'org.spongepowered:configurate-hocon:3.7-SNAPSHOT'
+        force 'org.spongepowered:configurate-gson:3.7-SNAPSHOT'
+        force 'org.spongepowered:configurate-yaml:3.7-SNAPSHOT'
```
gabizou: I don't like this being in `Minecraft.gradle` because it's not really related to being in Minecraft.gradle, but I see why it's being applied. That being said, I'd just override the dependency in SpongeCommon explicitly and exclude the dependency inherited from spongeAPI.

src/main/java/org/spongepowered/common/config/DuplicateRemovalVisitor.java
```diff
@@ -0,0 +1,122 @@
...
+import ninja.leaping.configurate.ConfigurationVisitor;
+import ninja.leaping.configurate.Types;
+
+import java.util.Objects;
+
+/**
+ * Given a configuration node, and a node at the same position in the tree of a parent configuration, remove values from the child node that don't
+ * change anything from the parent node.
+ */
+class DuplicateRemovalVisitor implements ConfigurationVisitor.Safe<ConfigurationNode, Void> {
```
gabizou: So what's this for exactly?

src/main/java/org/spongepowered/common/data/SpongeDataManager.java
```diff
@@ -70,15 +68,9 @@
...
-    private static final TypeToken<CatalogType> catalogTypeToken = TypeToken.of(CatalogType.class);
-    private static final TypeToken<DataSerializable> dataSerializableTypeToken = TypeToken.of(DataSerializable.class);
-
     static {
-        TypeSerializers.getDefaultSerializers().registerPredicate(
-            // We have a separate type serializer for CatalogTypes, so we explicitly discount them here.
-            // See https://github.com/SpongePowered/SpongeCommon/issues/1348
-            x -> dataSerializableTypeToken.isSupertypeOf(x) && !catalogTypeToken.isSupertypeOf(x), new DataSerializableTypeSerializer()
-        );
+        // TODO: centralize registration
```
gabizou, line 72: Is this something that can be done in sponge, or is it something that needs Configurate 4.x?

src/main/java/org/spongepowered/common/data/persistence/ConfigurateTranslator.java
```diff
@@ -24,118 +24,255 @@
...
-
-    private static final ConfigurateTranslator instance = new ConfigurateTranslator();
+    private static final ConfigurateTranslator INSTANCE = new ConfigurateTranslator();
     private static final TypeToken<ConfigurationNode> TOKEN = TypeToken.of(ConfigurationNode.class);
-
-    private ConfigurateTranslator() {
-    }
+    private static final ConfigurationOptions DEFAULT_OPTS = ConfigurationOptions.defaults()
+            .withNativeTypes(ImmutableSet.of(Map.class, List.class, Double.class,
+                    Long.class, Integer.class, Boolean.class, String.class));
```
gabizou:
```suggestion
    private static final ConfigurationOptions DEFAULT_OPTS = ConfigurationOptions.defaults()
            .withNativeTypes(ImmutableSet.of(Map.class, 
                    List.class, 
                    Double.class,
                    Long.class, 
                    Integer.class, 
                    Boolean.class, 
                    String.class
                    )
            );
```

gabizou (changes requested):

gradle/minecraft.gradle
zml2008: This is an old attempt at trying to make the dependency version override -- but nothing I've tried has worked unfortunately. I'll poke at it some more -- I'm sure it's just a matter of fighting gradle + fg until they behave

src/main/java/org/spongepowered/common/config/DuplicateRemovalVisitor.java
zml2008: This is the logic that removes values from, say a `dimension.conf` that are the same as values in `global.conf`, but migrated from `ConfigurationNodeWalker` to `ConfigurationVisitor`

src/main/java/org/spongepowered/common/data/SpongeDataManager.java
zml2008: Most can be done with sponge, but moving from adding to the global collection at various points around Common to adding to a Sponge-specific collection (Following a [Kyori text-like pattern](https://github.com/KyoriPowered/text/blob/master/serializer-gson/src/main/java/net/kyori/text/serializer/gson/GsonComponentSerializer.java#L58)) in one specific spot would be an api-breaking change.

What will change in 4.0 is that modifying existing TypeSerializerCollections will be forbidden when it is currently just discouraged.

Registration could be moved to be in one spot fairly easily, but there isn't much benefit to that when serializers are still being added to the global collection.

gradle/minecraft.gradle
gabizou: I'll try to give a better solution, just might need to wait till I can spend some time on it.

src/main/java/org/spongepowered/common/config/DuplicateRemovalVisitor.java
gabizou: Just might want to include that as a more specific case of how it's being used in the class java doc.

gradle/minecraft.gradle
dualspiral: I might be missing something totally obvious here, but why aren't you just bumping the dep in the API layer? Given I'd like 7.3 to include Configurate 3.7, if this ends up merged in I want it available in our API 7.3 snapshots anyway.

zml2008: I thought part of the plan was to have 3.7 available for 7.2 builds as well? but if there aren't more of those, then sure, we can just have the dep in-api and i'll strip out all the buildscript changes

dualspiral: The future 7.2.x builds will include the 7.3 snapshot, it's just we guarantee the 7.2 portion of the API. So yeah, just stick the dep in the API.

gabizou:
@zml2008 Seeing as configurate is now 3.7 on [maven central](https://repo1.maven.org/maven2/org/spongepowered/), anything left to do?

dualspiral:
I've given this a whirl and it all seems okay to me. I've bumped Configurate in the API too. If we come across any issues, we can always patch them. Best to get this to the people now though!

dualspiral:
e06bc948637a5b7108f442ffde9ca8694d6d9d07

PR #2605, opened 2020-05-10
Init collision state lazily

ImMorpheus:
Don't init collision state during entity construction but do it lazily during entity collisions.

`maxEntityCramming` is checked first since `listToFill.size() < capability.collision$getMaxCollisions()` is enough to make it happen (regardless of the collision state).


src/main/java/org/spongepowered/common/mixin/entitycollisions/EntityMixin_Collisions.java
```diff
@@ -77,7 +77,7 @@
 
             this.collision$entityModId = ((SpongeEntityType) ((Entity) this).getType()).getModId();
             if (!this.world.isRemote) {
-                collision$initializeCollisionState(this.world);
+                collision$requiresCollisionsCacheRefresh(true);
```
gabizou, line 80:
```suggestion
                this.collision$requiresCollisionsCacheRefresh(true);
```

PR #2567, opened 2020-04-14
Fix missing ShouldFires

ImMorpheus:
[SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/3131) | **SpongeCommon**

The PhaseTracker was the tricky part, some phases were using ShouldFire correctly while others were not using the flag at all.

PR #2607, opened 2020-05-11
Fix a few issues with shouldfires and use specific flags

ImMorpheus:
Splitting bugs and enhancements from actually "[missing shouldfires](https://github.com/SpongePowered/SpongeCommon/pull/2567)" 

Namely:
https://github.com/SpongePowered/SpongeCommon/blob/a4f4c117ad89266a57762eaf27cd01d30251a7bb/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java#L960-L966
Listening to `RotateEntityEvent` without listening to `MoveEntityEvent` would fire an event even when `!oldPositionVector.equals(currentPositionVector)`.
This is breaking the API contract (a `RotateEntityEvent` is only fired when the position is unchanged).

`(!ShouldFire.MOVE_ENTITY_EVENT && !ShouldFire.MOVE_ENTITY_EVENT_POSITION && !ShouldFire.ROTATE_ENTITY_EVENT)` is always false since that's handled before calling the factory method.

`ShouldFire.RIDE_ENTITY_EVENT_MOUNT || ShouldFire.RIDE_ENTITY_EVENT` is not good (this also applies to move events). Only the flag for the specific event should be used (when listening to a generic event the flag for the specific event is updated). 

Additionally a `shouldFire()` method is added to block change to enable targeted flag check instead of just `ShouldFire.CHANGE_BLOCK_EVENT`

### DIFF
--- a/src/main/java/org/spongepowered/common/event/ShouldFire.java
+++ b/src/main/java/org/spongepowered/common/event/ShouldFire.java
@@ -59,6 +59,7 @@ public class ShouldFire {
 
     public static boolean CHANGE_BLOCK_EVENT = false;
     public static boolean CHANGE_BLOCK_EVENT_PRE = false;
+    public static boolean CHANGE_BLOCK_EVENT_DECAY = false;
     public static boolean CHANGE_BLOCK_EVENT_MODIFY = false;
     public static boolean CHANGE_BLOCK_EVENT_BREAK = false;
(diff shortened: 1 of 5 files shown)


dualspiral (approved):
I see nothing contriversial here - well reasoned changes.

PR #2610, opened 2020-05-13
Init entity activation state lazily

ImMorpheus:
Same as https://github.com/SpongePowered/SpongeCommon/pull/2605 but for entity activation


src/main/java/org/spongepowered/common/mixin/entityactivation/entity/EntityMixin_Activation.java
```diff
@@ -54,7 +54,7 @@
     @Inject(method = "<init>", at = @At("RETURN"))
     private void activation$InitActivationRanges(@Nullable final World world, final CallbackInfo ci) {
         if (world != null && !((WorldBridge) world).bridge$isFake() && ((WorldInfoBridge) world.getWorldInfo()).bridge$isValid()) {
-            EntityActivationRange.initializeEntityActivationState((net.minecraft.entity.Entity) (Object) this);
+            activation$requiresActivationCacheRefresh(true);
```
gabizou:
```suggestion
            this.activation$requiresActivationCacheRefresh(true);
```

gabizou (changes requested):
Fix the `this.` and it is good

PR #2611, opened 2020-05-17
Reduce iterations in TileEntities activation

### LINKED ISSUE #2780: TileEntityActivation Performance
**I am currently running**

- SpongeForge version: 3720
- Forge version: 2836
- Java version: 1.8_201
- Operating System: Windows Server 2016

- Plugins/Mods: FTBRevelations 3.0.1 + https://pastebin.com/raw/QVBKZm3A

**Issue Description**

Performance impact from tile activation when controlling tick rates of entities.

I am using tile activation to control tick rates. I have the file populated and many of the mods disabled by default but still experiencing a hit from tile activation. 

Spark: https://sparkprofiler.github.io/#ThhUaANL8X
Sponge global.conf: https://pastebin.com/raw/Cxd4Nf5P

I can also provide a nps report of the same behavior. I can also provide a test build with the server configuration, this map is less than 5 days old with 30 players average, not many tile entities present.

Thanks.

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeForge/issues/2780

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/plugin/tileentityactivation/TileEntityActivation.java
+++ b/src/main/java/org/spongepowered/common/mixin/plugin/tileentityactivation/TileEntityActivation.java
@@ -175,10 +175,11 @@ public class TileEntityActivation {
                     spongeTileEntity.activation$requiresActivationCacheRefresh(false);
                 }
 
-                final int bbActivationRange = ((ActivationCapability) tileEntity).activation$getActivationRange();
+                final int activationRange = ((ActivationCapability) tileEntity).activation$getActivationRange();
+                final int bbActivationRange = (int) Math.pow(activationRange, 2);
                 for (EntityPlayerMP player : players) {
                     final Vector3i playerPos = VecHelper.toVector3i(player.getPosition());
-                    final int blockDistance = Math.round(tilePos.distance(playerPos));
+                    final int blockDistance = tilePos.distanceSquared(playerPos);
                     if (blockDistance <= bbActivationRange) {
                         ((ActivationCapability) tileEntity).activation$setActivatedTick(currentTick);
                         break;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/plugin/tileentityactivation/TileEntityActivation.java
```diff
@@ -177,9 +176,13 @@ private static void activateChunkTileEntities(final EntityPlayer player, final C
                 }
 
                 final int bbActivationRange = ((ActivationCapability) tileEntity).activation$getActivationRange();
-                final int blockDistance = Math.round(tilePos.distance(playerPos));
-                if (blockDistance <= bbActivationRange) {
-                    ((ActivationCapability) tileEntity).activation$setActivatedTick(currentTick);
+                for (EntityPlayerMP player : players) {
+                    final Vector3i playerPos = VecHelper.toVector3i(player.getPosition());
+                    final int blockDistance = Math.round(tilePos.distance(playerPos));
```
dualspiral: While we're looking at performance, I wonder if we should also look at making this dependent on distance squared. I realise that it makes some difference to getting `bbActivationRange` and `Math.round()` - but the activation range would then be `Math.pow(bbActivationRange - 0.5, 2)` due to the current rounding, and such rounding would not be required.

dualspiral (approved):
It does what it says on the tin, but there is a question here as to whether we could do something else and avoid square root calls.

We could pull as is if we think it's not worth doing that right now, or if you want to split that off into another PR.

src/main/java/org/spongepowered/common/mixin/plugin/tileentityactivation/TileEntityActivation.java
ImMorpheus: Good point.
I've switched to `Vector3i#distanceSquared` (which returns an int, so there's no need for rounding).

gabizou (approved):
*eventually* can have improved code, but this is improvement already.

PR #2620, opened 2020-05-22
API8 Fixing Data Impl

Faithcaio:
for reference
https://github.com/SpongePowered/SpongeAPI/pull/1996
https://gist.github.com/Faithcaio/4e0f85887c0921a96160603a6b8c555f

PR #2622, opened 2020-05-24
fix API8 datafixers

Faithcaio:
after #2620


src/main/java/org/spongepowered/common/data/fixer/entity/EntityTrackedUser.java
```diff
@@ -0,0 +1,69 @@
...
+
+        return TypeRewriteRule.seq(this.fixTracked(forgeDataFinder, spongeDataFinder, spongeDataType, Constants.Sponge.SPONGE_ENTITY_CREATOR),
+                                   this.fixTracked(forgeDataFinder, spongeDataFinder, spongeDataType, Constants.Sponge.SPONGE_ENTITY_NOTIFIER));
+    }
+
+    private TypeRewriteRule fixTracked(OpticFinder<?> forgeDataFinder, OpticFinder<?> spongeDataFinder, Type<?> spongeDataType, String name) {
+        final Type<?> trackedType = spongeDataType.findFieldType(name);
+        final OpticFinder<?> trackedFinder = DSL.fieldFinder(name, trackedType);
+
+        return this.fixTypeEverywhereTyped("Entity" + name + "UserFix", this.getInputSchema().getType(TypeReferences.ENTITY),
```
i509VCB, line 61: This is where things go down hill...

So `TypeReferences.ENTITY` does not exist on any schemas of Sponge's datafixer.

This means this will fail to fix, hard fail. The solution to this is the include minecraft's own schema into your datafixer, at a lower version than sponge's current schema version so the type reference exists.

Even then, there is a whole extra can of beans with the entity type reference, but that is closer to how modded entities are handled during registration.

invalid/main/java/org/spongepowered/common/data/fixer/entity/player/PlayerRespawnData.java
```diff
@@ -35,12 +38,13 @@
 import java.util.Map;
 import java.util.UUID;
 
-public class PlayerRespawnData implements IFixableData {
+// TODO this is actually doing nothing?
+public class PlayerRespawnData extends DataFix {
```
i509VCB, line 43: You'd probably want to call this `PlayerRespawnDataFix`

i509VCB (commented):
This is definitely an odd section of the codebase, if you need some help with DFU, just give me a ping.

src/main/java/org/spongepowered/common/data/fixer/entity/EntityTrackedUser.java
i509VCB: However that itself can be a mess, and I need to sleep. So I can clarify tomorrow.

gabizou:
Merged into api-8.

PR #2626, opened 2020-05-27
Log chunk load queue

ImMorpheus:
[SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/3165) | **SpongeCommon**

### DIFF
--- a/src/main/java/org/spongepowered/common/util/SpongeHooks.java
+++ b/src/main/java/org/spongepowered/common/util/SpongeHooks.java
@@ -194,6 +194,19 @@ public class SpongeHooks {
         }
     }
 
+    public static void logChunkQueueLoad(final World world, final Vector3i chunkPos) {
+        if (world.isRemote) {
+            return;
+        }
+
+        final SpongeConfig<WorldConfig> configAdapter = ((WorldInfoBridge) world.getWorldInfo()).bridge$getConfigAdapter();
+        if (configAdapter.getConfig().getLogging().chunkLoadLogging()) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/util/SpongeHooks.java
```diff
@@ -194,6 +194,19 @@ public static void logBlockAction(final World world, @Nullable final BlockChange
         }
     }
 
+    public static void logChunkQueueLoad(final World world, final Vector3i chunkPos) {
+        if (world.isRemote) {
```
gabizou, line 198: `((WorldBridge) world).bridge$isFake()`?

gabizou (approved):
LGTM

src/main/java/org/spongepowered/common/util/SpongeHooks.java
ImMorpheus: I'm using the same condition used for chunkload logging. I'd need to modify that as well

https://github.com/SpongePowered/SpongeCommon/blob/a865595b36ab95a876c8ebd46b464b23a73b09aa/src/main/java/org/spongepowered/common/util/SpongeHooks.java#L197-L199

That would disable chunkload logging for fake worlds. Is it intended ?

gabizou: Some fake worlds don't even use the normal chunk loader, some are designed around a lot of different reasons, but I'm not 100% sure whether it's worth to actually bother sanity checking even more on this case since all this is doing is logging.

gabizou (approved):

PR #2636, opened 2020-06-10
Inline micro optimize some bits in MultiBlockCaptureSupplier

gabizou:
Reviewing some old bit of code.

No real statistics on the micro optimization here, but can't hurt.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/context/MultiBlockCaptureSupplier.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/context/MultiBlockCaptureSupplier.java
@@ -74,7 +74,7 @@ import javax.annotation.Nullable;
 @SuppressWarnings("rawtypes")
 public final class MultiBlockCaptureSupplier implements ICaptureSupplier {
 
-    public static final boolean PRINT_TRANSACTIONS = Boolean.valueOf(System.getProperty("sponge.debugBlockTransactions", "false"));
+    public static final boolean PRINT_TRANSACTIONS = Boolean.parseBoolean(System.getProperty("sponge.debugBlockTransactions", "false"));
 
     @Nullable private LinkedListMultimap<BlockPos, SpongeBlockSnapshot> multimap;
     @Nullable private ListMultimap<BlockPos, BlockEventData> scheduledEvents;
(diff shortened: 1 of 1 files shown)


dualspiral:
It makes sense to take this class in full, rather than just the changed bits, to get a sense of what's going on here. It also is important to note that these objects are potentially pooled, in order to reduce object creation. However, there seems to be a fair bit of object creation within the supplier itself and a lot of null checking.

I was interested to have a look at `null` vs `empty` and realised that I couldn't quickly write a detailed benchmark that accurately portrays this. I did write something like this:

https://gist.github.com/dualspiral/af51df8f9535c0b2c8dc1ee9a7e48b64

and found that persisting the list and clearing it, rather than null checking, was much faster over 100k runs - four runs from a warmish start in my IDE on my laptop (i5-8279U processor, 2.4GHz, macOS, openjdk version "1.8.0_242" (AdoptOpenJDK)) within IntelliJ - this is the per "run" statistic.

```
null: 229ns
empty: 133ns

null: 210ns
empty: 125ns

null: 219ns
empty: 114ns

null: 223ns
empty: 119ns
```

I will say again that I don't think this is the most representative thing I could have done, but it was good enough for me to get an idea of the cost of null checking each time we run through a method vs being able to assume an object is there. With many potential calls a tick, we might not want to ignore this.

The reason I bring this up is the nullable collections. My suspicion is that they are nullable from before pooling was a thing - that makes sense, create only when necessary. However, I think the following sort of thing is probably unnecessary if a `null` list is effectively an empty list and we can keep that list around between uses of the supplier:

https://github.com/SpongePowered/SpongeCommon/blob/5310c0312182d9e6fcf23d30deb5fa502e049730/src/main/java/org/spongepowered/common/event/tracking/context/MultiBlockCaptureSupplier.java#L176-L178

If we can ditch an if check and object creation with clearing (rather than nulling) the list in `reset`, then we've got a quick gain by removing a handful of operations. [This operation can then remain without the nulling operation](https://github.com/SpongePowered/SpongeCommon/pull/2636/files#diff-e86193e9bdc968b7b08a56956f090d2bL844) and we should see a slight gain.

I think it's probably the same for `multimap` and the others. I did a quick check to see whether a `null` check or an `isEmpty` check was faster, `isEmpty` seemed to edge it but that's far from conclusive. The cost of `clear` is also likely higher than nulling the variable, so I'd be less worried about that because we're performing other logic based on that if statement.

So, I think we should not be using null for collections where we're continuously running the code `if null then create` on them, unless there is a significant reason why `null` is different from `empty` in our scenarios, and even then, I'd probably suggest having a separate boolean instead (but I'm not going to benchmark that right now).

dualspiral (approved):
While I've said what I had, and it was an interesting thought - I'm not going to insist on it. It's going to be a very minor thing and, truth be told, it'll gain very little, if nothing at all, in the long run. Who knows what the JVM will do for us here?

So - if you want to merge this in as is, I say go for it.

src/main/java/org/spongepowered/common/event/tracking/context/MultiBlockCaptureSupplier.java
```diff
@@ -750,25 +750,8 @@ public boolean processTransactions(final List<Transaction<BlockSnapshot>> transa
...
-            this.processingWorlds.clear();
-            for (BlockTransaction transaction = this.head; transaction != null; ) {
-                final BlockTransaction next = transaction.next;
-                transaction.previous = null;
-                transaction.next = null;
-                transaction = next;
-            }
-            this.head = null;
-            this.tail = null;
+            clearProxies();
```
Zidane, line 753: I like this things.

dualspiral:
@gabizou Shall we just go for this as is (Zidane liking this things aside)?

gabizou:
Yeah, sure

PR #2640, opened 2020-06-14
New Explosion Event for BlockExplosionResistance

### LINKED ISSUE #2093: Event to alter block blast resistance during explosion calculations
This is a feature that I intend to make use of and would like to open a PR in the future.
However, I am not sure what would be the best way to implement this feature.

Pulverizer:
[ [SpongeAPI ](https://github.com/SpongePowered/SpongeAPI/pull/2130)| **SpongeCommon** ]

Resolves https://github.com/SpongePowered/SpongeAPI/issues/2093

The intention of this PR is to allow plugins to intercept the fetching of a block's explosion resistance and adjust the value dynamically based on the block's location and type.

Example Usage:
```
    public void changeBlockResistance(ExplosionEvent.FetchBlockExplosionResistance event) {

// Check if plugin has a hard number in config for BlockType
        if (Settings.DurabilityOverride != null && Settings.DurabilityOverride.containsKey(event.getBlockState().getType())) {
            event.setResistance(Settings.DurabilityOverride.get(event.getBlockState().getType()));
        }

// Check if location is within plugin managed structure
        float newResistance = event.getResistance();
        for (Craft craft : CraftManager.getInstance().getCraftsFromLocation(event.getLocation())) {
            float testNum = event.getResistance() * (1F + ((float) craft.getSize() / 50000F));

            if (testNum > newResistance) {
                newResistance = testNum;
            }
        }

// Set new block explosion resistance
        event.setResistance(newResistance);
    }
```

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/api/mcp/world/ExplosionMixin_API.java
+++ b/src/main/java/org/spongepowered/common/mixin/api/mcp/world/ExplosionMixin_API.java
@@ -30,15 +30,14 @@ import org.spongepowered.api.entity.explosive.Explosive;
 import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
 import org.spongepowered.api.world.explosion.Explosion;
+import org.spongepowered.api.world.explosion.ResistanceCalculator;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.common.bridge.world.ExplosionBridge;
-import org.spongepowered.api.world.explosion.ResistanceCalculator;
-
-import java.util.Optional;
 
 import javax.annotation.Nullable;
+import java.util.Optional;
 
 @Mixin(net.minecraft.world.Explosion.class)
 public abstract class ExplosionMixin_API implements Explosion {
@@ -98,17 +97,17 @@ public abstract class ExplosionMixin_API implements Explosion {
 
     @Override
     public int getResolution() {
-        return  ((ExplosionBridge) this).bridge$getResolution();
+        return ((ExplosionBridge) this).bridge$getResolution();
     }
 
     @Override
     public float getRandomness() {
-        return  ((ExplosionBridge) this).bridge$getRandomness();
+        return ((ExplosionBridge) this).bridge$getRandomness();
     }
 
     @Override
     public double getKnockback() {
-        return  ((ExplosionBridge) this).bridge$getKnockback();
+        return ((ExplosionBridge) this).bridge$getKnockback();
     }
 
     @Override
--- a/src/main/java/org/spongepowered/common/mixin/api/mcp/world/ExplosionMixin_API.java
+++ b/src/main/java/org/spongepowered/common/mixin/api/mcp/world/ExplosionMixin_API.java
@@ -36,8 +36,9 @@ import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.common.bridge.world.ExplosionBridge;
 
-import javax.annotation.Nullable;
 import java.util.Optional;
+
+import javax.annotation.Nullable;
 
 @Mixin(net.minecraft.world.Explosion.class)
 public abstract class ExplosionMixin_API implements Explosion {
--- a/src/main/java/org/spongepowered/common/mixin/api/mcp/world/ExplosionMixin_API.java
+++ b/src/main/java/org/spongepowered/common/mixin/api/mcp/world/ExplosionMixin_API.java
@@ -113,7 +113,7 @@ public abstract class ExplosionMixin_API implements Explosion {
 
     @Override
     public Optional<ResistanceCalculator> getResistanceCalculator() {
-        return Optional.of(((ExplosionBridge) this).bridge$getResistanceCalculator());
+        return Optional.ofNullable(((ExplosionBridge) this).bridge$getResistanceCalculator());
     }
 
 }
--- a/src/main/java/org/spongepowered/common/mixin/api/mcp/world/ExplosionMixin_API.java
+++ b/src/main/java/org/spongepowered/common/mixin/api/mcp/world/ExplosionMixin_API.java
@@ -34,6 +34,7 @@ import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.common.bridge.world.ExplosionBridge;
+import org.spongepowered.api.world.explosion.ResistanceCalculator;
 
 import java.util.Optional;
 
@@ -108,6 +109,11 @@ public abstract class ExplosionMixin_API implements Explosion {
     @Override
     public double getKnockback() {
         return  ((ExplosionBridge) this).bridge$getKnockback();
+    }
+
+    @Override
+    public Optional<ResistanceCalculator> getResistanceCalculator() {
+        return Optional.of(((ExplosionBridge) this).bridge$getResistanceCalculator());
     }
 
 }
--- a/src/main/java/org/spongepowered/common/mixin/core/world/ExplosionMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/ExplosionMixin.java
@@ -103,8 +103,8 @@ public abstract class ExplosionMixin implements ExplosionBridge {
 
     @Inject(method = "<init>*", at = @At("RETURN"))
     private void onConstructed(final net.minecraft.world.World world, final Entity entity, final double originX, final double originY,
-                               final double originZ, final float radius, final boolean isFlaming, final boolean isSmoking,
-                               final CallbackInfo ci) {
+            final double originZ, final float radius, final boolean isFlaming, final boolean isSmoking,
+            final CallbackInfo ci) {
         // In Vanilla and Forge, 'damagesTerrain' controls both smoke particles and block damage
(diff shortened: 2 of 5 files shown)


src/main/java/org/spongepowered/common/mixin/api/mcp/world/ExplosionMixin_API.java
```diff
@@ -110,4 +111,9 @@ public double getKnockback() {
         return  ((ExplosionBridge) this).bridge$getKnockback();
     }
 
+    @Override
+    public Optional<ResistanceCalculator> getResistanceCalculator() {
```
dualspiral, line 115: Given that there is always a `ResistanceCalculator`, this shouldn't be optional.

src/main/java/org/spongepowered/common/mixin/core/world/ExplosionMixin.java
```diff
@@ -146,10 +149,11 @@ public void doExplosionA() {
                                 final IBlockState iblockstate = this.world.getBlockState(blockpos);
 
                                 if (iblockstate.getMaterial() != Material.AIR) {
-                                    final float f2 = this.exploder != null
-                                               ? this.exploder.getExplosionResistance((net.minecraft.world.Explosion) (Object) this
-                                            , this.world, blockpos, iblockstate)
-                                               : iblockstate.getBlock().getExplosionResistance((Entity) null);
+                                    final float f2 = impl$resistanceCalculator.calculateResistance(
```
dualspiral: `this.impl$resistanceCalculator.`...

Also, because this is an `Overwrite`, you MUST add `// Sponge start - (reason)` and `// Sponge end` comments where you make a change. It is crucial we know what the diffs are. You should also add to the javadoc on this method to explain what change you've made and why.

src/main/java/org/spongepowered/common/world/SpongeExplosionBuilder.java
```diff
@@ -28,15 +28,29 @@
...
 import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
 import org.spongepowered.api.world.explosion.Explosion;
 import org.spongepowered.common.bridge.world.ExplosionBridge;
+import org.spongepowered.common.util.VecHelper;
+import org.spongepowered.api.world.explosion.ResistanceCalculator;
 
 public class SpongeExplosionBuilder implements Explosion.Builder {
 
+    private static final ResistanceCalculator DEFAULT_RESISTANCE_CALCULATOR =  ((blockState, blockPosition, explosion) ->
```
dualspiral: More a remark: I'm not sure where this would go, or how I'd do this, but I feel like there should be a way to get the default calculator from the API...

dualspiral (changes requested):

src/main/java/org/spongepowered/common/mixin/api/mcp/world/ExplosionMixin_API.java
Pulverizer: I believe there are other projects that use the SpongeAPI as a base or so I was told with my last PR.

src/main/java/org/spongepowered/common/mixin/core/world/ExplosionMixin.java
Pulverizer: oops, will get right on that.

src/main/java/org/spongepowered/common/world/SpongeExplosionBuilder.java
Pulverizer: Good point, is there anyone you can think to ping for advice? I'm also not comfortable with how I implemented this.

dualspiral: Based on what I wrote in the API, I'd inline this again and then allow any calculator that exists to alter the value afterwards. Then the default doesn't need to be exposed and plugins can inspect what's going on before making a decision if necessary.

Pulverizer: If I inline it gets duplicated over two or three locations. That's why I created a static and just reference it inline.

src/main/java/org/spongepowered/common/mixin/core/world/ExplosionMixin.java
Pulverizer: Think I've corrected this problem?

src/main/java/org/spongepowered/common/world/SpongeExplosionBuilder.java
dualspiral: No, it wouldn't. 

> then allow any calculator that exists to alter the value afterwards

You'd ditch the default calculator, revert https://github.com/SpongePowered/SpongeCommon/pull/2640/files#diff-8b417570b299753ff70d99886c28d1a6L149-R159, allow the calculator to be nullable and do something like this after the initial calculation:

```
if (this.impl$resistance != null) {
    f2 = Math.min(0, Math.max(1, this.impl$resistance(f2, ...))); // (I get f2 is final at the moment, but this is the gist of what I was getting at)
}
```

Then a `null` resistance means default explosion resistance and it doesn't need to be exposed.

src/main/java/org/spongepowered/common/mixin/core/world/ExplosionMixin.java
```diff
@@ -67,24 +69,19 @@
...
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Random;
-import java.util.Set;
-
 import javax.annotation.Nullable;
+import java.util.*;
```
dualspiral: No star imports please.

src/main/java/org/spongepowered/common/mixin/core/world/ExplosionMixin.java
```diff
@@ -146,10 +149,14 @@ public void doExplosionA() {
                                 final IBlockState iblockstate = this.world.getBlockState(blockpos);
 
                                 if (iblockstate.getMaterial() != Material.AIR) {
-                                    final float f2 = this.exploder != null
-                                               ? this.exploder.getExplosionResistance((net.minecraft.world.Explosion) (Object) this
-                                            , this.world, blockpos, iblockstate)
-                                               : iblockstate.getBlock().getExplosionResistance((Entity) null);
+                                    // Sponge Start - Allows the insertion of custom block resistance calculations
+                                    final float f2 = impl$resistanceCalculator
```
dualspiral: Still missing `this.` on the call...

src/main/java/org/spongepowered/common/world/SpongeExplosionBuilder.java
Pulverizer:
```java
f2 = this.impl$resistanceCalculator.isPresent() ? this.impl$resistanceCalculator.get().calculateResistance( ... ) : <default behaviour>
```

src/main/java/org/spongepowered/common/mixin/core/world/ExplosionMixin.java
Pulverizer: No clue how that happened.

src/main/java/org/spongepowered/common/world/SpongeExplosionBuilder.java
dualspiral: No. Can you please read what I said in the API issue (and above), reproduced below.

>    What about if a plugin wants to affect obsidian blocks only? How would they determine the resistance for other blocks?
>
>    Earlier, I said:
>
>  >    a ResistanceCalculator which takes a Location, the current resistance as a float, and returns the new resistance as a float.
>
>    That's the reason why I suggested it.

I'm not saying replace default behaviour. I'm saying do the default, then allow a plugin to modify it because a plugin might only want to affect a specific block and keep everything else as the default. Or they might want to multiply the resistance by some value.

Please follow what I've asked here, because I'm trying to keep as much flexibility for plugins as possible.

Pulverizer: So many individual parts over two threads got a little hard to keep track. Once I've figured out the solution I'll make a commit and we can try again.

It was intended that plugins could get the default if required but I can see now that I messed up somewhere while coding and I've got to spend a bit of time figuring out a solution.

dualspiral: ...please just do as I asked, calculate the default as it does now, and just add a step where it is passable to the calculator.

Pulverizer: But what if a plugin doesn't want to chain and instead ignore previous Calculators and use default resistance as it's base, maybe I've just gone down the rabbit hole and got lost?

dualspiral: <del>Just ignore the parameter? There's only going to be one calculator registered...</del>

Okay, I misread what you said, but we already said that there's going to be one calculator as far as the API is concerned. At the end of the day, some conflicts could occur but we can't catch them all.

It's not worth worrying about.

Pulverizer: We did? I think I have a solution to it so that chaining can be done while still offering default values where required.

src/main/java/org/spongepowered/common/world/SpongeExplosionBuilder.java
```diff
@@ -47,6 +50,7 @@
     private int resolution = 16;
     private float randomness = 1.0F;
     private double knockback = 1;
+    private Optional<ResistanceCalculator> resistanceCalculator = Optional.empty();
```
dualspiral: Never _store_ an `Optional`, this should be a `@Nullable` field.

Note, you were actually leading up to an NPE if someone `reset()` then `build()` the builder.

PR #2643, opened 2020-06-18
Fix ChunkRegenerateFlagRegistryModules using an excessively large map

ImMorpheus:
The initial capacity of `70` was probably copied from `BlockChangeFlagRegistryModule`


Zidane:
rofl.

PR #2646, opened 2020-06-21
Fix duplicate empty chunk

ImMorpheus:
There's an empty chunk in the `World` class and one in `ChunkProviderServer`. 
This PR keeps a single instance around for each world.

### DIFF
--- a/src/main/java/org/spongepowered/common/bridge/world/WorldBridge.java
+++ b/src/main/java/org/spongepowered/common/bridge/world/WorldBridge.java
@@ -25,6 +25,7 @@
 package org.spongepowered.common.bridge.world;
 
 import net.minecraft.world.World;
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/bridge/world/WorldBridge.java
```diff
@@ -46,4 +47,6 @@
     void bridge$clearFakeCheck();
 
     boolean bridge$isAreaLoaded(int xStart, int yStart, int zStart, int xEnd, int yEnd, int zEnd, boolean allowEmpty);
+
+    SpongeEmptyChunk getEmptyChunk();
```
dualspiral, line 51: `bridge$getEmptyChunk()`

PR #2650, opened 2020-06-23
Fix ShouldFire optimization not working for LoadChunkEvent

ImMorpheus:
`SpongeImpl` is registering an internal listener for `LoadChunkEvent`. 
This is killing the related `ShouldFire` flag which will always be true, no matter how many plugin listeners are registered.

This PR moves the content of the listener (incrementing the chunk count for the chunk gc option) right after the hook we use to fire LoadChunkEvent.


Zidane:
@ImMorpheus Even better, we should not be listening to our own events we are firing....................

PR #2653, opened 2020-06-27
PoC for service registration

dualspiral:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2133) | **SpongeCommon**

@Zidane Also for discussion. I'm sorry for the large type definitions, it's just a proof of concept for now.

### DIFF
--- a/src/main/java/org/spongepowered/common/config/category/ServicesCategory.java
+++ b/src/main/java/org/spongepowered/common/config/category/ServicesCategory.java
@@ -1,0 +1,96 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.config.category;
+
+import ninja.leaping.configurate.objectmapping.Setting;
+import ninja.leaping.configurate.objectmapping.serialize.ConfigSerializable;
+
+@ConfigSerializable
+public final class ServicesCategory extends ConfigCategory {
+
+    @Setting(value = "service-plugin",
+            comment = "Services specified here can be implemented by plugins. To ensure that a"
+                    + "specific plugin implements a given service, set the relevant option to its"
+                    + "plugin ID. If you wish to use Sponge's default for a given service, use"
+                    + "'sponge' as the ID.\n\n"
+                    + "If the plugin ID is unknown, or the option is set to '?', all plugins will"
+                    + "be given the opportunity to register their service. If multiple plugins"
+                    + "attempt to register, one will be picked in an implementation dependent way."
+                    + "If no plugins attempt to register a service, the Sponge default will be used"
+                    + "if one exists.\n\n"
+                    + "No Sponge default service exists for the Economy service.")
+    private ServicePluginSubCategory servicePlugin = new ServicePluginSubCategory();
+
+    public ServicePluginSubCategory getServicePlugin() {
+        return this.servicePlugin;
+    }
+
+    @ConfigSerializable
+    public static final class ServicePluginSubCategory {
+
+        @Setting(value = "ban-service", comment = "Specifies the plugin that will provide the ban service")
+        private String banService = "?";
+
+        @Setting(value = "economy-service", comment = "Specifies the plugin that will provide the economy service")
+        private String economyService = "?";
+
+        @Setting(value = "pagination-service", comment = "Specifies the plugin that will provide the pagination service")
+        private String paginationService = "?";
+
+        @Setting(value = "permission-service", comment = "Specifies the plugin that will provide the permission service")
+        private String permissionService = "?";
+
+        @Setting(value = "user-storage-service", comment = "Specifies the plugin that will provide the user storage service")
+        private String userStorageService = "?";
+
+        @Setting(value = "whitelist-service", comment = "Specifies the plugin that will provide the whitelist service")
+        private String whitelistService = "?";
+
+        public String getBanService() {
+            return this.banService;
+        }
+
+        public String getEconomyService() {
+            return this.economyService;
+        }
+
+        public String getPaginationService() {
+            return this.paginationService;
+        }
+
+        public String getPermissionService() {
+            return this.permissionService;
+        }
+
+        public String getUserStorageService() {
+            return this.userStorageService;
+        }
+
+        public String getWhitelistService() {
+            return this.whitelistService;
+        }
+    }
+
+}
--- a/src/main/java/org/spongepowered/common/service/SpongeServiceProvider.java
+++ b/src/main/java/org/spongepowered/common/service/SpongeServiceProvider.java
@@ -24,7 +24,17 @@
  */
 package org.spongepowered.common.service;
 
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.common.reflect.TypeToken;
+import com.google.inject.Inject;
+import com.google.inject.Injector;
 import com.google.inject.Singleton;
+import org.apache.logging.log4j.Level;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.spongepowered.api.Game;
+import org.spongepowered.api.event.lifecycle.ProvideServiceEvent;
 import org.spongepowered.api.service.ServiceProvider;
 import org.spongepowered.api.service.ServiceRegistration;
 import org.spongepowered.api.service.ban.BanService;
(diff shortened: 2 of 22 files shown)


src/main/java/org/spongepowered/common/config/category/ServicesCategory.java
```diff
@@ -0,0 +1,79 @@
...
+    private ServicePluginSubCategory servicePlugin = new ServicePluginSubCategory();
+
+    public ServicePluginSubCategory getServicePlugin() {
+        return this.servicePlugin;
+    }
+
+    @ConfigSerializable
+    public static class ServicePluginSubCategory {
+
+        @Setting("ban-service")
```
ST-DDT: Please include a comment here before merging.

src/main/java/org/spongepowered/common/service/SpongeServiceProvider.java
```diff
@@ -0,0 +1,250 @@
...
+        // This does NOT support third party service interfaces, only impls.
+        for (final Map.Entry<Class<?>, Tuple<Function<ServicesCategory.ServicePluginSubCategory, String>, Supplier<?>>> candidate :
+                AVAILABLE_SERVICES.entrySet()) {
+
+            // If the configration file has a specific plugin ID, we look for it. If it's there, we will ONLY query that
+            // plugin.
+            final Optional<PluginContainer> specificPluginContainer =
+                    Sponge.getPluginManager().getPlugin(candidate.getValue().getFirst().apply(servicePluginSubCategory));
+            final Collection<PluginContainer> toQuery =
+                    specificPluginContainer.<Collection<PluginContainer>>map(ImmutableList::of).orElseGet(Sponge.getPluginManager()::getPlugins);
```
ST-DDT: IMO this should fail, if the specified plugin does not exist or does not provide the service.
The admin probably selected that plugin for a reason.

ST-DDT (commented):
Do I get it right that it is no longer possible to use the service provider for custom/plugin services?
Is there a special reason to it?

dualspiral:
> Do I get it right that it is no longer possible to use the service provider for custom/plugin services?

Correct. While the idea behind the service manager in 1-7 was noble, in reality it wasn't really used, or was particularly usable, by other plugins.

Say plugin A offers service IA. Plugin B comes along and has an implementation to IA, say CB. Good-o, this is a happy situation. What if plugin C comes along and wants to use service IA? Well, it needs a dep on A, so C could just ask A to provide a copy of IA.

In the sad path scenario, B comes along, doesn't properly declare dep on A, boom, `ClassNotFoundException`. Similar issue with C. In the case of having to properly rely on A, however, it generally becomes more obvious as to why a plugin failed.

All the service manager does is provide a very light dependency injection like container - which is mostly useful for Sponge that controls the service interfaces. When you drill down to it, there is no real value in enabling this functionality for non-Sponge impl environments.

Given we also want to lock in services during init, adding third party services adds complexity to the system. If we decide to do it and use this way to do it, fine, we just need to add _another_ event to the lifecycle to collect the service interfaces. I just don't see the point or need to do so.

vanilla/invalid/java/org/spongepowered/server/SpongeVanilla.java
```diff
@@ -179,9 +174,17 @@ public void onServerStopping() {
...
         SpongeCommon.postState(GameState.SERVER_STOPPED, SpongeEventFactory.createGameStoppedServerEvent(Sponge.getCauseStackManager().getCurrentCause()));
         ((SqlServiceImpl) this.game.getServiceManager().provideUnchecked(SqlService.class)).close();
         SpongeCommon.getConfigSaveManager().flush();
+=======
+        SpongeImpl.postState(GameState.SERVER_STOPPED, SpongeEventFactory.createGameStoppedServerEvent(Sponge.getCauseStackManager().getCurrentCause()));
+        if (this.game.getServiceProvider().sqlService() instanceof SqlServiceImpl) {
+            ((SqlServiceImpl) this.game.getServiceProvider().sqlService()).close();
+        }
+        SpongeImpl.getConfigSaveManager().flush();
+>>>>>>> 1762dc36f... Add entry point to service provider, make changes to support this
```
ImMorpheus: mmmh

src/mixins/java/org/spongepowered/common/mixin/core/service/permission/SubjectMixin.java
```diff
@@ -70,15 +70,14 @@ private void subjectConstructor(final CallbackInfo ci) {
     @Override
     public Optional<SubjectReference> bridge$resolveReferenceOptional() {
         if (this.impl$subjectReference == null) {
-            final Optional<PermissionService> serv = SpongeCommon.getGame().getServiceManager().provide(PermissionService.class);
-            serv.ifPresent(permissionService -> new SubjectSettingCallback(this).test(permissionService));
+            new SubjectSettingCallback(this).test(SpongeCommon.getGame().getServiceProvider().permissionService());
```
ImMorpheus: Now that callbacks are gone can this be simplified to a bipredicate (as a lambda I mean) ?

src/main/java/org/spongepowered/common/service/SpongeServiceProvider.java
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    @NonNull
+    @SuppressWarnings("unchecked")
+    private <T> T provideUnchecked(final Class<T> serviceClass) {
+        final Registration<T> registration = (Registration<T>) this.services.get(serviceClass);
+        if (registration != null) {
+            return registration.service();
+        }
+        throw new IllegalStateException("");
```
ImMorpheus: `"No provider is registered for the service '" + service.getName() + "'", service"` (this is the current errore message).

Shouldn't this thrown a `ProvisioningException` ?

vanilla/invalid/java/org/spongepowered/server/SpongeVanilla.java
dualspiral: At least it's invalid, though I thought I'd picked these all up.

src/mixins/java/org/spongepowered/common/mixin/core/service/permission/SubjectMixin.java
dualspiral: More likely to be a static method, the callback isn't really a callback anymore and we don't need to pass a method reference around.

src/main/java/org/spongepowered/common/service/SpongeServiceProvider.java
dualspiral: It's probably worth noting that this is a private method, so the rationale was simply that this never should happen. So ISE is fine - the object is in an illegal state if this is called and the service isn't there.

Given that exception does not exist, no.

ImMorpheus: I missed the exception was remove in the API PR

src/main/java/org/spongepowered/common/service/SpongeServiceProvider.java
dualspiral: Just asked @Zidane about this, we're going to print a huge warning instead.

vanilla/invalid/java/org/spongepowered/server/SpongeVanilla.java
```diff
@@ -179,9 +174,11 @@ public void onServerStopping() {
     }
 
     public void onServerStopped() throws IOException {
-        SpongeCommon.postState(GameState.SERVER_STOPPED, SpongeEventFactory.createGameStoppedServerEvent(Sponge.getCauseStackManager().getCurrentCause()));
-        ((SqlServiceImpl) this.game.getServiceManager().provideUnchecked(SqlService.class)).close();
-        SpongeCommon.getConfigSaveManager().flush();
+        SpongeImpl.postState(GameState.SERVER_STOPPED, SpongeEventFactory.createGameStoppedServerEvent(Sponge.getCauseStackManager().getCurrentCause()));
```
Zidane: You reverted back to SpongeImpl.

src/main/java/org/spongepowered/common/service/permission/SpongePermissionService.java
```diff
@@ -66,8 +67,8 @@
     private final SpongeSubjectCollection defaultCollection;
     private final SpongeSubject defaultData;
 
-    public SpongePermissionService(Game game) {
-        this.game = game;
+    public SpongePermissionService() {
```
Zidane: Would it be worth injecting the SpongeGame instance? Would require turning these into Singletons and creating them via the injector only if you need too.

src/main/java/org/spongepowered/common/event/service/ProvideServiceEventImpl.java
```diff
@@ -0,0 +1,74 @@
...
+import com.google.common.reflect.TypeToken;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.spongepowered.api.event.cause.Cause;
+import org.spongepowered.api.event.lifecycle.ProvideServiceEvent;
+
+import java.util.function.Supplier;
+
+// Specialised logic is required for this.
+public class ProvideServiceEventImpl<T> implements ProvideServiceEvent<T> {
```
Zidane: I like final things.

src/main/java/org/spongepowered/common/config/category/ServicesCategory.java
```diff
@@ -0,0 +1,103 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.config.category;
+
+import ninja.leaping.configurate.objectmapping.Setting;
+import ninja.leaping.configurate.objectmapping.serialize.ConfigSerializable;
+
+@ConfigSerializable
+public class ServicesCategory extends ConfigCategory {
```
Zidane: I like final things.

vanilla/invalid/java/org/spongepowered/server/SpongeVanilla.java
dualspiral: Given GameState is also gone, that line might as well go.

src/main/java/org/spongepowered/common/service/permission/SpongePermissionService.java
dualspiral: Yes, if we're okay with passing an injector to the service provider.

Zidane: I don't see an issue with it.

src/main/java/org/spongepowered/common/config/category/ServicesCategory.java
```diff
@@ -0,0 +1,103 @@
...
+        @Setting(value = "economy-service", comment = "Specifies the plugin that will provide the economy service")
+        private String economyService = "?";
+
+        @Setting(value = "pagination-service", comment = "Specifies the plugin that will provide the pagination service")
+        private String paginationService = "?";
+
+        @Setting(value = "permission-service", comment = "Specifies the plugin that will provide the permission service")
+        private String permissionService = "?";
+
+        @Setting(value = "sql-service", comment = "Specifies the plugin that will provide the SQL service")
```
ImMorpheus: SqlService is dead

src/mixins/java/org/spongepowered/common/mixin/tracker/world/chunk/ChunkMixin_OwnershipTracked.java
```diff
@@ -88,7 +88,9 @@
             + "/function/Consumer;)V", at = @At("RETURN"))
     private void tracker$setUpUserService(final World world, final ChunkPos pos, final Biome[] biomes, final UpgradeData data, final ITickList<Block> blockTickList,
                                           final ITickList<Fluid> fluidTickList, final long inhabitedTime, final ChunkSection[] sections, final Consumer<Chunk> chunkConsumer, final CallbackInfo ci) {
-        this.tracker$userStorageService = world != null && !((WorldBridge) world).bridge$isFake() ? null : Sponge.getServiceProvider().userStorageService();
+        this.tracker$userStorageService = world != null && !((WorldBridge) world).bridge$isFake()
+                                  ? null
+                                  : SpongeCommon.getGame().getServiceProvider().userStorageService();
```
ImMorpheus, line 93: This was causing a big performance issue in api7.

It should use the user storage service when the world is **not** fake (not the opposite).
(see `ChunkMixin_Tracker`)

dualspiral: I'm not interested in making logic changes outside the replacement of the service system in this PR because otherwise it'll never end.

src/main/java/org/spongepowered/common/config/category/ServicesCategory.java
dualspiral: Good spot.

PR #2654, opened 2020-06-27
Don't bypass attackEntityFrom if LivingHurtEvent is cancelled

### LINKED ISSUE #3178: Compatibility with FirstAid since RC4029 (Death Event Cancelling)
**I am currently running**

- SpongeForge version: spongeforge-1.12.2-2838-7.2.2-RC4029 (happens on latest, too)
- Forge version: 1.12.2-2847
- Java version: 1.8.0_241 64bit
- Operating System: Ubuntu 18.04

- Plugins/Mods:
firstaid-1.6.14
spongeforge-1.12.2-2838-7.2.2-RC4029.jar

**Issue Description**
(Conversation begins at: https://canary.discordapp.com/channels/142425412096491520/142425412096491520/726006967591501845)

It seems that since updating our Sponge to a version past RC4029 (such as RB 7.2.2), there's an issue where deaths aren't functioning normally where death messages aren't showing, and mods that use deaths in any way aren't able to function as intended. This will only occur when Firstaid is installed alongside Sponge (the Sponge version must be past RC4029, otherwise all is well). Running the server with `-Dsponge.shouldFireAll=true` also didn't change anything.

After talking with @dualspiral it seems that FirstAid isn't firing the death event, and the only indicators we have is that FA is cancelling the `LivingHurtEvent` at:
- https://github.com/ichttt/FirstAid/blob/master/src/main/java/ichttt/mods/firstaid/common/EventHandler.java#L130
- https://github.com/ichttt/FirstAid/blob/master/src/main/java/ichttt/mods/firstaid/common/EventHandler.java#L111

It was also mentioned that Sponge has adjusted some things in the recent updates that could contribute to why it's only occuring while Sponge is present and why it only occurs since then.

Thank you!

https://gist.github.com/TheFlash787/061dff055b0f9e2f1a0d33203d52a43b

dualspiral:
Don't bypass the rest of the attackEntityFrom method if damageEntityHook returns `false`

This brings our method much closer to what Forge does and should fix SpongePowered/SpongeForge#3178

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/EntityLivingBaseMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/EntityLivingBaseMixin.java
@@ -464,9 +464,7 @@ public abstract class EntityLivingBaseMixin extends EntityMixin implements Livin
                     if (this.entityType.isModdedDamageEntityMethod) {
                         this.damageEntity(source, amount - this.lastDamage);
                     } else {
-                        if (!this.bridge$damageEntityHook(source, amount - this.lastDamage)) {
-                            return false;
-                        }
+                        this.bridge$damageEntityHook(source, amount - this.lastDamage);
                     }
 
                     // this.damageEntity(source, amount - this.lastDamage); // handled above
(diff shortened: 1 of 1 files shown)


dualspiral:
Test build: [spongeforge-1.12.2-2838-7.2.3-RC0.jar.zip](https://github.com/SpongePowered/SpongeCommon/files/4840541/spongeforge-1.12.2-2838-7.2.3-RC0.jar.zip)

ImMorpheus (approved):
Working as intended. 
I like it when we don't break Forge contract.

PR #2619, opened 2020-05-22
Add pathfinding chunkload optimization

clienthax:
Noticed pathfinding was causing chunkloading while investigating another issue.
```

Chunks loaded 24   [-945 -1656, -942 -1659, -942 -1656, -946 -1658, -945 -1659, -945 -1656, -944 -1655, -943 -1655, -942 -1655, -942 -1656, -942 -1656, -942 -1655, -942 -1656, -944 -1655, -942 -1656, -942 -1655, -944 -1655, -942 -1656, -943 -1655, -944 -1656, -942 -1656, -942 -1656, -944 -1656, -942 -1656]
	at org.spongepowered.common.util.SpongeHooks.logStack(SpongeHooks.java:107) [SpongeHooks.class:1.12.2-2838-7.2.2-RC4027]
	at org.spongepowered.common.util.SpongeHooks.logChunkLoad(SpongeHooks.java:206) [SpongeHooks.class:1.12.2-2838-7.2.2-RC4027]
	at net.minecraft.world.chunk.Chunk.handler$zmb000$impl$UpdateNeighborsOnLoad(Chunk.java:3551) [axw.class:?]
	at net.minecraft.world.chunk.Chunk.onLoad(Chunk.java:863) [axw.class:?]
	at net.minecraftforge.common.chunkio.ChunkIOProvider.syncCallback(ChunkIOProvider.java:109) [ChunkIOProvider.class:?]
	at net.minecraftforge.common.chunkio.ChunkIOExecutor.syncChunkLoad(ChunkIOExecutor.java:94) [ChunkIOExecutor.class:?]
	at net.minecraft.world.gen.ChunkProviderServer.loadChunk(ChunkProviderServer.java:118) [on.class:?]
	at net.minecraft.world.gen.ChunkProviderServer.loadChunk(ChunkProviderServer.java:89) [on.class:?]
	at net.minecraft.world.gen.ChunkProviderServer.redirect$zme000$impl$ProvideChunkForced(ChunkProviderServer.java:1642) [on.class:?]
	at net.minecraft.world.gen.ChunkProviderServer.provideChunk(ChunkProviderServer.java:135) [on.class:?]
	at net.minecraft.world.World.getChunk(World.java:310) [amu.class:?]
	at net.minecraft.world.World.getChunk(World.java:305) [amu.class:?]
	at net.minecraft.world.WorldServer.getBlockState(WorldServer.java:5062) [oo.class:?]
[...]

PR #3033, opened 2020-06-28
Add raytracing chunkload optimization

clienthax:
Same kind of thing as https://github.com/SpongePowered/SpongeCommon/pull/2619 but for raytracing.


Zidane:
@clienthax This is a good change. Thanks for porting this patch.

I'll leave it to @dualspiral to review but it is a good from me.

dualspiral:
Looks fine to me. As it's off by default I'll just pull this now for next RB...

PR #3039, opened 2020-07-03
fixes #2803

### LINKED ISSUE #2803: Update BedBlockMixin from core:invalid
Located in: [core:invalid](https://github.com/SpongePowered/SpongeCommon/blob/6dc6982ad7d3ed44eb5ab542f1fcff51e2e1af38/invalid/main/java/org/spongepowered/common/mixin/invalid/core/block/BedBlockMixin.java)

Part of: block

Advice and tips include:

mikucat0309:
fixes #2803 
change super class to BlockMixin


src/mixins/java/org/spongepowered/common/mixin/core/block/BedBlockMixin.java
```diff
@@ -33,7 +33,7 @@
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
 
 @Mixin(BedBlock.class)
-public abstract class BedBlockMixin extends HorizontalBlockMixin {
+public abstract class BedBlockMixin extends BlockMixin {
 
     @Inject(method = "hasRoomForPlayer", at = @At(value = "RETURN"), cancellable = true)
```
gabizou: This injection doesn't exist in 1.14 either, it's not just the class hierarchy that is broken here.

gabizou (changes requested):

src/mixins/java/org/spongepowered/common/mixin/core/block/BedBlockMixin.java
gabizou: It seems like the intent of the original mixin was to specifically adjust the position exiting the bed to omit positions available if the positions were not within the world border.  

Looking at the comparison of 1.12's mixin to 1.14, the most accurate target for this would be updating the injection point to target `func_220175_a` which takes a few different arguments, but mostly returns the same thing (instead of nullable, it's an optional).

mikucat0309: I didn't find `func_220175_a` in version 20200119-1.14.3 of mcp mapping
but I found that: `func_213828_b` also checks the blcok whether obstruct player sleeping or not in `PlayerEntity`, maybe it is the alternative of `hasRoomForPlayer` ?
```
private boolean func_213828_b(BlockPos p_213828_1_, Direction p_213828_2_) {
   BlockPos blockpos = p_213828_1_.up();
   return !this.isNormalCube(blockpos) || !this.isNormalCube(blockpos.offset(p_213828_2_.getOpposite()));
}
```

gabizou: It most definitely still exists if you open up `BedBlock` in the workspace:
```java
   protected static Optional<Vec3d> func_220175_a(EntityType<?> p_220175_0_, IWorldReader p_220175_1_, BlockPos p_220175_2_) {
      VoxelShape voxelshape = p_220175_1_.getBlockState(p_220175_2_).getCollisionShape(p_220175_1_, p_220175_2_);
      if (voxelshape.getEnd(Direction.Axis.Y) > 0.4375D) {
         return Optional.empty();
      } else {
         BlockPos.MutableBlockPos blockpos$mutableblockpos = new BlockPos.MutableBlockPos(p_220175_2_);

         while(blockpos$mutableblockpos.getY() >= 0 && p_220175_2_.getY() - blockpos$mutableblockpos.getY() <= 2 && p_220175_1_.getBlockState(blockpos$mutableblockpos).getCollisionShape(p_220175_1_, blockpos$mutableblockpos).isEmpty()) {
            blockpos$mutableblockpos.move(Direction.DOWN);
         }

         VoxelShape voxelshape1 = p_220175_1_.getBlockState(blockpos$mutableblockpos).getCollisionShape(p_220175_1_, blockpos$mutableblockpos);
         if (voxelshape1.isEmpty()) {
            return Optional.empty();
         } else {
            double d0 = (double)blockpos$mutableblockpos.getY() + voxelshape1.getEnd(Direction.Axis.Y) + 2.0E-7D;
            if ((double)p_220175_2_.getY() - d0 > 2.0D) {
               return Optional.empty();
            } else {
               float f = p_220175_0_.getWidth() / 2.0F;
               Vec3d vec3d = new Vec3d((double)blockpos$mutableblockpos.getX() + 0.5D, d0, (double)blockpos$mutableblockpos.getZ() + 0.5D);
               return p_220175_1_.areCollisionShapesEmpty(new AxisAlignedBB(vec3d.x - (double)f, vec3d.y, vec3d.z - (double)f, vec3d.x + (double)f, vec3d.y + (double)p_220175_0_.getHeight(), vec3d.z + (double)f)) ? Optional.of(vec3d) : Optional.empty();
            }
         }
      }
   }
```

to which the injection would be pointing to the top of the method to check the world border belonging to the world reader is still existing.

PR #3040, opened 2020-07-04
Implement new trade offer logic

i509VCB:
And some whitespace for the builder.

### DIFF
--- a/src/main/java/org/spongepowered/common/item/merchant/SpongeTradeOfferBuilder.java
+++ b/src/main/java/org/spongepowered/common/item/merchant/SpongeTradeOfferBuilder.java
@@ -28,6 +28,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 
+import com.google.common.base.Preconditions;
 import net.minecraft.item.MerchantOffer;
 import org.spongepowered.api.data.persistence.DataBuilder;
 import org.spongepowered.api.data.persistence.DataView;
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/common/item/merchant/SpongeTradeOfferBuilder.java
```diff
@@ -53,9 +54,12 @@
     private int useCount;
     private int maxUses;
     private boolean allowsExperience;
+    private int merchantExperienceGranted;
+    private double priceGrowthMultiplier;
+    private int demandBonus;
 
     public SpongeTradeOfferBuilder() {
-        super(TradeOffer.class, 1);
+        super(TradeOffer.class, 2);
```
gabizou: What's the difference here?

src/main/java/org/spongepowered/common/item/merchant/SpongeTradeOfferBuilder.java
```diff
@@ -68,6 +72,7 @@ public SpongeTradeOfferBuilder() {
 
     @Override
     public TradeOffer.Builder secondBuyingItem(final ItemStack item) {
+        Preconditions.checkNotNull(item, "Second buying item cannot be null");
```
gabizou: Can the second item not be null? I thought we allowed it to be null implicitly?

gabizou (approved):
Null check aside, I'd be willing to make the second buying item as a null setter as well, but nitpicks are nitpicks.

src/main/java/org/spongepowered/common/item/merchant/SpongeTradeOfferBuilder.java
i509VCB: The item can be empty.

PR #3041, opened 2020-07-04
fixes #2987

### LINKED ISSUE #2987: Update ChatUtil from core:invalid
Located in: [core:invalid](https://github.com/SpongePowered/SpongeCommon/blob/6dc6982ad7d3ed44eb5ab542f1fcff51e2e1af38/invalid/main/java/org/spongepowered/common/text/chat/ChatUtil.java)

Part of: chat

Advice and tips include:

mikucat0309:
Replace `CommandSource` to `CommandCause`

### DIFF
--- a/src/main/java/org/spongepowered/common/text/chat/ChatUtil.java
+++ b/src/main/java/org/spongepowered/common/text/chat/ChatUtil.java
@@ -47,7 +47,7 @@ public final class ChatUtil {
     private ChatUtil() {
     }
 
-    public static void sendMessage(ITextComponent component, MessageChannel channel, CommandCause source, boolean isChat) {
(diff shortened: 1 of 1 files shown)


dualspiral:
The `CommandCause` is not the same as what a `CommandSource` was. This will get fired when a `ServerPlayer` or `Server` sends a message, for example, and there will be no `CommandCause` involved. 

What you really want is who the sender of a chat message is - so actually, this might just simply be an `Object`, I don't think we have a type to indicate a message sender. It may instead be a `MessageReceiver` - [but looking at message channels as they stand, I think the sender is indeed intended to be an `Object`](https://github.com/SpongePowered/SpongeAPI/blob/api-8/src/main/java/org/spongepowered/api/text/channel/MessageChannel.java#L216).

Zidane:
Closing as Adventure has changed this.

PR #3044, opened 2020-07-05
Update pagination service from invalid

### LINKED ISSUE #2990: Update SpongePaginationList from core:invalid
Located in: [core:invalid](https://github.com/SpongePowered/SpongeCommon/blob/6dc6982ad7d3ed44eb5ab542f1fcff51e2e1af38/invalid/main/java/org/spongepowered/common/service/pagination/SpongePaginationList.java)

Part of: pagination

Advice and tips include:

RedNesto:
Includes a couple of fixes.

From a couple of simple tests it seems to work correctly.

Closes #2990 #2991 #2992 #2993 #2994 #2995 #2996

### DIFF
--- a/src/main/java/org/spongepowered/common/command/manager/SpongeCommandManager.java
+++ b/src/main/java/org/spongepowered/common/command/manager/SpongeCommandManager.java
@@ -65,6 +65,7 @@ import org.spongepowered.api.event.SpongeEventFactory;
 import org.spongepowered.api.event.Cause;
 import org.spongepowered.api.event.EventContextKeys;
 import org.spongepowered.api.event.command.ExecuteCommandEvent;
+import org.spongepowered.api.service.pagination.PaginationService;
 import org.spongepowered.api.service.permission.Subject;
 import org.spongepowered.api.util.TextMessageException;
 import org.spongepowered.common.SpongeCommon;
@@ -521,11 +522,14 @@ public final class SpongeCommandManager implements CommandManager {
             throw new RuntimeException("Failed to create root Sponge command!", ex);
         }
         try {
-            SpongeParameterizedCommandRegistrar.INSTANCE.register(
-                    Launcher.getInstance().getCommonPlugin(),
-                    ((SpongePaginationService) Sponge.getServiceProvider().paginationService()).createPaginationCommand(),
-                    "pagination", "page"
-            );
+            final PaginationService paginationService = Sponge.getServiceProvider().paginationService();
+            if (paginationService instanceof SpongePaginationService) {
+                SpongeParameterizedCommandRegistrar.INSTANCE.register(
+                        Launcher.getInstance().getCommonPlugin(),
+                        ((SpongePaginationService) paginationService).createPaginationCommand(),
+                        "pagination", "page"
+                );
+            }
         } catch (final CommandFailedRegistrationException ex) {
             throw new RuntimeException("Failed to create pagination command!", ex);
         }
--- a/src/main/java/org/spongepowered/common/command/manager/SpongeCommandManager.java
+++ b/src/main/java/org/spongepowered/common/command/manager/SpongeCommandManager.java
@@ -51,6 +51,7 @@ import org.apache.logging.log4j.Level;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.spongepowered.api.Game;
+import org.spongepowered.api.Sponge;
 import org.spongepowered.api.command.CommandCause;
 import org.spongepowered.api.command.CommandResult;
 import org.spongepowered.api.command.exception.CommandException;
@@ -77,6 +78,7 @@ import org.spongepowered.common.event.ShouldFire;
 import org.spongepowered.common.event.lifecycle.RegisterCommandEventImpl;
 import org.spongepowered.common.event.tracking.PhaseTracker;
 import org.spongepowered.common.launch.Launcher;
+import org.spongepowered.common.service.pagination.SpongePaginationService;
 import org.spongepowered.common.util.PrettyPrinter;
 import org.spongepowered.plugin.PluginContainer;
 
@@ -517,6 +519,15 @@ public final class SpongeCommandManager implements CommandManager {
             );
         } catch (final CommandFailedRegistrationException ex) {
             throw new RuntimeException("Failed to create root Sponge command!", ex);
+        }
+        try {
+            SpongeParameterizedCommandRegistrar.INSTANCE.register(
+                    Launcher.getInstance().getCommonPlugin(),
+                    ((SpongePaginationService) Sponge.getServiceProvider().paginationService()).createPaginationCommand(),
+                    "pagination", "page"
+            );
+        } catch (final CommandFailedRegistrationException ex) {
+            throw new RuntimeException("Failed to create pagination command!", ex);
         }
         final Set<TypeToken<?>> usedTokens = new HashSet<>();
         for (final CommandRegistrar<?> registrar : this.game.getRegistry().getCatalogRegistry().getAllOf(CommandRegistrar.class)) {
--- a/src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java
+++ b/src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java
@@ -263,7 +263,7 @@ final class PaginationCalculator {
      * @return The finalized, properly styled text.
      */
     private Component finalizeBuilder(final Component text, final TextComponent.Builder build) {
-        return build.style(text.style().toBuilder().build()).build();
+        return build.style(text.style()).build();
     }
 
     /**
@@ -281,7 +281,7 @@ final class PaginationCalculator {
                 .append(spaces)
                 .append(text)
                 .append(spaces)
-                .style(text.style().toBuilder().build())
+                .style(text.style())
                 .build();
     }
 
--- a/src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java
+++ b/src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java
@@ -25,7 +25,6 @@
 package org.spongepowered.common.service.pagination;
 
 import com.google.common.annotations.VisibleForTesting;
-import net.kyori.adventure.audience.Audience;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.TextComponent;
 import net.minecraft.util.text.ITextComponent;
@@ -92,7 +91,7 @@ class PaginationCalculator {
         }
     }
 
-    int getLinesPerPage(Audience source) {
+    int getLinesPerPage() {
         return this.linesPerPage;
     }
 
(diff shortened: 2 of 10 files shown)


src/main/java/org/spongepowered/common/command/manager/SpongeCommandManager.java
```diff
@@ -524,6 +526,15 @@ public void init() {
         } catch (final CommandFailedRegistrationException ex) {
             throw new RuntimeException("Failed to create root Sponge command!", ex);
         }
+        try {
+            SpongeParameterizedCommandRegistrar.INSTANCE.register(
+                    Launcher.getInstance().getCommonPlugin(),
+                    ((SpongePaginationService) Sponge.getServiceProvider().paginationService()).createPaginationCommand(),
```
dualspiral: This should only run if the service wasn't replaced, this will throw a class cast exception otherwise.

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationService.java
```diff
@@ -143,105 +131,94 @@ private SourcePaginations getPaginationStateForPlayer(Player source, boolean cre
...
+                .setShortDescription(TextComponent.of("Helper command for paginations occurring"))
                 .build();
     }
 
-    private class ActivePaginationCommandElement extends CommandElement {
-
-        protected ActivePaginationCommandElement(@Nullable Text key) {
-            super(key);
-        }
+    private class ActivePaginationParser implements ValueParser<ActivePagination> {
```
dualspiral: Any reason that this is split into `ValueParser` and `ValueCompleter`? In general, they should just be one `ValueParameter` if this is the case ([the docs say if a `ValueParameter` is passed it will use the completer there](https://github.com/SpongePowered/SpongeAPI/blob/api-8/src/main/java/org/spongepowered/api/command/parameter/Parameter.java#L979-L983)).

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationService.java
```diff
@@ -143,105 +131,94 @@ private SourcePaginations getPaginationStateForPlayer(Player source, boolean cre
...
+                .setExecutor(pageExecutor)
+                .build();
 
         //We create the child manually in order to force that paginationElement is required for all children + fallback
         //https://github.com/SpongePowered/SpongeAPI/issues/1272
-        return CommandSpec.builder().arguments(paginationElement, firstParsing(childDispatcher, pageArgs))
-                .executor(childDispatcher)
-                .description(t("Helper command for paginations occurring"))
+        return Command.builder()
+                .parameters(paginationIdParameter, firstOf(pageParameter,
```
dualspiral: All static calls must be qualified - no static imports for this.

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationService.java
```diff
@@ -143,105 +131,94 @@ private SourcePaginations getPaginationStateForPlayer(Player source, boolean cre
...
-                    args.<ActivePagination>getOne("pagination-id").get().previousPage();
+        Command.Parameterized prev = Command.builder()
+                .setShortDescription(TextComponent.of("Go to the previous page"))
+                .setExecutor((context) -> {
+                    context.requireOne(paginationIdParameter).previousPage();
                     return CommandResult.success();
                 }).build();
 
-        CommandElement pageArgs = integer(t("page"));
+        Parameter.Value<Integer> pageParameter = integerNumber().setKey("page").build();
```
dualspiral: @Zidane likes final things. 

(When possible, variables and classes should be final)

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationBuilder.java
```diff
@@ -28,23 +28,27 @@
...
-    private Text title;
+    private Component title;
     @Nullable
-    private Text header;
+    private Component header;
     @Nullable
-    private Text footer;
-    private Text paginationSpacer = Text.of("=");
+    private Component footer;
+    private Component paginationSpacer = EQUALS_TEXT;
```
dualspiral: Static acesses from the same class should also be qualified with the class name.

dualspiral (changes requested):
Mostly style, but a potential crash here too.

src/main/java/org/spongepowered/common/command/manager/SpongeCommandManager.java
RedNesto: Oh right, I assumed it can't be replaced for some reason.

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationService.java
RedNesto: I missed that part, will do !

src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java
```diff
@@ -218,82 +213,57 @@ int getWidth(Component text) {
...
      * Finalizes the builder used in centering text.
      *
      * @param text The text to get the style from
      * @param build The work in progress text builder
      * @return The finalized, properly styled text.
      */
-    private Component finalizeBuilder(Component text,TextComponent.Builder build) {
-        return build.style(text.style()).build();
+    private Component finalizeBuilder(final Component text, final TextComponent.Builder build) {
+        return build.style(text.style().toBuilder().build()).build();
```
dualspiral: Is there a reason for this `toBuilder().build()`? The following compiles fine:

```
final Component component = TextComponent.empty();
TextComponent.builder().style(component.style()).build();
```

I haven't run it though - if there is a problem that you've worked around, we need to report it to Adventure.

src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java
```diff
@@ -306,12 +276,12 @@ private Component finalizeBuilder(Component text,TextComponent.Builder build) {
...
      * @return The text with the added spaces
      */
-    private Component addSpaces(Component spaces, Component text) {
+    private Component addSpaces(final Component spaces, final Component text) {
         return TextComponent.builder()
                 .append(spaces)
                 .append(text)
                 .append(spaces)
-                .style(text.style())
+                .style(text.style().toBuilder().build())
```
dualspiral: As above.

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationList.java
```diff
@@ -46,19 +45,23 @@
...
+public final class SpongePaginationList implements PaginationList {
 
     private final SpongePaginationService service;
-    private final Iterable<Text> contents;
-    private final Optional<Text> title;
-    private final Optional<Text> header;
-    private final Optional<Text> footer;
-    private final Text paginationSpacer;
+    private final Iterable<Component> contents;
+    private final Optional<Component> title;
```
dualspiral: More of a note for the future - but storing an optional just feels _wrong_. It's being used to avoid optional (re)creation in getters, but... how often are the getters called? Not often I'm sure! The `sendTo` method actually unwraps the optional into a null.

Perhaps this is a good time to ditch the optionals and just have nullable fields.

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationService.java
```diff
@@ -0,0 +1,216 @@
...
+            return CommandResult.success();
+        };
+
+        final Command.Parameterized page = Command.builder()
+                .setShortDescription(TextComponent.of("Go to a specific page"))
+                .parameter(pageParameter)
+                .setExecutor(pageExecutor)
+                .build();
+
+        //We create the child manually in order to force that paginationElement is required for all children + fallback
```
dualspiral: Comment should be ditched.

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationService.java
```diff
@@ -0,0 +1,216 @@
...
+            final ActivePagination pagination = paginations.get(id);
+            if (pagination == null) {
+                throw reader.createException(TextComponent.of("No pagination registered for id " + id));
+            }
+            return Optional.ofNullable(paginations.get(id));
+        }
+
+        @Override
+        public List<String> complete(final CommandContext context) {
+            return context.getCause().first(Audience.class)
```
dualspiral: `.first(Audience.class)` -> `.getAudience()` (audiences do have special handling in command causes)

src/main/java/org/spongepowered/common/service/pagination/SpongePaginationService.java
```diff
@@ -0,0 +1,216 @@
...
+                .build();
+    }
+
+    private final class ActivePaginationParameter implements ValueParameter<ActivePagination> {
+
+        @Override
+        public Optional<? extends ActivePagination> getValue(final Parameter.Key<? super ActivePagination> parameterKey,
+                final ArgumentReader.Mutable reader,
+                final CommandContext.Builder context) throws ArgumentParseException {
+            final Audience source = context.getCause().first(Audience.class)
```
dualspiral: `.first(Audience.class)` -> `.getAudience()`

src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java
RedNesto: Oh I didn't put a comment there to explain.
It seems to be needed to avoid StackOverflowExceptions that happen later for some reason. I mentioned it to kashike and AFAIK we're not sure why this happens.

dualspiral: Is there an issue reported for this? I imagine it's ~~probably on the Adventure end~~ but it needs to be reported somewhere. (edit: it wasn't)

Once done, add a comment linking to that issue so if some poor soul wants more information (and perhaps wants to have a stab at fixing it), they can do so.

RedNesto: Well I don't even know if the issue lies in the library or the SpongeCommon glue code. I should talk to kashike about it again I think.

dualspiral: Oh, I just looked and I know exactly what's going on. It's basically the fact that the Minecraft `Style` object is being cached for a given Adventure `Style` object, but when you use Minecraft's `Style#setStyle` it also sets the parent of the child's style. In this case, the parent and the child have been supplied with the same Style reference, so when the parent style is set... it's itself.

Sponge error. We basically should not be caching mutable objects (edit: without copying them when we use them).

dualspiral: https://github.com/SpongePowered/SpongeCommon/commit/44be61c0dbc79069f82c174b63d7a68aa254c1b7

RedNesto: Amazing, thank you ! I hope this is the last force push I have to do for this PR.

PR #3047, opened 2020-07-11
[API-8] Implement data keys related to pandas

Lignium:


src/main/java/org/spongepowered/common/data/provider/entity/EntityDataProviders.java
```diff
@@ -592,22 +592,22 @@ private void registerFox() {
     }
 
     private void registerPanda() {
-//        register(PandaEntity.class, Keys.EATING_TIME, getter, setter); // TODO
+        register(new PandaEntityEatingTimeProvider());
```
Cybermaxke: It's possible to use `PandaEntityAccessor` as a target type, so you can just use method references.

Lignium: Thanks, I'll know

Zidane:
Superseded by Grinch who updated it to use new system.

Thanks again!

PR #3053, opened 2020-07-14
Make some Jar in Jar tasks for distribution

gabizou:
So, there's something that I've been wanting to tackle for a minute, something that is closer to the end-goal of our wonderful 1.14 update: Building artifacts of our implementations with the new setup...

# Goals

We've wanted to be able to support jar-in-jar for a moment, but previous toolchains did not allow for such a thing. Now we're at the start with our newest toolchain that does support jar-in-jars, and with our split up source sets, we're in a prime case of being able to do exactly that: Build SpongeCommon/SpongeVanilla/SpongeForge with jar-in-jars.

# Anti-goals

We aren't setting up an installer ~~yet~~, but this will allow us to carefully manage our dependencies and reduce delegating libraries in a module centric world (Java 9+).

# Output

Currently, we support creating the following:
- [x] `sourcesJar`
- [x] `javadocJar`
- [x] `launchJar`
- [x] `accessorsJar`
- [x] `mixinsJar`
- [x] `fatJar`*

Caveats:
- [ ] Verify whether mixing configs can exist within these jars in the fat jar
- [ ] plugin-spi detection of plugins within jar-in-jars
- [ ] Manifest generation for these different jars?
- [ ] Should sourcesJar build all source sets?
- [ ] modlauncher questions?

<details name="Some Pretty Pictures">
Screen_Shot_2020-07-13_at_10 47 02_PM
</details>

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -34,6 +34,10 @@ val minecraftDep: String by project
 val minecraftVersion: String by project
 val recommendedVersion: String by project
 
+base {
+    archivesBaseName = "spongecommon"
+}
+
 minecraft {
     mappings(mcpType, mcpMappings)
     project.sourceSets["main"].resources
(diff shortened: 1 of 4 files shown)


build.gradle.kts
```diff
@@ -75,6 +79,44 @@ val modlauncherConfig by configurations.register("modlauncher") {
...
+val mixinsJar by tasks.registering(Jar::class) {
+    group = "build"
+    archiveClassifier.set("mixins")
+}
+
+val accessorsJar by tasks.registering(Jar::class) {
+    group = "build"
+    archiveClassifier.set("accessors")
+}
+val javadocJar by tasks.registering(Jar::class) {
```
dualspiral, line 111: Do we really need the JD task for Common? We don't really write them for the impl and so I don't see why we need to potentially generate such a jar.

PR #3054, opened 2020-07-14
Create permission nodes for commands that do op checks.

dualspiral:
(aka I never thought I'd miss the `canUseCommand(String)` method)

As it says in the title, what this attempts to do is determine permissions for given commands based on their node path and if a permission level check is made.

Unlike in API-7 which just slaps a permission on a command, we have to think about this some more. Because of Brig's node like structure, _any_ node (so, parameter) could do a check. We therefore need to make sure we have permissions for all nodes. I do this by just incorporating the node path into the permission path. `/execute` has some pretty long permission nodes. Illegal characters in node names are stripped, and if the entire string was illegal (e.g. `<=`) I just replace it with `node`.

As it stands, if you have `/foo bar` owned by `baz` and they both have permission checks, using the current scheme as a base, you get:

`/foo` -> `baz.command.foo`
`/foo bar` -> `baz.command.foo.bar`

...which is a problem because if you grant `/foo`, you grant `/foo bar`, but you need `/foo`'s permission to get to `/foo bar` anyway, making the second permission useless unless you're willing to go ungrant swathes of permissions. So we need to alter the scheme slightly, we'll have to add an identifier at the end of the string to avoid this from happening. If we were to take a leaf out of Nucleus' playbook, we'd go for `base`, so you get:

`/foo` -> `baz.command.foo.base`
`/foo bar` -> `baz.command.foo.bar.base`

and now, granting `/foo` does not grant `/foo bar`.

Permissions are calculated the first time an op check is made, then cached on a per node basis. They are logged to info, this will get downgraded to debug once this looks okay.

~At the moment, it looks like non-minecraft permissions are reporting `true` on the integrated server. I need to have a look as to why that is. It seems things are running fine for Minecraft commands though - op levels seem to be respected there so the right subject collections are being read.~ I forgot to assign permissions to nodes, that's fixed here.

Thoughts, people?

---

In other fixes:

* Setting up redirects for client completion depends on alphabetical order of the nodes in vanilla. This causes no end of problems. We patch that to become insertion order dependent, though I will likely replace that so that redirect nodes get sorted last (but this works for now).
* Some commands/redirects that we build pointed to nodes not actually in the tree. The `BrigadierCommandRegistrar` has been fixed to do it properly now.

---

Other things we need to think about - particularly for SF:

* What happens when a mod does a permission value check outside of commands - what pemission do we use then? We may need to just use permissions like `modid.level.<x>` (0 <= x <=4). I assume we get the mod out of the cause stack. 
* What if a mod does a core mod transformation and does an op level check without going through their own classes? That might be a little harder to do.
* I need to see if Forge has a permissions handler like they did in 1.12 - we'll want to hook into that.

Much of this is worst case scenarios, but it's worth thinking about them.


src/main/java/org/spongepowered/common/command/brigadier/dispatcher/SpongeNodePermissionCache.java
```diff
@@ -0,0 +1,132 @@
...
+    private final static WeakHashMap<CommandNode<CommandSource>, Supplier<String>> PERMISSION_MAP = new WeakHashMap<>();
+
+    public static boolean canUse(
+            final boolean isRoot,
+            final CommandDispatcher<CommandSource> dispatcher,
+            final CommandNode<CommandSource> node,
+            final CommandSource source
+    ) {
+        final Supplier<String> supplier;
+        if (SpongeNodePermissionCache.PERMISSION_MAP.containsKey(node)) {
```
ImMorpheus: Can be avoided by checking if get return null

src/main/java/org/spongepowered/common/command/brigadier/dispatcher/SpongeNodePermissionCache.java
```diff
@@ -0,0 +1,132 @@
...
+            ((CommandSourceBridge) source).bridge$setPotentialPermissionNode(null);
+        }
+    }
+
+    public static String createFromNode(
+            final CommandDispatcher<CommandSource> dispatcher,
+            final CommandNode<CommandSource> node) {
+        final String permission;
+        if (node.getRedirect() != null && !(node.getRedirect() instanceof RootCommandNode) && node.getCommand() == null) {
+            if (!SpongeNodePermissionCache.PERMISSION_MAP.containsKey(node)) {
```
ImMorpheus: Same as above

src/main/java/org/spongepowered/common/command/manager/SpongeCommandManager.java
```diff
@@ -134,7 +140,11 @@ public CommandMapping registerAlias(
             @NonNull final String primaryAlias,
             @NonNull final String @NonNull ... secondaryAliases)
             throws CommandFailedRegistrationException {
-        final CommandMapping mapping = this.registerAliasInternal(registrar, container, primaryAlias, secondaryAliases);
+        final List<String> aliases = new ArrayList<>();
+        aliases.add(primaryAlias);
+        aliases.addAll(Arrays.asList(secondaryAliases));
```
ImMorpheus: `Collections#addAll` to avoid list creation

src/main/java/org/spongepowered/common/service/permission/SpongePermissions.java
```diff
@@ -65,4 +68,22 @@ public static void populateNonCommandPermissions(final SubjectData data, final B
         }
     }
 
+    public static void registerPermission(final String permissionNode, final int opLevel) {
+        if (!REGISTERED_PERMISSIONS.contains(permissionNode)) {
```
ImMorpheus: if (!contains) add -> if (add)

src/main/java/org/spongepowered/common/command/brigadier/tree/SpongeArgumentCommandNode.java
```diff
@@ -201,4 +206,14 @@ public final void parse(final StringReader reader, final CommandContextBuilder<C
...
 
+    @Override
+    public void addChild(final CommandNode<CommandSource> node) {
+        super.addChild(node);
+        this.childNodes.add(node);
+    }
+
+    @Override
+    public Collection<CommandNode<CommandSource>> getUnsortedChildren() {
+        return this.childNodes;
```
ImMorpheus: This is returning a mutable collection but `SpongeRootCommandNode` is returning an immutable one. Is it intended ?

dualspiral: The intention was to just return the mutable collection for the simple reason that it's only being used in an iteration in one of the mixins to therefore avoid a copy of what could end up being a large collection. I may wrap it in `Collections.unmodifiableCollection(...)` or something.

As for the root collection, probably me in auto-mode.

PR #3062, opened 2020-07-15
Share the PluginLoader instance

jamierocks:


vanilla/src/launch/java/org/spongepowered/vanilla/launch/SpongeLaunch.java
```diff
@@ -0,0 +1,51 @@
...
+import org.spongepowered.plugin.PluginEnvironment;
+import org.spongepowered.vanilla.launch.plugin.PluginLoader;
+
+/**
+ * Shared concepts between Sponge's launching code and the running
+ * game.
+ *
+ * @author Jamie Mansfield
+ */
+public final class SpongeLaunch {
```
Zidane: This is the point of Launcher, both of our discrete platforms must extend Launcher.

Launcher needs to get the two of these passed up to it.

vanilla/src/modlauncher/java/org/spongepowered/vanilla/modlauncher/bootstrap/ClientDevLaunchHandler.java
```diff
@@ -41,7 +41,7 @@ public String name() {
 
     @Override
     protected void launchService0(final String[] arguments, final ITransformingClassLoader launchClassLoader) throws Exception {
-        final PluginEnvironment launchPluginEnvironment = Main.getLaunchPluginEnvironment();
+        final PluginEnvironment launchPluginEnvironment = SpongeLaunch.pluginEnvironment;
```
Zidane: Good example here, practice OOP and pass off that plugin environment from Main to the launch method of Client/ServerLauncher.

PR #3063, opened 2020-07-15
Implement Attributes API (#3036)

i509VCB:
This needs to be merged first of course:
https://github.com/SpongePowered/SpongeAPI/pull/2168


jamierocks (commented):
not enough final

src/mixins/java/org/spongepowered/common/mixin/api/mcp/item/ItemStackMixin_API.java
```diff
@@ -117,6 +128,37 @@ public void setRawData(final DataView container) throws InvalidDataException {
         return (ItemStack) (Object) this.shadow$copy();
     }
 
+    public Collection<AttributeModifier> apiStack$getAttributeModifiers(final AttributeType attributeType, final EquipmentType equipmentType) {
```
gabizou: What's with this method? IS it public for other consumers? is it prefixed because it's an `api` adding method?

gabizou (approved):

PR #3065, opened 2020-07-15
Implement the new network channels.

Cybermaxke:


src/accessors/java/org/spongepowered/common/accessor/network/play/CustomPayloadPacketAccessor.java
```diff
@@ -0,0 +1,39 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.accessor.network.play;
+
+import net.minecraft.network.PacketBuffer;
+import net.minecraft.util.ResourceLocation;
+
+public interface CustomPayloadPacketAccessor {
```
gabizou: Nope. Same deal with the net handler bridge, should be a bridge.

src/main/java/org/spongepowered/common/network/channel/ChannelConstants.java
```diff
@@ -0,0 +1,48 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.network.channel;
+
+import org.spongepowered.api.ResourceKey;
+
+public final class ChannelConstants {
```
gabizou: Move to common's constants class.

src/main/java/org/spongepowered/common/network/channel/RegisterChannelHelper.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * A minecraft channel used to register channels keys.
+     */
+    public static final ResourceKey REGISTER_KEY = ResourceKey.minecraft("register");
+
+    /**
+     * A minecraft channel used to unregister channels keys.
+     */
+    public static final ResourceKey UNREGISTER_KEY = ResourceKey.minecraft("unregister");
```
gabizou: Move to constants

src/main/java/org/spongepowered/common/network/channel/SpongeChannelRegistry.java
```diff
@@ -24,29 +24,298 @@
...
+
+        private final CompletableFuture<Void> future;
+
+        private ChannelRegistrationsResult(final CompletableFuture<Void> future) {
+            this.future = future;
+        }
+    }
+
+    public void postRegistryEvent() {
+        final Cause cause = Cause.of(EventContext.empty(), this);
```
gabizou, line 151: Can use the PhaseTracker as is now. But if this whole class shebang is usable from multiple threads, you may have to change the map used to at least ConcurrentHashMap.

src/main/java/org/spongepowered/common/network/channel/TransactionStore.java
```diff
@@ -0,0 +1,130 @@
...
+    private final AtomicInteger counter = new AtomicInteger();
+    private final ConcurrentMap<Integer, Entry> lookup = Caffeine.newBuilder()
+            .expireAfterAccess(1, TimeUnit.MINUTES)
+            .removalListener((RemovalListener<Integer, Entry>) (key, value, cause) -> {
+                if (cause == RemovalCause.EXPIRED && value != null) {
+                    value.getChannel().handleTransactionResponse(
+                            getConnection(), value.getData(), TransactionResult.failure(new TimeoutException()));
+                }
+            })
+            .build().asMap();
```
gabizou, line 55: Does `asMap` render the cache as invalid? Haven't read the javadoc on caffeine in a moment, but I'd figure it would make a copy as a Map.

src/main/java/org/spongepowered/common/network/channel/packet/AbstractPacketChannel.java
```diff
@@ -0,0 +1,249 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+public abstract class AbstractPacketChannel extends SpongeChannel implements TransactionalPacketRegistry {
+
+    protected final Map<Class<?>, PacketBinding<?>> byType = new HashMap<>();
+    protected final Int2ObjectMap<PacketBinding<?>> byOpcode = new Int2ObjectOpenHashMap<>();
```
gabizou: Concurrent?

gabizou (changes requested):
Questions about thread safety. I know some of this isn't explicitly used asynchronously, but I do worry about maps being used as registers if they are mutable (somewhat why I made the old GameRegistry have a temp mutable and a loaded immutable variant)

src/main/java/org/spongepowered/common/network/channel/SpongeChannelRegistry.java
Zidane: Depends where you call this in the lifecycle.

Part of the lifecycle is before the engine(s) are even created. If so, cannot use PhaseTracker.

Even if you could, don't see a point. We explicitly know what the cause will always be here, no reason to grab one from the CauseStack.

Zidane:
As an overall note Cyber, fully qualify static access please.

src/main/java/org/spongepowered/common/network/channel/TransactionStore.java
Cybermaxke: `asMap` returns a view of the cache

PR #3067, opened 2020-07-15
Replace UserDiscoverer with stripped back ServerUserProvider.

### LINKED ISSUE #3059: Re-visit UserDiscoverer/Implement streamAllProfiles()

dualspiral:
The idea of the UserManager is to manage User objects, not GameProfiles. While we do some operations with GameProfiles, they mirror what Users we think have been created - we're not caching `GameProfile`s.

The file system watching is in, but all GameProfile caching is delegated to the PlayerProfileCache, which we just get profiles from when required.

It needs some testing, but it seems to at least grab the right user file and I can see it stores the UUIDs of the files that exist - which is the point really.

Note that I have not brought over the notion that banned/whitelisted profiles necessarily have a `User` attached - they don't! If you want a users associated with a banned player, get the banned user's profile and create it.

Closes #3059


src/main/java/org/spongepowered/common/user/ServerUserProvider.java
```diff
@@ -0,0 +1,336 @@
...
+    Optional<User> getUser(final String lastKnownName) {
+        final com.mojang.authlib.GameProfile gameProfile = this.server.getPlayerProfileCache().getGameProfileForUsername(lastKnownName);
+        if (gameProfile == null) {
+            return Optional.empty();
+        }
+        return this.getUser((GameProfile) gameProfile, true);
+    }
+
+    Optional<User> getUser(final UUID uuid) {
+        this.pollFilesystemWatcher();
```
ImMorpheus: This method (getUser) is equivalent to `this.getUser(this.server.getPlayerProfileCache().getProfileByUUID(uuid), true);`

PR #3078, opened 2020-07-20
Add more Sponge commands:

dualspiral:
* /sponge which (not witch, as some people were led to believe!)
* /sponge version
* /sponge plugin <plugin>

Also ensure suggestions that are returned get filtered by startsWith

Also also: generics are a bit better now

Requires https://github.com/SpongePowered/SpongeAPI/pull/2181

### DIFF
--- a/src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
+++ b/src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
@@ -26,16 +26,20 @@ package org.spongepowered.common.command.sponge;
 
 import co.aikar.timings.Timings;
 import net.kyori.adventure.text.TextComponent;
+import net.kyori.adventure.text.event.ClickEvent;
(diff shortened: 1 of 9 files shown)


src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -26,83 +26,123 @@
...
 import java.io.File;
+import java.net.URL;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
+import java.util.Collection;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
-public final class SpongeCommand {
+public class SpongeCommand {
```
dualspiral, line 61: @Zidane I know you like final things, but this is intended to be overriden by SpongeForge to add any additional commands it sees fit, like `/sponge mods`, and injected via Guice (so it'll need an explicit binding on SF).

PR #3087, opened 2020-07-22
Add config option to disable neighbor notification to unloaded chunks

### LINKED ISSUE #3188: deny-chunk-requests not fully working
**I am currently running**

- SpongeForge version: spongeforge-1.12.2-2838-7.2.2
- Forge version: forge-1.12.2-14.23.5.2847-universal
- Java version: 1.8

**Issue Description**
The `deny-chunk-requests` doesn't fully work. There are still some tile entities causing the chunks to get loaded without any player or force loaded chunk around it.

This is causing some chunks to be in a load and unload loop. Which can cause frequent lag spikes depending on what's in the chunks. For example AE2 ME Interfaces.

ImMorpheus:
Close https://github.com/SpongePowered/SpongeForge/issues/3188

NB: isChunkAvailable is a lie. It doesn't return what the name imply.

### DIFF
--- a/src/main/java/org/spongepowered/common/bridge/world/WorldServerBridge.java
+++ b/src/main/java/org/spongepowered/common/bridge/world/WorldServerBridge.java
@@ -129,6 +129,8 @@ public interface WorldServerBridge {
 
     long bridge$getChunkUnloadDelay();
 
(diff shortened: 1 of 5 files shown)


dualspiral:
Once this is doing what its supposed to be doing, just merge in, the code itself looks fine and I don't see that changing.

PR #3089, opened 2020-07-24
Migrate Data Providers to a common registration

Grinch:


src/main/java/org/spongepowered/common/data/provider/block/state/AbstractFurnaceData.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    // @formatter:off
+    public static void register(final DataProviderRegistrator registrator) {
+        registrator
+                .asImmutable(BlockState.class)
+                    .create(Keys.DIRECTION)
+                        .get(h -> DirectionUtils.getFor(h.get(AbstractFurnaceBlock.FACING)))
+                        .set((h, v) -> DirectionUtils.set(h, v, AbstractFurnaceBlock.FACING))
+                        .supports(AbstractFurnaceBlock.class::isInstance)
```
gabizou: Is this calling  is instance on the BlockState or the BlockState.getBlock?

src/main/java/org/spongepowered/common/data/provider/block/state/AbstractButtonData.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    // @formatter:off
+    public static void register(final DataProviderRegistrator registrator) {
+        registrator
+                .asImmutable(BlockState.class)
+                    .create(Keys.IS_POWERED)
+                        .get(h -> h.get(AbstractButtonBlock.POWERED))
+                        .set((h, v) -> h.with(AbstractButtonBlock.POWERED, v))
+                        .supports(AbstractButtonBlock.class::isInstance);
```
gabizou: Do buttons also store their attached direction in their state? Or just powered?

Grinch: Appears to just be POWERED

gabizou (approved):

PR #3090, opened 2020-07-24
Reorganize config, make it work in mixin plugins

zml2008:
A bunch of rearrangement to better sort out which options are available
per-world, and which options are global only.

Adds a usable versioning system.

Configuration actually saves.

Per-dimension type configurations are gone

Still TODO (but for a future PR):

- better way to pass base directory info and isVanilla flag to the configuration classes
- figure out the logic in Block/Entity type constructors and make it actually correct
- commands to reload + edit the configuration


src/main/java/org/spongepowered/common/config/common/CommonConfig.java
```diff
@@ -22,41 +22,38 @@
...
+public class CommonConfig implements Config {
+
+    public static final String FILE_NAME = "sponge.conf";
 
-public class GlobalConfig extends GeneralConfigBase {
+    @Setting
+    private GeneralCategory general = new GeneralCategory();
 
-    @Setting(comment = "Configuration options related to the Sql service, including connection aliases etc")
+    @Setting(comment = "Configuration options related to the SQL service, including connection aliases etc")
```
Zidane: SQL manager

src/main/java/org/spongepowered/common/world/SpongeWorldArchetypeBuilder.java
```diff
@@ -266,7 +266,7 @@ public SpongeWorldArchetypeBuilder reset() {
         this.enabled = true;
         this.loadOnStartup = true;
         this.keepSpawnLoaded = null;
-        this.generateSpawnOnLoad = this.dimensionType.getConfigAdapter().getConfig().getWorld().getGenerateSpawnOnLoad();
+        this.generateSpawnOnLoad = true; // TODO this.dimensionType.getConfigAdapter().getConfig().getWorld().getGenerateSpawnOnLoad();
```
Zidane: This is actually a fine change with no TODO needed. DimensionType specifying if I'll generate spawn on load is just a Forge feature. This call will likely be changed to do...

ModWorldHooks.shouldGenerateSpawnOnLoad(DimensionType dimensionType);

zml2008:
rebased

PR #3095, opened 2020-07-27
Use a double for damage related keys.

Cybermaxke:
[API](https://github.com/SpongePowered/SpongeAPI/pull/2174) | Common


dualspiral:
I rebased and pushed it up: 97dfb90

PR #3115, opened 2020-08-16
Api8/custom world

Zidane:


src/main/java/org/spongepowered/common/command/parameter/managed/standard/SpongeWorldPropertiesValueParameter.java
```diff
@@ -60,7 +60,7 @@ public ResourceKey getKey() {
     @Override
     @NonNull
     public List<String> complete(@NonNull final CommandContext context) {
-        return SpongeCommon.getGame().getServer().getWorldManager().getAllProperties().stream().map(WorldProperties::getKey)
+        return SpongeCommon.getGame().getServer().getWorldManager().getAllProperties().stream().filter(properties -> properties.getWorld().isPresent()).map(WorldProperties::getKey)
```
Zidane, line 63: @dualspiral I feel we should discuss this because one can make an argument for this parameter giving you all the properties as much as someone arguing it should be only ones that are "online".

Maybe we need two builtin API parameters? ONLINE and ALL?

dualspiral: I'm glad you asked. 

If we are to have one element, it should get all worlds, unloaded or otherwise. Having two elements does make sense in that regard because there are (obviously) reasons why you might just want to target online worlds.

This then pertains to a wider question - should the location argument only allow for online worlds to be targetted? I'd be inclined to say yes, but it might be that we want to have both here too.

vanilla/src/main/java/org/spongepowered/vanilla/world/VanillaWorldManager.java
```diff
@@ -163,64 +157,295 @@ public ResourceKey getDefaultPropertiesKey() {
...
+
+        this.performPostLoadWorldLogic(world, this.createDefaultSettings(null, false, worldInfo.getSeed(), worldInfo.getGenerator(), null), worldDirectory, chunkStatusListener);
+
+        return CompletableFuture.completedFuture(Optional.of((org.spongepowered.api.world.server.ServerWorld) world));
     }
 
     @Override
-    public CompletableFuture<Optional<org.spongepowered.api.world.server.ServerWorld>> loadWorld(WorldProperties properties) throws IOException {
-        return null;
+    public CompletableFuture<Optional<org.spongepowered.api.world.server.ServerWorld>> loadWorld(WorldProperties properties) {
```
dualspiral: I think someone else said this previously, not sure if it was about this specifically or somewhere else, but in all cases in this method an empty Optional is actually an exceptional condition. So, we could make this return `CompletableFuture<ServerWorld>` and call [`CompletableFuture#completeExceptionally`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#completeExceptionally-java.lang.Throwable-), which would then be caught by plugin developers via the checked exception from `get()` (or if they use `join`, then that's their lookout).

The downside of this is creating the exception (could be expensive due to stack trace generation), and knowing what exception we should be throwing - at least that way a plugin can get more context as to why a world did not load.

src/main/java/org/spongepowered/common/config/common/WorldCategory.java
```diff
@@ -36,7 +36,7 @@
 
     @Setting(value = "auto-player-save-interval", comment = "The auto-save tick interval used when saving global player data. (Default: 900) \n"
```
ImMorpheus:
```suggestion
    @Setting(value = "auto-player-save-interval", comment = "The auto-save tick interval used when saving global player data. (Default: 6000) \n"
```

dualspiral:
This is in mainline now.

PR #3121, opened 2020-08-22
Early configuration

zml2008:
Common configurations now have full access to the `PluginEnvironment`, and are put into an isolated source set to make sure they don't accidentally access game classes.

lots of file moves in here, sorry :(


gabizou (approved):
I'm debating on the package but as long as the sourceSet is good, looks good to me.

PR #3141, opened 2020-08-29
Add string based parsing for states

dualspiral:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2219) | **SpongeCommon**

The initial intention was to remove `CatalogType` from `BlockState` and `FluidState` and use Minecraft's inbuilt `BlockStateParser` to allow us to rebuild strings that represent block states. FluidState has a toned down version of the parser.

There is a bigger issue in that the schematic system heavily relies on the fact that `BlockState` is a catalog type. I've not touched this - I'd rather someone else looked at that.


src/main/java/org/spongepowered/common/fluid/SpongeFluidStateBuilder.java
```diff
@@ -0,0 +1,195 @@
...
+import org.spongepowered.api.data.value.Value;
+import org.spongepowered.api.fluid.FluidState;
+import org.spongepowered.api.fluid.FluidType;
+import org.spongepowered.api.fluid.FluidTypes;
+import org.spongepowered.common.util.Constants;
+
+import java.util.Objects;
+import java.util.Optional;
+
+public class SpongeFluidStateBuilder implements FluidState.Builder {
```
Zidane: You know what is about to happen right?

dualspiral:
I think this is ready to go. The actual matcher logic is at https://github.com/SpongePowered/SpongeCommon/pull/3141/files#diff-23397afc63a4c03b7461df0f8f3db2c0eb3348cc521a89fb16a6728abf817e39R62-R83 - someone please take a look at that and make sure I've got this right before merging.

PR #3142, opened 2020-08-29
Fix/verify overwrites

ImMorpheus:
Verify overwrites


ImMorpheus:
@Zidane this should be ready. There's still an issue with the death message being shown twice but that's gonna be fixed in a separate PR.

Now the overwrites are verified and updated to 1.15.2.

Note: I haven't checked the overwrites in the `invalid` folder,

PR #3149, opened 2020-08-31
Bring back ProjectileLauncher

RedNesto:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2222) | SpongeCommon

This updates everything related to the ProjectileLauncher from invalid.

It is not perfect and needs more work, notably:
- `launchProjectileTo` only works for shulkers as it was only made for them initially. As it has been generalized I'll see how it can be implemented for all entities later
- using a dispenser as source may not work for some projectiles like wither skulls (they seem to spawn right in the block), I also had similar issues with explosive fireballs but I don't feel like looking more into it right now


src/mixins/java/org/spongepowered/common/mixin/api/mcp/entity/LivingEntityMixin_API.java
```diff
@@ -74,4 +81,19 @@ public Component getTeamRepresentation() {
         return values;
     }
 
+    @Override
+    public <T extends Projectile> Optional<T> launchProjectile(final EntityType<T> projectileType) {
+        return ProjectileLauncher.launch(checkNotNull(projectileType, "projectileType"), this, null);
```
Zidane: Objects#requireNotNull

(We're trying to cut down on Guava usage if not eliminate it).

Zidane:
Excellent PR @RedNesto and thanks so much for helping us out in bringing back our features.

src/mixins/java/org/spongepowered/common/mixin/api/mcp/entity/LivingEntityMixin_API.java
RedNesto: Oh right I didn't pay attention to it

Faithcaio:
manually merged

PR #3156, opened 2020-09-09
production manifest generation

zml2008:
dependency json mostly generates correctly

the accesstransformers lib is weird, don't know if we're using the correct artifact for it here.

also made the installer download/verify deps in parallel -- it's not super pretty, but makes launch a good bit faster


buildSrc/build.gradle.kts
```diff
@@ -6,11 +6,13 @@ plugins {
 repositories {
     gradlePluginPortal()
     maven("https://repo-new.spongepowered.org/repository/maven-public/")
+    mavenCentral()
```
gabizou: don't need this since repo-new proxies central

build.gradle.kts
```diff
@@ -763,44 +771,89 @@ project("SpongeVanilla") {
...
+                    // Get file input stream for reading the file content
+                    val md5hash = dependency.file.inputStream().use {
+                        DigestUtils.md5Hex(it)
+                    }
+
+                    // create descriptor
+                    DependencyDescriptor(
+                            group = ident.group,
+                            module = ident.name,
+                            version = version, // TODO: can we get timestamped snapshot versions?
```
gabizou: don't need the timestamped snapshot versions, the sonatype search uses the maven.version specifically, not the timestamp'ed snapshot.

gabizou (approved):

PR #3160, opened 2020-09-16
Another Round of Data (Impl Edition)

Grinch:
Relies on SpongePowered/SpongeAPI#2231


gabizou (approved):
Looks good, got a compilation issue with an import it seems but that's minor.

Grinch:
> 
> 
> Looks good, got a compilation issue with an import it seems but that's minor.

I believe that is because my PR doesn't include the changes for API.

PR #3163, opened 2020-09-20
Implement handling for Adventure's translation registry

zml2008:
This PR, when complete, will process any text sent to render localization keys provided on the server using Adventure's global TranslationRegistry.

I'm unsure about the placement of the client-only `TranslationText` mixin -- It's necessary for localization to happen on the client, and has to be client-only due to its references to the client's options, but it breaks the package structure a bit.


kashike (approved):
lgtm after initial review

Issue #2513, opened 2020-02-23
No hooks for Forge generic.reachDistance AttributeModifier in NetHandlerPlayServer.processUseEntity


PolyacovYury:
**I am currently running**
- SpongeForge version: 1.12.2-2838-7.1.10-RC3997
- Java version: 1.8

A recent need brought me into the depths of the debugger again.
I were trying to make a whatever-compatible fix for entity interactions not actually taking any note of player's `generic.reachDistance` attribute.
There was a (sadly staled and closed) PR into Forge that fixed this issue, and I was looking into it, as the one present in PotionCore completely broke GriefDefender's interact-entity protection module and needed to be disabled.
I essentially re-implemented the fix, but instead of trying to come up with a new value for d0 variable in `NetHandlerPlayServer.processUseEntity()`, I came up with a function that gets player's reach distance, calculates the distance between player's eye position and entity's bounding box and returns true if either the eye pos is inside the bbox or if the distance to the entity is less or equal to player's reach distance.
Running the code in a clean Forge environment had the desired effect - when wearing a diamond chestplate with a `generic.reachDistance` AttributeModifier of 10, I could break blocks and interact with entities 15 blocks away. But on the server the game seemed to not care.
I took to the debugger and discovered this gem.
```java
@Mixin(NetHandlerPlayServer.class)
public abstract class NetHandlerPlayServerMixin implements NetHandlerPlayServerBridge {
    @Overwrite
    public void processUseEntity(final CPacketUseEntity packetIn) {
    ...
    }
}
```
So, my changes to Forge code are completely ignored by a server if it is run with SpongeForge.
That's why I'm here.

```java
        final WorldServer worldserver = this.server.getWorld(this.player.dimension);
        final Entity entity = packetIn.getEntityFromWorld(worldserver);
        this.player.markPlayerActive();

        if (entity != null) {
            final boolean flag = this.player.canEntityBeSeen(entity);
            double d0 = 36.0D; // 6 blocks

            if (!flag) {
                d0 = 9.0D; // 1.5 blocks
            }

            if (this.player.getDistanceSq(entity) < d0) {
                // do the rest
```
I mean, really? A player can possibly ever reach an entity if it is 6 blocks away from him, or even 3 blocks, if they don't have a direct line of sight between them?

AstralSorcery has an ASM transformer designed to mitigate this issue (by setting the d0 value to Double.fromInt(Int.MAX_VALUE)), but that transformation has no effect due to the overwrite Mixin above.

[...]

PR #3166, opened 2020-09-21
Add SpongeImplHook for entity reach distance (#2513)

PolyacovYury:
What title says.

Also fixes [MC-107103](https://bugs.mojang.com/browse/MC-107103) by removing eye sight check in interaction packet handler.

### DIFF
--- a/src/main/java/org/spongepowered/common/SpongeImplHooks.java
+++ b/src/main/java/org/spongepowered/common/SpongeImplHooks.java
@@ -176,6 +176,13 @@ public final class SpongeImplHooks {
         return 5.0d;
     }
 
+    /**
+     * @author Polyacov_Yury
+     * @reason Forge reachDistance attribute compatibility
+     * @param player the player whose reach is being checked
+     * @param entity the entity that is being reached
+     * @return square of maximum player reach distance
+     */
     public static double getEntityReachDistanceSq(final EntityPlayerMP player, Entity entity) {
         double d0 = 36.0d; // 6 blocks
         if (!player.canEntityBeSeen(entity)) {  // TODO: this check introduces MC-107103
--- a/src/main/java/org/spongepowered/common/SpongeImplHooks.java
+++ b/src/main/java/org/spongepowered/common/SpongeImplHooks.java
@@ -176,6 +176,10 @@ public final class SpongeImplHooks {
         return 5.0d;
     }
 
+    public static double getEntityReachDistanceSq(final EntityPlayerMP player) {
+        return 36.0d; // 6 blocks
+    }
+
     // Entity registry
 
     @Nullable
--- a/src/main/java/org/spongepowered/common/SpongeImplHooks.java
+++ b/src/main/java/org/spongepowered/common/SpongeImplHooks.java
@@ -176,8 +176,12 @@ public final class SpongeImplHooks {
         return 5.0d;
     }
 
-    public static double getEntityReachDistanceSq(final EntityPlayerMP player) {
-        return 36.0d; // 6 blocks
+    public static double getEntityReachDistanceSq(final EntityPlayerMP player, Entity entity) {
+        double d0 = 36.0d; // 6 blocks
+        if (!player.canEntityBeSeen(entity)) {  // TODO: this check introduces MC-107103
+            d0 = 9.0D; // 3 blocks
+        }
+        return d0;
     }
 
     // Entity registry
--- a/src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
@@ -811,12 +811,16 @@ public abstract class NetHandlerPlayServerMixin implements NetHandlerPlayServerB
         this.player.markPlayerActive();
 
         if (entity != null) {
-            final boolean flag = this.player.canEntityBeSeen(entity);
-            double d0 = 36.0D; // 6 blocks
+            // Sponge start - flag is not necessary, see below
+            // final boolean flag = this.player.canEntityBeSeen(entity);
+            // compatibility with Forge reachDistance attribute
+            double d0 = SpongeImplHooks.getEntityReachDistanceSq(this.player);
 
-            if (!flag) {
-                d0 = 9.0D; // 1.5 blocks
-            }
+            // this check introduces MC-107103
+            // if (!flag) {
+            //     d0 = 9.0D; // 3 blocks
+            // }
+            // Sponge end
 
             if (this.player.getDistanceSq(entity) < d0) {
                 // Sponge start - Ignore CPacketUseEntity.Action.INTERACT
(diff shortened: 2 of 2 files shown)


src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
```diff
@@ -811,12 +811,16 @@ public void processUseEntity(final CPacketUseEntity packetIn) {
...
-
-            if (!flag) {
-                d0 = 9.0D; // 1.5 blocks
-            }
+            // Sponge start - flag is not necessary, see below
+            // final boolean flag = this.player.canEntityBeSeen(entity);
+            // compatibility with Forge reachDistance attribute
+            double d0 = SpongeImplHooks.getEntityReachDistanceSq(this.player);
+
+            // this check introduces MC-107103
```
dualspiral: I'm actually not really sure about "adding (removing)" this straight up. It feels to me that this was intended to work as a way to ensure "line of sight" for an interaction. The removal of this check means that I could, in theory, reach though a few blocks to interact with a switch on the other side, when I don't think that's intended.

I don't think this "fix" is correct and that the real bug is client side. I can see hack clients possibly making use of this to activate things though walls and such.

PolyacovYury: This line of sight check is only applied to interactions with entities, not blocks.
If you look at the bug description on Mojang's tracker, they talk about this exact if-statement.

Imagine this scenario. You are using a mob farm, where you 1) need to click at mobs' legs, while mobs are standing above you by 1 block, 2) there's a block that blocks line of sight between you and mobs' eyes, 3) you need to stand back 4 blocks (for one reason or another).
What happens is when you do the click - interaction only happens client-side. E.g. if you name a mob in that situation - you'll get a "ghost empty slot" in place of the name tag and a mob that shows a name but doesn't have one.

Again, this check only works on entities, as blocks are handled in a separate method.

dualspiral: It's the same principle though. Should this line of sight check be removed for these entity checks? I'm not convinced that it's a server bug, but a client one - the client is what is causing the ghosting.

src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
```diff
@@ -811,12 +811,16 @@ public void processUseEntity(final CPacketUseEntity packetIn) {
         this.player.markPlayerActive();
 
         if (entity != null) {
-            final boolean flag = this.player.canEntityBeSeen(entity);
-            double d0 = 36.0D; // 6 blocks
-
-            if (!flag) {
```
dualspiral, line 817: So I went ahead and looked at the [Forge patch for this class](https://github.com/MinecraftForge/MinecraftForge/blob/1.12.x/patches/minecraft/net/minecraft/network/NetHandlerPlayServer.java.patch) and this is _not_ one of the changes they make, `d0 = 36.0D` here is left alone. Why should we change the behaviour away from Forge here?

src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
PolyacovYury: Well, the only thing line of sight check does is it halves the distance the player can reach that entity from.
And remember, it doesn't check if player sees entity's hitbox. It checks if player sees entity's eye point.

src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
PolyacovYury: Because there was this PR - https://github.com/MinecraftForge/MinecraftForge/pull/4736/ - that was headed at exactly this issue. Without the change the player sees that they can reach that entity, they swing the sword and.. nothing happens.
That PR wasn't even properly discussed - just staled and closed due to inactivity.

src/main/java/org/spongepowered/common/mixin/core/network/NetHandlerPlayServerMixin.java
dualspiral: I know what it does.

My point is that the bug seems to be client side to me (the client ignores the fact this check exists) rather than this check exists in the first place. If we were to "fix" this, it really should be checking to see if we can see the hitbox rather than removing the check altogether, the check is clearly there for a reason. This is why I'm not sure we should accept this.

PolyacovYury: Well, I can only pull out one last argument: actual Minecraft sources for 1.16.
```java
   public void processUseEntity(CUseEntityPacket packetIn) {
      PacketThreadUtil.checkThreadAndEnqueue(packetIn, this, this.player.getServerWorld());
      ServerWorld serverworld = this.player.getServerWorld();
      Entity entity = packetIn.getEntityFromWorld(serverworld);
      this.player.markPlayerActive();
      this.player.setSneaking(packetIn.func_241792_e_());
      if (entity != null) {
         double d0 = 36.0D;
         if (this.player.getDistanceSq(entity) < 36.0D) {
            // do the rest
```
That `if` statement was so gone that they even forgot to remove the d0 variable. :) 
I guess, if we are worrying about hacked clients, there could be some sort of RayTrace check, but the if-removal at least won't make Sponge less secure than Vanilla is.

dualspiral:
I spoke to people within the Sponge team about this:

* The MC eye sight check is out of scope, at least for this PR. _If_ we were to accept it (and we may not anyway), it needs to be a separate PR and I would insist on it being configurable, defaulting to original behaviour. Please remove it from this PR.
* For the reach distance - I generally dislike moving away from what Forge does but It has been expressed that this is valid. I'd much rather see this fixed in Forge first, but I am aware that Forge is also EOL for 1.12. My concern, [as was also noted on the Forge PR](https://github.com/MinecraftForge/MinecraftForge/pull/4736#issuecomment-460073937) is that there may be mods (for some reason) relying on this. The next comment does downplay that and I am inclined to agree that it's not widespread.

  I'd be inclined to say that it should be configurable again, however my suspicion is that it probably won't affect much, if anything. So let's just put it in and if we have hordes of server owners coming after us, we can add a configuration check in there.

PolyacovYury:
My last reply before doing as I'm told above:  
In AstralSorcery, there are a bunch of skills that boost player's reach distance using the attributeModifier. Mod author might have noticed the problem, as was written in "next comment" in the PR, so they added an ASM transformer that replaces the value of `36.0D`/`9.0D` with `Int.MAX_VALUE`. Not exactly ideal, might've been better off replacing them with the attribute lookup that I suggested, but it sure worked :)

src/main/java/org/spongepowered/common/SpongeImplHooks.java
```diff
@@ -176,6 +176,14 @@ public static double getBlockReachDistance(final EntityPlayerMP player) {
         return 5.0d;
     }
 
+    public static double getEntityReachDistanceSq(final EntityPlayerMP player, Entity entity) {
```
gabizou, line 186: Can you just javadoc the method for the reasoning why it's a spongeimplhook (aside from the place that it's used) since in a SpongeVanilla workspace, this won't make much sense as to why it exists if not for SpongeForge ;)

PolyacovYury: Oh. Will do, I didn't see any javadocs for other ImplHooks and decided that it's fine :)

gabizou (approved):

PR #3167, opened 2020-09-21
Add max-chunk-lifetime setting used by chunk gc

PolyacovYury:
- mostly useful for those who need to disable full auto-saves controlled by auto-save-interval
- spreads out server load induced by chunk saving
- allows to periodically save chunks kept active by players for a long time

As I said in Discord:

If I only use chunk GC and set `auto-save-interval` to 0, then if a player stays afk in the same place for hours - chunks around them would never be saved, and if the server crashes - everything will be rolled back.
The reason why I want to turn auto-save off is - when it happens, the server completely stops processing anything for up to 10 seconds, since it usually needs to save up to 5k chunks, sometimes even up to 10k.
```
        # The auto-save tick interval used to save all loaded chunks in a world.
        # Set to 0 to disable. (Default: 900)
        # Note: 20 ticks is equivalent to 1 second.
        auto-save-interval=12000
```
```
[08:20:26] [Server thread/WARN] [minecraft/MinecraftServer]: Can't keep up! Did the system time change, or is the server overloaded? Running 4570ms behind, skipping 91 tick(s) 
[08:30:30] [Server thread/WARN] [minecraft/MinecraftServer]: Can't keep up! Did the system time change, or is the server overloaded? Running 6581ms behind, skipping 131 tick(s) 
```
I'd like to propose a setting for chunk GC. Something like `max-chunk-lifetime`. So, if a chunk has been loaded for e.g. 5 minutes, it gets saved to disk by the chunk GC code, even if it is still queried by a player.
This way the chunks wouldn't be saved all at once, but they will still get saved regularly.

### DIFF
--- a/src/main/java/org/spongepowered/common/bridge/world/chunk/ChunkBridge.java
+++ b/src/main/java/org/spongepowered/common/bridge/world/chunk/ChunkBridge.java
@@ -106,6 +106,10 @@ public interface ChunkBridge {
 
     boolean bridge$isQueuedForUnload();
 
+    long bridge$getLastSaveTime();
+
+    boolean bridge$getChunkDirty();
+
     void bridge$markChunkDirty();
 
     boolean bridge$isActive();
--- a/src/main/java/org/spongepowered/common/bridge/world/chunk/ChunkBridge.java
+++ b/src/main/java/org/spongepowered/common/bridge/world/chunk/ChunkBridge.java
@@ -108,7 +108,7 @@ public interface ChunkBridge {
 
     long bridge$getLastSaveTime();
 
-    boolean bridge$getChunkDirty();
+    boolean bridge$isChunkDirty();
 
     void bridge$markChunkDirty();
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/common/bridge/world/chunk/ChunkBridge.java
```diff
@@ -106,6 +106,10 @@
 
     boolean bridge$isQueuedForUnload();
 
+    long bridge$getLastSaveTime();
+
+    boolean bridge$getChunkDirty();
```
Zidane: bridge$isChunkDirty

src/main/java/org/spongepowered/common/mixin/core/world/chunk/ChunkMixin.java
```diff
@@ -155,11 +156,23 @@
         this.dirty = true;
     }
 
+    @Override
+    public boolean bridge$getChunkDirty()
+    {
```
Zidane: Formatting

PR #3170, opened 2020-10-02
API8 Advancements

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2238) | **SpongeCommon**


src/main/java/org/spongepowered/common/advancement/SpongeAdvancementProvider.java
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.commons.io.FileUtils;
+import org.spongepowered.api.advancement.Advancement;
+
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class SpongeAdvancementProvider {
```
Zidane: If there is Spooktober, is there Fintober?

PR #3178, opened 2020-10-05
Generify Palettes Implementation

gabizou:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2244) | **SpongeCommon**

This will bring the implementation of Palettes to be more inline with their mutability/immutability aspects, though I still have some questions to answer:
- [ ] How do we want to handle a `Palette` being built by referring to a Minecraft owned Palette (like the case of BlockStates)
- [ ] How do we want to make a `Palette` based on a Minecraft registry?
- [ ] Should we always make immutable palettes when presented with a registry for the string serialization handling?

More to come as it gets fleshed out.

### DIFF
--- a/src/main/java/org/spongepowered/common/world/schematic/ImmutableBimapPalette.java
+++ b/src/main/java/org/spongepowered/common/world/schematic/ImmutableBimapPalette.java
@@ -25,36 +25,35 @@
 package org.spongepowered.common.world.schematic;
 
 import com.google.common.collect.BiMap;
-import com.google.common.collect.HashBiMap;
-import org.spongepowered.api.CatalogType;
+import com.google.common.collect.ImmutableBiMap;
 import org.spongepowered.api.world.schematic.Palette;
 import org.spongepowered.api.world.schematic.PaletteType;
 
-import java.util.BitSet;
-import java.util.Collection;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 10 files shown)


src/main/java/org/spongepowered/common/world/schematic/ImmutableBimapPalette.java
```diff
@@ -0,0 +1,107 @@
...
+
+public class ImmutableBimapPalette<T> implements Palette.Immutable<T> {
+
+    private final ImmutableBiMap<Integer, T> ids;
+    private final ImmutableBiMap<T, Integer> idsr;
+    private final PaletteType<T> paletteType;
+    private final int maxId;
+
+    public ImmutableBimapPalette(final PaletteType<T> paletteType, final BiMap<Integer, T> reference) {
+        final ImmutableBiMap.Builder<Integer, T> builder = ImmutableBiMap.builder();
```
Cybermaxke, line 45: Why not use an `Int2ObjectMap` and `Object2IntMap` if you're going to make a copy?

kashike: :+1: for using fastutil here

Zidane: Agreed

PR #3173, opened 2020-10-03
Bump Mixin to 0.8.1 with AP fixes. Bump Modlauncher

gabizou:
**Some AP issues exist from the changes in Mixin 0.8.1 DO NOT MERGE YET**

This is a preliminary branch PR to test the build with Mixin 0.8.1

Unfortunately, the jar does not run successfully in production due to a new AP issue related to the [Mixin commit](https://github.com/SpongePowered/Mixin/commit/943894557ebb193e5f81dc772656e62183b7261d) that fixed the AP resolving hierarchical methods on targets.

PR #3180, opened 2020-10-08
[API-8] Bump modlauncher and mixin

ItsDoot:
Bumped:
* `modlauncher` from `4.1.x` to `7.0.x`
* `mixin` from `0.8.1-SNAPSHOT` to `0.8.2`

Also bumped these because Modlauncher bumped them:
* `jopt-simple` from `5.0.3` to `5.0.4`
* `grossjava9hacks` from `1.1.x` to `1.3.x`

It doesn't seem to matter that Modlauncher depends on ASM v7 while Mixin depends on v6, because mixins are still correctly applied. Also no other work needs to be done on our part for applaunch, everything still compiles with this update.

I've also gone ahead and checked that one of my plugins works without the `ServiceLoader` workaround, so `ServiceLoader`s work now as well (also plugin lang services shouldn't be blocked by that now either).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -549,7 +549,7 @@ project("SpongeVanilla") {
         add(vanillaLaunch.implementationConfigurationName, vanillaMinecraftConfig)
 
         vanillaInstallerConfig("com.google.code.gson:gson:2.8.0")
-        vanillaInstallerConfig("net.sf.jopt-simple:jopt-simple:5.0.3")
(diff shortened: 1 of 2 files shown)


zml2008:
were the AP issues in <https://github.com/SpongePowered/SpongeCommon/pull/3173> resolved?

Zidane:
This cannot be bumped unless production is confirmed to be working.

ItsDoot:
Proof of production jars working with modlauncher 7 and mixin 0.8.2: https://gist.github.com/ItsDoot/839fc05611a66b053b9d66fe3d53504e

PR #3181, opened 2020-10-09
Allow GitHub Actions to archive built artifacts for builds

gabizou:
This is more of an interim (but maybe as well not) solution to see about archiving the builds for each job of PR's as well as pushes to the repo in the `api-8` branch.

I'm convinced that we can take advantage of having the jars available not just on sonatype, but as well on individual builds (which is pretty nifty to be able to download the jar for a particular PR instead of having to clone and build the code yourself)

Pinging @Zidane and @dualspiral to get some thoughts on it.

Here's the proof of concept:
Screen Shot 2020-10-09 at 2 12 59 AM
Effectively can be viewed, downloaded, and ran with the understanding that the job specifically builds the code from the PR only.

### DIFF
--- a/.github/workflows/build-project.yaml
+++ b/.github/workflows/build-project.yaml
@@ -34,6 +34,15 @@ jobs:
         run: chmod +x gradlew
       - name: Build with Gradle
         run: ./gradlew -PenableTestPlugins=true -PenableSpongeForge=true build
+      - name: Archive artifacts for build
+        uses: actions/upload-artifact@v2
+        with:
+          name: Sponge Jars
+          path: |
+            ${{ github.workspace }}/SpongeAPI/build/libs/*.jar
+            ${{ github.workspace }}/build/libs/*.jar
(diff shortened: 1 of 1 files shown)


dualspiral:
I see no issue with this.

I'm more interested in letting it build for PRs - something that I can just download and test quickly if I've got five minutes and no time to open an IDE, clone, build, yada yada. That gets a big tick from me.

I'm less bothered about this for mainline builds as if they're in a `api-x` branch we consider it "stable-ish" and so people should get it from our downloads pages, but I wouldn't go out of my way to disable that function.

[Do note that the artefacts are deleted after 90 days](https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration#artifact-and-log-retention-policy), so this should not be considered an alternative to Sonatype, but a handy "here is a recent build from this PR/code change".

dualspiral (approved):
LGTM

PR #3187, opened 2020-10-13
Make SystemSubject ICommandSource, add SystemSubject to cause when a Server command source is used.

dualspiral:
If you execute a command via the console, in vanilla miencraft, it is executed by `MinecraftServer`, in the API, `Server`. `Server` implements `ForwardingAudience`, which sends a message to everyone on the server, which means that if a plugin sends a message to the server when it executes a command, everyone will know about it. This is obviously a problem.

There are three potential solutions:

* In API 7/1.12.2, `MinecraftServer` implemented the `ConsoleSource`, and `sendMessage` just sent a message to the server console, not the entire server. Doing this is perhaps the simplest way to do it and will have the fewest surprises for mods, however, this means that `Server` would not be able to implement `ForwardingAudience`, @Faithcaio suggested creating a getter to do this instead if we go this way.
* Alternatively, this PR forces the system subject onto the stack if the Server is executing a command, but this is special cased. This is an idea, and I'm really not sure I want to do this.
* There was a suggestion that the `CommandSource` could implement `Audience`, but the notion of a `CommandSource` isn't exposed to the system, only the `CommandCause`. I _guess_ we could make `CommandCause` implement `Audience` and make it use `sendMessage(ITextCompoenent)` on `CommandSource`. I might do this, but that means that `CommandCause#getAudience()` would be different and the point of the `getAudience()` method was to get the audience from the context or the cause stack.

I'm really leaning towards wanting to do point 1, but that breaks the idea behind `sendMessage()` on `Server`, so unless we move the `ForwardingAudience` to getters on `Server` and `World`, we can't do that. Point 2 is this what is here... I just don't like it.

Open for discussion. All options are on the table.

### DIFF
--- a/src/main/java/org/spongepowered/common/SpongeGame.java
+++ b/src/main/java/org/spongepowered/common/SpongeGame.java
@@ -72,7 +72,7 @@ public final class SpongeGame implements Game {
     private Client client;
     private Server server;
 
-    private SystemSubject systemSubject;
+    private ServerConsoleSystemSubject systemSubject;
 
     @Inject
     public SpongeGame(final Platform platform, final GameRegistry registry, final DataManager dataManager, final PluginManager pluginManager,
@@ -100,7 +100,7 @@ public final class SpongeGame implements Game {
     }
 
     @Override
-    public SystemSubject getSystemSubject() {
+    public ServerConsoleSystemSubject getSystemSubject() {
         if (this.systemSubject == null) {
             this.systemSubject = new ServerConsoleSystemSubject();
         }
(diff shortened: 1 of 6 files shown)


dualspiral:
https://github.com/SpongePowered/SpongeCommon/commit/d214fe66ec0b711580ec76457b6ae1534b5e97c9

I went for option 2 (what was this PR), however I'm still slightly uneasy at it. This commit basically suggests that any server executed commands is done by the system subject instead. I'd prefer that the server was considered the system subject itself, but with the current setup, we can't do that.

PR #3189, opened 2020-10-14
API8 Commands: Optional parameters fixes

Faithcaio:
Fixes missing non-terminal optional/defaulted nodes (because the SpongeCommandContextBuilder cloning ignored previous nodes)
Fixes completion suggesting base-commands
Attempting to fix optional Sequence parameters.


dualspiral:
What are we missing at this point, so I can get a clearer idea of what I might need to do vs. what might go here to get things working?

src/main/java/org/spongepowered/common/command/brigadier/context/SpongeCommandContextBuilder.java
```diff
@@ -93,6 +93,7 @@ public SpongeCommandContextBuilder(final SpongeCommandContextBuilder original, f
         this.modifier = original.modifier;
         this.forks = original.forks;
         this.currentTargetCommand = original.currentTargetCommand;
+        this.getNodes().addAll(original.getNodes());
```
dualspiral, line 96: This makes me feel a little stupid, forgetting to do this!

dualspiral (approved):

PR #3191, opened 2020-10-15
Fix #3190: fix mob breeding cooldown

### LINKED ISSUE #3190: Animals breeding has no cooldown
**I am currently running**

- SpongeForge version: 1.12.2-2838-7.2.4-RC4072

**Issue Description**
https://github.com/SpongePowered/SpongeCommon/commit/104fb62e1440e2e81f791d6d4f290c64123eb4ba
Before this commit, the code took into account the fact that age can be positive (which it is after breeding - it tracks breeding cooldown).
After that, anything bigger than 0 gets `Math.min()`'d. So animals can re-enter "love mode" **immediately** after a baby spawns.

PolyacovYury:
Fixes issue with unrestricted animal breeding introduced in https://github.com/SpongePowered/SpongeCommon/commit/104fb62e1440e2e81f791d6d4f290c64123eb4ba by correctly decreasing the absolute value of new entity's age.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/realtime/entity/EntityAgeableMixin_RealTime.java
+++ b/src/main/java/org/spongepowered/common/mixin/realtime/entity/EntityAgeableMixin_RealTime.java
@@ -45,10 +45,13 @@ public abstract class EntityAgeableMixin_RealTime extends EntityMixin_RealTime {
         }
         // Subtract the one the original update method added
         final int diff = (int) ((RealTimeTrackingBridge) this.world).realTimeBridge$getRealTimeTicks() - 1;
-        // age is positive ? subtract : add
-        final int new_age = age - (int) Math.signum(age) * diff;
-        // if sign changed - return 0
-        this.setGrowingAge((age ^ new_age) < 0 ? 0 : new_age);
+        if (diff == 0) {
+            this.setGrowingAge(age);
+        } else if (age > 0) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/realtime/entity/EntityAgeableMixin_RealTime.java
```diff
@@ -45,7 +45,10 @@
         }
         // Subtract the one the original update method added
         final int diff = (int) ((RealTimeTrackingBridge) this.world).realTimeBridge$getRealTimeTicks() - 1;
-        this.setGrowingAge(Math.min(0, age + diff));
+        // age is positive ? subtract : add
+        final int new_age = age - (int) Math.signum(age) * diff;
```
dualspiral: Rather than do this which is confusing to read, I'd much rather make this a simple if statement - something like:

```java
if (age > 0) {
    this.setGrowingAge(Math.max(0, age - diff));
} else {
    this.setGrowingAge(Math.min(0, age + diff));
}
```

That way it's a lot clearer about what's happening. Could also have a branch that bypasses these things completely if `diff = 0`

PolyacovYury: Agreed. Didn't think of that for some reason :D

dualspiral (approved):
Thanks!

PR #3193, opened 2020-10-18
fixed an incompatible type assignment

abdullahcevik:


dualspiral:
Why do our gradle scripts need a change like this? They work fine as is. This, in fact, _breaks_ the build.

There is no rationale or need for this PR.

PR #3194, opened 2020-10-30
Fix case-sensitivity bug in SpongeUserStorageService#match(String)

alexstaeding:
The current implementation of `SpongeUserStorageService#match(String)` converts the provided `lastKnownName` to lowercase but does not convert the `GameProfile` name to lowercase before comparing it. This means that `GameProfile` names with uppercase characters will never match the lowercase `lastKnownName`.

### DIFF
--- a/src/main/java/org/spongepowered/common/service/user/SpongeUserStorageService.java
+++ b/src/main/java/org/spongepowered/common/service/user/SpongeUserStorageService.java
@@ -113,7 +113,8 @@ public class SpongeUserStorageService implements UserStorageService {
         Collection<GameProfile> allProfiles = UserDiscoverer.getAllProfiles();
         Collection<GameProfile> matching = Sets.newHashSet();
         for (GameProfile profile : allProfiles) {
-            if (profile.getName().isPresent() && profile.getName().get().startsWith(lastKnownName)) {
+            Optional<String> name = profile.getName();
+            if (name.isPresent() && name.get().toLowerCase(Locale.ROOT).startsWith(lastKnownName)) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/service/user/SpongeUserStorageService.java
```diff
@@ -113,7 +113,8 @@ public boolean delete(User user) {
         Collection<GameProfile> allProfiles = UserDiscoverer.getAllProfiles();
         Collection<GameProfile> matching = Sets.newHashSet();
         for (GameProfile profile : allProfiles) {
-            if (profile.getName().isPresent() && profile.getName().get().startsWith(lastKnownName)) {
+            Optional<String> name = profile.getName().map(n -> n.toLowerCase(Locale.ROOT));
+            if (name.isPresent() && name.get().startsWith(lastKnownName)) {
```
zml2008: You're changing this from immediately breaking out of optional-world to being half optional-style and half not.

Either do something like `name.get().toLowerCase(Locale.ROOT).startsWith(lastKnownName)`, or continue the Optional chain so it terminates with an `orElse(false)`, removing any use of `isPresent` or `get`.

alexstaeding: Initially I was going to continue with `.filter(n -> n.startsWith(lastKnownName))` but variables in the lambda need to be effectively final. I'll just change it to the first option you suggested.

zml2008 (approved):

PR #3205, opened 2020-11-19
Add missing scoreboard criteria

thibaulthenry:
Hey,

Here is the fix for SpongeAPI issue : https://github.com/SpongePowered/SpongeAPI/issues/2187
Linked PR on SpongeAPI: https://github.com/SpongePowered/SpongeAPI/pull/2258

Thank you for reading this PR.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/api/mcp/scoreboard/CriterionMixin_API.java
+++ b/src/main/java/org/spongepowered/common/mixin/api/mcp/scoreboard/CriterionMixin_API.java
@@ -28,19 +28,40 @@ import com.google.common.base.CaseFormat;
 import net.minecraft.scoreboard.IScoreCriteria;
 import net.minecraft.scoreboard.ScoreCriteria;
 import net.minecraft.scoreboard.ScoreCriteriaColored;
+import net.minecraft.util.text.TextFormatting;
 import org.spongepowered.api.scoreboard.critieria.Criterion;
+import org.spongepowered.api.text.format.TextColor;
 import org.spongepowered.asm.mixin.Implements;
 import org.spongepowered.asm.mixin.Interface;
 import org.spongepowered.asm.mixin.Intrinsic;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.Surrogate;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import org.spongepowered.common.registry.type.text.TextColorRegistryModule;
 
 import javax.annotation.Nullable;
+import java.util.Optional;
 
 @Mixin(value = {ScoreCriteriaColored.class, ScoreCriteria.class})
 @Implements(@Interface(iface = Criterion.class, prefix = "criterion$"))
 public abstract class CriterionMixin_API implements IScoreCriteria { // Trick to allow avoid shadowing, since multiple targets are used
 
-    @Nullable private String spongeId;
+    @Nullable
+    private String spongeId;
+    @Nullable
+    private TextFormatting format;
+
+    @Inject(method = "<init>", at = @At(value = "RETURN"))
+    @SuppressWarnings("InvalidInjectorMethodSignature")
+    private void onConstructed(String name, CallbackInfo ci) {
+    }
+
+    @Surrogate
+    private void onConstructed(String name, TextFormatting format, CallbackInfo ci) {
+        this.format = format;
+    }
 
     @Intrinsic
     public String criterion$getName() {
@@ -52,5 +73,9 @@ public abstract class CriterionMixin_API implements IScoreCriteria { // Trick to
             this.spongeId = "minecraft:" + CaseFormat.LOWER_CAMEL.to(CaseFormat.LOWER_UNDERSCORE, this.getName().replace("count", "s"));
         }
         return this.spongeId;
+    }
+
+    public Optional<TextColor> getTeamColor() {
+        return Optional.ofNullable(TextColorRegistryModule.enumChatColor.get(this.format));
     }
 }
--- a/src/main/java/org/spongepowered/common/registry/type/scoreboard/CriteriaRegistryModule.java
+++ b/src/main/java/org/spongepowered/common/registry/type/scoreboard/CriteriaRegistryModule.java
@@ -24,7 +24,6 @@
  */
 package org.spongepowered.common.registry.type.scoreboard;
 
-import com.google.common.collect.Maps;
 import net.minecraft.scoreboard.IScoreCriteria;
 import net.minecraft.util.text.TextFormatting;
 import org.spongepowered.api.registry.util.RegisterCatalog;
@@ -35,6 +34,7 @@ import org.spongepowered.common.registry.type.AbstractPrefixAlternateCatalogType
 import org.spongepowered.common.registry.type.text.TextColorRegistryModule;
 import org.spongepowered.common.text.format.SpongeTextColor;
 
+import java.util.HashMap;
 import java.util.Map;
 
 @RegistrationDependency(TextColorRegistryModule.class)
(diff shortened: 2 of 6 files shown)


src/main/java/org/spongepowered/common/registry/type/scoreboard/CriteriaRegistryModule.java
```diff
@@ -24,17 +24,32 @@
...
 @RegisterCatalog(Criteria.class)
 public final class CriteriaRegistryModule extends AbstractPrefixAlternateCatalogTypeRegistryModule<Criterion> {
 
-    public CriteriaRegistryModule() {
-        super("minecraft", new String[] {"minecraft"}, id -> id.replace("_count", "s"));
+    public static CriteriaRegistryModule getInstance() {
+        return CriteriaRegistryModule.Holder.INSTANCE;
+    }
+
+    public final Map<String, Criterion> teamKillMappings = Maps.newLinkedHashMap();
```
dualspiral: * I realise you're copying `DisplaySlotRegistryModule` here but I don't like this being public - I'd prefer getters on this module that do the lookup (but I'll let someone else overrule me if we're okay with leaving it like this).
* Why are these linked hash maps? I don't think these are iterated over and the order isn't important, so these should be standard hash maps.
* Use `new HashMap<>()` instead of the `Maps.new<blah>` methods, we're trying to get away from Guava as much as possible and while that's _probably_ not so important for API 7, it's good to follow that now.

src/main/java/org/spongepowered/common/registry/type/scoreboard/CriteriaRegistryModule.java
```diff
@@ -45,6 +60,33 @@ public void registerDefaults() {
...
+            final int colorIndex = entry.getKey().getColorIndex();
+
+            if (colorIndex < 0) {
+                continue;
+            }
+
+            final String id = entry.getValue().getId();
+
+            final Criterion teamKillCriterion = (Criterion) IScoreCriteria.TEAM_KILL[colorIndex];
+            register(teamKillCriterion);
```
dualspiral: I realise it wasn't like this before, but can you prefix all instance methods and fields with `this.` please? Again, enforced for API 8, but I'd like to ensure that anything new going into API 7 also follows it so we're already in the swing of things!

dualspiral (commented):
I understand you've copied `DisplaySlotRegistryModule` so these comments can equally be directed at that module, but some food for thought, particularly for any help you give us for API 8 where what I'm pointing out will be important.

It might be that @Zidane is fine accepting these PRs as due to the similarity with the aforementioned module, but these are things that should be kept in mind for the future.

thibaulthenry:
I would like to ask you guys if I should also add the registration of all statistic criteria in CriteriaRegistryModule ?

At the moment, in my plugin, I have to do this extra-process to retrieve every scoreboard criteria : 

```java
Collection<Criterion> criteria =
            Stream.concat(
                    Sponge.getRegistry().getAllOf(Criterion.class).stream(),
                    Sponge.getRegistry().getAllOf(Statistic.class)
                            .stream()
                            .map(Statistic::getCriterion)
                            .filter(Optional::isPresent)
                            .map(Optional::get)
            )
                    .collect(Collectors.toList());
```


EDIT: After some tests, I truely think that it's mandatory to include Statistic Criteria into the CriteriaRegistryModule. Otherwise, there's no way to retrieve a statistic criterion via GameRegistry#getType

thibaulthenry:
Well, the build is failing because he needs to know the change in the API : 
https://github.com/SpongePowered/SpongeAPI/pull/2258

(getTeamColor method in Criterion.class)

src/main/java/org/spongepowered/common/mixin/api/mcp/scoreboard/CriterionMixin_API.java
```diff
@@ -28,19 +28,40 @@
...
 
 @Mixin(value = {ScoreCriteriaColored.class, ScoreCriteria.class})
 @Implements(@Interface(iface = Criterion.class, prefix = "criterion$"))
 public abstract class CriterionMixin_API implements IScoreCriteria { // Trick to allow avoid shadowing, since multiple targets are used
 
-    @Nullable private String spongeId;
+    @Nullable
+    private String spongeId;
+    @Nullable
+    private TextFormatting format;
```
dualspiral: `api$` (also for `spongeId`, which I know wasn't your fault!)

src/main/java/org/spongepowered/common/mixin/api/mcp/scoreboard/CriterionMixin_API.java
```diff
@@ -28,19 +28,40 @@
...
+
+    @Inject(method = "<init>", at = @At(value = "RETURN"))
+    @SuppressWarnings("InvalidInjectorMethodSignature")
+    private void onConstructed(String name, CallbackInfo ci) {
```
dualspiral: I really don't like this - this is a complicated "hack" to basically avoid doing two mixins instead of one.

I would much prefer that this logic is split into two separate mixins. This mixin, before your changes, can stay for the common logic, but using `@Surrogate` to support two constructors in two separate classes when the real intention is to use it for potential variences in the LVT for the same method is just... bad.

How Mixin works is black magic to some. Our Mixin classes should not be. If there needs to be an injection to a specific class, give it its own mixin.

(Also, this is not an API implementing method, so `api$`)

src/main/java/org/spongepowered/common/mixin/api/mcp/scoreboard/CriterionMixin_API.java
```diff
@@ -53,4 +74,8 @@
         }
         return this.spongeId;
     }
+
+    public Optional<TextColor> getTeamColor() {
```
dualspiral: This implements the `@Interface`, so should use the prefix in that annotation - `criterion$`

src/main/java/org/spongepowered/common/registry/type/scoreboard/CriteriaRegistryModule.java
```diff
@@ -30,21 +30,47 @@
...
+            });
+        });
+    }
+
+    private static final class Holder {
+        static final CriteriaRegistryModule INSTANCE = new CriteriaRegistryModule();
+    }
+
+    public Map<String, Criterion> getTeamKillMappings() {
+        return teamKillMappings;
```
dualspiral: `this.`

src/main/java/org/spongepowered/common/registry/type/scoreboard/CriteriaRegistryModule.java
```diff
@@ -30,21 +30,47 @@
...
+    }
+
+    public Map<String, Criterion> getKilledByTeamMappings() {
+        return killedByTeamMappings;
```
dualspiral: `this.`

src/main/java/org/spongepowered/common/registry/type/scoreboard/CriteriaRegistryModule.java
```diff
@@ -30,21 +30,47 @@
...
-        register((Criterion) IScoreCriteria.PLAYER_KILL_COUNT);
-        register((Criterion) IScoreCriteria.TOTAL_KILL_COUNT);
-        register((Criterion) IScoreCriteria.DEATH_COUNT);
+        IScoreCriteria.INSTANCES.values().forEach(scoreCriterion -> {
+            final Criterion criterion = (Criterion) scoreCriterion;
+
+            this.register(criterion);
+
+            criterion.getTeamColor().ifPresent(color -> {
+                teamKillMappings.put(color.getId(), criterion);
```
dualspiral: `this.`

src/main/java/org/spongepowered/common/registry/type/scoreboard/CriteriaRegistryModule.java
```diff
@@ -30,21 +30,47 @@
...
+                killedByTeamMappings.put(color.getId(), criterion);
```
dualspiral: `this.`

dualspiral (changes requested):

src/main/java/org/spongepowered/common/mixin/api/mcp/scoreboard/CriterionMixin_API.java
thibaulthenry: Can I implement the same interface with 2 different mixins ?

dualspiral: Yes

thibaulthenry:
Well, I rebased to be up-to-date, I shouldn't sorry

dualspiral:
Manually merged, thanks!

PR #3206, opened 2020-11-20
Add max-chunk-lifetime setting used by chunk gc (#3167) - fixed

PolyacovYury:
Fixed a bug in my previous edition of this PR.

> If the chunk is not queued for unload, and the max-lifetime setting is disabled - the chunk gets saved to disk. Every tick.
> That else if should look like `else if (max-lifetime <= 0 || !delay_expired || !isDirty) {continue;}`.

Now that else if looks like it should've. :)
(If only GitHub allowed to reopen merged pull-requests by reverting it's merge commit or something...)

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/world/gen/ChunkProviderServerMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/world/gen/ChunkProviderServerMixin.java
@@ -83,6 +83,7 @@ public abstract class ChunkProviderServerMixin implements ChunkProviderServerBri
     private boolean impl$forceChunkRequests = false;
     private long impl$chunkUnloadDelay = Constants.World.DEFAULT_CHUNK_UNLOAD_DELAY;
     private int impl$maxChunkUnloads = Constants.World.MAX_CHUNK_UNLOADS;
+    private int impl$maxChunkLifetime = Constants.World.MAX_CHUNK_LIFETIME;
 
     @Shadow @Final private WorldServer world;
     @Shadow @Final private IChunkLoader chunkLoader;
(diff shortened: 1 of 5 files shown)


thibaulthenry:
As far as I can test, the server does not crash anymore as it used to in your previous commit.

src/main/java/org/spongepowered/common/mixin/core/world/gen/ChunkProviderServerMixin.java
```diff
@@ -268,22 +270,34 @@ public boolean tick()
...
                             continue;
                         }
                         spongeChunk.bridge$setScheduledForUnload(-1);
                     }
                     chunk.onUnload();
-                    this.saveChunkData(chunk);
-                    this.saveChunkExtraData(chunk);
+                }
+                // max lifetime only applies to chunks that should stay loaded but still need saving once in a while
+                else if (this.impl$maxChunkLifetime <= 0 // if setting is disabled
```
dualspiral, line 290: Just clarify something for me here: this seems to change the logic such that chunk data and extra chunk data will never be saved if the max chunk lifetime is disabled because this check will then always succeed and fall through to the continue? Shouldn't this check only pass if the chunk isn't due to be unloaded in the first place?

dualspiral: Oh, wait, I see, else if. Got it.

PolyacovYury: Exactly. If the chunk is scheduled for unload - it's lifetime doesn't need to be checked - we just fall through to the saving code.

PR #3212, opened 2020-11-28
[API7] [WIP] Expose portal agents

thibaulthenry:
**SpongeCommon | [SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2264)**

In this PR I'm trying to expose the management of portal agents directly from the API.

There is a bit of logic in both methods WorldInfoMixin_API, based on these facts : 

- From the overworld, going through a Nether portal will trigger the "minecraft:default_the_nether" portal agent.
- From the overworld, going through a End portal will trigger the "minecraft:default_the_end" portal agent.
- From the Nether, going through a Nether/End portal will trigger the "minecraft:default_overworld" portal agent.
- From the End, going through a Nether/End portal will trigger the "minecraft:default_overworld" portal agent.

There is a third portal type known as END_GATEWAY in Minecraft, but since this type of portal is not even firing MoveEntityEvent, I guess it deserves its own PR.

I have a TODO to find a way to retrieve the default world name in the server as this world is the default world redirection if no portal-agent is set, or if destination worlds are not loaded.

Thanks for reading this PR.

### DIFF
--- a/src/main/java/org/spongepowered/common/config/category/WorldCategory.java
+++ b/src/main/java/org/spongepowered/common/config/category/WorldCategory.java
@@ -27,6 +27,7 @@ package org.spongepowered.common.config.category;
 import net.minecraft.launchwrapper.Launch;
 import ninja.leaping.configurate.objectmapping.Setting;
 import ninja.leaping.configurate.objectmapping.serialize.ConfigSerializable;
+import org.spongepowered.api.Sponge;
 
 import java.io.IOException;
 import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/common/config/category/WorldCategory.java
```diff
@@ -208,6 +209,15 @@ public void setMobSpawnRange(int range) {
         return this.portalAgents;
     }
 
+    public String getPortalDestination(String dimensionName) {
+        // TODO I would like to know how to retrieve the default world name from here, in a better way than below
```
JBYoshi: I don't think there is any better way to get the default world name - it's configurable on its own. The server uses the `level-name` value in `server.properties`, and the client uses the name of the world folder.

thibaulthenry: Is this safe to use this method here ?

ryantheleach:
@thibaulthenry It was my (naive) assumption that portal agents were restricted for handling multi-dimensional travel. Which is why end gateways fail to trigger them.

An aside on the information you provide about the mixins:

It's surprising that the return from the end to the overworld uses the same agent as nether to overworld. I would have thought these would have been different agents, as the behavior between them is different. e.g. one hunts for a bed, the other a portal.

thibaulthenry:
You are right about PortalAgent. There are aimed for multi-dimensional travel.
I was introducing END_GATEWAY as a PortalType which differs from PortalAgent. Even though I guess no more implementation about Portals will occur in API-7, I was attempting to draw @Zidane attention because he wrote the Portal API in 8, and END_GATEWAY is missing still.

The return from Nether or End dimensions is, in Vanilla, always towards the overworld, this is why `minecraft:default_overworld` is the responsible PortalAgent.

dualspiral:
See API PR for closure reason.

PR #3213, opened 2020-11-28
Implement data keys and event for client settings

zml2008:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2265) | **SpongeCommon**

This implements the existing event, and the new data keys added in the API PR -- feedback is appreciated

### DIFF
--- a/src/main/java/org/spongepowered/common/data/provider/entity/ServerPlayerData.java
+++ b/src/main/java/org/spongepowered/common/data/provider/entity/ServerPlayerData.java
@@ -29,6 +29,8 @@ import net.minecraft.stats.Stat;
 import net.minecraft.world.GameType;
 import org.spongepowered.api.data.Keys;
 import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.entity.living.player.chat.ChatVisibilities;
+import org.spongepowered.api.entity.living.player.chat.ChatVisibility;
 import org.spongepowered.api.entity.living.player.gamemode.GameMode;
 import org.spongepowered.api.profile.property.ProfileProperty;
 import org.spongepowered.api.statistic.Statistic;
@@ -37,6 +39,7 @@ import org.spongepowered.common.bridge.entity.player.ServerPlayerEntityBridge;
 import org.spongepowered.common.bridge.stats.StatisticsManagerBridge;
 import org.spongepowered.common.data.provider.DataProviderRegistrator;
 import org.spongepowered.common.util.Constants;
+import org.spongepowered.common.util.LocaleCache;
 
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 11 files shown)


src/main/java/org/spongepowered/common/data/provider/entity/ServerPlayerData.java
```diff
@@ -76,7 +91,12 @@ public static void register(final DataProviderRegistrator registrator) {
                             h.bridge$setHealthScale(v);
                             return true;
                         })
-                        .resetOnDelete(Constants.Entity.Player.DEFAULT_HEALTH_SCALE);
+                        .resetOnDelete(Constants.Entity.Player.DEFAULT_HEALTH_SCALE)
+                    .create(Keys.VIEW_DISTANCE)
+                        .get(ServerPlayerEntityBridge::bridge$getViewDistance)
+                    .create(Keys.SKIN_PARTS)
+                        .get(ServerPlayerEntityBridge::bridge$getSkinParts);
+                        // .set(ServerPlayerEntityBridge::bridge$setSkinParts);
```
Zidane: ?

src/main/java/org/spongepowered/common/registry/SpongeCatalogRegistry.java
```diff
@@ -618,6 +620,7 @@ public void registerDefaultRegistries() {
             .generateRegistry(RailDirection.class, ResourceKey.minecraft("rail_direction"), Arrays.stream(RailShape.values()), true, false)
             .generateRegistry(SelectorSortAlgorithm.class, ResourceKey.minecraft("selector_sort_algorithm"), SelectorSortAlgorithmStreamGenerator.stream(), true, false)
             .generateRegistry(SelectorType.class, ResourceKey.minecraft("selector_type"), SelectorTypeStreamGenerator.stream(), true, false)
+            .generateRegistry(SkinPart.class, ResourceKey.minecraft("skin_part"), SkinPartStreamGenerator.stream(), true, false)
```
Zidane, line 627: If skin parts are not keyed in Vanilla, use Sponge namespace.

Zidane: nvm, read the below!

src/mixins/java/org/spongepowered/common/mixin/core/entity/EntityMixin.java
```diff
@@ -169,6 +169,8 @@
 
     @Shadow public abstract void shadow$setLocationAndAngles(double x, double y, double z, float yaw, float pitch);
 
+    @Shadow public abstract EntityDataManager getDataManager();
```
Zidane: shadow$ prefix on shadow methods is our standards. At least we try to do so..

src/main/java/org/spongepowered/common/data/provider/entity/ServerPlayerData.java
zml2008: I did a basic implementation for this one, so it theoretically works since it is just a data tracker field, but because the client can update the skin parts option at any time the behaviour is unpredictable so I didn't actually activate the set operation.

Zidane: Probably should do it as read-only then.

Zidane:
Manually merged, thanks!

PR #3214, opened 2020-11-28
Expose entity tags

thibaulthenry:
**SpongeCommon | [SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2266)**

Hey,

Here is a small PR trying to expose entity tags which can be used as selector arguments in commands.
Until now they were not exposed in the API.

Learn more about tags: https://minecraft.gamepedia.com/Scoreboard#Tags

Thanks for reading this PR.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/api/mcp/entity/EntityMixin_API.java
+++ b/src/main/java/org/spongepowered/common/mixin/api/mcp/entity/EntityMixin_API.java
@@ -106,6 +106,7 @@ import org.spongepowered.common.util.VecHelper;
 import org.spongepowered.common.world.WorldManager;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Optional;
@@ -141,6 +142,7 @@ public abstract class EntityMixin_API implements org.spongepowered.api.entity.En
     @Shadow protected EntityDataManager dataManager;
     @Shadow public int dimension;
     @Shadow protected UUID entityUniqueID;
+    @Shadow public Set<String> tags;
 
     @Shadow public abstract void setPosition(double x, double y, double z);
     @Shadow public abstract void setDead();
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/common/mixin/api/mcp/entity/EntityMixin_API.java
```diff
@@ -682,4 +684,32 @@ public EntityArchetype createArchetype() {
         this.get(VelocityData.class).ifPresent(manipulators::add);
     }
 
+    @Override
+    public Set<String> getTags() {
```
dualspiral: `Entity#getTags` returns the original set. While I don't see any vanilla code that makes use of the fact that the backing set is mutable, mods very well might. It would be stupid for them to do so, but they might.

Thankfully, there is a really easy way to sidestep this: make this method return a `Collection<String>`! While Java doesn't let you vary only by return type, no such restriction exists on the JVM. That way our method and the vanilla `Set<String> getTags()` method can co-exist peacefully.

src/main/java/org/spongepowered/common/mixin/api/mcp/entity/EntityMixin_API.java
```diff
@@ -682,4 +684,32 @@ public EntityArchetype createArchetype() {
...
+        return Collections.unmodifiableSet(this.tags);
+    }
+
+    @Override
+    public boolean addTag(final String tag) {
```
dualspiral: I think this would be better off being `@Intrinsic` and calling through to the original method so it doesn't interfere with other vanilla calls.

```
@Intrinsic
public boolean entity$addTag(final String tag) {
    Objects.requireNonNull(tag, "tag");
    return this.shadow$addTag(tag);
}
```

The difference between this method and that one is that you return the result of `Set#add`, vanilla just returns `true` - is it a problem if `addTag` just reports whether the entry is in the set, which it will be if it reports true, regardless?

src/main/java/org/spongepowered/common/mixin/api/mcp/entity/EntityMixin_API.java
```diff
@@ -682,4 +684,32 @@ public EntityArchetype createArchetype() {
...
+
+        if (this.tags.size() >= 1024) {
+            return false;
+        }
+
+        return this.tags.add(tag);
+    }
+
+    @Override
+    public boolean removeTag(final String tag) {
```
dualspiral: Intrinsic, as above.

src/main/java/org/spongepowered/common/mixin/api/mcp/entity/EntityMixin_API.java
```diff
@@ -682,4 +684,32 @@ public EntityArchetype createArchetype() {
...
+        checkNotNull(tag);
+        return this.tags.remove(tag);
+    }
+
+    @Override
+    public boolean hasTag(final String tag) {
```
dualspiral, line 705: Intrinsic, as above.

dualspiral (changes requested):

src/main/java/org/spongepowered/common/mixin/api/mcp/entity/EntityMixin_API.java
thibaulthenry: My original intent on this was to prevent developers to put a null value in the set.
But I agree we can report true as vanilla does

src/main/java/org/spongepowered/common/mixin/api/mcp/entity/EntityMixin_API.java
thibaulthenry: On this one it's not a vanilla method

dualspiral:
Manually merged (with minor updates), thanks!

PR #3215, opened 2020-11-29
Implement MatterState.

JBYoshi:
Other code for this is already written using the constants in `MatterStates`, but nothing actually implements the `MatterState` interface - this should take care of that. Let me know if there's anything I'm missing.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/type/SpongeMatterState.java
+++ b/src/main/java/org/spongepowered/common/data/type/SpongeMatterState.java
@@ -1,0 +1,12 @@
+package org.spongepowered.common.data.type;
+
+import org.spongepowered.api.ResourceKey;
+import org.spongepowered.api.data.type.MatterState;
+import org.spongepowered.common.SpongeCatalogType;
+
+public final class SpongeMatterState extends SpongeCatalogType implements MatterState {
+
+    public SpongeMatterState(ResourceKey key) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/common/data/type/SpongeMatterState.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.spongepowered.common.data.type;
+
+import org.spongepowered.api.ResourceKey;
+import org.spongepowered.api.data.type.MatterState;
+import org.spongepowered.common.SpongeCatalogType;
+
+public final class SpongeMatterState extends SpongeCatalogType implements MatterState {
+
+    public SpongeMatterState(ResourceKey key) {
```
Zidane, line 33: `final`

dualspiral:
I realise that you're just implementing an interface, but I'm a little iffy that the API is named `MatterState`, as it's not a [`State`](https://github.com/SpongePowered/SpongeAPI/blob/303691899f1b8a288febfa36cda971a5cffb0a1d/src/main/java/org/spongepowered/api/state/State.java#L39) in the same way `BlockState` and `FluidState` is. Could we rename the API to something else, maybe `MatterType`?

Zidane:
Manually merged, thanks!

PR #3219, opened 2020-12-01
Proxy PhaseState method access to avoid raw typing

gabizou:
Mostly a QOL change, as I've been thinking about it, the proxy access by PhaseContext is handy, and can be kept tidy in PhaseStateProxy.

Demonstration of changes can be seen in ContainerUtil:
```diff
-       final IPhaseState<@NonNull ?> currentState = context.state;
-        if (((IPhaseState) currentState).doesBlockEventTracking(context)) {
+        if (context.doesBlockEventTracking()) {
```

I've not given thought to profile the difference as it is replacing an explicit raw type cast to a method invocation by proxy. Seeing how it's already relatively small, I can't imagine this would be an expensive change.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
@@ -130,7 +130,6 @@ import org.spongepowered.common.bridge.world.chunk.ChunkBridge;
 import org.spongepowered.common.entity.EntityUtil;
 import org.spongepowered.common.entity.PlayerTracker;
 import org.spongepowered.common.entity.projectile.UnknownProjectileSource;
-import org.spongepowered.common.event.tracking.IPhaseState;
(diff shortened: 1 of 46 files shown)


dualspiral:
I don't see why this would be contentious really. It's basically hiding "get the context from the state, then feed the context into the state" which, let's face it, does seem a little odd at first glance. There is a readability win here.

Profiling would be interesting, but I'm sure this is the sort of thing the JVM will optimise anyway, particularly for those methods that don't rely on the context, and therefore are constants. At that point - this could even be a win for performance. It does make me wonder if there should be some inching towards state and context merging into one object, we're already blurring the lines here. Food for thought, not convinced it's the right thing to do, but it might be worth consideration - though with the current pooling we have, I'm not sure I'd bother because the gain will be miniscule.

(We do pool both states and contexts for the server and client threads, right?)

Zidane:
I don't see the issues either, update this and get it merged Gabe.

PR #3226, opened 2020-12-06
[API7] [WIP] Implement ChangeWorldTimeEvent

thibaulthenry:
**SpongeCommon | [SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2273)**

Hey, a small PR to implement the ChangeWorldTimeEvent as @BrainStone  was looking for it in Discord.

I would like to cancel the result message after using the vanilla time command if the event is cancelled, but I can't find a way to retrieve this event as I'm not throwing it in the CommandTimeMixin_MultiWorldCommand class.

Do someone has a solution for this ?

Thanks for reading this PR.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/api/mcp/world/storage/WorldInfoMixin_API.java
+++ b/src/main/java/org/spongepowered/common/mixin/api/mcp/world/storage/WorldInfoMixin_API.java
@@ -36,16 +36,20 @@ import net.minecraft.world.GameRules;
 import net.minecraft.world.GameType;
 import net.minecraft.world.WorldType;
 import net.minecraft.world.storage.WorldInfo;
+import org.spongepowered.api.Sponge;
 import org.spongepowered.api.data.DataContainer;
 import org.spongepowered.api.data.DataQuery;
 import org.spongepowered.api.data.DataView;
 import org.spongepowered.api.data.persistence.DataFormats;
 import org.spongepowered.api.entity.living.player.gamemode.GameMode;
+import org.spongepowered.api.event.SpongeEventFactory;
+import org.spongepowered.api.event.world.ChangeWorldTimeEvent;
 import org.spongepowered.api.world.DimensionType;
 import org.spongepowered.api.world.GeneratorType;
 import org.spongepowered.api.world.PortalAgentType;
 import org.spongepowered.api.world.SerializationBehavior;
 import org.spongepowered.api.world.SerializationBehaviors;
+import org.spongepowered.api.world.World;
 import org.spongepowered.api.world.difficulty.Difficulty;
 import org.spongepowered.api.world.gen.WorldGeneratorModifier;
 import org.spongepowered.api.world.storage.WorldProperties;
@@ -113,7 +117,6 @@ public abstract class WorldInfoMixin_API implements WorldProperties {
     @Shadow public abstract void setDifficulty(EnumDifficulty newDifficulty);
     @Shadow public abstract NBTTagCompound cloneNBTCompound(@Nullable NBTTagCompound nbt);
     @Shadow public abstract String shadow$getWorldName();
-    @Shadow public abstract void shadow$setWorldTime(long time);
 
     private SerializationBehavior api$serializationBehavior = SerializationBehaviors.AUTOMATIC;
 
@@ -160,9 +163,28 @@ public abstract class WorldInfoMixin_API implements WorldProperties {
         return this.worldTime;
     }
 
-    @Intrinsic
-    public void worldproperties$setWorldTime(final long time) {
-        this.shadow$setWorldTime(time);
+    @Override
+    public void setWorldTime(final long time) {
+        final Optional<World> optionalTargetWorld = Sponge.getServer().getWorld(getUniqueId());
+
+        if (!optionalTargetWorld.isPresent()) {
+            this.worldTime = time;
+            return;
+        }
+
+        final ChangeWorldTimeEvent event = SpongeEventFactory.createChangeWorldTimeEvent(
+                Sponge.getCauseStackManager().getCurrentCause(),
+                worldTime,
+                time,
+                optionalTargetWorld.get(),
+                worldTime
+        );
+
+        if (Sponge.getEventManager().post(event)) {
+            return;
+        }
+
+        this.worldTime = event.getTime();
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/mixin/api/mcp/world/storage/WorldInfoMixin_API.java
```diff
@@ -160,9 +163,28 @@ public long getTotalTime() {
         return this.worldTime;
     }
 
-    @Intrinsic
-    public void worldproperties$setWorldTime(final long time) {
-        this.shadow$setWorldTime(time);
+    @Override
+    public void setWorldTime(final long time) {
+        final Optional<World> optionalTargetWorld = Sponge.getServer().getWorld(getUniqueId());
```
dualspiral, line 182: There are two things that I'm a little iffy about here - one should be easy to fix, the other, not so much (a similar issue exists with your previous PR):

* The event construction and firing should be wrapped in `ShouldFire.CHANGE_WORLD_TIME` (I think?), as this is going to fire 20 times a second _at least_. Search for `ShouldFire` in your IDE and see how we do it with other big hitting events - this should avoid this event being called if no-one is listening for it.
* This returns `void`. A caller will set the time but there is no way to know what happened with that request - we've had this before with cancellable events. This directly feeds into your question about the multi-world command. Thus, I'm a little iffy on this... annoyingly because this would be a nice PR to accept.

For your multi world command stuff, you could just check what the world time is as it's been set at this point - and if it's the same as before, you know nothing changed. But my concern is that plugins won't know about this and could cause weird behaviour.

Then again, this might be an edge case and users who install plugins that do this will be fully aware of the consequences. If this is the case, we _must_ update the Javadocs.

There is also the option of introducing a new API method, but naming it would be annoying as `setWorldTime` is a good name already. Could be `long setTime(long)`, of course...

Consider how this would fit into API 8. In particular, I introduced `MinecraftDayTime` - this is going to change anyway so it would be great if you could at least specify the method changes you think you'll need so we don't get caught unawares later!

BrainStone:
Is there a particular reason you undid what dual commited recently?  
That's actually a pretty important change.

thibaulthenry:
> 
> 
> Is there a particular reason you undid what dual commited recently?
> That's actually a pretty important change.

If I write code in a method annoted by `@Intrisic`, it is not reached.

If I don't delete the shadowed method WolrdInfo#setWorldTime and override the API method WorldProperties#setWorldTime, I got an infinite loop

BrainStone:
> If I write code in a method annoted by @Intrisic, it is not reached.

Correct

Just Inject into the vanilla ``setWorldTime``. The purpose of what dual did was to make the Sponge ``setWorldTime`` call the vanilla ``setWorldTime``.

dualspiral:
> If I write code in a method annoted by `@Intrisic`, it is not reached.

The problem is that there is a `setWorldTime` method on both our `WorldProperties` interface, and on MCP's `WorldInfo`. In dev, they are the same name, and so `@Intrinsic` discards our method, favouring the one on MCP's. In production, the MCP method name is different, so we need to redirect our `setWorldTime` to the obfuscated version - that's what my change does (before we just duplicated the logic).

While what you've done works in dev, put it into production and you'll find that it won't work on every tick, only when a plugin sets it. That's why you need to inject into the vanilla method instead.

BrainStone:
Anything missing or is this good to go?

src/main/java/org/spongepowered/common/mixin/api/mcp/world/storage/WorldInfoMixin_API.java
```diff
@@ -165,6 +173,31 @@ public long getTotalTime() {
...
+
+        final ChangeWorldTimeEvent event = SpongeEventFactory.createChangeWorldTimeEvent(
+                Sponge.getCauseStackManager().getCurrentCause(),
+                time,
+                time,
+                optionalTargetWorld.get(),
+                worldTime
+        );
+
+        if (Sponge.getEventManager().post(event)) {
```
dualspiral, line 196: The result of `setTime` is ignored, it shouldn't be if it's part of the event.

src/main/java/org/spongepowered/common/mixin/command/multiworld/CommandTimeMixin_MultiWorldCommand.java
```diff
@@ -29,36 +29,75 @@
...
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.Redirect;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
 @Mixin(CommandTime.class)
 public abstract class CommandTimeMixin_MultiWorldCommand {
 
+    private boolean modificationCancelled = false;
```
dualspiral, line 39: `multiWorldCommand$modificationCancelled`

src/main/java/org/spongepowered/common/mixin/command/multiworld/CommandTimeMixin_MultiWorldCommand.java
```diff
@@ -29,36 +29,75 @@
...
+    private boolean incrementCancelled = false;
```
dualspiral, line 40: `multiWorldCommand$incrementCancelled`

dualspiral (changes requested):
I'm still not ready to pull this because we need to decide what to do about the problem of having a `void` return type on `WorldProperties#setTime` and cancellations.

There are also other issues...

thibaulthenry:
> 
> 
> I'm still not ready to pull this because we need to decide what to do about the problem of having a `void` return type on `WorldProperties#setTime` and cancellations.
> 
> There are also other issues...

Maybe we should give up the implementation of this event in API-7. The solution you introduced about having 2 methods that does the same thing but one return void and one return boolean, is really confusing I guess.

If it is not possible to change anything on API-7, we should focus more on API-8 and see if we can avoid **void** returns on setters, especially if those setters may potentially have a linked event.
I don't think I'm aware enough about API-8 to make this decision, maybe you should discuss about it with the team

dualspiral:
I know it's been a year but we did discuss it later.

Ultimately, due to the challenge it poses and the lateness in the API 7 cycle, I'm going to close this - but it did give us things to think about for API 8 - so thank you for that.

PR #3239, opened 2020-12-18
Porting the portals

dualspiral:
For portal discussion with @Zidane


src/mixins/java/org/spongepowered/common/mixin/core/entity/player/ServerPlayerEntityMixin.java
```diff
@@ -468,26 +495,144 @@ public void teleportTo(net.minecraft.world.server.ServerWorld world, double x, d
...
+            if (registrykey == World.END && targetWorld.dimension() == World.OVERWORLD && spongeTeleporter.isVanilla()) { // avoids modded dimensions
+                this.shadow$unRide();
+                this.shadow$getLevel().removePlayerImmediately((ServerPlayerEntity) (Object) this);
+                if (!this.wonGame) {
+                    this.wonGame = true;
+                    this.connection.send(new SChangeGameStatePacket(SChangeGameStatePacket.WIN_GAME, this.seenCredits ? 0.0F : 1.0F));
+                    this.seenCredits = true;
+                }
+
+                // TODO: Figure out the reposition step (though that might just be a respawn)
```
Zidane: This is a respawn and makes sense to not do anything with the logic here.

src/mixins/java/org/spongepowered/common/mixin/core/entity/player/ServerPlayerEntityMixin.java
```diff
@@ -468,26 +495,144 @@ public void teleportTo(net.minecraft.world.server.ServerWorld world, double x, d
...
+                return (Entity) (Object) this;
+            } else {
+                IWorldInfo iworldinfo = targetWorld.getLevelData();
+                this.connection.send(new SRespawnPacket(targetWorld.dimensionType(), targetWorld.dimension(),
```
Zidane: Once the server can run, I can figure out if we'll need to send fake respawn packets or not (to clear Vanilla client chunk caches).

May need to keep in mind, if not, will need to call the packet hook for this.

src/mixins/java/org/spongepowered/common/mixin/core/entity/player/ServerPlayerEntityMixin.java
```diff
@@ -468,26 +495,144 @@ public void teleportTo(net.minecraft.world.server.ServerWorld world, double x, d
...
+                final PortalInfo portalinfo = spongeTeleporter.getPortalInfo((Entity) (Object) this, targetWorld, this::shadow$findDimensionEntryPoint);
+                // Sponge End
+                if (portalinfo != null) {
+                    // Sponge Start: wrap the teleportation logic within a function to allow for modification
+                    // of the teleporter, as well as calling events(?)
+                    final Vector3d originalDestination = new Vector3d(portalinfo.pos.x, portalinfo.pos.y, portalinfo.pos.z);
+                    final Entity e = spongeTeleporter.performTeleport((Entity) (Object) this, serverworld, targetWorld, this.xRot, spawnPortal -> {
+                        // Sponge End
+                        serverworld.getProfiler().push("moving");
+                        if (registrykey == World.OVERWORLD && targetWorld.dimension() == World.NETHER) {
```
Zidane: I wonder if we need to run this if the type is NETHER and not Vanilla's NETHER.

It may only be used for the dimension triggers which we wouldn't then, been a while since I read up on this.

src/mixins/java/org/spongepowered/common/mixin/core/entity/player/ServerPlayerEntityMixin.java
```diff
@@ -468,26 +495,144 @@ public void teleportTo(net.minecraft.world.server.ServerWorld world, double x, d
...
+                // TODO: This could house the Reposition Event, with some flag on SpongeTeleporter that can be set to cause this to happen later
+                //  for mods.
+                //  This can then hold our custom portal logic (basically any edits to findDimensionEntryPoint that we might need to do)
+                final PortalInfo portalinfo = spongeTeleporter.getPortalInfo((Entity) (Object) this, targetWorld, this::shadow$findDimensionEntryPoint);
+                // Sponge End
+                if (portalinfo != null) {
+                    // Sponge Start: wrap the teleportation logic within a function to allow for modification
+                    // of the teleporter, as well as calling events(?)
+                    final Vector3d originalDestination = new Vector3d(portalinfo.pos.x, portalinfo.pos.y, portalinfo.pos.z);
+                    final Entity e = spongeTeleporter.performTeleport((Entity) (Object) this, serverworld, targetWorld, this.xRot, spawnPortal -> {
```
Zidane: Where is spawnPortal being fed from?

src/mixins/java/org/spongepowered/common/mixin/core/entity/player/ServerPlayerEntityMixin.java
```diff
@@ -468,26 +495,144 @@ public void teleportTo(net.minecraft.world.server.ServerWorld world, double x, d
...
+                        // Sponge End
+                        serverworld.getProfiler().push("moving");
+                        if (registrykey == World.OVERWORLD && targetWorld.dimension() == World.NETHER) {
+                            this.enteredNetherPosition = this.shadow$position();
+                        // Sponge Start: only enter this branch if we're going through portal logic
+                        } else if (spawnPortal && targetWorld.dimension() == World.END) {
```
Zidane: I heavily debated running this for Vanilla's END or END dimension types (since they will use the END generator).

The problem is the Vanilla code, no matter what, puts the platform there. Someone could get angry in a custom end having a build at  spawn get overwrote due to this. For Vanilla END, we can state that is game mechanics.

src/mixins/java/org/spongepowered/common/mixin/core/entity/player/ServerPlayerEntityMixin.java
```diff
@@ -468,26 +495,144 @@ public void teleportTo(net.minecraft.world.server.ServerWorld world, double x, d
...
+                        this.shadow$setRot(portalinfo.yRot, portalinfo.xRot);
+                        this.shadow$moveTo(portalinfo.pos.x, portalinfo.pos.y, portalinfo.pos.z);
+                        serverworld.getProfiler().pop();
+                        // Sponge Start: return this entity
+                        return (Entity) (Object) this;
+                    });
+                    // Sponge End
+                    // This is done here (it's not in Forge) to ensure that it's done. This is to ensure
+                    // advancements are triggered.
+                    this.shadow$triggerDimensionChangeTriggers(serverworld);
```
Zidane: Should run this after instance check (589)

src/mixins/java/org/spongepowered/common/mixin/core/entity/player/ServerPlayerEntityMixin.java
dualspiral: Forge - it's to allow mod `ITeleporter`s to prevent the creation of an end platforms.

src/mixins/java/org/spongepowered/common/mixin/core/entity/player/ServerPlayerEntityMixin.java
```diff
@@ -468,26 +495,144 @@ public void teleportTo(net.minecraft.world.server.ServerWorld world, double x, d
...
+                    if (e != (Object) this) {
+                        throw new IllegalArgumentException(String.format("Teleporter %s "
+                                + "did not return the expected player entity: got %s, expected PlayerEntity %s", spongeTeleporter, e, this));
+                    }
+
+                    // If we need to reposition: well... reposition.
+                    // Downside: portals won't come with us, but with how it's implemented in Forge,
+                    // not sure how we'd do this
+                    final Vector3d destination = VecHelper.toVector3d(this.shadow$position());
+                    final ChangeEntityWorldEvent.Reposition reposition = SpongeEventFactory.createChangeEntityWorldEventReposition(
```
Zidane: Here is the big difference between my version of this and yours.

In my version, I run reposition logic by finding where the portal will put you but I don't place any portal blocks. If the event succeeds, I run the logic again based on the event's position and put the blocks there. I did this to prevent cancelling reposition would end up having to roll back blocks in the PhaseTracker.

Basically I run it in the lamdba which, truthfully, was a giant giant pain...

dualspiral: This is the big problem I have.

For vanilla, that's okay. We can do some magic with `getExitPortal` (or just copy/modify a redirection of it), and that's what my one of my TODOs is about. So if this is a route we wanted to go down, I would have spent some time doing that for vanilla logic - this wouldn't remain here if vanilla logic is running.

For Forge, there's very little we can do because there's no way to feed in the requested position to the mod.

Zidane: You are correct, in my version if a mod passed in their own lambda we would call no events. It was definitely a flaw but I accepted it cause the whole thing is asinine to begin with.............

src/mixins/java/org/spongepowered/common/mixin/core/entity/player/ServerPlayerEntityMixin.java
```diff
@@ -468,26 +495,144 @@ public void teleportTo(net.minecraft.world.server.ServerWorld world, double x, d
...
+                    this.gameMode.setLevel(targetWorld);
+                    this.connection.send(new SPlayerAbilitiesPacket(this.abilities));
+                    playerlist.sendLevelInfo((ServerPlayerEntity) (Object) this, targetWorld);
+                    playerlist.sendAllPlayerInfo((ServerPlayerEntity) (Object) this);
+
+                    for (EffectInstance effectinstance : this.shadow$getActiveEffects()) {
+                        this.connection.send(new SPlayEntityEffectPacket(this.shadow$getId(), effectinstance));
+                    }
+
+                    this.connection.send(new SPlaySoundEventPacket(1032, BlockPos.ZERO, 0, false));
```
Zidane: This should only be called if it is actually a portal else API usages of NetherPortalType will have the swoosh sound which would be odd imo.

Zidane: Assuming this code is copied to a util method for handling post teleport logic.

vanilla/src/mixins/java/org/spongepowered/vanilla/mixin/core/entity/player/ServerPlayerEntityMixin_Vanilla.java
```diff
@@ -81,6 +83,18 @@
         }
     }
 
+    /**
+     * @author dualspiral - 18th December 2020 - 1.16.4
+     * @reason This is effectively an overwrite, redirecting the vanilla
```
Zidane: If it is effectively an overwrite, make it one. With auditing, these overwrites like up like a Christmas tree, ensures we don't overlook it.

src/mixins/java/org/spongepowered/common/mixin/core/entity/player/ServerPlayerEntityMixin.java
```diff
@@ -468,26 +495,144 @@ public void teleportTo(net.minecraft.world.server.ServerWorld world, double x, d
...
+                            (org.spongepowered.api.world.server.ServerWorld) originalDestinationWorld,
+                            originalDestination,
+                            (org.spongepowered.api.world.server.ServerWorld) targetWorld
+                    );
+                    if (reposition.isCancelled()) {
+                        // send them back to the original destination
+                        this.shadow$setPos(originalDestination.getX(), originalDestination.getY(), originalDestination.getZ());
+                    } else if (reposition.getDestinationPosition() != destination) {
+                        final Vector3d finalPosition = reposition.getDestinationPosition();
+                        this.shadow$setPos(finalPosition.getX(), finalPosition.getY(), finalPosition.getZ());
```
Zidane: So if we keep the strategy of doing reposition like this, we need to rollback the portal blocks in the PhaseTracker, issue a ticket for the chunk pos the player is going to, and call moveTo instead.

dualspiral:
So I think we're almost there with this. I haven't yet added a phase state for portal teleport (not sure what type of phase this will be) and the associated rollback of any created portals for the modded side. 

Other than that, what this has done is deduplicated the code in both `Entity` and `ServerPlayer` changeDimension methods, added the reposition even around `getExitPortal` for vanilla mechanics, and after portal placement for non-vanilla (which is what the phase state rollback would be needed for, unless we decide we're not running the event for mod controlled teleports).

src/mixins/java/org/spongepowered/common/mixin/core/entity/EntityMixin.java
```diff
@@ -161,14 +171,21 @@
...
     @Shadow public abstract void shadow$setDeltaMovement(net.minecraft.util.math.vector.Vector3d motion);
+    @Shadow public abstract void shadow$unRide();
+    @Shadow protected abstract Optional<TeleportationRepositioner.Result> shadow$getExitPortal(
+            net.minecraft.world.server.ServerWorld targetWorld, BlockPos targetPosition, boolean isNether);
+    @Shadow protected abstract net.minecraft.util.math.vector.Vector3d shadow$getRelativePortalPosition(Direction.Axis direction$axis,
+            TeleportationRepositioner.Result teleportationrepositioner$result);
+    @Shadow protected abstract void shadow$removeAfterChangingDimensions();
     // @formatter:on
 
+    @Shadow public abstract EntityType<?> getType();
```
ImMorpheus: missing `shadow$` prefix

PR #3241, opened 2020-12-21
WIP API8 Plugin-Datapacks for Advancements/Recipes

Faithcaio:


src/main/java/org/spongepowered/common/SpongeLifecycle.java
```diff
@@ -72,6 +76,17 @@ public SpongeLifecycle(final Game game, final Injector injector) {
         this.injector = injector;
     }
 
+    // Called before loading datapacks
+    public final void earlyInit(Path datapackDir) {
```
kashike:
```suggestion
    public final void earlyInit(final Path datapackDir) {
```

kashike (changes requested):

src/main/java/org/spongepowered/common/SpongeLifecycle.java
```diff
@@ -94,7 +109,7 @@ public void establishGlobalRegistries() {
         // Need to do this here to prevent classloading Registry too early...
         holder.setRootMinecraftRegistry((Registry<Registry<?>>) Registry.REGISTRY);
 
-        SpongeRegistries.registerGlobalRegistries((SpongeRegistryHolder) Sponge.getGame().registries());
+        SpongeRegistries.registerGlobalRegistries((SpongeRegistryHolder)  this.game.registries());
```
kashike: Hmm...

src/main/java/org/spongepowered/common/registry/VanillaRegistryLoader.java
```diff
@@ -246,6 +256,12 @@ private void loadEnumRegistries() {
             l.add(Triggers.USED_ENDER_EYE, k -> (Trigger) CriteriaTriggers.USED_ENDER_EYE);
             l.add(Triggers.USED_TOTEM, k -> (Trigger) CriteriaTriggers.USED_TOTEM);
             l.add(Triggers.VILLAGER_TRADE, k -> (Trigger) CriteriaTriggers.TRADE);
+            final DefaultedRegistryReference<Trigger<?>> dummyKey =
```
Zidane, line 259: Why?

Faithcaio: how?

src/main/java/org/spongepowered/common/registry/SpongeRegistryType.java
```diff
@@ -52,6 +52,14 @@ public ResourceKey location() {
         return this.location;
     }
 
+    @Override
+    public String toString() {
```
Zidane, line 57: Please use the StringJoiner toString()s found elsewhere.

src/main/java/org/spongepowered/common/registry/SpongeRegistryLoaders.java
```diff
@@ -877,5 +882,13 @@ private static WorldArchetype newWorldArchetype(final ResourceKey key, final Reg
         return (WorldArchetype) (Object)  archetype;
     }
 
+    public static RegistryLoader<DataFormat> dataFormat() {
+        return RegistryLoader.of(l -> {
```
Zidane, line 886: Are `DataFormat`s really Registry...?

src/main/java/org/spongepowered/common/event/lifecycle/RegisterDataPackValueEventImpl.java
```diff
@@ -43,6 +46,8 @@
     public RegisterDataPackValueEventImpl(final Cause cause, final Game game) {
         super(cause, game);
         this.serializables = new Object2ObjectOpenHashMap<>();
+        this.serializables.computeIfAbsent((SpongeDataPackType) DataPackTypes.RECIPE, k -> new ArrayList<>());
```
Zidane: Shouldn't this be fixed in a getter call? Would rather not hard code like this especially when more datapack things are added...

src/main/java/org/spongepowered/common/datapack/SpongeDataPackManager.java
```diff
@@ -44,7 +48,8 @@
 @Singleton
 public final class SpongeDataPackManager {
 
-    public static SpongeDataPackManager INSTANCE;
+    // TODO @Zidane where do you want to instantiate this?
+    public static SpongeDataPackManager INSTANCE = new SpongeDataPackManager(Sponge.getGame());
```
Zidane, line 49: This can't be done like this. The class is Singleton which means that if it is injected elsewhere, it will be a different object ref.

So likely toss @Singleton then.

src/main/java/org/spongepowered/common/datapack/DataPackSerializer.java
```diff
@@ -44,12 +44,17 @@ public DataPackSerializer(final String token, final String typeDirectoryName) {
         this.typeDirectoryName = typeDirectoryName;
     }
 
-    protected void serialize(final Path dataPackDirectory, final List<T> objects) throws IOException {
-        FileUtils.deleteDirectory(dataPackDirectory.toFile());
+    protected boolean serialize(final Path dataPackDirectory, final List<T> objects) throws IOException {
+        final Path datapackDir = dataPackDirectory.resolve(this.getPackName());
```
Zidane, line 48: Should plugin generated content really be outside others? What is the reasoning for this? Datapack content is namespaced...

src/main/java/org/spongepowered/common/registry/SpongeRegistryLoaders.java
Faithcaio: probably not - I just made it work

src/main/java/org/spongepowered/common/datapack/DataPackSerializer.java
Faithcaio: wdym?
atm. this creates 2 datapacks:
`plugin-advancements` and `plugin-recipes` which contains the pack.mcmeta
and under `data/<namespace>/<type>/<value>.json` the recipe/advancement files

src/main/java/org/spongepowered/common/event/lifecycle/RegisterDataPackValueEventImpl.java
Faithcaio: Its so that both datapacks are actually deleted when no plugin registers anything for it.
There is no getter call for a specific DataPackType its just `event.getSerializables()` but I'll move it to `SpongeDataPackManager`

PR #3242, opened 2020-12-25
Implement enhanced InteractEvents [API8]

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2278) | **SpongeCommon**

Notes for Implementation:
`CPlayerDiggingPacket` is sent multiple times with different actions (start digging/cancel digging/finish digging) in survival when digging a block
creative only uses the start digging action.
`CAnimateHandPacket` is sent for a bunch of different actions. 
We need to filter out packets that are not for just swinging in air to use in InteractItemEvent.Primary. (e.g. while digging, after throwing out an item)
`CUseEntityPacket` is sent twice: first with `INTERACT_AT` then `INTERACT`


ImMorpheus (approved):
Working as intended.

PR #3245, opened 2020-12-28
Rebuild the SpongeParameterTranslator

dualspiral:
Mostly here as a final sanity check, and for the chance for any _final_ things to be checked out.

The original design of the system was designed to build a tree from the end so we know about optionals and terminating parameters ahead of time using a recursive algorithm. It turns out that the idea was quite frankly insane and has led to numerous issues with optional and default parameters, as well as indicating when something is terminal. The idea was to not add to Brig as much as possible, but it turned out that we actually just went and made many changes anyway. It was unsustainable.

This algorithm modifies the nodes such that we can force an executor or redirection on our nodes only. That way, we can build a node and retroactively apply these properties, meaning less messing about in the Brig core code and an easier time maintaining things... maybe (such updating core Brig code we've copied into Sponge is for another commit).

This should fix most of the issues people have been having - most of the test commands (those that I've tested that don't either need a little bit more thought on defaults handling, or is world stuff and that means we're waiting for that to be done)

### DIFF
--- a/src/main/java/org/spongepowered/common/command/SpongeParameterizedCommand.java
+++ b/src/main/java/org/spongepowered/common/command/SpongeParameterizedCommand.java
@@ -26,7 +26,6 @@ package org.spongepowered.common.command;
 
 import com.mojang.brigadier.ParseResults;
 import com.mojang.brigadier.StringReader;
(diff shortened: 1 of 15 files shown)


dualspiral:
ec10a35c5

PR #3262, opened 2021-01-22
Some fixes for commands

zml2008:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2288) | **Sponge**

This fixes some minor issues with commands, from things that came up in Discord and a few things that I noticed while reading through the implementation.

I'd still like to look at:

- [x] Dispatcher lifecycle
      The function loading changes in 1.16 have meant that commands are initialized extremely early, and are re-initialized on every datapack reload. I'm thinking it'll probably make sense to convert the dispatcher registry into a dispatcher *factory* registry, and match the Mojang datapack lifecycle more closely with our registration sequence. Since Mojang's strategy is to fully register commands and load datapacks with the new set of commands before applying those commands to the active server, I think we'll want to just re-register commands on a whole new dispatcher, and remove all the singletons used.
- [x] The possibility of passing through an `ArgumentReader` to `Raw` commands rather than just a plain String
      This seems like the closest we can get to allowing custom commands systems to match vanilla behaviour without binding them to constructing their own `Parameterized` commands. Unfortunately, I think `CommandContext` takes in too much detail specific to parameterized/Brigadier commands for it to be exposed to raw commands without major design changes. I think the `ArgumentReader` can also expose enough low-level parsing to easily cover converting everything that is a `ClientCompletionType`

### DIFF
--- a/src/main/java/org/spongepowered/common/command/manager/SpongeCommandManager.java
+++ b/src/main/java/org/spongepowered/common/command/manager/SpongeCommandManager.java
@@ -27,6 +27,7 @@ package org.spongepowered.common.command.manager;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Multimap;
+import io.leangen.geantyref.GenericTypeReflector;
 import io.leangen.geantyref.TypeToken;
 import com.google.inject.Inject;
 import com.google.inject.Provider;
@@ -73,7 +74,6 @@ import org.spongepowered.api.event.command.ExecuteCommandEvent;
 import org.spongepowered.api.registry.RegistryTypes;
 import org.spongepowered.api.service.pagination.PaginationService;
 import org.spongepowered.api.service.permission.Subject;
-import org.spongepowered.api.util.ComponentMessageException;
 import org.spongepowered.common.SpongeCommon;
 import org.spongepowered.common.adventure.SpongeAdventure;
 import org.spongepowered.common.bridge.command.CommandsBridge;
@@ -107,6 +107,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
 public final class SpongeCommandManager implements CommandManager.Mutable {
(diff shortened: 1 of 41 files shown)


src/main/java/org/spongepowered/common/command/manager/SpongeCommandManager.java
```diff
@@ -493,49 +511,64 @@ private void prettyPrintThrowableError(final Throwable thr, final String command
...
-            } else {
-                SpongeCommon.getLogger()
-                        .warn("Command type '{}' has already been collected, skipping request from {}",
-                                handledType.toString(),
-                                registrar.getClass());
-            }
-        });
-        SpongeParameterizedCommandRegistrar.INSTANCE.register(
+
+        // TODO: this used to be done after the event, do we care?
```
dualspiral: It doesn't matter. Callbacks in Sponge should be calling `sponge:callback` anyway, so as long as that registers, it should be fine.

src/main/java/org/spongepowered/common/command/parameter/SpongeParameterValueBuilder.java
```diff
@@ -90,7 +90,7 @@ public SpongeParameterValueBuilder(@NonNull final Type token) {
     @Override
     public Parameter.Value.@NonNull Builder<T> setRequiredPermission(@Nullable final String permission) {
         if (permission == null) {
-            return this.setUsage(null);
+            return this.setRequirements(null);
```
dualspiral, line 93: ...heh

src/main/java/org/spongepowered/common/command/registrar/BrigadierCommandRegistrar.java
```diff
@@ -84,36 +87,16 @@
...
-            }
-        }
-        this.vanilla.clear();
-    }
-
     @Override
-    public TypeToken<LiteralArgumentBuilder<CommandSource>> handledType() {
-        return BrigadierCommandRegistrar.COMMAND_TYPE;
+    public @NonNull CommandRegistrarType<LiteralArgumentBuilder<CommandSource>> type() {
+        return TYPE;
```
dualspiral: To do a @Zidane - this should be `BrigadierCommandRegistrar.TYPE`, fully qualifying all static access, even if in the same class.

src/main/java/org/spongepowered/common/command/registrar/SpongeParameterizedCommandRegistrar.java
```diff
@@ -58,13 +58,25 @@
...
+
+    public SpongeParameterizedCommandRegistrar(final CommandManager.Mutable commandManager) {
+        this.commandManager = commandManager;
+    }
 
     @Override
-    public TypeToken<Command.Parameterized> handledType() {
-        return SpongeParameterizedCommandRegistrar.COMMAND_TYPE;
+    public @NonNull CommandRegistrarType<Command.Parameterized> type() {
+        return TYPE;
```
dualspiral: As above

src/main/java/org/spongepowered/common/command/registrar/SpongeRawCommandRegistrar.java
```diff
@@ -48,15 +49,19 @@
...
+    SpongeRawCommandRegistrar(final CommandManager.Mutable manager) {
+        this.manager = manager;
+    }
 
     @Override
-    @NonNull
-    public TypeToken<Command.@NonNull Raw> handledType() {
-        return SpongeRawCommandRegistrar.COMMAND_TYPE;
+    public @NonNull CommandRegistrarType<Command.Raw> type() {
+        return TYPE;
```
dualspiral: As above

dualspiral (approved):
Thanks for doing this. I don't really see any issues with this beyond anything stylistic.

PR #3283, opened 2021-02-04
Fix incorrect use of display slot indexes

thibaulthenry:
**Sponge | [SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2295)**

In this PR I'm attempting to fix the bug I found updating display slots in API-8.

The use of the display slot indexes was wrong. So for example when I tried to display an objective to the sidebar, it actually appeared in the sidebar.aqua.

I decided to put an id field on the SpongeDisplaySlot. I tested every cases when display slots were used, and it seems to fix the bug.

Also, I noticed that some the parameters `ChatFormatting color, @Nullable final Function<ChatFormatting, DisplaySlot> withColorFunction` are always null, even for colored display slots.

I can update this PR to fill the `color` parameter, but for `withColorFunction` I don't know what I need to put into.
Need your feedbacks for this.

Also, I made a little change in the API, like we talked about a while ago with Zidane and Masa.

Thanks for reading this PR

### DIFF
--- a/src/main/java/org/spongepowered/common/bridge/scoreboard/ServerScoreboardBridge.java
+++ b/src/main/java/org/spongepowered/common/bridge/scoreboard/ServerScoreboardBridge.java
@@ -49,6 +49,8 @@ public interface ServerScoreboardBridge {
 
     void bridge$updateDisplaySlot(@Nullable Objective objective, DisplaySlot displaySlot) throws IllegalStateException;
 
+    void bridge$updateDisplaySlot(@Nullable Objective objective, int slot) throws IllegalStateException;
+
     Optional<Objective> bridge$getObjective(DisplaySlot slot);
 
     Set<Objective> bridge$getObjectivesByCriterion(Criterion criterion);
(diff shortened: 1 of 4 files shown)


dualspiral:
Hi @thibaulthenry - I'm sorry this hasn't been looked at until now.

`withColorFunction` is really weird here. It's meant to be used as a function to get a `DisplaySlot` based on the supplied colour. For some reason, it's been made an instance method on `DisplaySlot` when really, it should not be.

That said, is there a use for getting an objective by colour? I'm not sure there is. If not, the `withTeamColor` method should be removed, and with it, the `withColorFunction` parameter. If so, that method should be static and (probably) call into a factory that does the selection for you.

thibaulthenry:
Hey, I'll not be able to look at it until this week-end but yeah, I'll try to fix things with your feedbacks

dualspiral:
Supersceded by #3405

Thanks for the initial contribution which I have taken forward (what you had was good but that "withTeamColor" thing has proved to be a bit of a headache).

PR #3288, opened 2021-02-06
Bring back the optimization module and update the mixins

ImMorpheus:
Still need to bring back eigen redstone and review if some of the optimizations are still valid.


ImMorpheus:
I'll reopen this for api-10

PR #3296, opened 2021-02-14
Introduce services for observability

pschichtel:
API PR: SpongePowered/SpongeApi#2298

So far this is only a few random metrics thrown in here to show how the prometheus metrics could be used here. I'm currently only using the global default registry (`CollectorRegistry.defaultRegistry()`, implicity used by the `.register()` calls.

I would expect a sane service implementation to scrape the default registry, so this should just work. Alternatively you could setup a new registry and add it to the `MetricsService` once it becomes available.


src/main/java/org/spongepowered/common/event/tracking/TrackingUtil.java
```diff
@@ -95,6 +97,14 @@
 @SuppressWarnings({"unchecked", "rawtypes"})
 public final class TrackingUtil {
 
+    private static final Counter TRACKING_COUNTER = Counter.build()
+            .name("sponge_tracking_tileentity_count")
+            .register();
+
+    private static final Histogram TILEENTITY_TICK_TIMER = Histogram.build()
+            .name("sponge_tracking_tileentity_tick_seconds")
```
kashike:
```suggestion
            .name("sponge_tracking_blockentity_tick_seconds")
```

src/main/java/org/spongepowered/common/event/tracking/TrackingUtil.java
```diff
@@ -95,6 +97,14 @@
 @SuppressWarnings({"unchecked", "rawtypes"})
 public final class TrackingUtil {
 
+    private static final Counter TRACKING_COUNTER = Counter.build()
+            .name("sponge_tracking_tileentity_count")
```
kashike:
```suggestion
            .name("sponge_tracking_blockentity_count")
```

src/main/java/org/spongepowered/common/event/tracking/TrackingUtil.java
```diff
@@ -95,6 +97,14 @@
...
+            .register();
+
+    private static final Histogram TILEENTITY_TICK_TIMER = Histogram.build()
```
kashike:
```suggestion
    private static final Histogram BLOCKENTITY_TICK_TIMER = Histogram.build()
```

src/main/java/org/spongepowered/common/event/tracking/TrackingUtil.java
```diff
@@ -95,6 +97,14 @@
 @SuppressWarnings({"unchecked", "rawtypes"})
 public final class TrackingUtil {
 
+    private static final Counter TRACKING_COUNTER = Counter.build()
```
kashike: Should mention "blockentity" in the field name to match the counter name.

kashike (changes requested):

Zidane:
@pschichtel Could you update this PR?

pschichtel:
> @pschichtel Could you update this PR?

done. Not sure on where the dependency needs to go, I simply placed it where timings already was.

Besides the dependency, this PR was mostly intended to give an example of how to use this stuff. I have no clue if these metrics are actually useful or what else could be useful. My intention was to leave that up to you. I don't have the time to dig into this.

gabizou:
I can take over the remaining bits of this PR to add a few various histograms and such. I've a fair idea on how to add more pieces of information that may help with observability.

pschichtel:
that would be awesome. I thought about extracting the observer plugin out of the CubeEngine project to act as a reference implementation for plugin devs, maybe that could even be moved to the existing observer repo we have here?

gabizou:
> that would be awesome. I thought about extracting the observer plugin out of the CubeEngine project to act as a reference implementation for plugin devs, maybe that could even be moved to the existing observer repo we have here?

Do you have a link to said project? I'm curious what metrics you've found would be beneficial

pschichtel:
https://github.com/CubeEngine/modules-extra/blob/master/observe/src/main/java/org/cubeengine/module/observe/metrics/impl/SpongeCollector.java

It's still using the old interfaces and not the observer library, but since you asked about the metrics.

PR #3300, opened 2021-02-18
Reimplement Schematic Serialization

gabizou:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2355) | **Sponge** | [Specification](https://github.com/SpongePowered/Schematic-Specification/pull/20)

This is a pending implementation (needs to go through an RFC process with the Schematic Format to be sure) of Schematics with relation to 1.16 and version 3 of the schematic handling.

Due to the nature of 1.15+ Biomes are very dynamic, so having a pointer to the registry is necessary for the Biome palette, but pending some debate, we should be able to settle on whether Archetype/Schematic volumes only have a Palette of RegistryKey<Biome> types.


src/main/java/org/spongepowered/common/world/schematic/SchematicTranslator.java
```diff
@@ -0,0 +1,565 @@
...
+        final int length = updatedView.getShort(Constants.Sponge.Schematic.LENGTH)
+            .orElseThrow(() -> new InvalidDataException("Missing value for: " + Constants.Sponge.Schematic.LENGTH));
+        if (width <= 0
+            || height <= 0
+            || length <= 0) {
+            throw new InvalidDataException(String.format(
+                "Schematic is larger than maximum allowable size (found: (%d, %d, %d) max: (%d, %<d, %<d)",
+                width, height, length, Constants.Sponge.Schematic.MAX_SIZE
+            ));
+        }
```
octylFractal, line 188: This is incorrect, the spec uses `unsigned short` here, which means the negative values should be mapped to positive integers. Depending on how `getShort` works, you may need to first extract an `int`, then validate that against the `[0, 65535]` bounds, and finally encode that as a `short`.

gabizou: Should be fixed

src/main/java/org/spongepowered/common/world/volume/buffer/archetype/AbstractReferentArchetypeVolume.java
```diff
@@ -303,7 +303,7 @@ public int highestYAt(final int x, final int z) {
                         // Order of operations matters here, block states need to be mirrored first, then rotated
                         .mirror(mirror)
                         .rotate(this.transformation.rotation()),
-                    this.transformStreamBlockPosition(e.position())
+                    this.transformStreamBlockPosition(e.position().add(VolumePositionTranslators.BLOCK_OFFSET)).sub(VolumePositionTranslators.BLOCK_OFFSET)
```
dualspiral: @gabizou This was the fix. If you look at `transformStreamBlockPosition`, the comment says:

```
        // we assume that the position is already center adjusted, but since
        // we're not going to want to "flatten" the positions, we need to correctly
        // round and then re-add the offset post transformation
```

It wasn't.

This might not be the best place to fix it, but I'm not sure where it should go in the grand scheme of things.

PR #3076, opened 2020-07-20
Restore IP forwarding for Sponge 8

astei:
A few notable changes:

* Velocity modern forwarding support has been added (in fact, it's the recommended solution)
* Simplified the configuration a bit - if the user enables the BungeeCord module, we assume they do want forwarding

I'm creating this as a draft as this is a bit of functionality that requires configuration to work, but configuration is currently busted.

Thanks to @Zidane for answering my n00b questions about Mixins, this is the first time I've ever made a mixin. I unfortunately needed to use an access transformer to accomplish the Velocity modern forwarding support. I talked to Zidane and he said this was my only option currently.

TODO:
- [ ] Make sure configuration works
- [ ] Move to just an ipforwarding module and reorganize the configuration along those lines
- [ ] Evaluate how #3065 impacts this

PR #3302, opened 2021-02-21
Restore BungeeCord and Velocity support

zml2008:
This restores IP forwarding from api-7, picking up from #3076.

Main changes from that PR are:

- Use Sponge's networking API where possible
- Merge the BungeeCord and Velocity mixin plugins into one toggle
- Add a note before the vanilla offline mode warning listing out the active forwarding mode, and force offline mode when IP forwarding is enabled, since offline mode is required for IP forwarding to work.

The configuration is tweaked somewhat from that PR. Now, to configure Velocity support, the following steps are required:

1. In the `modules` section of `sponge.conf`, set `ip-forwarding` to true.
2. Update the `ip-forwarding` section of the same file to match the following:
```hocon
ip-forwarding {
    # The IP forwarding mode to use with a proxy. Supported values:
    #  - NONE: Do not forward IP addresses
    #  - LEGACY: Use the BungeeCord/pre-1.13 protocol for IP forwarding (CAUTION: This protocol is insecure)
    #  - MODERN: Use the Velocity protocol for IP forwarding
    # When any forwarding mode but NONE is selected, the server will be put into offline mode and will only accept connections from proxies.
    mode=MODERN
    # The player info forwarding secret from your Velocity configuration.
    # Only used with 'MODERN' forwarding mode.
    secret="<velocity secret here>"
}
```

/cc @astei your review would be appreciated


astei (approved):
Looks good to me!

PR #3313, opened 2021-03-07
vanilla: Set up basic integration testing in CI

zml2008:
This check verifies that Mixins apply properly in the production environment. More may be added later.

This is currently included in the production server jar -- if there are any more complex tests needed it may make sense to build a separate source set for tests.


.github/workflows/build-project.yaml
```diff
@@ -43,17 +52,73 @@ jobs:
...
-          ORG_GRADLE_PROJECT_spongePassword: ${{ secrets.SPONGE_MAVEN_REPO_PASSWORD }}
-
-
+          CI_SYSTEM: Github Actions
+          GITHUB_USERNAME: "${{ github.actor }}"
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          ORG_GRADLE_PROJECT_spongeSnapshotRepo: "${{ secrets.SPONGE_MAVEN_SNAPSHOT_REPO_URL }}"
+          ORG_GRADLE_PROJECT_spongeReleaseRepo: "${{ secrets.SPONGE_MAVEN_RELEASE_REPO_URL }}"
+          ORG_GRADLE_PROJECT_spongeUsername: "${{ secrets.SPONGE_MAVEN_REPO_USER }}"
+          ORG_GRADLE_PROJECT_spongePassword: "${{ secrets.SPONGE_MAVEN_REPO_PASSWORD }}"
```
gabizou: These shouldn't be double quoted (or rather, there's no obvious reason why these are being double quoted)

vanilla/src/installer/java/org/spongepowered/vanilla/installer/InstallerMain.java
```diff
@@ -227,9 +228,21 @@ private Path remapMinecraft(final Path librariesDirectory, final Path inputJar,
...
-                    new LorenzRemapper(mappings, ctx.inheritanceProvider())
-            ));
+            atlas.install(ctx -> new JarEntryRemappingTransformer(new LorenzRemapper(mappings, ctx.inheritanceProvider())) {
+                @Override
+                public JarClassEntry transform(final JarClassEntry entry) {
+                    // Skip shaded classes that we know are non-obf
+                    final String name = entry.getName();
+                    if (name.startsWith("it/unimi")
+                        || name.startsWith("com/google")
+                        || name.startsWith("com/mojang/datafixers")
```
gabizou, line 239: might as well list also brigadier as well

.github/workflows/build-project.yaml
```diff
@@ -43,17 +52,73 @@ jobs:
...
+        with:
+          java-version: "${{ matrix.java }}"
+      - uses: actions/cache@v2
+        with:
+          path: "${{ github.workspace}}/libraries"
+          key: "${{runner.os}}-${{matrix.java}}-it-libraries"
+      - name: Download SpongeVanilla server
+        uses: actions/download-artifact@v2
+        with:
+          name: SpongeVanilla Production Jar
```
gabizou, line 87: You can add the caching of the `libraries` folder as well here:
```yaml
- name: Check out repository to use the build.gradle.kts as a hash file
  uses: actions/checkout@v2
  with:
    path: code
- name: Cache dependency libraries folder
  uses: actions/cache@v2
  with:
    path: ./libraries
    key: ${{ runner.os }}-${{ hashFiles('code/build.gradle.kts') }}
```

PR #3314, opened 2021-03-07
notification-ticket

gabizou:
[API](https://github.com/SpongePowered/SpongeAPI/pull/2304) | **Sponge**

Implements the new NotifyNeighborBlockEvent

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/context/transaction/NeighborNotification.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/context/transaction/NeighborNotification.java
@@ -26,32 +26,34 @@ package org.spongepowered.common.event.tracking.context.transaction;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMultimap;
+import net.minecraft.core.BlockPos;
+import net.minecraft.server.level.ServerLevel;
+import net.minecraft.world.level.block.Block;
+import net.minecraft.world.level.block.entity.BlockEntity;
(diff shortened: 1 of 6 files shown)


src/main/java/org/spongepowered/common/event/tracking/context/transaction/NeighborNotification.java
```diff
@@ -106,11 +130,9 @@ public String toString() {
 
     @Override
     public Optional<BiConsumer<PhaseContext<@NonNull ?>, CauseStackManager.StackFrame>> getFrameMutator(
-        @Nullable GameTransaction<@NonNull ?> parent
+        @Nullable final GameTransaction<@NonNull ?> parent
     ) {
-        return Optional.of((context, frame) -> {
-            frame.pushCause(this.locatableBlock.get());
-        });
+        return Optional.of((context, frame) -> frame.pushCause(this.ticketSupplier.get()));
```
gabizou: This is where I'm debating whether there should be more to the frame mutator, but as it stands, it's suiting nicely

gabizou: Like, something I could imagine being useful is:
```java
        return Optional.of((context, frame) -> {
            if (parent instanceof ChangeBlock) {
                frame.pushCause(((ChangeBlock) parent).original);
            }
            frame.pushCause(this.ticketSupplier.get());
        });
```
 to populate the original changeblock being changed (may look at potentially providing the parent transaction as a BlockTransaction accessor if available)

PR #3323, opened 2021-03-10
Enable depedency caching for travis

Taher-Ghaleb:
Would be interested to know why gradle dependencies haven't been cached on Travis. Thank you.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 sudo: required
 dist: trusty
 
+cache:
+  directories:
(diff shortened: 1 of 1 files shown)


zml2008:
We don't use travis anymore, we've moved over to GitHub Actions. Seems like we've just forgotten to get rid of the old `.travis.yml` file. Sorry for the confusion!

Taher-Ghaleb:
I see. Thanks for your feedback.

PR #3333, opened 2021-03-12
Bump guice to 5.0.1

zml2008:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2311) | **Sponge**

This bumps Guice to an [API-compatible fork](https://github.com/zml2008/guice/compare/5.0.1...zml2008:backport/5.0.1) off of Guice 5.0.1

# Why bump?

- Plugin developers want to use Java versions newer that Java 8 for their own plugins
- Guice 4.1.0 has compatibility issues with Java 16 due to its use of internal JDK APIs. Curretly we use instrumentation to add extra opens for modules, but Guice 5 would let us avoid that.
- There are API breaks in Guice 5, and we might as well introduce them to plugins at the same time as the rest of the API breaks in API 8

# Why use a fork?

Unfortunately, the combination of upstream choices and the fact that we are fixed to the version of Guava shipped with Minecraft mean that we are unable to use standard Guice 5.0.1 at runtime

The dependency chain API on ProvisionListeners has been removed upstream, which is used for our SpongeInjectionPoint, and there is currently no suitable alternative API in guice.

Guice also shades its own copy of ASM -- in 4.1.0 that was an outdated version that caused problems when working with plugins. Google exposes shading as a build option, but does not publish unshaded Guice versions.

While Guice appears to compile on an older Guava version without changes, there is still the risk of unexpected binary changes causing problems if we use upstream guice.

For all of those reasons, I've produced a minimal Guice fork with the necessary changes to make Guice 5.0.1 work for Sponge. This fork is API-compatible, so it can be used in implementations without exposing the fact that we are using a fork to SpongeAPI. 

# Alternatives

- Stay with Guice 4.1.0 -- Drawbacks discussed above
- Find an alternative dependency injection solution -- perhaps the best long-term solution, but no obvious choice exists.
- Get upstream to provide an API alternative to the dependency chain system -- there is an open issue at https://github.com/google/guice/issues/1121 that has not had a response in the almost **4 years** since it was opened.

Given the alternatives, I think forking guice is our best way forward for now. It will keep us compatible with modern Java, remain closer to upstream than 4.1.0 was, and not require someone to go to the effort of finding a new DI system and migrating Sponge to use it.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -153,7 +153,6 @@ dependencies {
     implementation("org.mariadb.jdbc:mariadb-java-client:2.0.3")
     implementation("com.h2database:h2:1.4.196")
     implementation("org.xerial:sqlite-jdbc:3.20.0")
-    implementation("com.google.inject:guice:4.1.0")
     implementation("javax.inject:javax.inject:1")
 
     // ASM - required for generating event listeners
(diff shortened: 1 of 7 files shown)


kashike:
The only main concern I have here, is that external implementations, such as Lantern, would be required to either use the same fork, or fork it themselves, to offer the same functionality.

PR #3348, opened 2021-03-24
Fix /xp command not being able to take away experience completely

### LINKED ISSUE #3346: Taking experience levels with /xp -5L  command does not match vanilla behavior
**I am currently running**

```
    SpongeVanilla
    Minecraft: 1.12.2
    SpongeAPI: 7.3-7127e0882
    Sponge: 1.12.2-7.3.0
    SpongeVanilla: 1.12.2-7.3.0
    JVM: 1.8.0_252/64-bit (AdoptOpenJDK)
    OS: Windows 10 (10.0/amd64)
```

- Plugins (6): Minecraft, Mod Coder Pack, SpongeAPI, Sponge, SpongeVanilla

# Issue Description

In vanilla, experience levels can be taken away from players by providing negative values to the `/xp` command (e.g., `/xp -1L`),  If the amount of experience levels taken away would make the player have a negative level, the player's experience is set to 0.

Excerpt from `EntityPlayer`
```java
    public void addExperienceLevel(int levels) {
        this.experienceLevel += levels;
        if (this.experienceLevel < 0) {
            this.experienceLevel = 0;
            this.experience = 0.0F;
            this.experienceTotal = 0;
        }
    ...
```

However, the EntityPlayer mixin overrides this behavior (specifically, the `Math.max`) https://github.com/SpongePowered/Sponge/blob/aa86ca4ec0d768a17c5daebdb9aebce748663be0/src/main/java/org/spongepowered/common/mixin/core/entity/player/EntityPlayerMixin.java#L230-L241 

Thus, when a player has under 1 level of total experience points, `/xp -1L` should set the total XP to 0, but it does not, leaving the total experience points unchanged.

Tommsy64:
Fixes #3346.

### DIFF
--- a/src/main/java/org/spongepowered/common/mixin/core/entity/player/EntityPlayerMixin.java
+++ b/src/main/java/org/spongepowered/common/mixin/core/entity/player/EntityPlayerMixin.java
@@ -233,8 +233,8 @@ public abstract class EntityPlayerMixin extends EntityLivingBaseMixin implements
             return;
         }
         if (!this.dontRecalculateExperience) {
-            final int newLevel = Math.max(this.experienceLevel + levels, 0);
-            postEventAndUpdateExperience(ExperienceHolderUtils.xpAtLevel(newLevel)
+            final int newLevel = this.experienceLevel + levels;
+            postEventAndUpdateExperience(newLevel < 0 ? 0 : ExperienceHolderUtils.xpAtLevel(newLevel)
                     + (int) (this.experience * ExperienceHolderUtils.getExpBetweenLevels(newLevel)));
             ci.cancel();
         }


src/main/java/org/spongepowered/common/mixin/core/entity/player/EntityPlayerMixin.java
```diff
@@ -233,8 +233,8 @@ private void onAddExperienceLevels(final int levels, final CallbackInfo ci) {
             return;
         }
         if (!this.dontRecalculateExperience) {
-            final int newLevel = Math.max(this.experienceLevel + levels, 0);
-            postEventAndUpdateExperience(ExperienceHolderUtils.xpAtLevel(newLevel)
+            final int newLevel = this.experienceLevel + levels;
+            postEventAndUpdateExperience(newLevel <= 0 ? 0 : ExperienceHolderUtils.xpAtLevel(newLevel)
```
dualspiral: Shouldn't this just be `<`, not `<=`? That would be consistent with https://github.com/SpongePowered/Sponge/issues/3346 then.

Tommsy64: Yes, good catch.

PR #3354, opened 2021-03-29
Implement Favicon.Factory

jpenilla:
Depends on https://github.com/SpongePowered/SpongeAPI/pull/2323

### DIFF
--- a/src/main/java/org/spongepowered/common/network/status/SpongeFavicon.java
+++ b/src/main/java/org/spongepowered/common/network/status/SpongeFavicon.java
@@ -155,4 +155,32 @@ public class SpongeFavicon implements Favicon {
             base64.release();
         }
     }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/network/status/SpongeFavicon.java
```diff
@@ -155,4 +155,32 @@ private static BufferedImage decode(String encoded) throws IOException {
             base64.release();
         }
     }
+
+    public static final class FactoryImpl implements Favicon.Factory {
+
+        @Override
```
zml2008, line 139: Could you just move these static `load` methods into the Factory? seems like they're unused.

jpenilla: Yeah, sure.

zml2008 (approved):

PR #3365, opened 2021-04-11
reactivate entityactivation

Faithcaio:
Ported most of the invalid entityactivation code.
But still has some problems:
 - [x] `LevelChunkMixin` active chunk check
 - [x] `WorldInfoBrige#bridge$isValid()`
 - [x] fake Player check in `EntityActivationRange#initializeEntityActivationState` and `checkIfActive`
 - [x] Villager isMating check in `EntityActivationRange#checkEntityImmunities`
 - [x] global config usage in `EntityActivationRange#addEntityToConfig`
 - [x] usage of final fields in `addEntityToConfig` e.g. local `autopopulate` variable is false even if the actual value in the config is true.


Faithcaio:
https://github.com/SpongePowered/Sponge/commit/0e33d585b6a4e991ae56e31f4673ebd43d060acb

PR #3375, opened 2021-04-24
Implement recipe, biome, and sound ClientCompletionKeys

jpenilla:
Sponge | [SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2333)

Implements the `all_recipes`, `available_biomes`, and `available_sounds` `ClientCompletionKey`s. These are not specific argument types, only suggestion providers. In Vanilla, these suggestion providers are always used in combination with the `ResourceLocationArgument`, so that is how they are used here.

One thing I was kind of unsure about in implementing this is that the suggestion providers are of type `SuggestionProvider<CommandSourceStack>`, not `SuggestionProvider<SharedSuggestionProvider>`. To work around this I used an unchecked cast, but like I said I am unsure if it's the right approach.

This PR also fixes an issue where the `ENTITY_SUMMON` `ClientCompletionKey` would not provide any suggestions.

### DIFF
--- a/src/main/java/org/spongepowered/common/command/registrar/tree/builder/ArgumentCommandTreeNode.java
+++ b/src/main/java/org/spongepowered/common/command/registrar/tree/builder/ArgumentCommandTreeNode.java
@@ -25,12 +25,15 @@
 package org.spongepowered.common.command.registrar.tree.builder;
 
 import com.mojang.brigadier.arguments.ArgumentType;
+import com.mojang.brigadier.suggestion.SuggestionProvider;
 import com.mojang.brigadier.tree.CommandNode;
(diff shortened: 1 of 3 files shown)


dualspiral:
Supersceded by https://github.com/SpongePowered/Sponge/pull/3376 - but thanks for taking a stab, it was just the wrong direction.

To answer your point here:

> One thing I was kind of unsure about in implementing this is that the suggestion providers are of type SuggestionProvider<CommandSourceStack>, not SuggestionProvider<SharedSuggestionProvider>. To work around this I used an unchecked cast, but like I said I am unsure if it's the right approach.

We do that all the time - it was the right approach.

PR #3380, opened 2021-04-25
Fix per player scoreboard team friendly fire

aromaa:
The #getTeam was using the per world instance of scoreboard instead of the per player one.

Also change the #canHarmPlayer to use the attacker's scoreboard instead of the attacked player one to detect whatever its allowed to attack the player.

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/core/world/entity/player/PlayerMixin.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/core/world/entity/player/PlayerMixin.java
@@ -63,6 +63,7 @@ import net.minecraft.world.item.enchantment.EnchantmentHelper;
 import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.state.pattern.BlockInWorld;
 import net.minecraft.world.scores.Scoreboard;
+import net.minecraft.world.scores.Team;
 import org.spongepowered.api.block.BlockSnapshot;
 import org.spongepowered.api.entity.living.Humanoid;
 import org.spongepowered.api.entity.living.Living;
@@ -93,6 +94,7 @@ import org.spongepowered.asm.mixin.injection.Redirect;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
 import org.spongepowered.common.SpongeCommon;
 import org.spongepowered.common.bridge.authlib.GameProfileHolderBridge;
+import org.spongepowered.common.bridge.server.level.ServerPlayerBridge;
 import org.spongepowered.common.bridge.world.entity.PlatformEntityBridge;
 import org.spongepowered.common.bridge.world.entity.player.PlayerBridge;
 import org.spongepowered.common.bridge.server.level.ServerLevelBridge;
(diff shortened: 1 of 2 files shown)


src/mixins/java/org/spongepowered/common/mixin/core/world/entity/player/PlayerMixin.java
```diff
@@ -616,4 +618,18 @@ public void attack(final Entity targetEntity) {
...
+    @Inject(method = "canHarmPlayer", at = @At("HEAD"), cancellable = true)
+    private void impl$onCanHarmPlayer(final net.minecraft.world.entity.player.Player other, final CallbackInfoReturnable<Boolean> cir) {
+        if (!(other instanceof org.spongepowered.api.entity.living.player.server.ServerPlayer)) {
+            return;
+        }
+
+        //Check whatever the other player can hit this player
+        //Fixes per player scoreboards which could have different team set up for each player
+        final Team otherTeam = other.getTeam();
+        final Team thisTeam = ((net.minecraft.world.scores.Scoreboard) ((ServerPlayerBridge) other).bridge$getScoreboard()).getPlayersTeam(this.shadow$getScoreboardName());
```
ImMorpheus, line 638: Still not convinced by this.

I wanted to setup a test case for this the other day but it seems adding/removing a player from a team is broken.

aromaa: I made a sample plugin to test this: https://gist.github.com/aromaa/3c43abec63a8a800e056246c532e3cf5

Connect two clients to the server and run /scoreboardteambug. When this patch is applied, the player who sees the prefix [FF TRUE] can hit the other player but not the player with [FF FALSE]. With the current api-8 branch BOTH players can hit each other, which is incorrect.

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/player/PlayerMixin.java
```diff
@@ -623,4 +625,18 @@ public void attack(final Entity targetEntity) {
     public boolean impl$canCallIgniteEntityEvent() {
         return !this.shadow$isSpectator() && !this.shadow$isCreative();
     }
+
+    @Inject(method = "canHarmPlayer", at = @At("HEAD"), cancellable = true)
+    private void impl$onCanHarmPlayer(final net.minecraft.world.entity.player.Player other, final CallbackInfoReturnable<Boolean> cir) {
+        if (!(other instanceof org.spongepowered.api.entity.living.player.server.ServerPlayer)) {
```
dualspiral, line 631: `canHarmPlayer` also exists on `ServerPlayer` - why did you not inject at an appropriate point there instead, rather than do a type check here?

aromaa: To note, this is the paramater of the method, not `this`. I had it previously in the `ServerPlayer` but I decided to change its location.

Whenever you attack another player, you are the attacker (`this`) and the attacked is the `other`. So if the attacker isn't instance of ServerPlayer but the attacked is, then it wouldn't properly check against it.

dualspiral: Yes, `other` - skipped over that detail, got it.

dualspiral:
@ImMorpheus Can you explain what you think might be wrong with this? If there is nothing wrong, can this be merged?

ImMorpheus:
> @ImMorpheus Can you explain what you think might be wrong with this? If there is nothing wrong, can this be merged?

This PR is using `other`'s scoreboard to get both teams. As pointed out [here](https://github.com/SpongePowered/Sponge/pull/3380/files#diff-41442ce6837720fb0cff117700634ad0dd107da493d494f2309293decd55819eR636) per-player scoreboard may setup different teams for each player. It doesn't seem right to me we're ignoring `this` scoreboard since it could have a team that disallow attacking `other`.  
I wanted to setup a reproducible test case but adding/retrieving a player from a team wasn't working as intended (something to do with teamRepresentations, but I haven't dig further).

aromaa:
There are two different use cases for sure;

- Use the attackers view of the world: Allows attacking other players while preventing them from attacking you. This allows the plugin developer to more heavily customize the behavior they want. My sample plugin demonstrates this by allowing only one of the players to attack the other. Also the plugin developer can use the same team setup on both players to allow them attacking each other (or to prevent).
- Use the attackers and attacked view of the world: When the attacker is allowed to attack but the attacked is not, prevent the action. This has less control over what happens, and the sample plugin I provided is now impossible to do without listening to damage events.

In my opinion, the first option makes sense as the plugin developer can fallback to both behaviors.

Zidane:
I think @aromaa 's explanation makes sense. We can certainly circle back to this if someone brings this back to a head.

PR #3383, opened 2021-04-26
Fix wrong cast in SpongeColorValueParameter

jpenilla:
Builds a command context instead of casting CommandCause to CommandContext. Tested to fix suggestions.

It's worth noting that two of the accepted formats get highlighted red on the client (`#ffffff` and `111,111,111`), however I'm not sure there is anything we can really do about that beyond allowing quoted input.


dualspiral:
Opted to sidestep this completely by just getting the names directly - as we're adding additional ways to create a colour we shouldn't feel beholden to using that argument (I was really hoping that Mojang would make that argument support all colours, alas).

https://github.com/SpongePowered/Sponge/commit/088b603afe28d62e718d6cd4ed58b46e36608181#diff-9470d0be082b1b8b7baba999f5d41bd9d94d48cc0d83cc27ceb17ae388e63b36R134-R138

Thanks for pointing out the issue though.

PR #3389, opened 2021-04-30
Filter degenerate AABBs in World#getIntersectingCollisionBoxes

Yeregorix:
Someone reported to me [an issue](https://github.com/Yeregorix/GhostBuster/issues/1) where `World#getIntersectingCollisionBoxes` throws an internal IllegalArgumentException. This PR fixes this issue.


kashike:
Is there a *good* reason we don't just allow creating AABBs that are `0` for everything?

Yeregorix:
It would certainly be simpler to allow it in API. It may also avoid future issues. Currently `VecHelper#toSpongeAABB` must always be wrapped in a try-catch.

Before writing this PR, I tried to sound out people opinion on Discord but I got only one answer: https://discord.com/channels/142425412096491520/539537645152239626/833622843539324958.

kashike:
I think it might be worth changing the API, and allowing this instead.

PR #3390, opened 2021-05-01
Small fixes to multi-block captures

gabizou:
This aims to resolve the case of how some blocks perform multiple changes as a reaction to being broken by a player, specifically beds. 
 
The resulting transaction lists look like:
- ChangeBlockEvent for the first half of the bed
- NeighborNotifications for said first half being broken
- ChangeBlockEvent for the second half of the bed
- NeighborNotifications for the said second half being broken


src/main/java/org/spongepowered/common/event/tracking/context/transaction/TransactionalCaptureSupplier.java
```diff
@@ -176,6 +177,17 @@ public boolean logTileRemoval(final @Nullable BlockEntity tileentity, final Supp
             if (newRecorded) {
                 return true;
             }
+            if (this.tail.hasChildTransactions()) {
```
gabizou, line 186: @dualspiral this is where I'd say it's a little hairy to do a reverse traversal, but I feel it should do a reverse traversal  to try and find the "latest" transaction that would accept a tile replacement, but maybe I wonder if there's the remote chance that it could occur as a step of a previous transaction than tail and not through tail's side effects.

PR #3391, opened 2021-05-01
improve PlayerChatEvent

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2340) | **SpongeCommon**


src/main/java/org/spongepowered/common/adventure/SpongeAdventure.java
```diff
@@ -731,6 +732,13 @@ public static SoundSource asVanilla(final Sound.Source source) {
         return SpongeAdventure.asVanilla(source);
     }
 
+    public static Iterable<? extends Audience> unpackAudiences(Audience audience) {
```
kashike, line 735: This will not work as-is in some situations - audiences can be nested.

kashike (changes requested):

PR #3393, opened 2021-05-01
Implement DataSerializable for ServerLocation

ItsDoot:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2341) | **Sponge**


src/main/java/org/spongepowered/common/world/server/SpongeServerLocationBuilder.java
```diff
@@ -0,0 +1,30 @@
+package org.spongepowered.common.world.server;
```
dualspiral, line 25: Missing the licence here (easily done!)

ItsDoot: Woops, fixed!

dualspiral:
8bf35817def81dabb85f0c6c078795d54d820221

PR #3395, opened 2021-05-02
Add editor config and sort imports

gabizou:


.editorconfig
```diff
@@ -0,0 +1,364 @@
...
+ij_java_field_annotation_wrap = split_into_lines
+ij_java_finally_on_new_line = false
+ij_java_for_brace_force = always
+ij_java_for_statement_new_line_after_left_paren = false
+ij_java_for_statement_right_paren_on_new_line = false
+ij_java_for_statement_wrap = on_every_item
+ij_java_generate_final_locals = true
+ij_java_generate_final_parameters = true
+ij_java_if_brace_force = always
+ij_java_imports_layout = $*,|,*,|,org.spongepowered.**,|,com.mojang.**,net.minecraft.**,|,javax.**,java.**
```
kashike, line 120: I'm not a fan of these separate import groups, can we just have imports normally please?

kashike (changes requested):

.editorconfig
```diff
@@ -0,0 +1,364 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+max_line_length = 120
```
ImMorpheus, line 9: This should be 150.

.editorconfig
ImMorpheus, line 120: > Imports must be grouped in the following order, where each group is separated by an empty line
> 
>     Static imports
> 
>     All other imports
> 
>     java imports
> 
>     javax imports

ImMorpheus (changes requested):
This is not consistent with what we are doing in the API/checkstyle and what's written in the docs.

https://docs.spongepowered.org/stable/en/contributing/implementation/codestyle.html

I think we should PR this to the API to discover any issue with the current checkstyle config before tackling the implementation.
If needed, the docs should be updated to reflect the new guidelines.

Zidane:
Closing this as we need to have a much more inclusive discussion as a dev team on editor formatting changes.

PR #3400, opened 2021-05-04
Implemented ChangeDataHolderEvent.ValueChange for FOOD_LEVEL, SATURATION & EXHAUSTION

aromaa:


src/mixins/java/org/spongepowered/common/mixin/core/world/entity/player/PlayerMixin.java
```diff
@@ -616,4 +617,13 @@ public void attack(final Entity targetEntity) {
     public boolean impl$canCallIgniteEntityEvent() {
         return !this.shadow$isSpectator() && !this.shadow$isCreative();
     }
+
+    @Redirect(method = "<init>", at = @At(value = "NEW", target = "net/minecraft/world.food/FoodData"))
```
aromaa: I'm not really sure about this one..

src/mixins/java/org/spongepowered/common/mixin/core/world/food/FoodDataMixin.java
```diff
@@ -0,0 +1,121 @@
...
+    // @formatter: off
+    @Shadow private int foodLevel;
+    @Shadow private float saturationLevel;
+    @Shadow private float exhaustionLevel;
+    // @formatter: on
+
+    private Player impl$player;
+
+    @Override
+    public void bridge$setPlayer(Player player) {
```
dualspiral: All parameters should be `final`

src/mixins/java/org/spongepowered/common/mixin/core/world/food/FoodDataMixin.java
```diff
@@ -0,0 +1,121 @@
...
+    private void impl$addExhaustion(FoodData self, float value) {
+        this.exhaustionLevel = this.fireEventAndGetValue(Keys.EXHAUSTION, (double) this.exhaustionLevel, (double) value).floatValue();
+    }
+
+    @Redirect(method = "tick", at = @At(value = "FIELD", target = "Lnet/minecraft/world/food/FoodData;exhaustionLevel:F", opcode = Opcodes.PUTFIELD))
+    private void impl$tickDrainExhaustion(FoodData self, float value) {
+        this.exhaustionLevel = this.fireEventAndGetValue(Keys.EXHAUSTION, (double) this.exhaustionLevel, (double) value).floatValue();
+    }
+
+    private <E> E fireEventAndGetValue(Key<? extends Value<E>> key, E currentValue, E value) {
```
dualspiral: `impl$fireEventAndGetValue`

I'm also a little concerned about the spamminess of this. This needs to make sure that, at the very least, if `Objects.equals(currentValue, value)` is `true`, then this all gets short circuited because nothing does change. We might also want to wrap this in a `ShouldFire` so it doesn't fire if no-one is listening to it (see the ShouldFire class for more details on how that works).

aromaa: The exhaustion one is spammy as that one is called every tick. I could drop that but implemented it for completeness. What do you think?

dualspiral: No, don't drop it right this second, right now I'm just putting my thoughts out there. I imagine some of this can be mitigated in the ways I outline above though - if the value doesn't change, we don't need to fire the event.

aromaa: Sure, though it is modified every time player moves or interacts with the world. I myself don't see the exhaustion as necessary due the common case being for food related events to prevent it from draining, and the saturation is the last "visible" part that that the client sees. So in my eyes, cancelling saturation for that purpose is enough.

dualspiral: > Sure, though it is modified every time player moves or interacts with the world

There are points in your injections where there might not be a change - `addExhaustion` may remain capped at 40 and it's possible on `tick` that `foodLevel` will remain at zero. Quite possible with `setFoodLevel` too.

> I myself don't see the exhaustion as necessary

But it is a data value and conceviably someone might want to monitor it.

src/mixins/java/org/spongepowered/common/mixin/core/world/food/FoodDataMixin.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        for (final Value.Immutable<?> immutableValue : event.endResult().successfulData()) {
+            if (immutableValue.key() == key) {
+                return (E) immutableValue.get();
+            }
+        }
+
+        return currentValue;
+    }
```
dualspiral, line 127: @gabizou Just want to check with you to make sure this is the right way to do `DataHolderEvent.ValueChange` events for specific keys. If it looks okay, this is good to go, if it doesn't, let us know how to sort it (@aromaa I can take over if you're unable to complete this, I know you said you have things coming up!)

PR #3413, opened 2021-05-23
Adventure Pointers

kashike:
https://github.com/KyoriPowered/adventure/pull/352

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/api/mcp/server/level/ServerPlayerMixin_API.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/api/mcp/server/level/ServerPlayerMixin_API.java
@@ -29,6 +29,8 @@ import net.kyori.adventure.audience.MessageType;
 import net.kyori.adventure.bossbar.BossBar;
 import net.kyori.adventure.identity.Identity;
 import net.kyori.adventure.inventory.Book;
+import net.kyori.adventure.permission.PermissionChecker;
+import net.kyori.adventure.pointer.Pointer;
 import net.kyori.adventure.sound.Sound;
 import net.kyori.adventure.sound.SoundStop;
 import net.kyori.adventure.text.Component;
@@ -116,6 +118,8 @@ public abstract class ServerPlayerMixin_API extends PlayerMixin_API implements S
 
     @Shadow public abstract net.minecraft.server.level.ServerLevel shadow$getLevel();
     // @formatter:on
+
+    private PermissionChecker adventure$permissions;
 
     private final TabList api$tabList = new SpongeTabList((net.minecraft.server.level.ServerPlayer) (Object) this);
     @Nullable private PlayerChatFormatter api$chatRouter;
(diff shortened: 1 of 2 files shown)


src/mixins/java/org/spongepowered/common/mixin/api/mcp/server/level/ServerPlayerMixin_API.java
```diff
@@ -117,6 +121,8 @@
     @Shadow public abstract net.minecraft.server.level.ServerLevel shadow$getLevel();
     // @formatter:on
 
+    private net.kyori.adventure.permission.PermissionChecker adventure$permissions;
+
```
gabizou, line 123: NIT: Extra space and unnecessary qualified import

src/mixins/java/org/spongepowered/common/mixin/api/mcp/server/level/ServerPlayerMixin_API.java
```diff
@@ -384,6 +390,34 @@ public void setWorldBorder(final @Nullable WorldBorder border) {
 
     // Audience
 
+    @Override
+    @SuppressWarnings("unchecked") // safe casts
+    public <T> Optional<T> get(final Pointer<T> pointer) {
+        if (pointer == Identity.NAME) {
```
gabizou, line 402: I'd rather these are just shoved into a map of Pointer to function, having this large amount of if statements is silly.

src/mixins/java/org/spongepowered/common/mixin/api/mcp/server/level/ServerPlayerMixin_API.java
```diff
@@ -384,6 +390,34 @@ public void setWorldBorder(final @Nullable WorldBorder border) {
...
+                        return TriState.FALSE;
+                    } else if (value == Tristate.TRUE) {
+                        return TriState.TRUE;
+                    }
+                    throw new IllegalArgumentException(value.name());
+                };
+            }
+            return Optional.of((T) this.adventure$permissions);
+        }
+        return super.get(pointer);
```
gabizou: What is super.get(pointer) going to default to?

gabizou (changes requested):

src/mixins/java/org/spongepowered/common/mixin/api/mcp/server/level/ServerPlayerMixin_API.java
zml2008: tbh I'd like to see this become a switch expression once Java supports those properly, but for now I'm not sure I see the benefit of a map *unless* the set of supported pointers is extendable.

kashike: I see no reason for a map of functions to exist - that is just going to introduce unnecessary complexity here. There is **far** worse code in the repository.

gabizou: Well sure, there might be more pointers in the future, but like from the standpoint of "if your if else is more than three, then it's silly to make it an if else..." But I'll admit that when java supports proper switches, that'll be nice.

dualspiral: I don't expect that there will be more pointers though because these are generally used in _identity_ and there really doesn't need to be any more.

For what it's worth, if this does expand - from a performance point of view, a hash map may be better. For a four item if statement - I don't think it matters. Personally, if I was to do this I'd have done a map because that's how we did it in some of our previous professional projects and it made it super easy to extend. It's not that complicated but it might not be obvious to young developers. Java doing switch properly would put all of this to rest because then it'll decide the right way to do it.

An if statement is fine here. It's not like this is going to get hit all that often, if at all really, and I don't see it being extended, as I said.

> There is far worse code in the repository.

As a note, we shouldn't be trying to race to the bottom as a baseline for the standards we should uphold. This code isn't actually bad, but we should _never_ justify a design decision based on the worst in the repo. It's a big project, there is bad code. Readable, or at least commented (especially if it's "clever") code is what we aim for here.

src/mixins/java/org/spongepowered/common/mixin/api/mcp/server/level/ServerPlayerMixin_API.java
```diff
@@ -117,6 +119,8 @@
     @Shadow public abstract net.minecraft.server.level.ServerLevel shadow$getLevel();
     // @formatter:on
 
+    private PermissionChecker adventure$permissions;
```
dualspiral, line 122: This should be `api$`, not `adventure$` - all fields in an API mixin should be prefixed that way, not with the library it comes from.

PR #3414, opened 2021-05-24
Initial update of block & block entity data

zml2008:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2357) | **Sponge**


src/main/java/org/spongepowered/common/data/provider/block/state/SimpleWaterloggedBlockData.java
```diff
@@ -24,24 +24,26 @@
...
-                .asImmutable(BlockState.class)
-                    .create(Keys.IS_WATERLOGGED)
-                        .get(h -> h.getValue(CrossCollisionBlock.WATERLOGGED))
-                        .set((h, v) -> h.setValue(CrossCollisionBlock.WATERLOGGED, v))
-                        .supports(h -> h.getBlock() instanceof CrossCollisionBlock);
+            .asImmutable(BlockState.class)
+            .create(Keys.IS_WATERLOGGED)
+            .get(h -> h.getValue(SeaPickleBlock.WATERLOGGED))
+            .set((h, v) -> h.setValue(SeaPickleBlock.WATERLOGGED, v))
+            .supports(h -> h.getBlock() instanceof SimpleWaterloggedBlock);
```
gabizou: This doesn't make sense, is it WATERLOGGED from SimpleWaterloggedBlock? Or from SeaPickleBlock?

zml2008: oh rip, that was a bad copy-paste -- `SimpleWaterloggedBlock` refers to `BlockStateProperties.WATERLOGGED` which this should use instead.

PR #3419, opened 2021-06-03
Ensure mapIdUUIDIndex is initialized for getting a mapinfo.

tyhdefu:
Its possible (though very unlikely) for a NPE if you try to retrieve a MapInfo by UUID, before any maps have been loaded in the world (e.g in the player's hand). Just ensure its grabbed in this rare case.


dualspiral:
I took the moment to go over the various map impl classes and fixed up the formatting in them, and pulled this in with it.

PR #3421, opened 2021-06-06
Add SteerVehicleEvent

### LINKED ISSUE #2283: Add SteeringEvent
We've been listening [STEER_VEHICLE packet](https://wiki.vg/Protocol#Steer_Vehicle) with [ProtocolLib's STEER_VEHICLE](https://github.com/aadnk/ProtocolLib/blob/master/src/main/java/com/comphenix/protocol/PacketType.java#L322) in our vehicle plugin.

 Since Sponge does not expose packets, I suggest adding some sort of SteeringEvent that could allow listening to steering events.

avaruus1:
Sponge | [SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2364)

Added SteerVehicleEvent. Resolves SpongePowered/SpongeAPI#2283

While having the event be cancellable would be ideal, I’m not sure how that could be done or if it’s even possible, since Mojang changed the client to handle dismounting fully client-side in 1.16.

Regardless of that, I hope this event makes it to API-8, since I believe it would be of great use when creating vehicle plugins.


dualspiral:
Gonna need an API PR too!

src/mixins/resources/mixins.sponge.core.json
```diff
@@ -1,248 +1,249 @@
...
+    "commands.CommandsMixin",
+    "commands.CommandSourceMixin",
+    "commands.CommandSourceStackMixin",
+    "commands.arguments.ComponentArgumentMixin",
+    "commands.arguments.DimensionArgumentMixin",
+    "commands.arguments.selector.EntitySelectorParserMixin",
+    "core.RegistryAccessMixin",
+    "core.Vec3iMixin",
+    "core.dispenser.AbstractProjectileDispenseBehaviorMixin",
+    "data.SpongeDataHolderMixin",
```
gabizou: Don't change the formatting here on a whim please.

src/mixins/java/org/spongepowered/common/mixin/core/network/ServerGamePacketListenerImplMixin.java
```diff
@@ -0,0 +1,58 @@
...
+import org.spongepowered.common.event.tracking.PhaseTracker;
+
+@Mixin(ServerGamePacketListenerImpl.class)
+public abstract class ServerGamePacketListenerImplMixin implements ServerGamePacketListener {
+
+    @Shadow public ServerPlayer player;
+
+    @Inject(method = "handlePlayerInput", at = @At("HEAD"))
+    public void onHandlePlayerInput(ServerboundPlayerInputPacket packet, CallbackInfo ci) {
+        SpongeCommon.postEvent(SpongeEventFactory.createSteerVehicleEvent(
```
gabizou: Is the event cancellable? If so, you can make the injection cancellable

src/mixins/java/org/spongepowered/common/mixin/core/network/ServerGamePacketListenerImplMixin.java
```diff
@@ -0,0 +1,58 @@
...
+import org.spongepowered.common.SpongeCommon;
+import org.spongepowered.common.event.tracking.PhaseTracker;
+
+@Mixin(ServerGamePacketListenerImpl.class)
+public abstract class ServerGamePacketListenerImplMixin implements ServerGamePacketListener {
+
+    @Shadow public ServerPlayer player;
+
+    @Inject(method = "handlePlayerInput", at = @At("HEAD"))
+    public void onHandlePlayerInput(ServerboundPlayerInputPacket packet, CallbackInfo ci) {
```
gabizou: Make these variables final please 
```suggestion
    public void onHandlePlayerInput(final ServerboundPlayerInputPacket packet, final CallbackInfo ci) {
```

src/mixins/java/org/spongepowered/common/mixin/core/network/ServerGamePacketListenerImplMixin.java
```diff
@@ -0,0 +1,58 @@
...
+        SpongeCommon.postEvent(SpongeEventFactory.createSteerVehicleEvent(
+                PhaseTracker.getCauseStackManager().currentCause(),
```
gabizou: Is it guaranteed that we'll always be on the main thread here?

src/mixins/java/org/spongepowered/common/mixin/core/network/ServerGamePacketListenerImplMixin.java
avaruus1: Pre 1.16 that would work fine, but that is not the case in newer vesions since dismounting is handled fully client-side.

src/mixins/resources/mixins.sponge.core.json
avaruus1: Not exactly sure how I managed to do that, but it should be fixed now :sweat_smile:

src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerPlayerMixin.java
```diff
@@ -903,6 +904,19 @@ public void sendMessage(final net.minecraft.network.chat.Component p_241151_1_,
             }
         }
     }
+
+    @Inject(method = "setPlayerInput", at = @At("HEAD"))
+    public void onSetPlayerInput(final float sway, final float surge, final boolean jumping, final boolean dismount, final CallbackInfo ci) {
+        SpongeCommon.postEvent(SpongeEventFactory.createSteerVehicleEvent(
```
dualspiral: This is going to fire if the Player Input Packet is sent, regardless of whether the player is actually mounted. It would make more sense to redirect `isPassenger` and fire there if `isPassenger` would return `true`. You could then return `false` if you wanted to cancel this packet, which kind of leads me into my next question:

> Pre 1.16 that would work fine, but that is not the case in newer vesions since dismounting is handled fully client-side.

...how does that make sense? The server still has to know about this and validate it. Is it because the client dismounts without validation from the server?

(I'll test it at some point, just haven't looked into it yet)

avaruus1: > This is going to fire if the Player Input Packet is sent, regardless of whether the player is actually mounted. It would make more sense to redirect isPassenger and fire there if isPassenger would return true.

That's probably a good idea :sweat_smile: 

> Is it because the client dismounts without validation from the server?

That's exactly what I meant! Sorry for being unclear.

src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerPlayerMixin.java
```diff
@@ -830,6 +831,30 @@ public void sendMessage(final net.minecraft.network.chat.Component p_241151_1_,
             }
         }
     }
+
+    @Inject(method = "setPlayerInput", at = @At("HEAD"), cancellable = true)
```
dualspiral: This should be a redirect on `isPassenger`, so you can avoid calling it twice, and just return `false` here to short circuit the if statement.

src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerPlayerMixin.java
```diff
@@ -830,6 +831,30 @@ public void sendMessage(final net.minecraft.network.chat.Component p_241151_1_,
...
+    public void onSetPlayerInput(final float sway, final float surge, final boolean jumping, final boolean dismount, final CallbackInfo ci) {
```
dualspiral: `private boolean impl$onSetPlayerInput(final ServerPlayer player, final float sway, final float surge, final boolean jumping, final boolean dismount)`

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
```diff
@@ -204,6 +204,8 @@
     @Shadow protected abstract boolean shadow$fireImmune();
     // @formatter:on
 
+    @Shadow public abstract boolean isPassenger();
```
dualspiral: This should be put in the `// @formatter` block above, and prefixed with `shadow$`, so 

```
@Shadow public abstract boolean shadow$isPassenger();
```

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/LivingEntityMixin.java
```diff
@@ -170,6 +170,9 @@
     @Shadow public abstract void shadow$clearSleepingPos();
     // @formatter:on
 
+    @Shadow public float xxa;
+    @Shadow public float zza;
+    @Shadow protected boolean jumping;
```
dualspiral: Put these in the `// @formatter` block above.

src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerPlayerMixin.java
```diff
@@ -830,6 +831,30 @@ public void sendMessage(final net.minecraft.network.chat.Component p_241151_1_,
...
+        if(isPassenger()) {
```
dualspiral: We prefix all same class accesses with `this.` (but this would become `this.shadow$isPassenger()`)

Also, space after `if`

src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerPlayerMixin.java
```diff
@@ -830,6 +831,30 @@ public void sendMessage(final net.minecraft.network.chat.Component p_241151_1_,
...
+            if(sway == this.xxa && surge == this.zza && jumping == this.jumping) {
+                return;
```
dualspiral: This would `return true;` as this is a success condition - everything is the same but the action isn't being cancelled, so continue.

src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerPlayerMixin.java
```diff
@@ -830,6 +831,30 @@ public void sendMessage(final net.minecraft.network.chat.Component p_241151_1_,
...
+                    jumping,
+                    surge,
+                    sway
+            );
+
+            Sponge.eventManager().post(event);
+
+            if(event.isCancelled()) {
+                ci.cancel();
+            }
```
dualspiral: You could replace this with 

```suggestion
            return !Sponge.eventManager().post(event);
```

src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerPlayerMixin.java
```diff
@@ -830,6 +831,30 @@ public void sendMessage(final net.minecraft.network.chat.Component p_241151_1_,
...
+        }
+    }
```
dualspiral, line 874: You'll want to return false at the end.

ImMorpheus:

___

PR #3422, opened 2021-06-11
Tag API Implementation

tyhdefu:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2365) | Sponge


tyhdefu:
I've created a registry that will can be "refreshed", which is refreshed when datapacks are. I think this is a good way to deal with refreshing datapacks, and it seems to work, but disabling plugin datapacks doesn't seem to workfully, though they are being refreshed. I think this will probably be a good way to do advancements etc.

src/main/java/org/spongepowered/common/registry/RefreshableRegistry.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.Set;
+import java.util.function.Consumer;
+
+/**
+ * A registry that can be refreshed at runtime,
+ * reloading its contents. Does not support integer
+ * ids as those would be unstable.
+ * @param <T>
+ */
+public class RefreshableRegistry<T> extends Registry<T> {
```
dualspiral: Honestly, this seems wrong to me and works against how registries are meant to work. Engine and Level bound registries are examples where we could have used such a system instead, using the concept of "refreshing" to create Sponge registry "views" if you will, but we haven't.

If we were to go down this road, would we be better having a datapack scoped registry instead? That way, when the datapacks are reloaded the registry is thrown away and repopulated? 

That said, should we really be trying to shoehorn this tagging system into registries - is this a round peg, square hole sort of deal?

I'm not saying we should follow Mojang's lead on this, absolutely not, if we just followed what Mojang did all the time I'd argue that Sponge would be worthless, but this does feel like we're making the solution fit the problem.

tyhdefu: I've moved the registries to a new RegistryScope - DATA_PACK. They are now acessable via Sponge.game().dataPackManager().registries()
Currently the actual instance of DataPackManager doesn't change, just the registry inside it, but it functions the same within the API (assuming you don't store DataPackManager).
Let me know what you think of this. I think that scoping them under data pack is the correct way to go.
I also added loadedPacks() to demonstrate that there could easily be more things in here - Reloading it from a plugin, enabling/disabling datapacks?

dualspiral: Well, I threw out this as a thought and a question, not necessarily the route we should go down and wanted more of a discussion on it. Specifying a whole new registry holder is a much larger thing than adding a Tag API and it does need some thought.

@Zidane I'd appreciate you weighing in on this as someone who basically built the registry system on top of Vanilla. Do we want to have this notion of a datapack scoped registry? I don't necessarily mind the idea but it is another layer of complexity and as I said above, this does feel a bit like we're forcing the problem into a weird solution (there has to be a reason Mojang didn't go down the registry route here, instead opting for tag collections...) - though it is far better than the "RefreshableRegistry" that was there. 

If that's what we want, we need to think about how we represent this.

gabizou: Just throwing a quick two cents: DataPack scoped registries I think aren’t really feasible from the perspective of when they’re loaded/swapped. In the interest of KISS, a data pack does cause a reload, but it’s still effectively game/server based, right? From the server’s perspective, during a datapack reload, the server’s registries are reloaded for the datapack’s changed/missing options, not something the datapack can explicitly dictate individually right?

dualspiral: > In the interest of KISS, a data pack does cause a reload, but it’s still effectively game/server based, right?

Server based.

> From the server’s perspective, during a datapack reload, the server’s registries are reloaded for the datapack’s changed/missing options, not something the datapack can explicitly dictate individually right?

No - updating/reloading registries play no part in the vanilla datapack system, which is kind of my point here. Here, we're using registries to represent datapack registered values when perhaps we shouldn't be. Advancements, recipes and tags all use their own managers/collections that can be reloaded and so if we were to use the registry system for this, we'd need to replace the registry to rebuild it.

I just don't want to end up forcing this into the registry system when it might not fit well.

(edit, originally said we were using registries already for datapack stuff, we're not as far as I can tell)

Zidane: So this is the right idea but the wrong approach.

I do think the data pack stuff should have a place in our registry system but everyone needs to remember that data pack content eventually creates registry or registry-like content.

`WorldTypeTemplate` (Data Pack) -> `WorldType` (Registry)

The issue is follow through, making sure we keep doing that. Data Pack types in the API should only be used to build templates and pass to the engine for processing then discarded. Only the built types will matter for registry lookups.

With that out of the way, lets discuss this addition. Vanilla has no registry for tags...no traditional one that is. What they do have is `StaticTagHelper` which...is a stone age registry-like construct. Why Mojang didn't make it a full registry is a question for them...you could likely argue they didn't because going into the registry system now binds you to several contract mechanisms. While I feel that Mojang doesn't make enough registry, API-7 made too many things registry that should not be.

That said, Tags sit on the fence. If you read `StaticTagHelper`, you realize that all `TagTypes` have a `ResourceLocation` as well as all bound `Tag`s have a `ResourceLocation` via the `bind` and `StaticTagHelper.Wrapper` system. In fact, Mojang could have easily did a similar thing you did here...create a `Registry` subclass that can reset itself and register it in the system. Again, question for them why they didn't.

The question we need to answer in Sponge is simply: Do we have this go into the registry framework or simply add some `TagManager` to the API with typical lookups for `Tag` objects that wraps around `StaticTagHelper`. If we are going to stick with the former, your implementation of a `Registry` for them is in-correct, it actually needs to wrap around `StaticTagHelper` to retain mod support.

Zidane: In fact, wrapping around `StaticTagHelper` will basically invalidate any resetting/etc that we are doing as the base game does this anyways. As a bonus, the contract in api-8 for registry means it is legal for the underyling registry to be reset or changed at any time. This is why we do the reference system.

Good idea to keep in mind while discussing this.

src/mixins/java/org/spongepowered/common/mixin/api/mcp/tag/BlockTypeTaggableMixin_API.java
```diff
@@ -0,0 +1,51 @@
...
+import org.spongepowered.api.tag.TagType;
+import org.spongepowered.api.tag.TagTypes;
+import org.spongepowered.api.tag.Taggable;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.common.tag.TagUtil;
+
+import java.util.Collection;
+
+@Mixin(Block.class)
+public abstract class BlockTypeTaggableMixin_API implements Taggable<BlockType> {
```
Zidane: No, this goes onto the API mixin we already have.

src/main/java/org/spongepowered/common/registry/RefreshableRegistry.java
tyhdefu: I quite like the idea of wrapping around StaticTagHelper. That means there will be no reseting / invalidation needed.
I've made it so that sponge's tag mixes into Tag.Wrapper, which wraps a tag and allows it to reload (A vanilla design).
Forge seems to recommend the use of datapacks to add tags, so I wouldn't have too many concerns about mods.
Wrapping around StaticTagHelper is definitely the way to go, no need to refreshing registries. (Assuming we go for the registry design).

tyhdefu: Wrapped around StaticTagHelper.

src/main/java/org/spongepowered/common/registry/TagRegistry.java
```diff
@@ -0,0 +1,184 @@
...
+        final ResourceLocation valueLocation = this.getKey(var1);
+        if (valueLocation == null) {
+            return Optional.empty();
+        }
+        return Optional.of(net.minecraft.resources.ResourceKey.create(this.key(), valueLocation));
+    }
+
+    @Override
+    public int getId(@Nullable final Tag<T> var1) {
+        throw new UnsupportedOperationException("No ids not supported in TagRegistry!");
```
dualspiral: This error message doesn't make sense.

"IDs are not supported."

src/main/java/org/spongepowered/common/registry/TagRegistry.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    @Nullable
+    @Override
+    public Tag<T> byId(final int var1) {
+        throw new UnsupportedOperationException("No ids not supported in TagRegistry!");
```
dualspiral: As above.

src/main/java/org/spongepowered/common/registry/TagRegistry.java
```diff
@@ -0,0 +1,184 @@
...
+    public boolean containsKey(ResourceLocation var1) {
+        return this.staticTagHelper.getAllTags().getAllTags().containsKey(var1);
+    }
+
+    @Override
+    public boolean isDynamic() {
+        // Its dynamic, but this is not the place to register.
+        // I think its best to return true, because its possible a plugin
+        // would use this to determine whether it can be cached, and it can't.
+        return true;
```
dualspiral: No, this should return `false`. `isDynamic`'s javadocs say the following:

> Returns if this registry supports adding additional values

It actually does not - really all this "registry" is is a view. We also don't want to fire the `RegisterRegistryValueEvent` for this registry. 

If we want to indicate what you're asking for, we _might_ be inclined to add a "isView" method on `Registry` or have some `ViewRegistry`, though I don't think that's really necessary, plugins shouldn't be caching values in a registry at any stage, that's what a registry is for!

src/main/java/org/spongepowered/common/registry/TagRegistry.java
```diff
@@ -0,0 +1,208 @@
...
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@org.checkerframework.framework.qual.DefaultQualifier(org.checkerframework.checker.nullness.qual.NonNull.class)
+public class TagRegistry<T> extends net.minecraft.core.Registry<Tag<T>> implements Registry<Tag<T>> {
```
dualspiral: All classes should be final unless something inherits from them.

src/main/java/org/spongepowered/common/registry/TagRegistry.java
```diff
@@ -0,0 +1,208 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.registry;
+
+import com.google.common.collect.HashBiMap;
+import com.mojang.serialization.Lifecycle;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.tags.StaticTagHelper;
+import org.jetbrains.annotations.NotNull;
```
dualspiral: `org.checkerframework.checker.nullness.qual.NonNull`

src/main/java/org/spongepowered/common/registry/TagRegistry.java
```diff
@@ -0,0 +1,208 @@
...
+import org.jetbrains.annotations.Nullable;
```
dualspiral: `org.checkerframework.checker.nullness.qual.Nullable`

src/main/java/org/spongepowered/common/registry/TagRegistry.java
```diff
@@ -0,0 +1,208 @@
...
+import java.util.AbstractMap;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@org.checkerframework.framework.qual.DefaultQualifier(org.checkerframework.checker.nullness.qual.NonNull.class)
```
dualspiral: Import this annotation, don't fully qualify here

generator/src/main/java/org/spongepowered/vanilla/generator/TagGenerator.java
```diff
@@ -0,0 +1,115 @@
...
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.tags.Tag;
+
+import javax.lang.model.element.Modifier;
+import java.io.IOException;
+import java.util.Comparator;
+import java.util.List;
+import java.util.function.Supplier;
+
+public class TagGenerator implements Generator {
```
dualspiral: `final`

src/main/java/org/spongepowered/common/registry/TagRegistry.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    @Nullable
+    @Override
+    public ResourceLocation getKey(final Tag<T> var1) {
+        return this.staticTagHelper.getAllTags().getId((net.minecraft.tags.Tag<T>) var1);
+    }
+
+    @Override
+    public Optional<net.minecraft.resources.ResourceKey<Tag<T>>> getResourceKey(Tag<T> var1) {
```
dualspiral: All parameters should be `final`.

src/main/java/org/spongepowered/common/tag/SpongeTagTemplateBuilder.java
```diff
@@ -0,0 +1,136 @@
...
+    @Override
+    public TagTemplate.Builder<T> replace(final boolean replace) {
+        this.ensureTagTypeAdded();
+        this.replace = replace;
+        return this;
+    }
+
+    @Override
+    public TagTemplate.Builder<T> addChild(final Tag<T> childTag) {
+        ensureTagTypeAdded();
```
dualspiral: `this.ensureTagTypeAdded()`

src/main/java/org/spongepowered/common/tag/SpongeTagTemplateBuilder.java
```diff
@@ -0,0 +1,136 @@
...
+import org.spongepowered.api.tag.TagTemplate;
+import org.spongepowered.api.tag.TagType;
+import org.spongepowered.api.tag.Taggable;
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class SpongeTagTemplateBuilder<T extends Taggable<T>> implements TagTemplate.Builder<T> {
```
dualspiral: `final`

src/main/java/org/spongepowered/common/tag/SpongeTagType.java
```diff
@@ -0,0 +1,63 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.tag;
+
+import org.spongepowered.api.registry.RegistryType;
+import org.spongepowered.api.tag.Tag;
+import org.spongepowered.api.tag.TagType;
+import org.spongepowered.api.tag.Taggable;
+
+public class SpongeTagType<T extends Taggable<T>> implements TagType<T> {
```
dualspiral: `final`

src/main/java/org/spongepowered/common/tag/TagUtil.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.registry.RegistryType;
+import org.spongepowered.api.tag.Tag;
+import org.spongepowered.api.tag.Taggable;
+
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+public class TagUtil {
```
dualspiral: `final`

Also, I think all Util classes go in `org.spongepowered.common.util`.

src/mixins/java/org/spongepowered/common/mixin/api/mcp/tag/TagWrapperMixin_API.java
```diff
@@ -0,0 +1,52 @@
...
+    public Collection<T> all() {
+        return this.getValues();
+    }
+
+    @Override
+    public ResourceKey key() {
+        return (ResourceKey) (Object) this.name;
+    }
+
+    // contains(T value) already defined.
```
dualspiral: For visibility (and for Forge 1.16.5 which uses srg), `contains` should be soft-implemented with `@Intrinsic`.

testplugins/src/main/java/org/spongepowered/test/tag/TagTest.java
```diff
@@ -0,0 +1,326 @@
...
+import org.spongepowered.plugin.PluginContainer;
+import org.spongepowered.plugin.jvm.Plugin;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Plugin("tagtest")
+public class TagTest {
```
dualspiral: `final`

testplugins/src/main/java/org/spongepowered/test/tag/TagTest.java
```diff
@@ -0,0 +1,326 @@
...
+                .type(TagTypes.BLOCK_TYPE.get())
+                .addValue(BlockTypes.DIAMOND_BLOCK.get())
+                .build();
+
+        event.register(underwaterDiamond);
+    }
+
+    @Listener
+    public void registerCommands(final RegisterCommandEvent<Command.Parameterized> event) {
+        logger.info("Registering commands");
```
dualspiral: I know this is in a test plugin, but don't do this - we don't need to know this plugin is registering commands.

testplugins/src/main/java/org/spongepowered/test/tag/TagTest.java
```diff
@@ -0,0 +1,326 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Plugin("tagtest")
+public class TagTest {
+
+    @Inject
+    public PluginContainer pluginContainer;
```
dualspiral: I'd like to keep injection consistent across our test plugins - please do a constructor injection instead, make these fields final, and assign them as normal (see other test plugins for how we've done it).

src/main/java/org/spongepowered/common/datapack/TagDataPackSerializer.java
```diff
@@ -0,0 +1,86 @@
...
+import org.spongepowered.common.datapack.tag.TagSerializedObject;
+import org.spongepowered.common.tag.SpongeTagType;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class TagDataPackSerializer extends DataPackSerializer<TagSerializedObject> {
```
dualspiral: `final`

src/main/java/org/spongepowered/common/datapack/tag/TagSerializedObject.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package org.spongepowered.common.datapack.tag;
+
+import com.google.gson.JsonObject;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.spongepowered.api.ResourceKey;
+import org.spongepowered.api.tag.TagType;
+import org.spongepowered.common.datapack.DataPackSerializedObject;
+
+public class TagSerializedObject extends DataPackSerializedObject {
```
dualspiral: `final`

src/main/java/org/spongepowered/common/tag/SpongeTagTemplate.java
```diff
@@ -0,0 +1,109 @@
...
+import org.spongepowered.api.datapack.DataPackType;
+import org.spongepowered.api.datapack.DataPackTypes;
+import org.spongepowered.api.tag.TagTemplate;
+import org.spongepowered.api.tag.TagType;
+import org.spongepowered.common.SpongeCommon;
+
+import java.io.IOException;
+import java.util.List;
+
+public class SpongeTagTemplate implements TagTemplate {
```
dualspiral: `final`

src/main/java/org/spongepowered/common/datapack/tag/TagSerializedObject.java
```diff
@@ -0,0 +1,44 @@
...
+    private final TagType<@NonNull ?> tagType;
+
+    public TagSerializedObject(ResourceKey key, JsonObject object, TagType<@NonNull ?> tagType) {
+        super(key, object);
+        this.tagType = tagType;
+    }
+
+    public TagType<@NonNull ?> getTagType() {
+        return tagType;
```
dualspiral: `this.`

dualspiral (commented):
Some thoughts - most are things where changes need to be made to fit our style but there are a couple of things on substance that need to be thought about.

src/main/java/org/spongepowered/common/registry/TagRegistry.java
Zidane: I would go for `ReadOnlyRegistry` myself since any future writes to it are literally it being overwritten.

Additionally dual is correct, no registry value can be cached. It is not legal per API spec. I suspect people will do it anyways but we shouldn't contribute confusion. This should return false.

src/accessors/java/org/spongepowered/common/accessor/tags/BlockTagsAccessor.java
```diff
@@ -0,0 +1,39 @@
...
+import net.minecraft.tags.StaticTagHelper;
+import net.minecraft.world.level.block.Block;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+import org.spongepowered.common.UntransformedAccessorError;
+
+@Mixin(BlockTags.class)
+public interface BlockTagsAccessor {
+
+    @Accessor("HELPER") static StaticTagHelper<Block> accessor$helper() { throw new UntransformedAccessorError(); }
```
Zidane: All caps per the spec @kashike put into place

`accessor$HELPER`

src/main/java/org/spongepowered/common/datapack/SpongeDataPackType.java
```diff
@@ -154,5 +175,10 @@ public boolean persistent() {
         public DataPackType<WorldTemplate> world() {
             return world;
         }
+
+        @Override
+        public DataPackType tag() {
+            return tag;
```
Zidane: this.tag

(I know world is wrong above, missed it myself...)

src/main/java/org/spongepowered/common/registry/SpongeRegistryHolder.java
```diff
@@ -212,4 +214,30 @@ public void setRootMinecraftRegistry(final net.minecraft.core.Registry<net.minec
         }
         return (Registry<T>) registry;
     }
+
+    public <T> Registry<T> createRegistry(final RegistryType<Tag<T>> type, final StaticTagHelper<T> staticTagHelper) {
```
Zidane: wrapTagHelperAsRegistry

src/accessors/java/org/spongepowered/common/accessor/tags/BlockTagsAccessor.java
```diff
@@ -0,0 +1,39 @@
...
+import net.minecraft.tags.StaticTagHelper;
+import net.minecraft.world.level.block.Block;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+import org.spongepowered.common.UntransformedAccessorError;
+
+@Mixin(BlockTags.class)
+public interface BlockTagsAccessor {
+
+    @Accessor("HELPER") static StaticTagHelper<Block> accessor$HELPER() { throw new UntransformedAccessorError(); }
```
Zidane, line 37: We don't do one line blocks like this, could you put on a new line?

If not, I can do so.

src/main/java/org/spongepowered/common/SpongeLifecycle.java
```diff
@@ -117,8 +117,8 @@ public void establishDataKeyListeners() {
         ((SpongeDataManager) this.game.dataManager()).registerKeyListeners();
     }
 
-    public void callRegisterDataPackValueEvent(final Path dataPacksDirectory) {
-        SpongeDataPackManager.INSTANCE.callRegisterDataPackValueEvents(dataPacksDirectory);
+    public void callRegisterDataPackValueEvent(final Path datapackDir) {
```
Zidane, line 120: Un-needed change

src/main/java/org/spongepowered/common/datapack/SpongeDataPackManager.java
```diff
@@ -50,6 +51,7 @@
     public static SpongeDataPackManager INSTANCE = new SpongeDataPackManager(Sponge.game());
 
     private final Game game;
+    private RegistryHolder registryHolder;
```
Zidane, line 54: Un-needed change

src/main/java/org/spongepowered/common/registry/TagRegistry.java
```diff
@@ -0,0 +1,236 @@
...
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@DefaultQualifier(NonNull.class)
+public final class TagRegistry<T> extends net.minecraft.core.Registry<Tag<T>> implements Registry<Tag<T>> {
```
Zidane, line 51: So our RegistryMixin_API already implements this interface using the contract of what is in Minecraft's `Registry` class. I think you can toss the implements and remove implementation below.

src/main/java/org/spongepowered/common/util/TagUtil.java
```diff
@@ -0,0 +1,42 @@
...
+import org.spongepowered.api.registry.RegistryType;
+import org.spongepowered.api.tag.Tag;
+import org.spongepowered.api.tag.Taggable;
+
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+public final class TagUtil {
+
+    public static <T extends Taggable<T>> Collection<Tag<T>> getAssociatedTags(T taggable, RegistryType<Tag<T>> taggableRegistry) {
```
Zidane, line 37: `final` params, add a private constructor to this class as well.

dualspiral:
49048743efe65799a1ec8019afc07203556885cf
0c83e6bdd83a8c09714844791f667615dda0bfd2

PR #3428, opened 2021-06-12
Fix error when hooking entity

RedNesto:


src/mixins/java/org/spongepowered/common/mixin/core/world/entity/projectile/FishingHookMixin.java
```diff
@@ -85,11 +84,10 @@
 
     @Nullable private ProjectileSource impl$projectileSource;
 
-    @Inject(method = "setHookedEntity", at = @At("HEAD"), cancellable = true)
-    private void onSetHookedEntity(CallbackInfo ci) {
-        if (SpongeCommon
-            .postEvent(SpongeEventFactory.createFishingEventHookEntity(PhaseTracker.getCauseStackManager().currentCause(), (Entity) this.hookedIn, (FishingBobber) this))) {
-            this.hookedIn = null;
```
ImMorpheus, line 92: Shouldn't the hooked entity be set to null if the event is canceled ?

RedNesto: I am assuming the entity is already null (because it's injected before setting the field), but setting it to null doesn't hurt I guess

ImMorpheus: *stares at mods calling `setHookedEntity` twice*

ImMorpheus (approved):
Thanks

Issue #3429, opened 2021-06-13
[1.16.5] Error when launching


MrSlimeDiamond:
**I am currently running**

- SpongeVanilla version: spongevanilla-1.16.5-8.0.0-RC644-universal.jar
 Not using 1.17 because I wanted 1.16.5 :frog:
- Java version: openjdk version "1.8.0_292"
- Operating System: Ubuntu 20.04.2 LTS aarch64 

- Plugins/Mods: None

**Issue Description**
When I launch spongevanilla-1.16.5-8.0.0-RC644-universal.jar, I get the following error:
```
Exception in thread "main" java.lang.ExceptionInInitializerError
        at org.fusesource.jansi.AnsiConsole.ansiStream(AnsiConsole.java:239)
        at org.fusesource.jansi.AnsiConsole.initStreams(AnsiConsole.java:542)
        at org.fusesource.jansi.AnsiConsole.systemInstall(AnsiConsole.java:496)
        at org.spongepowered.vanilla.applaunch.Main.<clinit>(Main.java:46)
        at java.lang.Class.forName0(Native Method)
        at java.lang.Class.forName(Class.java:264)
        at org.spongepowered.vanilla.installer.InstallerMain.invokeMain(InstallerMain.java:148)
        at org.spongepowered.vanilla.installer.InstallerMain.downloadAndRun(InstallerMain.java:133)
        at org.spongepowered.vanilla.installer.InstallerMain.run(InstallerMain.java:75)
        at org.spongepowered.vanilla.installer.InstallerMain.main(InstallerMain.java:70)
Caused by: java.lang.RuntimeException: Unable to load jansi native library
        at org.fusesource.jansi.internal.JansiLoader.initialize(JansiLoader.java:62)
        at org.fusesource.jansi.internal.CLibrary.<clinit>(CLibrary.java:30)
        ... 10 more
Caused by: java.lang.Exception: No native library found for os.name=Linux, os.arch=aarch64, paths=[/usr/java/packages/lib/aarch64:/usr/lib/aarch64-linux-gnu/jni:/lib/aarch64-linux-gnu:/usr/lib/aarch64-linux-gnu:/usr/lib/jni:/lib:/usr/lib]
        at org.fusesource.jansi.internal.JansiLoader.loadJansiNativeLibrary(JansiLoader.java:333)
        at org.fusesource.jansi.internal.JansiLoader.initialize(JansiLoader.java:60)
        ... 11 more

```

[...]

PR #3431, opened 2021-06-15
Update jansi dep to support ARM64

dualspiral:
See #3429

I haven't got an ARM64 machine up and running to test this (will have to dust off my Raspberry Pi which I _think_ is 64 bit, but I'm hoping someone else will be able to tell me if it works before I get to that tomorrow)

This might not work, or it might not be the best way to do this...

[If someone wants to test this, give this a go - spongevanilla-1.16.5-8.0.0-RC0-universal.jar.zip](https://github.com/SpongePowered/Sponge/files/6658950/spongevanilla-1.16.5-8.0.0-RC0-universal.jar.zip)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -21,6 +21,9 @@ pluginSpiVersion=0.2.0
 guavaVersion=21.0
 junitVersion=5.7.0
 jlineVersion=3.20.0
+# If JLine is updated and it updates the jansi dep, remove the following
+# (and update the SV build.gradle)
+jansiVersion=2.3.3
 timingsVersion=1.0-SNAPSHOT
 tinyLogVersion=2.3.2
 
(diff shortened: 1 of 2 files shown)


dualspiral:
It turns out I have a 32bit distro on my Pi, simply as I'm using a model 3B. So, I won't be able to test it, but if someone can verify this works for me, then we can pull it. Alternatively, we wait until jline updates its dep and we just do it then.

MrSlimeDiamond:
> See #3429
> 
> I haven't got an ARM64 machine up and running to test this (will have to dust off my Raspberry Pi which I _think_ is 64 bit, but I'm hoping someone else will be able to tell me if it works before I get to that tomorrow)
> 
> This might not work, or it might not be the best way to do this...
> 
> [If someone wants to test this, give this a go - spongevanilla-1.16.5-8.0.0-RC0-universal.jar.zip](https://github.com/SpongePowered/Sponge/files/6658950/spongevanilla-1.16.5-8.0.0-RC0-universal.jar.zip)

The provided link works perfectly fine for me on my ARM machine!

MrSlimeDiamond:
You should probably do this for the api-9 branch too.

dualspiral:
It'll get done when someone next does a merge on it.

PR #3433, opened 2021-06-18
Add factory for Orientation

Lignium:
Add factory for Orientation and improve SpongeOrientation class (hashCode, equals and toString).

### DIFF
--- a/src/main/java/org/spongepowered/common/util/SpongeOrientation.java
+++ b/src/main/java/org/spongepowered/common/util/SpongeOrientation.java
@@ -26,16 +26,54 @@ package org.spongepowered.common.util;
 
 import org.spongepowered.api.util.orientation.Orientation;
 
+import java.util.Optional;
+
 public final class SpongeOrientation implements Orientation {
+
+    public static final class Factory implements Orientation.Factory {
+
+        @Override
+        public Optional<Orientation> fromDegrees(final int degrees) {
+            if (degrees % 45 == 0) {
+                return Optional.of(new SpongeOrientation(degrees % 360));
+            }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/util/SpongeOrientation.java
```diff
@@ -26,16 +26,54 @@
...
     }
+
+    @Override
+    public int hashCode() {
+        return this.angle;
+    }
+
+    @Override
+    public boolean equals(final Object anotherObject) {
+        if (anotherObject == null) return false;
```
Zidane, line 66: No shorthand ifs please.

src/main/java/org/spongepowered/common/util/SpongeOrientation.java
```diff
@@ -26,16 +26,54 @@
...
         return this.angle;
     }
+
+    @Override
+    public int hashCode() {
+        return this.angle;
+    }
+
+    @Override
+    public boolean equals(final Object anotherObject) {
```
Zidane, line 65: s/anotherObject/obj

src/main/java/org/spongepowered/common/util/SpongeOrientation.java
Zidane: Did not mean to resolve, instanceof check would handle the null check.

src/main/java/org/spongepowered/common/util/SpongeOrientation.java
kashike: or just "object" is fine too.

PR #3434, opened 2021-06-18
Add missing keys to EntityMixin_API#api$getVanillaValues

Lignium:
This PR fixes the same problem as the previous PR but for API-8 (https://github.com/SpongePowered/Sponge/pull/2493).
All data was taken from the `org.spongepowered.common.data.provider.entity` package.
In addition, the inheritance of some mixins was fixed, and two intermediate mixins were added.


src/main/java/org/spongepowered/common/data/provider/entity/ServerPlayerData.java
```diff
@@ -57,7 +60,14 @@ public static void register(final DataProviderRegistrator registrator) {
                         .get(h -> (GameMode) (Object) h.gameMode.getGameModeForPlayer())
                         .set((h, v) -> h.setGameMode((GameType) (Object) v))
                     .create(Keys.SKIN_PROFILE_PROPERTY)
-                        .get(h -> (ProfileProperty) h.getGameProfile().getProperties().get(ProfileProperty.TEXTURES).iterator().next())
+                        .get(h -> {
+                            final PropertyMap properties = h.getGameProfile().getProperties();
+                            final Collection<Property> values = properties.get(ProfileProperty.TEXTURES);
+
+                            if (values.isEmpty()) return null;
```
dualspiral: No one-line ifs like this.

Zidane:
@Lignium What is the status on this?

Lignium:
@Zidane Now it can be reviewed and merged. During the work on the API PR (https://github.com/SpongePowered/SpongeAPI/pull/2371), I found several issues.

Lignium:
@Zidane

Test code: https://hastebin.com/ebamekafus.java
Test results (output only the number of values): https://hastebin.com/raw/ofajapavev

There were several errors in the data implementation, but all of them were fixed in the last commit. Unimplemented keys that throw a `MissingImplementationException` have been disabled with `TODO` comments. The `FishingBobber` could not be tested because it cannot be spawned.

dualspiral:
Along with the API (https://github.com/SpongePowered/SpongeAPI/pull/2371) component of this, it's been sitting around long enough, another can through it and testing stuff on the server seems to be fine. As I imagine most, if not all, issues moving forward will be impl details, we can handle that as and when (that includes TODOs too, but I want to get this in before it gets bitrotted).

Thanks!

dualspiral:
https://github.com/SpongePowered/Sponge/commit/0be2e312b0504e8e7a76291b6aaf425feb331903

PR #3435, opened 2021-06-18
add shorthand aliases in /sponge

MrSlimeDiamond:


dualspiral:
Why do you think that adding shorthand aliases is important?

MrSlimeDiamond:
> Why do you think that adding shorthand aliases is important?

**i don't know**

Zidane:
Not seeing a lot of value in this PR as it stands.

Am open to better dialogue on this front before another attempt.

PR #3436, opened 2021-06-19
Fix IgniteEntityEvent

Lignium:
Fixed several bugs that prevented `IgniteEntityEvent` from working properly and caused exceptions.


src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
```diff
@@ -1165,14 +1167,14 @@ public ItemEntity entityDropItem(final ItemStack stack, final float offsetY) {
...
+     * Overridden method for Players to determine whether this entity is not immune to
+     * fire such that {@link IgniteEntityEvent}s are not needed to be thrown as they
+     * cannot take fire damage, nor do they light on fire.
      *
-     * @return True if this entity is immune to fire.
+     * @return True if this entity is not immune to fire.
      */
     protected boolean impl$canCallIgniteEntityEvent() {
-        return false;
+        return true;
```
ImMorpheus: Should be `return !this.shadow$fireImmune()`

ImMorpheus (changes requested):

PR #3437, opened 2021-06-20
Set mod name to Sponge with capital S

MrSlimeDiamond:
I know I'm being a little picky here but it annoys me a little lol.


dualspiral:
I'm going to close this, for a couple of reasons:

* We prefer that minor changes (including typos) are reported in #23, that way we can group them all together.
* This isn't really a change that I want to make. This is simply the "server branding" and generally isn't that visible - in Vanilla the brand is also lowercase.
* "Because it annoys you a little" is not really a reason that we accept changes like this. When proposing a change, please be prepared to back up why such a change is necessary. For things like this, feel free to ask us why we've chosen to do it in this way on Discord because there may be deeper reasons to consider (probably not the case here but changing the case of a string that is used as an ID can easily break things).

Thanks!

PR #3439, opened 2021-06-20
Another fix of IgniteEntityEvent

Lignium:
When the skeletons catch fire during the day, it still throws a negative tick exception, so I just raised the lower bound to zero. This time the final fix, there should be no more exceptions.


dualspiral:
Does a negative number mean indefinite?

Lignium:
Each tick for a entity which does not burning is executing: `setRemainingFireTicks(-getFireImmuneTicks())`. However when it ignites the value suddenly becomes positive. We cannot create negative ticks. And does it make sense?

dualspiral:
Yes, it makes sense, they're using that field for two different things, and a negative value means something slightly different to a positive one.

There is a case for a "fire_immunity" key to expose this, I guess, but not for now.

PR #3442, opened 2021-06-26
Implement Entity#setPosition(Vector3d) and fix Entity#setLocation(ServerLocation)

Lignium:


src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/entity/EntityMixin_API.java
```diff
@@ -120,6 +119,12 @@ public Vector3d position() {
         return new Vector3d(this.shadow$getX(), this.shadow$getY(), this.shadow$getZ());
     }
 
+    @Override
+    public boolean setPosition(final Vector3d position) {
+        Preconditions.checkNotNull(position, "The position was null!");
```
Zidane: Uses `Objects.requireNonNull`

Zidane: Also inline it in the return

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/entity/EntityMixin_API.java
```diff
@@ -167,7 +172,7 @@ public boolean setTransform(final Transform transform) {
         }
         Preconditions.checkNotNull(transform, "The transform cannot be null!");
         final Vector3d position = transform.position();
-        this.shadow$setPos(position.x(), position.y(), position.z());
+        this.setPosition(transform.position());
```
Zidane: Do not make API calls, the below code shouldn't either IF there is a bridge method.

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
```diff
@@ -240,6 +241,32 @@
...
+            Vector3d destinationPosition = position;
+            if (ShouldFire.MOVE_ENTITY_EVENT) {
+                final MoveEntityEvent event = SpongeEventFactory.createMoveEntityEvent(frame.currentCause(),
+                    (org.spongepowered.api.entity.Entity) this, VecHelper.toVector3d(this.shadow$position()), position, position);
+                if (SpongeCommon.postEvent(event)) {
+                    return false;
+                }
+                destinationPosition = event.destinationPosition();
+            }
+            this.shadow$teleportToWithTicket(destinationPosition.x(), destinationPosition.y(), destinationPosition.z());
```
dualspiral, line 261: `teleportToWithTicket` has been overwritten to fire the event anyway. Shouldn't you just call that and not fire the event here?

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
```diff
@@ -265,52 +293,42 @@
...
-            this.bridge$revive();
-            this.shadow$setLevel(destinationWorld);
-            destinationWorld.addFromAnotherDimension((Entity) (Object) this);
-
-            originalWorld.resetEmptyTime();
-            destinationWorld.resetEmptyTime();
-
-            final ChunkPos chunkPos = new ChunkPos((int) this.shadow$getX() >> 4, (int) this.shadow$getZ() >> 4);
-            destinationWorld.getChunkSource().addRegionTicket(TicketType.POST_TELEPORT, chunkPos, 1, ((Entity) (Object) this).getId());
+            this.shadow$teleportToWithTicket(destinationPosition.x(), destinationPosition.y(), destinationPosition.z());
```
dualspiral, line 328: Same problem again here - no need to fire the event above if it's going to happen in this method.

dualspiral (commented):
As well as the comments below (you're firing an event twice), you'll need to have different handling for `ServerPlayer` because of the way it works with packets. See how it's done in `TeleportCommandMixin` - you may want to unify the logic a little.

dualspiral:
#3484 - but thanks for this, it really helped me get this to where it is now!

PR #3449, opened 2021-07-04
[WIP] Instrument methods for REGISTRY fields (or lookups)

Zidane:
https://github.com/SpongePowered/SpongeAPI/pull/2372

This is only on Javapoet side currently. Need to do it on other side next.


zml2008 (approved):
Seems relatively straightforward

There is an issue with `AttributeTypes` -- you might just have to shift that one over to a validator at this point (I debated doing that earlier, but managed to coast by on the assumption that everything is a `RangedAttributeType`)

PR #3454, opened 2021-07-07
Fix Mixin obf environment

Logiquo:
Fix Mixin AP cannot correctly remap.


gabizou:
Was this the only issue?

PR #3464, opened 2021-07-10
Add status/uptime command

MrSlimeDiamond:
# Status/Uptime command
| Added Command | Description
| :--- | :---
| /sponge status| Show server status

image


src/main/java/org/spongepowered/common/util/DateUtil.java
```diff
@@ -0,0 +1,44 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.util;
+
+import java.time.Duration;
+
+public class DateUtil {
```
Lignium: Instead of writing utility classes, you can simply write code where it is used directly. Moreover, it is used only once. Or at worst make a static method.

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -564,6 +573,16 @@ private TextColor tpsColor(final double tps) {
         return CommandResult.success();
     }
 
+    private @NonNull CommandResult uptimeExecutor(final CommandContext context) {
+        String uptime = DateUtil.format(Duration.ofMillis(ManagementFactory.getRuntimeMXBean().getUptime()));
+
+        context.sendMessage(Identity.nil(),
+                Component.text("JVM uptime: "+uptime, NamedTextColor.YELLOW)
```
Lignium: Missing spaces around "+"

src/main/java/org/spongepowered/common/util/DateUtil.java
```diff
@@ -0,0 +1,44 @@
...
+    public static String format(Duration d) {
+        long days = d.toDays();
+        d = d.minusDays(days);
+        long hours = d.toHours();
+        d = d.minusHours(hours);
+        long minutes = d.toMinutes();
+        d = d.minusMinutes(minutes);
+        long seconds = d.getSeconds() ;
+        return
+                (days ==  0?"":days+" days, ")+
```
Lignium: Here too everything stuck together ugly. Always write spaces around operator symbols.

Lignium (changes requested):

src/main/java/org/spongepowered/common/util/DateUtil.java
MrSlimeDiamond: Done in last commit

src/main/java/org/spongepowered/common/util/DateUtil.java
MrSlimeDiamond: Done in last commit

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
MrSlimeDiamond: Done in last commit

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -191,6 +192,12 @@
             .addChild(reloadWorldCommand, "world")
             .build();
 
+        // /sponge uptime
+        final Command.Parameterized uptimeCommand = Command.builder()
+                .permission("sponge.command.uptime")
+                .shortDescription(Component.text("Shows Sponge JVM uptime."))
```
ItsDoot: nitpick: rather `Shows JVM uptime.` The property isn't anything specific to Sponge, just general JVM uptime.

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -564,6 +572,31 @@ private TextColor tpsColor(final double tps) {
         return CommandResult.success();
     }
 
+    public static String formatMilisDate(Duration d) {
```
ItsDoot: private method

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -564,6 +572,31 @@ private TextColor tpsColor(final double tps) {
...
+        long days = d.toDays();
```
ItsDoot: non-mutated variables should be `final`

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -564,6 +572,31 @@ private TextColor tpsColor(final double tps) {
...
+        long hours = d.toHours();
+        d = d.minusHours(hours);
+        long minutes = d.toMinutes();
+        d = d.minusMinutes(minutes);
+        long seconds = d.getSeconds();
+        return
+                (days ==  0?"":days+" days, ") +
+                (hours == 0?"":hours+" hours, ") +
+                (minutes ==  0?"":minutes+" minutes, ") +
+                (seconds == 0?"":seconds+" seconds");
```
ItsDoot: this formatting needs some fixing

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -564,6 +572,31 @@ private TextColor tpsColor(final double tps) {
...
+    }
+
+    private @NonNull CommandResult uptimeExecutor(final CommandContext context) {
+        String uptime = formatMilisDate(Duration.ofMillis(ManagementFactory.getRuntimeMXBean().getUptime()));
```
ItsDoot: `final String uptime`

ItsDoot (changes requested):

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
MrSlimeDiamond: Fixed in 1435d1b

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
MrSlimeDiamond: Fixed in 1435d1b

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
MrSlimeDiamond: Fixed in 1435d1b

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
MrSlimeDiamond: Fixed in 1435d1b

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
MrSlimeDiamond: Unsure how I could make this formatting better. :P

MrSlimeDiamond: fixed in d9b7d10

dualspiral:
I'm not interested in having `/sponge uptime` on its own. 

That said, I can see some value in a command like this being baked into Sponge, but it'd need to be more useful, so I think I'd be more in favour of a `/sponge status` command that contains various useful bits of information about the currently running instance, one of which is JVM uptime. There are also things like server uptime, memory usage, etc. that might be useful.

Some might argue that this belongs in a plugin ([and right now, I do this sort of stuff in Nucleus](https://github.com/NucleusPowered/Nucleus/blob/v3/nucleus-modules/src/main/java/io/github/nucleuspowered/nucleus/modules/misc/commands/ServerStatCommand.java#L47-L103)) but I could be convinced to have a status subcommand in Sponge if it's done well as this really is on a server management end.

Don't just copy what I've done, but go away, think about what might be useful to server owners about the current system environment, then suggest what you think would work in a `status` like subcommand.

MrSlimeDiamond:
> 
> 
> I'm not interested in having `/sponge uptime` on its own.
> 
> That said, I can see some value in a command like this being baked into Sponge, but it'd need to be more useful, so I think I'd be more in favour of a `/sponge status` command that contains various useful bits of information about the currently running instance, one of which is JVM uptime. There are also things like server uptime, memory usage, etc. that might be useful.
> 
> Some might argue that this belongs in a plugin ([and right now, I do this sort of stuff in Nucleus](https://github.com/NucleusPowered/Nucleus/blob/v3/nucleus-modules/src/main/java/io/github/nucleuspowered/nucleus/modules/misc/commands/ServerStatCommand.java#L47-L103)) but I could be convinced to have a status subcommand in Sponge if it's done well as this really is on a server management end.
> 
> Don't just copy what I've done, but go away, think about what might be useful to server owners about the current system environment, then suggest what you think would work in a `status` like subcommand.

That's a very cool idea!

MrSlimeDiamond:
@dualspiral `/sponge version` and `/sponge status` are similar commands (I based my status command off the version commmand) would it be a good idea if I merged them to make `/sponge info`?

dualspiral:
No.

MrSlimeDiamond:
> That said, I can see some value in a command like this being baked into Sponge, but it'd need to be more useful, so I think I'd be more in favour of a `/sponge status` command that contains various useful bits of information about the currently running instance, one of which is JVM uptime. There are also things like server uptime, memory usage, etc. that might be useful.

I added the status command as requested.

src/main/java/org/spongepowered/common/SpongeLifecycle.java
```diff
@@ -67,6 +68,7 @@
 @Singleton
 public final class SpongeLifecycle {
 
+    public static long tookStartedNanos;
```
dualspiral, line 71: No. Not here.

Reading your code, `tookStartedNanos` here is intended to be the time where the server is deemed to have started... I think. It's not what it is necessarily though, as `Util.getNanos()` may have a different meaning on the client. 

Further, within a game lifecycle, the server may be restarted multiple times - this is what happens on the client if you switch saves. This isn't handled here.

This _really_ shouldn't be on the lifecycle object, most certainly the wrong place. The server startup time should be on the `Server` object, and not as a static field.

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -564,6 +570,69 @@ private TextColor tpsColor(final double tps) {
...
+        final long maxMem = Runtime.getRuntime().maxMemory() / 1024 / 1024;
+        final long totalMem = Runtime.getRuntime().totalMemory() / 1024 / 1024;
+        final long freeMem = Runtime.getRuntime().freeMemory() / 1024 / 1024;
+        final int tps = Sponge.server().targetTicksPerSecond();
+        final long milisStartTime = SpongeLifecycle.tookStartedNanos / 1000000;
+        final String serverUptime = formatDuration(Duration.ofMillis(ManagementFactory.getRuntimeMXBean().getUptime() - milisStartTime));
+
+        builder.append(
+                Component.newline(),
+                Component.text("    Max memory", NamedTextColor.GRAY),
```
dualspiral: There is `SpongeCommand.INDENT` at the top, use that so we're consistent.

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -564,6 +570,69 @@ private TextColor tpsColor(final double tps) {
...
+                colon,
+                Component.text(totalMem + "MB", NamedTextColor.WHITE),
+                Component.newline(),
+                Component.text("    Free memory", NamedTextColor.GRAY),
+                colon,
+                Component.text(freeMem + "MB", NamedTextColor.WHITE),
+                Component.newline(),
+                Component.text("    TPS", NamedTextColor.GRAY),
+                colon,
+                Component.text(tps, tps >= 15 ? NamedTextColor.GREEN : tps >= 10 ? NamedTextColor.YELLOW : NamedTextColor.RED),
```
dualspiral, line 603: There is a function in this class, `tpsColor` that was added to do the colouring. Use that.

Note we have `/sponge tps`. Might be worth making this clickable causing it to run that command?

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -564,6 +570,69 @@ private TextColor tpsColor(final double tps) {
         return CommandResult.success();
     }
 
+    private @NonNull CommandResult statusExecutor(final CommandContext context) {
+        final TextComponent.Builder builder = Component.text()
+                .append(
+                        Component.text(Launch.instance().platformPlugin().metadata().name().get(), Style.style(NamedTextColor.YELLOW, TextDecoration.BOLD))
+                );
+        final Component colon = Component.text(": ", NamedTextColor.GRAY);
+        final String uptime = formatDuration(Duration.ofMillis(ManagementFactory.getRuntimeMXBean().getUptime()));
```
dualspiral: Any static access should be prefixed by the class name, even if it is the same class, so `SpongeCommand.formatDuration(...)`

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -564,6 +570,69 @@ private TextColor tpsColor(final double tps) {
...
+        final long maxMem = Runtime.getRuntime().maxMemory() / 1024 / 1024;
+        final long totalMem = Runtime.getRuntime().totalMemory() / 1024 / 1024;
+        final long freeMem = Runtime.getRuntime().freeMemory() / 1024 / 1024;
+        final int tps = Sponge.server().targetTicksPerSecond();
```
dualspiral, line 584: Look how it's done in `tpsExecutor`, you should follow that.

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -564,6 +570,69 @@ private TextColor tpsColor(final double tps) {
...
+    private static String formatDuration(Duration d) {
+        final long days = d.toDays();
+        d = d.minusDays(days);
+        final long hours = d.toHours();
+        d = d.minusHours(hours);
+        final long minutes = d.toMinutes();
+        d = d.minusMinutes(minutes);
+        final long seconds = d.getSeconds();
+        final StringJoiner joiner = new StringJoiner(", ");
+        if (days > 0) joiner.add(days + " days");
```
dualspiral: If statements must not be oneliners.

```java
if (days > 0) {
    joiner.add(days + "days");
}
```

src/mixins/java/org/spongepowered/common/mixin/core/server/dedicated/DedicatedServerMixin.java
```diff
@@ -50,9 +50,9 @@
 
     @Inject(method = "<init>", at = @At("TAIL"))
     private void impl$setServerOnGame(Thread p_i232601_1_, RegistryAccess.RegistryHolder p_i232601_2_, LevelStorageSource.LevelStorageAccess p_i232601_3_,
-            PackRepository p_i232601_4_, ServerResources p_i232601_5_, WorldData p_i232601_6_, DedicatedServerSettings p_i232601_7_,
-            DataFixer p_i232601_8_, MinecraftSessionService p_i232601_9_, GameProfileRepository p_i232601_10_, GameProfileCache p_i232601_11_,
-            ChunkProgressListenerFactory p_i232601_12_, CallbackInfo ci) {
+                                      PackRepository p_i232601_4_, ServerResources p_i232601_5_, WorldData p_i232601_6_, DedicatedServerSettings p_i232601_7_,
+                                      DataFixer p_i232601_8_, MinecraftSessionService p_i232601_9_, GameProfileRepository p_i232601_10_, GameProfileCache p_i232601_11_,
+                                      ChunkProgressListenerFactory p_i232601_12_, CallbackInfo ci) {
```
dualspiral, line 55: Unnecessary change, please don't change formatting and nothing else.

src/mixins/java/org/spongepowered/common/mixin/core/server/dedicated/DedicatedServerMixin.java
```diff
@@ -74,4 +74,4 @@ private void onSave(final GameProfileCache cache) {
     public boolean bridge$performAutosaveChecks() {
         return this.shadow$isRunning();
     }
-}
```
dualspiral, line 77: Unnecessary change, please don't change formatting if you're not touching the class or area of code (and style says new line at the end of a file anyway).

src/main/java/org/spongepowered/common/SpongeCommon.java
```diff
@@ -33,6 +33,7 @@
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.event.Event;
+import org.spongepowered.api.event.lifecycle.StartedEngineEvent;
```
dualspiral: Undo this (it's easily left in when testing I know, I miss these things myself, but the PR process lets us see these mistakes more easily...)

src/main/java/org/spongepowered/common/SpongeLifecycle.java
```diff
@@ -190,6 +192,7 @@ public void callStartingEngineEvent(final Engine engine) {
     public void callStartedEngineEvent(final Engine engine) {
         this.game.eventManager().post(SpongeEventFactory.createStartedEngineEvent(PhaseTracker.getCauseStackManager().currentCause(),
                 engine, this.game, (TypeToken<Engine>) TypeToken.get(engine.getClass())));
+        tookStartedNanos = Util.getNanos();
```
dualspiral, line 195: Though this should not be here at all, this would otherwise be `SpongeLifecycle.tookStartedNanos`.

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -564,6 +570,69 @@ private TextColor tpsColor(final double tps) {
...
+                        Component.text(Launch.instance().platformPlugin().metadata().name().get(), Style.style(NamedTextColor.YELLOW, TextDecoration.BOLD))
+                );
+        final Component colon = Component.text(": ", NamedTextColor.GRAY);
+        final String uptime = formatDuration(Duration.ofMillis(ManagementFactory.getRuntimeMXBean().getUptime()));
+        final long maxMem = Runtime.getRuntime().maxMemory() / 1024 / 1024;
+        final long totalMem = Runtime.getRuntime().totalMemory() / 1024 / 1024;
+        final long freeMem = Runtime.getRuntime().freeMemory() / 1024 / 1024;
+        final int tps = Sponge.server().targetTicksPerSecond();
+        final long milisStartTime = SpongeLifecycle.tookStartedNanos / 1000000;
+        final String serverUptime = formatDuration(Duration.ofMillis(ManagementFactory.getRuntimeMXBean().getUptime() - milisStartTime));
```
dualspiral: Qualify access.

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -564,6 +570,69 @@ private TextColor tpsColor(final double tps) {
...
+
+        builder.append(
+                Component.newline(),
+                Component.text("    Max memory", NamedTextColor.GRAY),
+                colon,
+                Component.text(maxMem + "MB", NamedTextColor.WHITE),
+                Component.newline(),
+                Component.text("    Total memory", NamedTextColor.GRAY),
```
dualspiral: What does Total/Free Memory mean? Are these terms clear to server owners?

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -71,12 +72,10 @@
...
 import java.text.DecimalFormat;
 import java.text.MessageFormat;
+import java.time.Duration;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
```
dualspiral, line 79: No star imports as per our style guide.

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -204,7 +209,8 @@
                 .addChild(tpsCommand, "tps")
                 .addChild(versionCommand, "version")
                 .addChild(whichCommand, "which")
-                .addChild(reloadCommand, "reload");
+                .addChild(reloadCommand, "reload")
+                .addChild(statusCommand, "status", "uptime");
```
dualspiral: Just status.

dualspiral (commented):
There are a lot of unnecessary changes and style problems. [You can see our style guide here](https://docs.spongepowered.org/stable/en/contributing/implementation/codestyle.html). There are a couple of things that are now different (don't use Google Preconditions, static access must _always_ be qualified with the class name), but it's worth knowing about them.

What I was aiming for in my previous comment was to have some discussion and ideas as to what might be useful from a server owner's point of view, and gave you my example as a springboard. If a server owner ran such a command, what information about the status of the server might be useful to them? Are all of these things useful?

I just want a bit of thought and reasoning as to why what you've done is useful, rather than pull in a command that has a jumble of stats that may or may not be useful.

src/main/java/org/spongepowered/common/SpongeLifecycle.java
MrSlimeDiamond: Yeah, for now I'll just keep it to JVM uptime.

src/mixins/java/org/spongepowered/common/mixin/core/server/dedicated/DedicatedServerMixin.java
MrSlimeDiamond: Didn't realize I did that. Oops.

src/mixins/java/org/spongepowered/common/mixin/core/server/dedicated/DedicatedServerMixin.java
MrSlimeDiamond: Again, accident

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
MrSlimeDiamond: It's a private method

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
MrSlimeDiamond: Done in 0656326

src/main/java/org/spongepowered/common/SpongeCommon.java
MrSlimeDiamond: Done in 0656326

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
MrSlimeDiamond: Done in 0656326

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
MrSlimeDiamond: Done in 0656326

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
MrSlimeDiamond: Done in 0656326

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
dualspiral: Doesn't matter. It's our style.

MrSlimeDiamond: image

dualspiral: ....what? That's not qualifying access, I was saying prefix the `formatDuration` call with the class name - `SpongeComponents.formatDuration`

MrSlimeDiamond: OH, I think I did that in 0656326.

dualspiral:
I'm going to close this.

While some server owners may like the idea of an uptime command, the problem here is that a lot of ideas have been put together and there isn't really a theme of what might be useful. I was always on the side of needing to be convinced rather than needing to be dissuaded anyway. I've also made many comments but most of them haven't really been resolved. I'm still not really sure about this and I think it might need a fresh approach if we are to do anything like this.

So, if server owners still want this, I think the better way to do this would be to start by opening an issue on the Sponge repo and adding discussion as to what might be useful.

Alternatively, I'm content leaving most of this to plugins.

PR #3468, opened 2021-07-12
Mark all soft-implement of sponge interface to be Remap.NONE

Logiquo:


Logiquo:
These two methods have a synthetic method with the same name in the binary Minecraft 1.16.5. Add Descriptor to distinguish them.

Logiquo:
Hi, is there anything stops this from being merged?

Zidane:
@LogicFan Can you explain why you are needing this?

Logiquo:
Suppress some warnings and correctly remap some method for https://github.com/FabricMC/tiny-remapper/pull/63 when I try to build Sponge against the fabric environment.

The first commit is to suppress warnings of `@Implements` since the warning will be generated if no mapping is found.

The second commit is to disambiguous the method since there is a method with same name in the class (they are synthetic methods).

Logiquo:
Any feedback, @Zidane ?

PR #3470, opened 2021-07-13
allow different mapping for reflection

Logiquo:
~~It's configured using json file.~~ I've included one for sponge vanilla.

### DIFF
--- a/src/main/java/org/spongepowered/common/util/ReflectionUtil.java
+++ b/src/main/java/org/spongepowered/common/util/ReflectionUtil.java
@@ -84,36 +84,6 @@ public final class ReflectionUtil {
         Entity.class
     };
 
-    private static final String REFLECTION_PROPERTY_FILE = "reflection_mappings.json";
-    private static Map<String, String> mappingCache;
-
-    private static String getMethodNameForEnvironment(final String methodName) {
-        if (Launch.instance().usesMojangMappings()) {
-            return methodName;
-        }
-
-        if (Launch.instance().developerEnvironment() || Launch.instance().usesMojangMappings()) {
-            return methodName;
-        }
-
-        if (mappingCache == null) {
-            try (InputStreamReader reader = new InputStreamReader(Objects.requireNonNull(ReflectionUtil.class.getClassLoader().getResourceAsStream(ReflectionUtil.REFLECTION_PROPERTY_FILE)))) {
-                mappingCache = new Gson().fromJson(reader, new TypeToken<Map<String, String>>(){}.getType());
-            } catch (IOException | NullPointerException e) {
-                mappingCache = Collections.emptyMap();
-            }
-        }
-
-        final String methodNameForEnvironment = mappingCache.get(methodName);
-
-        if (methodNameForEnvironment == null) {
-            SpongeCommon.logger().warn(ReflectionUtil.STUPID_REFLECTION, "Could not find reflection mapping for method {} in reflection property file {}", methodName, ReflectionUtil.REFLECTION_PROPERTY_FILE);
-            return methodName;
-        } else {
-            return methodNameForEnvironment;
-        }
-    }
-
     public static boolean isNeighborChangedDeclared(final Class<?> targetClass) {
         return ReflectionUtil.doesMethodExist(
             targetClass,
@@ -147,7 +117,7 @@ public final class ReflectionUtil {
         final String methodName,
         final Class<?>[] methodParameters
     ) {
-        final String targetMethodForEnvironment = ReflectionUtil.getMethodNameForEnvironment(methodName);
+        final String targetMethodForEnvironment = Launch.instance().mappingManager().toRuntimeMethodName(targetClass, methodName, methodParameters);
         try {
             final Class<?> declaringClass = targetClass.getMethod(targetMethodForEnvironment, methodParameters).getDeclaringClass();
             return !ignoredClass.equals(declaringClass);
--- a/src/main/java/org/spongepowered/common/util/ReflectionUtil.java
+++ b/src/main/java/org/spongepowered/common/util/ReflectionUtil.java
@@ -93,7 +93,7 @@ public final class ReflectionUtil {
         }
 
         if (mappingCache == null) {
-            try (InputStreamReader reader = new InputStreamReader(Objects.requireNonNull(ReflectionUtil.class.getClassLoader().getResourceAsStream(REFLECTION_PROPERTY_FILE)))) {
+            try (InputStreamReader reader = new InputStreamReader(Objects.requireNonNull(ReflectionUtil.class.getClassLoader().getResourceAsStream(ReflectionUtil.REFLECTION_PROPERTY_FILE)))) {
                 mappingCache = new Gson().fromJson(reader, new TypeToken<Map<String, String>>(){}.getType());
             } catch (IOException | NullPointerException e) {
                 mappingCache = Collections.emptyMap();
@@ -103,7 +103,7 @@ public final class ReflectionUtil {
         final String methodNameForEnvironment = mappingCache.get(methodName);
 
         if (methodNameForEnvironment == null) {
-            SpongeCommon.logger().fatal(ReflectionUtil.STUPID_REFLECTION, "Could not find reflection mapping for method {} in reflection property file {}", methodName, REFLECTION_PROPERTY_FILE);
+            SpongeCommon.logger().warn(ReflectionUtil.STUPID_REFLECTION, "Could not find reflection mapping for method {} in reflection property file {}", methodName, ReflectionUtil.REFLECTION_PROPERTY_FILE);
             return methodName;
         } else {
             return methodNameForEnvironment;
--- a/src/main/java/org/spongepowered/common/util/ReflectionUtil.java
+++ b/src/main/java/org/spongepowered/common/util/ReflectionUtil.java
@@ -24,23 +24,9 @@
  */
 package org.spongepowered.common.util;
 
-import java.io.IOException;
-import java.io.InputStreamReader;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Modifier;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-
 import static com.google.common.base.Preconditions.checkArgument;
 import static org.apache.commons.lang3.ClassUtils.isAssignable;
 
-import com.google.gson.Gson;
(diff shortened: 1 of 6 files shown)


src/main/java/org/spongepowered/common/util/ReflectionUtil.java
```diff
@@ -110,7 +139,7 @@ public static boolean doesMethodExist(
         final String methodName,
         final Class<?>[] methodParameters
     ) {
-        final String targetMethodForEnvironment = Launch.instance().developerEnvironment() ? methodName : methodName;
+        final String targetMethodForEnvironment = Launch.instance().developerEnvironment() ? methodName : getMethodNameForEnvironment(methodName);
```
Zidane: Fully qualify static access.

src/main/java/org/spongepowered/common/util/ReflectionUtil.java
```diff
@@ -77,6 +84,28 @@
...
+    private static String getMethodNameForEnvironment(final String methodName) {
+        if (mappingCache == null) {
+            try (InputStreamReader reader = new InputStreamReader(Objects.requireNonNull(ReflectionUtil.class.getClassLoader().getResourceAsStream(REFLECTION_PROPERTY_FILE)))) {
+                mappingCache = new Gson().fromJson(reader, new TypeToken<Map<String, String>>(){}.getType());
+            } catch (IOException | NullPointerException e) {
+                mappingCache = Collections.emptyMap();
+            }
+        }
+
+        String methodNameForEnvironment = mappingCache.get(methodName);
```
Zidane: `final`

src/main/java/org/spongepowered/common/util/ReflectionUtil.java
```diff
@@ -77,6 +84,28 @@
         Entity.class
     };
 
+    public static final String REFLECTION_PROPERTY_FILE = "reflection_mappings.json";
+    private static Map<String, String> mappingCache;
+
+    private static String getMethodNameForEnvironment(final String methodName) {
+        if (mappingCache == null) {
```
Zidane: So `Launch` knows if this is a developer environment or not.

I assume that the userdev you have for SpongeFabric exists with official class names and mappings? If so, gate all of this behind the `Launch` `isDeveloperEnvironment` check and just return the mapping. Only production is the issue.

src/main/java/org/spongepowered/common/util/ReflectionUtil.java
```diff
@@ -77,6 +84,28 @@
         Entity.class
     };
 
+    public static final String REFLECTION_PROPERTY_FILE = "reflection_mappings.json";
```
Zidane: Not used outside this class so make it `private`

src/main/java/org/spongepowered/common/util/ReflectionUtil.java
Logiquo: This method call is guarded by `isNeighborChangedDeclared` already. Do I move the if-condition from `isNeighborChangedDeclared` to inside this function? It does not make sense for me to guard it twice.

Zidane: Move it in here imo. It does increase the call stack once but does not rely on us remembering that we're already doing that.

Logiquo:
Should be fixed now.

Zidane:
@gabizou @zml2008 

Any changes? Looks straightforward to me.

src/main/java/org/spongepowered/common/util/ReflectionUtil.java
```diff
@@ -77,6 +84,32 @@
...
+                mappingCache = new Gson().fromJson(reader, new TypeToken<Map<String, String>>(){}.getType());
+            } catch (IOException | NullPointerException e) {
+                mappingCache = Collections.emptyMap();
+            }
+        }
+
+        final String methodNameForEnvironment = mappingCache.get(methodName);
+
+        if (methodNameForEnvironment == null) {
+            SpongeCommon.logger().fatal(ReflectionUtil.STUPID_REFLECTION, "Could not find reflection mapping for method {} in reflection property file {}", methodName, REFLECTION_PROPERTY_FILE);
```
dualspiral: I know this isn't your fault here - it's not really targeted at you Logic, but I'd rather not have the logger spit out "REFLECTION_BULLSHIT" every time this fails...

That said, do we _really_ need to add an entry into the Json file for every field that we use with reflection in vanilla? The reality is that in SV our dev and prod environments are mapped the same, and this will be the case in Forge for 1.17.x and beyond. I don't really see why we should be emitting a _fatal_ logger message when it doesn't matter in that environment.

Is there a way to only generate this logger message if the environment is obfuscated?

Also, `REFLECTION_PROPERTY_FILE` static access needs to be qualified.

edit: to make it clear, I mean in vanilla when it doesn't matter, why should this log entry be generated?

Logiquo: To hard code, some platforms do not sound right to me, since this is common code. I guess it either introduces a new property `isObfuscated` in somewhere so I can use it to detect if the environment is obf or not. Or just put the json file even if it is not in obf environment.

Zidane: @dualspiral 

This will not be the case for Forge 1.17.x as production will be SRG mapped with Official class names.

Anyways I agree with dual's comments, he brings up good points.

1. Per the REFLECTION_BULLSHIT comment...While funny, it isn't professional. Looking at you @gabizou 
2. We can add a `Launch` `requiresReflectionChecks` or something. `obfuscated` is not the right word here, some environments are just remapped vs. others.

dualspiral: > This will not be the case for Forge 1.17.x as production will be SRG mapped with Official class names.

Ah, my misunderstanding. Good to know.

Logiquo: > We can add a Launch requiresReflectionChecks

Should this be covered in this PR, or you will handle it later? Also, I guess `requiresReflectionRemap` is a better name?

Zidane: Chuck it into PR. Call it `Launch.usesThirdPartyMappings()`. Document method to state that this allows a platform to indicate that the codebase will be in mappings that are not supplied by Mojang. Something to that effect. Utilize that check and remove the mappings file from `vanilla`

Logiquo:
fixed.

zml2008:
This looks like a fairly straightforward implementation -- but it doesn't capture the situation of using a SpongeForge/SpongeFabric jar in a dev environment with different mappings. Forge has an (iirc) `INameMappingService`, fabric loader has some mapping API as well to convert from SRG/intermediary strings into the appropriate dev-time mappings. Could you look into exposing that, probably through `Launch` as well?

Logiquo:
> Could you look into exposing that, probably through Launch as well?

I looked at the fabric part. So, to remap a method, it also needs to also pass a descriptor. I also don't want this API to be restricted to method only. It might be necessary to use reflection on class name or field one day. I think it will be a little bit tricky to design an API that could facilitate potential future use & quite clean right now.

I think I will need to look at the forge part too to have a better understanding of what will be the best way.

Logiquo:
Hi, @zml2008 , do you think this new approach works better? or is it overkill?

vanilla/src/launch/java/org/spongepowered/vanilla/launch/mapping/VanillaMappingManager.java
```diff
@@ -0,0 +1,44 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.vanilla.launch.mapping;
+
+import org.spongepowered.common.launch.mapping.SpongeMappingManager;
+
+public class VanillaMappingManager implements SpongeMappingManager {
```
Zidane: `final`

vanilla/src/launch/java/org/spongepowered/vanilla/launch/mapping/VanillaMappingManager.java
```diff
@@ -0,0 +1,44 @@
...
+	@Override
+	public String toRuntimeClassName(String srcName) {
```
Zidane: `final srcName`

vanilla/src/launch/java/org/spongepowered/vanilla/launch/mapping/VanillaMappingManager.java
```diff
@@ -0,0 +1,44 @@
...
+		return srcName;
+	}
+
+	@Override
+	public String toRuntimeFieldName(Class<?> owner, String srcName) {
+		return srcName;
```
Zidane: You have indented too far (only 4 spaces please)

zml2008:
This is a step in the right direction -- however, taking a look at [Fabric's Mapping resolver API](https://github.com/FabricMC/fabric-loader/blob/master/src/main/java/net/fabricmc/loader/api/MappingResolver.java#L86), this API would fail when used in development environments, particularly if a user depended on your fabric implementation from a Yarn-based workspace (since Sponge will at most know about official names and intermediary). You'd have a mismatch between the class literals (remapped by TR), and the method names (string constants), unless you were in a prod environment using intermediary class names.

Logiquo:
> You'd have a mismatch between the class literals (remapped by TR), and the method names (string constants)

As you can see in the fabric mapping resolver, there is a `unmap` method, which can be used for this situation. If a user runs things on yarn environment and the jar is in an intermediary, then the process will be
```
owner (class literal) -> getName (yarn name) -> unmap (intermediary name)
desc (class literals) -> similar to above
method (mojname) -> some how make it intermediary name -> map (with the intermediary owner and desc described above) -> yarn name
```

zml2008:
Ah I see that, thanks for clairfying -- it's a touch non-traditional, but the class literal-based API is less error prone. I'm fine with that, but would be even better to see a class literal-based API exposed directly upstream in loader.

zml2008 (approved):

Logiquo:
styles fixed.

Logiquo:
Since SpongeForge is developed using forge-loom (which also uses tiny-remapper IIRC), https://github.com/FabricMC/tiny-remapper/issues/70 might be a better solution than this PR.

PR #3474, opened 2021-07-17
Make Users load async, ensure we don't load users in the impl

dualspiral:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2378) | **Sponge**

So - there's a lot going on here.

* `ServerUserProvider` no longer exists. It was originally there because the `UserManager` was `UserService` and was global... obviously now it's not. No point keeping the split.
* On the other hand, `SpongeUser` has been split into two, `SpongeUserView` and `SpongeUserData`. The former is a `User`, but is just a view to the current user or player. This means that when a player is online, we can just ditch the offline version and use this object to proxy to the player. We then don't need to store or think about data on the `User` object, and allows us to drop things like `forceRecreateUser` on the user manager - things that really were just a pain to keep synced up. We'll just drop user data on login.
  
  By doing this split, it means that if we drop the user data but someone has still kept hold of the view, we no longer have to worry about weird things going on without having to time things out.
* The Notifier and Creator on the phase tracker are now UUIDs in keeping with this. If a plugin wants the user, they can go fetch it themselves. We still cache the user data for an hour so if a plugin wants the data and uses it frequently, it won't be loaded every time.
* Some assumptions regarding users in the tracker are fixed - `User` is never an entity so such checks would just say "no".

~~Still got a couple of things to fix up so it compiles on CI (_it worked on my machine!_).~~


src/main/java/org/spongepowered/common/bridge/server/MinecraftServerBridge.java
```diff
@@ -46,4 +47,7 @@
     boolean bridge$performAutosaveChecks();
 
     ServerTimingsHandler bridge$timingsHandler();
+
+    SpongeUserManager bridge$userManager();
```
Zidane, line 51: Wonder if this should be put onto `SpongeServer`? I have a, albeit, weird separation going on there.

dualspiral: Actually, yes, that would be a perfect candidate to do so.

src/main/java/org/spongepowered/common/entity/player/SpongeUserData.java
```diff
@@ -417,34 +382,28 @@ public void markDirty() {
...
-            SpongeCommon.logger().warn("Failed to save user file [{}]!", file, e);
+        synchronized (this) {
+            final SpongeUserManager userManager = ((SpongeServer) SpongeCommon.server()).userManager();
+            final LevelStorageSource.LevelStorageAccess storageSource = ((MinecraftServerAccessor) Sponge.server()).accessor$storageSource();
+            final File file = storageSource.getLevelPath(LevelResource.PLAYER_DATA_DIR).resolve(this.uniqueId() + ".dat").toFile();
+            CompoundTag compound;
+            try {
+                compound = NbtIo.readCompressed(new FileInputStream(file));
+            } catch (final IOException ignored) {
+                // Nevermind
```
Zidane: We need to print this as it typically means the compression metadata in the file failed or a file lock. We should not continue at all, if Vanilla is they are playing dangerously...

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/server/MinecraftServerMixin_API.java
```diff
@@ -157,9 +155,9 @@ public MinecraftServerMixin_API(final String name) {
         this.api$scheduler = new ServerScheduler();
         this.api$playerDataHandler = new SpongePlayerDataManager(this);
         this.api$teleportHelper = new SpongeTeleportHelper();
-        this.api$userManager = new SpongeUserManager(this);
         this.api$mapStorage = new SpongeMapStorage();
         this.api$registryHolder = new SpongeRegistryHolder(p_i232576_2_);
+        this.api$spongeUserManager = new SpongeUserManager((MinecraftServer) (Object) this);
```
Zidane: just `api$userManager`

src/mixins/java/org/spongepowered/common/mixin/core/world/damagesource/IndirectEntityDamageSourceMixin.java
```diff
@@ -70,8 +80,8 @@ public String toString() {
             .add("Type", resourceKey)
             .add("Source", this.shadow$getDirectEntity())
             .add("IndirectSource", this.shadow$getEntity());
-        if (this.impl$creator != null) {
-            helper.add("SourceOwner", this.impl$creator);
+        if (this.impl$creatorUUID != null) {
+            helper.add("SourceOwner", this.impl$creatorUUID);
```
Zidane: SourceCreator...need to remove this flip flop of Owner/Creator in this codebase eventually...

src/main/java/org/spongepowered/common/entity/player/SpongeUserData.java
dualspiral: Well, this is actually a bit of a weird one because we've loaded the compound already, but on save we're loading it again and basically ignoring any changes to what we already have - this is the case with data and maybe inventories. So the correct approach is to just use what we have stored and write to that, I think.

We may want to add a way to invalidate any cache on disc in the user manager, but the use case for that will be few and far between, I'm sure.

Zidane: Ah you are performing a merge...I could see that being dangerous because who gets to say exactly what gets merged? Our code or what is on disk? What happens if a cached User is saved right before the hour mark is over yet a player has been playing the last 45 or so mins and the file is "more up to date"? I wonder if there is a way to "invalidate" User references but allow API consumers to know this happened? Trying to cover as many scenarios as possible.

dualspiral: It's difficult to say, however part of the point of this object is that any access will require a hit on the cache. That will then reset the timer.

That said, the biggest problem with the previous system was that we were holding data for a `User` and for a `Player` at the same time. This is no longer the case - if a player logs in, the `SpongeUserData` is saved, discarded, and then the `SpongeUserView` switches over to use the `ServerPlayer` as the data source automatically (though the opposite won't happen). We don't need to worry about keeping `User` and `ServerPlayer` in sync any more.

I'm not entirely sure how we could notify users that a `User` has expired, but I don't think it's going to be a significant problem.

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -273,34 +273,42 @@ protected void additionalActions(final Command.Builder builder) {
         return CommandResult.success();
     }
 
+    private CompletableFuture<Component> userIdToComponent(final @Nullable UUID uuid) {
+        if (uuid != null) {
+                return Sponge.server().userManager().load(uuid).handleAsync((userOptional, throwable) -> {
```
ImMorpheus: too much spacing ?

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -273,34 +273,42 @@ protected void additionalActions(final Command.Builder builder) {
...
-                    .append(notifier)
-                    .build());
+                final CompletableFuture<Component> creator = this.userIdToComponent(serverLocation.get(Keys.CREATOR).orElse(null));
+                final CompletableFuture<Component> notifier = this.userIdToComponent(serverLocation.get(Keys.NOTIFIER).orElse(null));
+                CompletableFuture.allOf(creator, notifier).thenAccept(x ->
+                        context.sendMessage(Identity.nil(), Component.text()
+                            .content("Block Info: ")
+                            .color(TextColor.color(SpongeCommand.GREEN))
+                            .append(Component.text(serverLocation.blockPosition().toString())
+                                    .hoverEvent(ItemStack.builder().fromBlockState(serverLocation.block()).build().createSnapshot())
```
ImMorpheus, line 303: This is not thread-safe. `ServerLocation#block` and `ItemStack#fromBlockState` should only be called from the server thread.

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -319,27 +327,24 @@ protected void additionalActions(final Command.Builder builder) {
...
+                        final CompletableFuture<Component> creator = this.userIdToComponent(locatableBlock.world()
+                                .get(locatableBlock.blockPosition(), Keys.CREATOR).orElse(null));
+                        final CompletableFuture<Component> notifier = this.userIdToComponent(locatableBlock.world()
+                                .get(locatableBlock.blockPosition(), Keys.CREATOR).orElse(null));
+                        CompletableFuture.allOf(creator, notifier).thenAccept(x ->
+                                context.sendMessage(Identity.nil(), Component.text()
+                                    .content("Block Info: ")
+                                    .color(TextColor.color(SpongeCommand.GREEN))
+                                    .append(Component.text(locatableBlock.blockPosition().toString())
+                                            .hoverEvent(ItemStack.builder().fromBlockState(locatableBlock.blockState()).build().createSnapshot())
```
ImMorpheus, line 339: Same as above.

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -357,23 +362,20 @@ protected void additionalActions(final Command.Builder builder) {
...
-                            .append(Component.text("Notifier: ", TextColor.color(SpongeCommand.MINT)))
-                            .append(notifier)
-                            .build());
+                        final CompletableFuture<Component> creator = this.userIdToComponent(entity.get(Keys.CREATOR).orElse(null));
+                        final CompletableFuture<Component> notifier = this.userIdToComponent(entity.get(Keys.NOTIFIER).orElse(null));
+                        CompletableFuture.allOf(creator, notifier).thenAccept(x ->
+                                context.sendMessage(Identity.nil(), Component.text()
+                                    .content("Entity Info: ")
+                                    .color(TextColor.color(SpongeCommand.GREEN))
+                                    .append(entity.type().asComponent().hoverEvent(entity))
```
ImMorpheus, line 371: Same as above, this will end up calling `EntityMixin_API#asHoverEvent`

src/main/java/org/spongepowered/common/entity/player/SpongeUserData.java
```diff
@@ -118,24 +113,53 @@
...
-    public SpongeUser(final GameProfile profile) {
+    public static SpongeUserData create(final GameProfile profile) throws IOException {
+        final ServerLevel world = SpongeCommon.server().overworld();
+        if (world == null) {
+            SpongeCommon.logger().warn("Overworld not initialized, cannot create users!");
+            throw new IllegalStateException("Overworld not initialized, cannot create users!");
+        }
+
+        final LevelStorageSource.LevelStorageAccess storageSource = ((MinecraftServerAccessor) Sponge.server()).accessor$storageSource();
+        final File file = storageSource.getLevelPath(LevelResource.PLAYER_DATA_DIR).resolve(profile.getId().toString() + ".dat").toFile();
```
ImMorpheus, line 137: Thought about migrating this to `Path` (java.nio) ?

src/main/java/org/spongepowered/common/entity/player/SpongeUserData.java
```diff
@@ -118,24 +113,53 @@
...
+            final int version = compound.contains("DataVersion", 3) ? compound.getInt("DataVersion") : -1;
+            NbtUtils.update(DataFixers.getDataFixer(), DataFixTypes.PLAYER, compound, version);
+            return new SpongeUserData(profile, compound);
+        } catch (final IOException e) {
+            SpongeCommon.logger().warn("Corrupt user file '{}'!", file, e);
+            throw e;
+        }
+    }
+
+    public SpongeUserData(final GameProfile profile, final CompoundTag tag) {
```
ImMorpheus: Should this be private ?

src/main/java/org/spongepowered/common/entity/player/SpongeUserData.java
```diff
@@ -417,34 +383,21 @@ public void markDirty() {
...
-        try (final FileOutputStream out = new FileOutputStream(file)) {
-            NbtIo.writeCompressed(compound, out);
-            SpongeUser.dirtyUsers.remove(this);
-            this.invalidate();
-        } catch (final IOException e) {
-            SpongeCommon.logger().warn("Failed to save user file [{}]!", file, e);
+        synchronized (this) {
+            final SpongeUserManager userManager = ((SpongeServer) SpongeCommon.server()).userManager();
+            final LevelStorageSource.LevelStorageAccess storageSource = ((MinecraftServerAccessor) Sponge.server()).accessor$storageSource();
+            final File file = storageSource.getLevelPath(LevelResource.PLAYER_DATA_DIR).resolve(this.uniqueId() + ".dat").toFile();
```
ImMorpheus, line 401: Same as above (migrate to path ?)

src/main/java/org/spongepowered/common/event/tracking/PhaseContext.java
```diff
@@ -396,6 +399,12 @@ protected P defensiveCopy(final PhaseTracker tracker) {
         return newCopy;
     }
 
+    private void checkNotCompleted() {
```
ImMorpheus, line 402: `requireNotCompleted` ?

src/main/java/org/spongepowered/common/user/SpongeUserManager.java
```diff
@@ -24,96 +24,404 @@
...
     @Override
-    public User findOrCreate(final GameProfile profile) {
-        return this.serverUserProvider.getOrCreateUser(this.ensureNonEmptyUUID(profile), false);
+    public CompletableFuture<Optional<User>> load(final String lastKnownName) {
+        Objects.requireNonNull(lastKnownName, "lastKnownName");
+        if (lastKnownName.isEmpty() || lastKnownName.length() <= 16) {
+            throw new IllegalArgumentException(String.format("Invalid username %s", lastKnownName));
+        }
+        final com.mojang.authlib.@Nullable GameProfile mcProfile =
+                this.server.getProfileCache().get(lastKnownName.toLowerCase(Locale.ROOT));
```
ImMorpheus: Since this is  using a `ConcurrentMap` under the hood and should be safe to call off thread, can we wrap this in a CompletableFuture and chain the result ?

ImMorpheus (changes requested):
WIP, I'm nearly done

testplugins/src/main/java/org/spongepowered/test/customdata/CustomDataTest.java
```diff
@@ -209,12 +211,14 @@ public void onJoin(final ServerSideConnectionEvent.Join event) {
...
-        if (user.isPresent()) {
-            final Integer integer = user.get().get(this.myDataKey).orElse(0);
-            this.plugin.logger().info("Custom data on user {}: {}", user.get().name(), integer);
-            user.get().offer(this.myDataKey, number);
-        }
+    private CompletableFuture<Void> customUserData(final UUID playerUUID, final int number) {
+        return Sponge.server().userManager().load(playerUUID)
+                .thenAccept(user -> {
+                    if (user.isPresent()) {
+                        final Integer integer = user.get().get(this.myDataKey).orElse(0);
```
ImMorpheus, line 218: This is not thread-safe, it's calling `PlayerList#getPlayer` under the hood

src/main/java/org/spongepowered/common/entity/player/SpongeUserData.java
dualspiral: I'm not doing that here. This is mirroring what Vanilla does and I don't really want to break from that, the comparison is easier then.

testplugins/src/main/java/org/spongepowered/test/customdata/CustomDataTest.java
```diff
@@ -135,11 +136,12 @@ public void onRegisterSpongeCommand(final RegisterCommandEvent<Command.Parameter
...
-                            final Scheduler scheduler = Sponge.server().scheduler();
-                            scheduler.submit(Task.builder().delay(Ticks.single()).execute(() -> this.customUserData(player.uniqueId(), number)).plugin(this.plugin).build());
-                            scheduler.submit(Task.builder().delay(Ticks.of(2)).execute(() -> this.customUserData(player.uniqueId(), number)).plugin(this.plugin).build());
+                            this.customUserData(player.uniqueId(), number).thenAcceptAsync(v -> {
+                                player.kick(Component.text("Setting User data..."));
+                                final Scheduler scheduler = Sponge.server().scheduler();
+                                scheduler.submit(Task.builder().delay(Ticks.single()).execute(() -> this.customUserData(player.uniqueId(), number).join()).plugin(this.plugin).build());
+                                scheduler.submit(Task.builder().delay(Ticks.of(2)).execute(() -> this.customUserData(player.uniqueId(), number).join()).plugin(this.plugin).build());
+                            }, Sponge.server().scheduler().createExecutor(this.plugin));
```
ImMorpheus, line 144: Since this is a testplugin it's a good idea to add `.exceptionally` and showcase exception handling

src/main/java/org/spongepowered/common/user/SpongeUserManager.java
dualspiral: I actually don't want it to go to Mojang for this so I'll just change the call.

src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerLoginPacketListenerImplMixin.java
```diff
@@ -94,36 +103,90 @@ public void handleAcceptedLogin() {
...
+
+                        try {
+                            this.server.getPlayerList()
+                                    .placeNewPlayer(this.connection, this.server.getPlayerList().getPlayerForLogin(this.gameProfile));
+                        } catch (final Exception e) {
+                            throw new RuntimeException(e);
+                        }
+                    }
+                    return null;
+                }, SpongeCommon.server()).handle((ignored, throwable) -> {
```
ImMorpheus: `.exceptionally` ?

src/main/java/org/spongepowered/common/user/SpongeUserManager.java
ImMorpheus: Fair enough

ImMorpheus (changes requested):

src/main/java/org/spongepowered/common/entity/player/SpongeUserData.java
ImMorpheus: Fair enough

testplugins/src/main/java/org/spongepowered/test/customdata/CustomDataTest.java
dualspiral: ...but it's not going to call an exception here, so why?

I'm not interested in changing the test plugins in this PR, only to get them working. Once this is merged, if you want to update the plugins, have at it.

ImMorpheus: > I'm not interested in changing the test plugins in this PR, only to get them working.

:+1:

PR #3476, opened 2021-07-17
Fix bell blockentity memory leak

ImMorpheus:
Note: optimization module/category needs to be reimplemented

EDIT: done

### DIFF
--- a/src/applaunch/java/org/spongepowered/common/applaunch/config/common/OptimizationCategory.java
+++ b/src/applaunch/java/org/spongepowered/common/applaunch/config/common/OptimizationCategory.java
@@ -50,6 +50,14 @@ public final class OptimizationCategory {
     @Comment("Uses theosib's redstone algorithms to completely overhaul the way redstone works.")
     public final EigenRedstoneCategory eigenRedstone = new EigenRedstoneCategory();
 
+    @Setting(value = "bell-leak-fix")
+    @Comment("Bells will store references of nearby entities when rang.\n"
+            + "The entity list is never cleared, thus leaking memory until\n"
+            + "the chunk is unloaded. Since the entity list is reused if the\n"
+            + "bell is rang again within 60 ticks, this provides an option to\n"
+            + "clear the list if it is not needed anymore.")
+    public boolean bellLeak = true;
+
     @Setting("faster-thread-checks")
     @Comment("If 'true', allows for Sponge to make better assumptions on single threaded\n"
                + "operations with relation to various checks for server threaded operations.\n"
--- a/src/mixins/java/org/spongepowered/common/mixin/optimization/world/level/block/entity/BellBlockEntityMixin_Optimization_BellLeak.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/optimization/world/level/block/entity/BellBlockEntityMixin_Optimization_BellLeak.java
@@ -1,0 +1,48 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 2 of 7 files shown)


src/applaunch/java/org/spongepowered/common/applaunch/config/common/OptimizationCategory.java
```diff
@@ -51,6 +51,9 @@
     @Comment("Uses theosib's redstone algorithms to completely overhaul the way redstone works.")
     public final EigenRedstoneCategory eigenRedstone = new EigenRedstoneCategory();
 
+    @Setting(value = "bell-leak-fix")
+    private boolean bellLeak = true;
```
dualspiral: Could you add a quick comment to explain what it does, just so people know what they are turning on/off?

Zidane:
Does this have a corresponding Mojira ticket that could be linked?

src/applaunch/java/org/spongepowered/common/applaunch/config/common/OptimizationCategory.java
ImMorpheus: I've added a small comment explaining the issue and how this option should fix it

ImMorpheus:
> Does this have a corresponding Mojira ticket that could be linked?

No, but I don't know if it's still an issue in 1.17

src/mixins/java/org/spongepowered/common/mixin/optimization/world/level/block/entity/BellBlockEntityMixin_Leak.java
```diff
@@ -0,0 +1,48 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import java.util.List;
+
+@Mixin(BellBlockEntity.class)
+public abstract class BellBlockEntityMixin_Leak {
```
dualspiral: `BellBlockEntityMixin_Optimization_BellLeak` - if we fix another leak like this the names should be separate.

src/mixins/java/org/spongepowered/common/mixin/optimization/world/level/block/entity/BellBlockEntityMixin_Leak.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Shadow private List<LivingEntity> nearbyEntities;
+
+    @Inject(method = "tick", at = @At(value = "FIELD", target = "Lnet/minecraft/world/level/block/entity/BellBlockEntity;resonating:Z", opcode = Opcodes.PUTFIELD, ordinal = 2))
```
dualspiral: I'd rather this used a slice rather than an ordinal for selection to ensure we have the right one.

dualspiral (commented):
Note: this is going to need adding to `mixins.sponge.optimization.json` and `OptimizationPlugin` to be effective (as suggested by the original comment).

Is this going to be considered an "optimisation" or should this be a "fix" category?

I've opened up the Minecraft sources to verify that this is the right fix. It **almost** is. The actual logic is a little more complicated than "just clear when we stop reasonating" and I think that for any optimisation and fix that we do there needs to be an explanation as to why it's needed and what it does - along with any logic gotchas here.

The gotcha is when `updateEntities()` is called and what happens. `updateEntities()` occurs when a block event triggers - in this case, the bell is triggered. This resets `ticks` to zero and, crucially, `resonationTicks`. It also resets `nearbyEntities` - so on the next tick, `areRaidersNearby` will work fine.

The time where it is safe to clear the nearby entities is after `showBellParticles` is called in `tick`, because it isn't used again until the bell is triggered.

**...except in one situation.** If we're less than 61 ticks since the last event (updateEntities, line 92), you've cleared the list, but the logic repopulating the nearby entities does not happen - so the bell operates on nothing, which is a change from the behaviour of this game. There is a half second window for this to happen - not negligible, so I think this is being cleared in the wrong place. Instead, I think it should clear at 60 ticks.

Interestingly, the lines below check for alive and removed entities, they're obviously aware of this fact - though I agree, there would have been better ways to handle this.

Zidane:
Additionally I think a mixin plugin needs to get added to Sponge that fixes bugs.

With the requirement that all bug fixing mixins *must* reference a Mojira so that we can consult it every Minecraft update and see if this is still even needed.

src/mixins/java/org/spongepowered/common/mixin/optimization/world/level/block/entity/BellBlockEntityMixin_Optimization_BellLeak.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.List;
+
+@Mixin(BellBlockEntity.class)
+public abstract class BellBlockEntityMixin_Optimization_BellLeak extends BlockEntityMixin {
+
+    @Shadow private List<LivingEntity> nearbyEntities;
+    @Shadow private long lastRingTimestamp;
+
+    @Inject(method = "tick", at = @At(value = "HEAD"))
+    private void impl$onStopResonating(final CallbackInfo ci) {
```
gabizou:
```suggestion
    private void bellleakfix$clearNearbyEntitiesWhenDoneResonating(final CallbackInfo ci) {
```

so that it's clearer when reviewing the method calls if this ever shows up in a stack trace ro something.

PR #3477, opened 2021-07-18
Add lifecycle state for stopped game, and hook in executors

### LINKED ISSUE #3472: Server shutdown hangs (SV 1.16.5 and 1.17.1)
The SCS is using:
    SpongeVanilla: 1.17.1-9.0.0-RC742
    JVM: 16.0.1/64-bit (Private Build)
    OS: Linux/5.4.0-66-generic (amd64)

with the usual gaggle of CubeEngine plugins. 

On shutdown, it looks good until the end, where it just hangs. I have to use ctrl-c to exit the jvm. 
This only happens *after* a player has logged in to the server - otherwise it shuts down cleanly.
This has been happening on 1.16.5 and 1.17.1 SpongeVanilla servers. I just tested SpongeVanilla 1.16.5 - RC 761 in the same environment and it did the same.

Stack traces for the hanging server thread:
1.16.5 (RC 761)
[stackdump16.txt](https://github.com/SpongePowered/Sponge/files/6834196/stackdump16.txt)
1.17.1 (RC 742)
[stackdump17.txt](https://github.com/SpongePowered/Sponge/files/6834197/stackdump17.txt)

zml2008:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2379)

Fixes #3472 (though today's earlier user refactor likely means that the issue will no longer be triggered)

Implements the new `StoppedGameEvent`

### DIFF
--- a/src/main/java/org/spongepowered/common/SpongeLifecycle.java
+++ b/src/main/java/org/spongepowered/common/SpongeLifecycle.java
@@ -31,6 +31,7 @@ import com.google.inject.Injector;
 import com.google.inject.Singleton;
 import io.leangen.geantyref.TypeToken;
 import net.minecraft.core.Registry;
+import org.checkerframework.checker.nullness.qual.Nullable;
 import org.spongepowered.api.Client;
 import org.spongepowered.api.Engine;
 import org.spongepowered.api.Game;
@@ -52,10 +53,12 @@ import org.spongepowered.common.event.lifecycle.RegisterFactoryEventImpl;
 import org.spongepowered.common.event.tracking.PhaseTracker;
 import org.spongepowered.common.launch.plugin.DummyPluginContainer;
 import org.spongepowered.common.network.channel.SpongeChannelManager;
+import org.spongepowered.common.profile.SpongeGameProfileManager;
 import org.spongepowered.common.registry.SpongeBuilderProvider;
 import org.spongepowered.common.registry.SpongeFactoryProvider;
 import org.spongepowered.common.registry.SpongeRegistries;
 import org.spongepowered.common.registry.SpongeRegistryHolder;
+import org.spongepowered.common.scheduler.AsyncScheduler;
 import org.spongepowered.common.service.SpongeServiceProvider;
 import org.spongepowered.common.service.server.permission.SpongeContextCalculator;
 import org.spongepowered.plugin.PluginContainer;
@@ -199,6 +202,20 @@ public final class SpongeLifecycle {
     public void callStoppingEngineEvent(final Engine engine) {
         this.game.eventManager().post(SpongeEventFactory.createStoppingEngineEvent(PhaseTracker.getCauseStackManager().currentCause(),
                 engine, this.game, (TypeToken<Engine>) TypeToken.get(engine.getClass())));
+        if (engine instanceof SpongeServer) {
+            final @Nullable SpongeGameProfileManager profileManager = ((SpongeServer) engine).gameProfileManagerIfPresent();
+            if (profileManager != null) {
+                profileManager.close();
+            }
+        }
+    }
+
+    public void callStoppedGameEvent() {
+        // Call an event for plugins to shut down any thread pools
+        this.game.eventManager().post(SpongeEventFactory.createStoppedGameEvent(PhaseTracker.getCauseStackManager().currentCause(), this.game));
+
+        // Then shut down our own thread pool
+        ((AsyncScheduler) this.game.asyncScheduler()).close();
     }
 
     private Collection<PluginContainer> filterInternalPlugins(final Collection<PluginContainer> plugins) {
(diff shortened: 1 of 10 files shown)


src/main/java/org/spongepowered/common/SpongeLifecycle.java
```diff
@@ -199,6 +202,20 @@ public void callLoadedGameEvent() {
...
+            }
+        }
+    }
+
+    public void callStoppingGameEvent() {
+        // Call an event for plugins to shut down any thread pools
+        this.game.eventManager().post(SpongeEventFactory.createStoppedGameEvent(PhaseTracker.getCauseStackManager().currentCause(), this.game));
+
+        // Then shut down our own thread pool
+        ((AsyncScheduler) this.game.asyncScheduler()).close();
```
Zidane, line 218: Would it be better to do this in a shutdown hook?

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/server/MinecraftServerMixin_API.java
```diff
@@ -350,6 +350,11 @@ public GameProfileManager gameProfileManager() {
         return this.api$profileManager;
     }
 
+    @Override
+    public SpongeGameProfileManager gameProfileManagerIfPresent() {
```
Zidane, line 354: Just `gameProfileManager`

src/main/java/org/spongepowered/common/SpongeLifecycle.java
zml2008: If we did that, the async scheduler's thread pool would have to use deamon threads, which loses us control, and the ability to warn on stuck tasks when the server is shutting down.

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/server/MinecraftServerMixin_API.java
zml2008: There's already a `gameProfileManager()` method (just above), that creates the game profile manager when its called. for shutdown purposes, we don't want to create a new manager instance though, which is why this method exists.

src/main/java/org/spongepowered/common/SpongeLifecycle.java
Zidane: Ohhh I see now, did not consider that.

src/main/java/org/spongepowered/common/SpongeLifecycle.java
```diff
@@ -199,6 +202,20 @@ public void callLoadedGameEvent() {
...
                 engine, this.game, (TypeToken<Engine>) TypeToken.get(engine.getClass())));
+        if (engine instanceof SpongeServer) {
+            final @Nullable SpongeGameProfileManager profileManager = ((SpongeServer) engine).gameProfileManagerIfPresent();
+            if (profileManager != null) {
+                profileManager.close();
+            }
+        }
+    }
+
+    public void callStoppingGameEvent() {
```
Zidane: `callStoppedGameEvent`

src/main/java/org/spongepowered/common/profile/SpongeGameProfileManager.java
```diff
@@ -171,12 +182,32 @@ public void lookupUserAsync(final UUID uniqueId) {
...
             }
         });
     }
+
+    public void close() {
+        // Shut down the executor
+        this.gameLookupExecutorService.shutdown();
+
+        try {
+            if (!this.gameLookupExecutorService.awaitTermination(10, TimeUnit.SECONDS)) {
```
Zidane: I wonder if this may be too high? Maybe 5 is more reasonable unless this value was sourced from elsewhere.

src/mixins/java/org/spongepowered/common/mixin/core/client/MinecraftMixin.java
```diff
@@ -103,8 +100,8 @@
     }
 
     @Inject(method = "close", at = @At(value = "INVOKE", target = "Lnet/minecraft/Util;shutdownExecutors()V"))
-    private void impl$shutdownAsyncScheduler(final CallbackInfo ci) {
-        SpongeCommon.game().asyncScheduler().close();
+    private void impl$callStoppingGame(final CallbackInfo ci) {
```
Zidane: `impl$callStoppedGame`

src/mixins/java/org/spongepowered/common/mixin/core/server/dedicated/DedicatedServerMixin.java
```diff
@@ -66,8 +67,8 @@ private void onSave(final GameProfileCache cache) {
     }
 
     @Inject(method = "stopServer", at = @At("TAIL"))
-    private void impl$shutdownAsyncScheduler(final CallbackInfo ci) {
-        SpongeCommon.game().asyncScheduler().close();
+    private void impl$callStoppingGame(final CallbackInfo ci) {
```
Zidane: See above

src/main/java/org/spongepowered/common/profile/SpongeGameProfileManager.java
zml2008: sure? It's a bit of a block i've copied around a few projects, haven't thought about the time in a while

PR #3484, opened 2021-07-24
Implement Entity#setPosition and clean up some related logic

dualspiral:
Superscedes https://github.com/SpongePowered/Sponge/pull/3442

With many thanks to @Lignium for starting this. The further changes attempts to unify some of the move event handling and reduce duplication. Some cleanup has occurred too, converting methods that didn't actually need to be bridge methods to `impl$` methods.

I want to test it once more tomorrow before I merge.

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
@@ -72,7 +72,6 @@ import org.spongepowered.api.event.entity.ChangeEntityWorldEvent;
 import org.spongepowered.api.event.entity.IgniteEntityEvent;
 import org.spongepowered.api.event.entity.MoveEntityEvent;
 import org.spongepowered.api.util.Ticks;
-import org.spongepowered.api.util.Transform;
 import org.spongepowered.api.world.portal.Portal;
 import org.spongepowered.api.world.portal.PortalTypes;
 import org.spongepowered.api.world.server.ServerLocation;
@@ -94,7 +93,6 @@ import org.spongepowered.common.bridge.data.DataCompoundHolder;
 import org.spongepowered.common.bridge.data.VanishableBridge;
 import org.spongepowered.common.bridge.world.entity.EntityBridge;
 import org.spongepowered.common.bridge.world.entity.PlatformEntityBridge;
-import org.spongepowered.common.bridge.server.level.ServerPlayerBridge;
 import org.spongepowered.common.bridge.world.damagesource.DamageSourceBridge;
 import org.spongepowered.common.bridge.world.level.PlatformServerLevelBridge;
 import org.spongepowered.common.bridge.world.level.LevelBridge;
@@ -150,6 +148,7 @@ public abstract class EntityMixin implements EntityBridge, PlatformEntityBridge,
     @Shadow public double xo;
     @Shadow public double yo;
     @Shadow public double zo;
+    @Shadow private int remainingFireTicks;
 
     @Shadow public abstract void shadow$setPos(double x, double y, double z);
     @Shadow public abstract double shadow$getX();
@@ -205,8 +204,6 @@ public abstract class EntityMixin implements EntityBridge, PlatformEntityBridge,
     @Shadow protected abstract boolean shadow$fireImmune();
     // @formatter:on
 
-    @Shadow private int remainingFireTicks;
-
     private boolean impl$isConstructing = true;
     private boolean impl$vanishPreventsTargeting = false;
     private boolean impl$isVanished = false;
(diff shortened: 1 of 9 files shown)


src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
```diff
@@ -237,6 +236,27 @@
...
     }
 
+    @Override
+    public boolean bridge$setPosition(final Vector3d position) {
+        if (this.removed) {
+            return false;
+        }
+
+        try (final CauseStackManager.StackFrame frame = PhaseTracker.getCauseStackManager().pushCauseFrame()) {
+            frame.pushCause(SpongeCommon.activePlugin());
```
Zidane: This should not be necessary as the plugin is set when we invoke a listener or scheduled task.

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
```diff
@@ -237,6 +236,27 @@
...
+            frame.addContext(EventContextKeys.MOVEMENT_TYPE, MovementTypes.PLUGIN);
+            final Vector3d destinationPosition = this.impl$fireMoveEvent(PhaseTracker.SERVER, position);
+            if (destinationPosition == null) {
+                return false;
+            }
+            if (this instanceof ServerPlayerBridge) {
```
Zidane: Despite the desire to remove duplicated code, I have always disliked these instance checks when it feels better to practice OOP. By having it in ServerPlayer. What do you think?

Zidane (approved):

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
```diff
@@ -247,70 +267,53 @@
...
-            this.bridge$revive();
-            this.shadow$setLevel(destinationWorld);
-            destinationWorld.addFromAnotherDimension((Entity) (Object) this);
-
-            originalWorld.resetEmptyTime();
-            destinationWorld.resetEmptyTime();
-
-            final ChunkPos chunkPos = new ChunkPos((int) this.shadow$getX() >> 4, (int) this.shadow$getZ() >> 4);
-            destinationWorld.getChunkSource().addRegionTicket(TicketType.POST_TELEPORT, chunkPos, 1, ((Entity) (Object) this).getId());
+            return this.impl$teleportToWithTicket(destinationPosition.x(), destinationPosition.y(), destinationPosition.z(), false);
```
Zidane: Here is a good discussion.

Does setting the location of a non player trigger a chunk ticket? Should we be doing that?

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
dualspiral: Sure - but I also want to make a policy clear here because it's not that clear at the moment, what is our policy on overriding bridge methods?

I was under the (recent) impression that overriding bridge methods in sub classes shouldn't be done, though I'm seeing plenty of examples in the codebase. Platform bridges are also inherently designed to be overridden. Am I wrong on that, did I get the wrong end of the stick?

That said, I'd just change this to be an `impl$...` method that is overridden in a subclass. That's what I did elsewhere.

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
dualspiral: If the chunk isn't loaded and the entity is moved there, where would the entity be saved to? Does the chunk need to be loaded to save the entity's location to it?

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
Zidane: Bridge methods should only exist if we need to invoke code from non-mixins against mixed classes. Otherwise it is `impl$`/etc. As for OOP, there is nothing that prevents overriding bridge methods or even `impl$` methods. In fact, I 100% encourage it.

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
Zidane: That is a good question dual. In fact, it is likely premature to discuss this now since entities still exist in Chunk files in 1.16.5. In 1.17, they exist outside chunk files and it likely makes a lot of sense to reposition them and do a save without triggering a chunk load.

dualspiral: That's what I thought, I knew a split happened in 1.17.x - so I suspect that the answer is we need the ticket in 1.16, revisit in 1.17...

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
dualspiral: Alright. I was overriding impl$ methods anyway, so that's likely what will happen here too.

PR #3486, opened 2021-07-26
SpongeFabric

Logiquo:
Currently WIP. Use a custom version of `fabric-loom` to build & use a custom version of `fabric-loader` & `mixin` to start. All mixin codes basically copy & paste from SpongeVanilla. Can only build a prod version; I am still trying to set up the dev environment.

Tasks List:
- [ ] Find an appropriate entry point to download the dependencies (currently use a modified `fabric-loader` to work around this).
- [ ] Waiting for a new Mixin release that fixes the LVT error.
- [ ] PluginEngine is incomplete.
- [ ] Need to test compatibility with other fabric mods (especially fabric api).


Zidane:
I am going to close this but I want to explain, for those onlookers and for Logic, exactly why.

Unless the person lives under a rock, everyone knows that Sponge ships 2 implementations: SpongeVanilla and SpongeForge. Sponge originally started out on SpongeForge (simply Sponge back then) and over time we absorbed a project called "Granite" which was basically SpongeVanilla. In hindsight, SpongeVanilla was inevitable...we have to be able to run our modifications clean of any 3rd party code.

So fast forward several years and now we arrive at these two distinct implementations. Vanilla and Forge. Our roots steep directly into Forge, our project was birthed on comingling plugins and mods. While as a team we consider SpongeVanilla to be our flagship in 1.16.5 and higher, SpongeForge is definitely not playing a second fiddle...it will get first class support as always.

So that brings us to Fabric. My team and my fellow leaders aren't ignorant; we know there is demand for Sponge on Fabric. Unfortunately if it was as easy as waving a magic wand, the Fabric implementation would exist already. With a project as large as Sponge, manpower across the devs are our most valuable and lacking commodity. We do not want to commit to an additional platform without staffing to back it up.

With this in mind, I want to thank Logic for the work done thus far. I would encourage the work to continue, despite the closure, as our stance may change down the line. As we have done so far, we will continue to provide the necessary support (via code review or specific changes) so that this project can continue. It just won't be under the Sponge umbrella at this time.

PS As for actual review, Logic you need to re-think the approach in this PR. You cannot be the one to bootstrap Fabric (rather you _shouldn't_) as this introduces logistics issues when running a Fabric implementation on clients. Additionally, you should make use of Fabric subsystems (where possible) and ensure that Fabric mods are Sponge plugins and vice versa. Basically how SF worked on 1.12.2 and will continue to work on 1.16.5

PR #3504, opened 2021-08-08
Rejig teleporting to fit with Forge.

dualspiral:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2386) | **Sponge**

* Need to put hooks subsitution into the right place.
* Syncs ChangeEntityWorldEvent.Pre/Post with Forge equivalents
* Remove hook for the Post event - because it's literally just calling the event - syncing handles it all.


forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/event/entity/EntityTravelToDimensionEventMixin_Forge.java
```diff
@@ -0,0 +1,76 @@
...
+import org.spongepowered.api.event.entity.ChangeEntityWorldEvent;
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Mutable;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.common.SpongeCommon;
+import org.spongepowered.common.event.tracking.PhaseTracker;
+import org.spongepowered.forge.launch.bridge.event.ForgeEventBridge_Forge;
+
+@Mixin(value = EntityTravelToDimensionEvent.class, remap = false)
```
gabizou, line 45: Does this disable remapping the obfuscated names for the native methods?

forge/src/mixins/java/org/spongepowered/forge/mixin/core/server/level/ServerPlayerMixin_Forge.java
```diff
@@ -0,0 +1,57 @@
...
+     *
+     *         This will get called on the nether dimension changes, as the
+     *         end portal teleport call itself has been redirected to provide
+     *         the correct type.
+     */
+    @Overwrite
+    @Nullable // should be javax.annotations.Nullable
+    public Entity changeDimension(final ServerLevel serverLevel, final ITeleporter teleporter) {
+        return ((EntityBridge) this).bridge$changeDimension(serverLevel, (PortalLogic) teleporter);
+    }
```
gabizou, line 55: technically you could inject at head and cancel if you wanted to eliminate the overwrite entirely

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/event/entity/EntityTravelToDimensionEventMixin_Forge.java
dualspiral: There shouldn't be any because the base class is a Forge class.

forge/src/mixins/java/org/spongepowered/forge/mixin/core/server/level/ServerPlayerMixin_Forge.java
dualspiral: I thought about that, but we use overwrites elsewhere too. I think if we wanted to inject and cancel, we should _always_ inject and cancel and discourage overwrites.

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/event/entity/EntityTravelToDimensionEventMixin_Forge.java
gabizou: no, but you call methods on objects that are returned, right? Like `Level.dimension()`?

gabizou: I just can't remember whether disabling remap on the target disables remap overall for the entire mixin

dualspiral: From what I remember, it's more concerned with the mixin _targets_, rather than the contents (so, I think this is fine).

PR #3507, opened 2021-08-14
Dependency resolving and sorting

dualspiral:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2390) | **Sponge**

Sorts dependencies, passes them to the load routine in order. This only affects plugin construction, not event firing order.

Also detects cyclic and missing deps, pretty prints an error (it needs to be noticible).

Wording can be changed.

image

(also this PR saw -spi bump to 0.3, -meta bump to 0.8 and a whole load of metadata file changes, as well as the rename of plugins.json to sponge_plugins.json... but it still has the dep order changes!)


vanilla/src/launch/java/org/spongepowered/vanilla/launch/plugin/VanillaPluginManager.java
```diff
@@ -86,7 +88,9 @@ public void loadPlugins(final VanillaPluginPlatform platform) {
             } catch (final InstantiationException | IllegalAccessException | ClassNotFoundException | NoSuchMethodException | InvocationTargetException e) {
                 throw new RuntimeException(e);
             }
-            for (final PluginCandidate<PluginResource> candidate : candidates) {
+            final ResolutionResult<PluginResource> resolutionResult = DependencyResolver.resolveAndSortCandidates(candidates);
+            resolutionResult.printErrorsIfAny(platform.logger());
+            for (final PluginCandidate<PluginResource> candidate : resolutionResult.sortedSuccesses()) {
                 PluginContainer plugin = this.plugins.get(candidate.metadata().id());
```
Zidane: I wonder if we should be keeping around the failed plugins in the manager?

vanilla/src/launch/java/org/spongepowered/vanilla/launch/plugin/resolver/DependencyResolver.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public final class DependencyResolver {
```
Zidane, line 48: I wonder if this would look cleaner overall as an instance on `VanillaPluginManager`? I try to avoid statics if I can.

vanilla/src/launch/java/org/spongepowered/vanilla/launch/plugin/VanillaPluginManager.java
dualspiral: The failed candidates are available in the result, so we can do that if you so wish, easily.

Zidane: I miss understood the code, perhaps.

Does `this.plugins` keep the failed plugins after the `DependencyResolver` runs?

vanilla/src/launch/java/org/spongepowered/vanilla/launch/plugin/resolver/DependencyResolver.java
dualspiral: I was on the edge about doing that - don't really have a preference. I do tend to try to avoid statics from a testing point of view, but it makes little to no difference here because it's written in a way that doesn't store a state (rather, it's passed around). If you were going to make the plugin manager testable, on the other hand, then there'd be something to gain, but I suspect that might not happen.

As for your cleaner argument, I would disagree - creating an instance and storing it as opposed to a direct call isn't really cleaner, it's less so if anything. That's most certainly personal preference though. Static methods do have their place.

Zidane: Fair enough, your counter arguments are difficult to rebut.

vanilla/src/launch/java/org/spongepowered/vanilla/launch/plugin/VanillaPluginManager.java
dualspiral: No, because the plugin containers are never constructed. I was just saying we can do that easily because I still have the failed candidates - we don't want to make the instances so we'd have to have logic for creating failed containers.

If we're going to do that, I think I'd want to make sure containers have a way of indicating their status.

Zidane: I was unclear with my original response, I didn't want them to stick around.

..that said I could see @Grinch wanting to show a plugin as *bad* on the GUI Plugins screen...it didn't load because of: <reason>. Granted it is hard for me to justify the additional backend work simply for that screen's benefit. I am not sure what, if anything else, would need to know what failed past your printout.

dualspiral: Ah, right. For the absence of doubt:

The resolver returns all the candidates, but separated. `resolutionResult.sortedSuccesses()` will only proceed with those we know have a good dependency tree, they are the only ones that become `PluginContainer`s.

If you'd like to store failed plugins, the candidates are available through the other methods on the result.

dualspiral:
We now do cross language-loader dependency management, and bailing out of loading plugins where a required dependency failed to be constructed.

Tested with the test plugins and causing the neighbour notify test to fail, with the custom data test depending on that. This is the result.

image

vanilla/src/launch/java/org/spongepowered/vanilla/launch/plugin/resolver/DependencyResolver.java
```diff
@@ -0,0 +1,228 @@
...
+
+    public static <T extends PluginResource> ResolutionResult<T> resolveAndSortCandidates(final Collection<PluginCandidate<T>> candidates) {
+        final Map<String, Node<T>> nodes = new HashMap<>();
+        final ResolutionResult<T> resolutionResult = new ResolutionResult<>();
+        for (final PluginCandidate<T> candidate : candidates) {
+            final String id = candidate.metadata().id();
+            // If we already have an entry, this is now a duplicate ID situation.
+            if (nodes.containsKey(id)) {
```
ImMorpheus, line 57: `containsKey` + `put` -> `putIfAbsent` ?

dualspiral: Don't want to create a new `Node` if we don't have to.

src/main/java/org/spongepowered/common/command/sponge/SpongeCommand.java
```diff
@@ -787,7 +787,7 @@ private Component createContainerMeta(final PluginMetadata pluginMetadata) {
             }
         }
 
-        this.appendPluginMeta(builder, "Main class", pluginMetadata.mainClass());
+        this.appendPluginMeta(builder, "Main class", pluginMetadata.entrypoint());
```
Zidane: "Main Class" -> "Entrypoint"

vanilla/src/launch/java/org/spongepowered/vanilla/launch/VanillaLaunch.java
```diff
@@ -144,23 +145,25 @@ protected final void createPlatformPlugins() {
...
                         }
                     }
                 }
             }
 
             if (read == null) { // other measures failed, fall back to directly querying the classpath
-                read = PluginMetadataHelper.builder().build().read(VanillaLaunch.class.getResourceAsStream(
-                        "/META-INF/" + JVMPluginResourceLocatorService.DEFAULT_METADATA_FILENAME));
+                final Path vanillaPath =
+                        Paths.get(VanillaLaunch.class.getResource("/META-INF/" + JVMPluginResourceLocatorService.DEFAULT_METADATA_FILENAME).toURI());
```
Zidane: https://github.com/SpongePowered/plugin-spi/blob/0.3.x/src/main/java/org/spongepowered/plugin/jvm/locator/JVMPluginResourceLocatorService.java#L37

vanilla/src/launch/java/org/spongepowered/vanilla/launch/plugin/JavaPluginLoader.java
```diff
@@ -25,37 +25,49 @@
...
 
 import java.util.Optional;
 
-public final class JavaPluginLoader extends JVMPluginLoader<JVMPluginContainer> {
+public final class JavaPluginLoader extends JVMPluginLoader<JVMPluginResource, VanillaJavaPluginContainer> {
 
     @Override
-    public Optional<JVMPluginContainer> createPluginContainer(final PluginCandidate<JVMPluginResource> candidate, final PluginEnvironment environment) {
+    public ArtifactVersion version() {
+        return null;
```
Zidane: I didn't plan to actually bump the version of this.

So 1.0 will be sufficient as the version.

PR #3508, opened 2021-08-14
Implement PotionEffect#showIcon

ImMorpheus:


src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/effect/MobEffectInstanceMixin_API.java
```diff
@@ -87,6 +93,7 @@ public DataContainer toContainer() {
                 .set(Constants.Item.Potions.POTION_DURATION, this.duration)
                 .set(Constants.Item.Potions.POTION_AMPLIFIER, this.amplifier)
                 .set(Constants.Item.Potions.POTION_AMBIANCE, this.ambient)
-                .set(Constants.Item.Potions.POTION_SHOWS_PARTICLES, this.visible);
+                .set(Constants.Item.Potions.POTION_SHOWS_PARTICLES, this.visible)
+                .set(Constants.Item.Potions.POTION_SHOWS_ICON, this.showIcon);
```
ImMorpheus, line 100: Do we need to bump the content version ?

dualspiral: @gabizou Am I right in thinking that we don't need to bump the version if we're just adding a field, or is it good practice to anyway?

gabizou: It's actually required to bump the version because it's a non-optional field whereas previous potions will not have the icon available.

src/main/java/org/spongepowered/common/effect/potion/SpongePotionBuilder.java
```diff
@@ -89,10 +89,12 @@ public SpongePotionBuilder() {
         final int amplifier = container.getInt(Constants.Item.Potions.POTION_AMPLIFIER).get();
         final boolean ambience = container.getBoolean(Constants.Item.Potions.POTION_AMBIANCE).get();
         final boolean particles = container.getBoolean(Constants.Item.Potions.POTION_SHOWS_PARTICLES).get();
+        final boolean showsIcon = container.getBoolean(Constants.Item.Potions.POTION_SHOWS_ICON).get();
```
gabizou: Gonna need a data version bump for potions to handle icons missing.

gabizou (changes requested):

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/effect/MobEffectInstanceMixin_API.java
ImMorpheus: I've made the field optional

src/main/java/org/spongepowered/common/effect/potion/SpongePotionBuilder.java
ImMorpheus: I've made the field optional, it's using the default value if the icon is missing

PR #3509, opened 2021-08-15
Fix World#min() and constants for biomes

Lignium:


dualspiral:
I'm happy to correct `World#max()` (not `min` as you state in your commit message) and will do so directly, but I'm less sure about the biome constants because they're not even used. No point assiging a constant to the same constant, especially if they're then not going to be used.

I'm just going to remove them - if someone wants them they can re-add them.

dualspiral:
https://github.com/SpongePowered/Sponge/commit/0433253eeddf2fba81c9da5f84a7391ccb541afd

PR #3511, opened 2021-08-27
Resolve Stack Overflow issues with Objectives, fixes #3510

### LINKED ISSUE #3510: [API 8] Subsequent Attempts to Update a Scoreboard Objective Result in Stack Overflow Exceptions
**I am currently running**

- SpongeVanilla version: 1.16.5-8.0.0-RC865
- Java version: 1.8.0_282
- Operating System: Ubuntu 18.04

- Plugins/Mods: Impactor, ImpactDev Interface

**Issue Description**
In the ObjectiveMixin routine for setDisplayName (onSetDisplayName as declared in the mixin), line 76 is incorrectly delegating back to the sponge implementation of objective to set the display name, which eventually roots back to this trouble line. This is causing a stack overflow, and the operation fails. 

https://github.com/SpongePowered/Sponge/blob/d52b62fdf840e49f100cc7230f4a405425fec05c/src/mixins/java/org/spongepowered/common/mixin/core/world/scores/ObjectiveMixin.java#L76

Coincidentally, it seems that a first attempt at updating the display name may pass through just fine, and the update is applied correctly. To test, I have join and disconnect listeners attached that are meant to update the objective display name with the player count per event. Seemingly, on a fresh start, the join listener is able to apply the change with no issue. However, when leaving the server, and subsequently joining, the stack overflow commences, running between the following four lines:
```
at net.minecraft.world.scores.Objective.handler$znj000$onSetDisplayName(SourceFile:1076) ~[?:?]
at net.minecraft.world.scores.Objective.setDisplayName(SourceFile) ~[?:?]
at org.spongepowered.common.scoreboard.SpongeObjective.updateDisplayName(SpongeObjective.java:200) ~[SpongeObjective.class:1.16.5-8.0.0-RC865]
at org.spongepowered.common.scoreboard.SpongeObjective.setDisplayName(SpongeObjective.java:86) ~[SpongeObjective.class:1.16.5-8.0.0-RC865]
```

For reference, the listeners attempting to update the objective are here:
```Java
Sponge.eventManager().registerListener(EventListenerRegistration.builder(ServerSideConnectionEvent.Join.class)
	.plugin(Plugin.container.getInstance().getPluginContainer())
	.listener(event -> {
		objective.setDisplayName(tpus.parse(title, Lists.newArrayList(() -> player)));
	})
	.order(Order.DEFAULT)
	.build());
Sponge.eventManager().registerListener(EventListenerRegistration.builder(ServerSideConnectionEvent.Disconnect.class)
	.plugin(Plugin.container.getInstance().getPluginContainer())
	.listener(event -> {
		objective.setDisplayName(tpus.parse(title, Lists.newArrayList(() -> player)));
	})
	.order(Order.DEFAULT)
	.build());
```
(Sidenote, curious on why order is a required parameter for these dynamic event listener registrations, would think they could just default to default)

NickImpact:
This removes callbacks to the sponge implementation that eventually recall the same mixin methods that were removed, ultimately ending in stack overflows. Verified the changes work as desired locally, and can provide the test plugin suite if need be for additional case proving.


Zidane:
So this is an interesting issue.

You are correct, I see the stack overflow and if we only cared about the API, you would be 100% right. The problem is if something in Vanilla/a mod end up calling setDisplayName/setRenderType we need to catch those changes to update the wrapper object.

What actually needs to happen is methods should be added to `SpongeObjective` that updates these fields in question without triggering the refresh call. Should be a simple change.

NickImpact:
Any name preference for these methods? Right now, I'm thinking simply `wrapX`, but figure I'd ask before going at that blind.

NickImpact:
Went ahead with storeDisplayName and storeDisplayMode.

PR #3515, opened 2021-08-29
Don't continuously check command nodes for permissions when we have already done so once

dualspiral:
This should hopefully fix the problems with lockups on flags - due to their cyclic nature nodes that returned `false` for `canUse` failed were not being cached, so `canUse` potentially was being called many, _many_ times, possibly causing a server to stall out due to the number of pathways to flags, especially when many are involved.

We cache nodes that have had any canUse check now - so we don't have to permission check a given node more than once and should hopefully improve performance.

Not tested - can't do so at the moment (and I need test code).

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/core/commands/CommandsMixin.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/core/commands/CommandsMixin.java
@@ -238,7 +238,9 @@ public abstract class CommandsMixin implements CommandsBridge {
             }
             // If empty, we have a node won't resolve (even if not complex), so we ignore it.
             return false;
-        } else if (!SpongeNodePermissionCache.canUse(
+        // If we have already processed this node and it appears in the suggestion node list, prevent a potentially costly
+        // canUse check as we know we can already use it.
+        } else if (!commandNodeToSuggestionNode.containsKey(commandNode) && !SpongeNodePermissionCache.canUse(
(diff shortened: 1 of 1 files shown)


src/mixins/java/org/spongepowered/common/mixin/core/commands/CommandsMixin.java
```diff
@@ -208,57 +208,57 @@
...
-                            if (argNode.getCustomSuggestions() != null) {
-                                // Rebuild the node without the custom suggestions.
-                                rootSuggestion.addChild(this.impl$cloneArgumentCommandNodeWithoutSuggestions(argNode));
-                                continue;
-                            }
-                        } else if (node instanceof ArgumentCommandNode && ((ArgumentCommandNode<?, ?>) node).getCustomSuggestions() != null) {
-                            hasCustomSuggestionsAlready = true; // no more custom suggestions
+        final ServerPlayer e = (ServerPlayer) sourceButTyped.getEntity();
+        final Map<CommandNode<CommandSourceStack>, List<CommandNode<SharedSuggestionProvider>>> playerNodes = this.impl$playerNodeCache.get(e);
+        // Note that the value for a key can be null, this we have to check to see if it contains a key (we've processed it)
```
Zidane: Is there a way to encapsulate a "We have processed this, here is the result" without triggering the double lookup?

You trade a slightly higher memory footprint for CPU which almost always wins.

dualspiral: I'll use an empty list.

PR #3520, opened 2021-09-04
Implement .entities and add test plugin

tyhdefu:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2391) | Sponge

I have added a test plugin that does basic automatic checks, and have also checked that the entities displayed in console during the test line up with the entities in the game.

### DIFF
--- a/src/main/java/org/spongepowered/common/world/volume/buffer/entity/ObjectArrayMutableEntityBuffer.java
+++ b/src/main/java/org/spongepowered/common/world/volume/buffer/entity/ObjectArrayMutableEntityBuffer.java
@@ -188,7 +188,7 @@ public class ObjectArrayMutableEntityBuffer extends AbstractBlockBuffer implemen
 
     @Override
     public Collection<? extends Entity> entities() {
-        return Collections.unmodifiableCollection(this.entities);
+        return ImmutableList.copyOf(this.entities);
     }
 
     @Override
--- a/src/main/java/org/spongepowered/common/world/volume/buffer/entity/ObjectArrayMutableEntityBuffer.java
+++ b/src/main/java/org/spongepowered/common/world/volume/buffer/entity/ObjectArrayMutableEntityBuffer.java
@@ -24,6 +24,7 @@
  */
 package org.spongepowered.common.world.volume.buffer.entity;
 
+import com.google.common.collect.ImmutableList;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.spongepowered.api.block.BlockState;
 import org.spongepowered.api.block.BlockType;
@@ -182,6 +183,11 @@ public class ObjectArrayMutableEntityBuffer extends AbstractBlockBuffer implemen
         return this.entities.stream()
             .filter(entity -> uuid.equals(entity.uniqueId()))
             .findFirst();
+    }
+
+    @Override
+    public Collection<? extends Entity> entities() {
+        return ImmutableList.copyOf(this.entities);
     }
 
     @Override
--- a/src/main/java/org/spongepowered/common/world/volume/buffer/entity/ObjectArrayMutableEntityBuffer.java
+++ b/src/main/java/org/spongepowered/common/world/volume/buffer/entity/ObjectArrayMutableEntityBuffer.java
@@ -48,6 +48,7 @@ import org.spongepowered.math.vector.Vector3i;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -187,7 +188,7 @@ public class ObjectArrayMutableEntityBuffer extends AbstractBlockBuffer implemen
 
     @Override
     public Collection<? extends Entity> entities() {
-        return ImmutableList.copyOf(this.entities);
+        return Collections.unmodifiableCollection(this.entities);
     }
 
     @Override
--- a/src/mixins/java/org/spongepowered/common/mixin/api/minecraft/client/multiplayer/ClientLevelMixin_API.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/api/minecraft/client/multiplayer/ClientLevelMixin_API.java
@@ -24,6 +24,7 @@
  */
 package org.spongepowered.common.mixin.api.minecraft.client.multiplayer;
 
+import com.google.common.collect.ImmutableList;
 import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.multiplayer.ClientLevel;
@@ -38,9 +39,11 @@ import org.spongepowered.common.world.storage.SpongeChunkLayout;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
 @Mixin(ClientLevel.class)
 public abstract class ClientLevelMixin_API implements org.spongepowered.api.world.client.ClientWorld {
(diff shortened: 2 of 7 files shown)


src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/level/EntityGetterMixin_API.java
```diff
@@ -52,6 +52,8 @@
     @Shadow List<? extends net.minecraft.world.entity.player.Player> shadow$players();
     //@formatter:on
 
+    @Shadow List<net.minecraft.world.entity.Entity> getEntities(@org.jetbrains.annotations.Nullable net.minecraft.world.entity.Entity param0, net.minecraft.world.phys.AABB param1);
```
dualspiral: * This needs to be in the `//formatter` block above
* Use `shadow$getEntities()`

testplugins/src/main/java/org/spongepowered/test/entity/EntityVolumeTest.java
```diff
@@ -0,0 +1,95 @@
...
+import org.spongepowered.api.world.server.ServerLocation;
+import org.spongepowered.plugin.PluginContainer;
+import org.spongepowered.plugin.jvm.Plugin;
+
+import java.util.Collection;
+
+@Plugin("entityvolumetest")
+public class EntityVolumeTest {
+
+    @Inject private PluginContainer plugin;
```
dualspiral: This should be constructor injections, in common with the rest of the test plugins.

testplugins/src/main/resources/META-INF/plugins.json
```diff
@@ -654,6 +654,27 @@
                 "id": "spongeapi",
                 "version": "8.0.0"
             }]
+        },
+        {
```
dualspiral: Note that once #3057 merges, this will need to change. I would imagine that #3057 will be merged very soon so this may need to be updated if that ends up being the case.

(see https://github.com/SpongePowered/Sponge/blob/92d325b201621af60ec978814777e1449e9b3e53/testplugins/src/main/resources/META-INF/plugins.json for the specific changes)

tyhdefu:
This is ready for merge and has the latest test plugin json changes now.

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/level/EntityGetterMixin_API.java
```diff
@@ -50,8 +50,10 @@
     @Shadow List<net.minecraft.world.entity.Entity> shadow$getEntities(@Nullable net.minecraft.world.entity.Entity p_175674_1_, net.minecraft.world.phys.AABB p_175674_2_, @Nullable Predicate<? super net.minecraft.world.entity.Entity> p_175674_3_);
     @Shadow<T extends net.minecraft.world.entity.Entity> List<T> shadow$getEntitiesOfClass(Class<? extends T> p_175647_1_, net.minecraft.world.phys.AABB p_175647_2_, @Nullable Predicate<? super T> p_175647_3_);
     @Shadow List<? extends net.minecraft.world.entity.player.Player> shadow$players();
+    @Shadow List<net.minecraft.world.entity.Entity> shadow$getEntities(@org.jetbrains.annotations.Nullable net.minecraft.world.entity.Entity param0, net.minecraft.world.phys.AABB param1);
     //@formatter:on
 
+
```
gabizou: NIT: extra space

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/level/EntityGetterMixin_API.java
```diff
@@ -50,8 +50,10 @@
     @Shadow List<net.minecraft.world.entity.Entity> shadow$getEntities(@Nullable net.minecraft.world.entity.Entity p_175674_1_, net.minecraft.world.phys.AABB p_175674_2_, @Nullable Predicate<? super net.minecraft.world.entity.Entity> p_175674_3_);
     @Shadow<T extends net.minecraft.world.entity.Entity> List<T> shadow$getEntitiesOfClass(Class<? extends T> p_175647_1_, net.minecraft.world.phys.AABB p_175647_2_, @Nullable Predicate<? super T> p_175647_3_);
     @Shadow List<? extends net.minecraft.world.entity.player.Player> shadow$players();
+    @Shadow List<net.minecraft.world.entity.Entity> shadow$getEntities(@org.jetbrains.annotations.Nullable net.minecraft.world.entity.Entity param0, net.minecraft.world.phys.AABB param1);
```
gabizou: Wrong `@Nullable` annotation

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/level/EntityGetterMixin_API.java
```diff
@@ -83,6 +85,10 @@ default boolean isAreaAvailable(final int x, final int y, final int z) {
         return Collections.unmodifiableCollection((List<? extends Player>) (List<?>) this.shadow$players());
     }
 
+    default Collection<? extends Entity> entities() {
```
gabizou, line 89: Is this an override?

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/level/LevelMixin_API.java
```diff
@@ -483,6 +484,13 @@ public boolean setBiome(final int x, final int y, final int z, final Biome biome
         return entityList;
     }
 
+    @Override
+    public Collection<? extends Entity> entities() {
+        ImmutableList.Builder<Entity> list = ImmutableList.builder();
+        this.loadedChunks().forEach(chunk -> list.addAll(chunk.entities()));
```
gabizou: is `loadedChunks()` an API method? I almost don't want to use it explicitly because I don't know what the semantics are of loadedChunks() by itself.

gabizou (changes requested):

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/level/LevelMixin_API.java
tyhdefu: I've removed this implementation in LevelMixin, instead it has implementations in Server and Client seperately, the server implementation already exists.

src/main/java/org/spongepowered/common/world/volume/buffer/entity/ObjectArrayMutableEntityBuffer.java
```diff
@@ -184,6 +186,11 @@ public boolean spawnEntity(final Entity entity) {
             .findFirst();
     }
 
+    @Override
+    public Collection<? extends Entity> entities() {
+        return Collections.unmodifiableCollection(this.entities);
```
dualspiral: Interesting point here - `Collections.unmodifiableCollection` doesn't return a copy of the wrapped collection, but a view of it where the modification methods are guarded. While this might seem kind of immutable to API consumers, it's kind of not really - because if an entity is removed from the backing buffer, it'll disappear from this returned collection.

Now, this is mitagated by the fact that users should not be storing the collection, but if we want a truly immutable collection insofar as the references themselves won't change, this collection needs to be copied, not simply wrapped.

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/client/multiplayer/ClientLevelMixin_API.java
```diff
@@ -66,6 +69,11 @@ public ChunkLayout chunkLayout() {
         return (Optional) this.entitiesById.values().stream().filter(x -> x.getUUID().equals(uuid)).findFirst();
     }
 
+    @Override
+    public Collection<Entity> entities() {
+        return (Collection< org.spongepowered.api.entity.Entity>) (Object) Collections.unmodifiableCollection(this.entitiesById.values());
```
dualspiral: As above.

dualspiral (commented):
Generally looks okay, though there is an interesting point here.

(Also, sorry it's been a little bit of time)

src/main/java/org/spongepowered/common/world/volume/buffer/entity/ObjectArrayMutableEntityBuffer.java
tyhdefu: Gotcha - I knew this but i didn't consider the implications.

dualspiral:
https://github.com/SpongePowered/Sponge/commit/0852eb1f66e763e0cddaab75a36561e327e3c4e8

PR #3523, opened 2021-09-06
feature/resources

Zidane:
For visibility


vanilla/src/main/java/org/spongepowered/vanilla/server/packs/PluginPackResources.java
```diff
@@ -0,0 +1,158 @@
...
+    private final String name;
+    private final PluginResource resource;
+    private final PackMetadataSection metadata;
+    private FileSystem fileSystem;
+
+    public PluginPackResources(final String name, final PluginResource resource) {
+        super(new File("ignore_me"));
+        this.name = name;
+        this.resource = resource;
+        this.metadata = new PackMetadataSection(new TextComponent("Plugin Resources"), 6);
```
dualspiral, line 64: Is there a constant in the Minecraft codebase that we can use here in place of 6, or is that all inlined? I guess the latter...

src/main/java/org/spongepowered/common/resource/SpongeResource.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public ResourcePath path() {
+        return this.path;
+    }
+
+    @Override
+    public InputStream inputStream() {
+        if (this.stream == null) {
```
dualspiral, line 57: So, this still has the inherent issue of people not closing the stream or this resource. However, I have an idea - which might seem a little complex.

How about we track each resource in a `PhantomReference`? When a `Resource` reference is enqueued, we can match up the reference with the input stream and force close it. We can't do `PhantomReference#get` to get the object being collected, but if we map this phantom reference to the input stream in question, we can ensure we close said stream.

It _might_ be overkill, but it is an option if we want to guard against plugins forgetting to close streams when they're done.

Zidane: That isn't a bad idea and should be explored...but that doesn't fix a "one and done" situation where a plugin asks for the stream, uses it, doesn't close it, never asked for again in the engine lifecycle.

Thoughts on that?

dualspiral: It depends how they get the stream, but if we're encouraging the `Resource` to be the object that we try-with-resources around then if a plugin discards the resource (without closing), when the garbage collector kills the `Resource` the stream should be closed at that point regardless - so if a plugin grabs the stream from the resource this is still coming in to play. I would think that would solve the issue.

Ultimately, the suggestion is to tie the stream lifetime to the resource lifetime. If a plugin keeps hold of the `Resource`... well, really not much we can (or should) do about it, the plugin is being _really_ stupid then.

We should probably annotate `Resource` with `@DoNotStore` with that in mind.

Zidane: It is already annotated with `@DoNotStore`

dualspiral: Oh... right, okay!

PR #3526, opened 2021-09-09
Split out our directory structure vs. Forge

Zidane:
After discussion with leadership, we have elected to split out the directory structure of Sponge to remain compatible (mostly) with 3rd party platforms (namely Forge). Current running idea is to do as follows:

* `../mods` -> `../plugins`
* Split where plugin configs and Sponge configs are stored
* Remove the alternate plugins directory configuration property. The only reason this was useful was SpongeForge co-mingling plugins and mods. In the high unlikelihood that an implementation is made that the root platform also has `plugins`, we can circle back at that point.
* Remove specifying where the config directory for plugins are located. Per the same reason as the prior bullet.
* Add `../plugin-configs`
* Add `../sponge`. This will become where Sponge's internal files will mostly reside (except for things specific to worlds)
  * Add `../sponge/config`
  * Add `../sponge/libraries`


ryantheleach:
I'm curious as to what this change means for what was previously called "hybrid" Sponge plugins. e.g. they contained both mods and plugins , usually in the form of mixins.

src/applaunch/java/org/spongepowered/common/applaunch/CorePaths.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * Core directory structure of the Sponge platform
+ *
+ * <base_directory>
+ *  - plugins
+ *  - plugin-configs
```
Yeregorix, line 36: IMO `plugins/configs/` would be more consistent with `sponge/configs/`.

Zidane: My goal was to limit drill downs and keep the most commonly accessed folders quick from the root.

Zidane:
> I'm curious as to what this change means for what was previously called "hybrid" Sponge plugins. e.g. they contained both mods and plugins , usually in the form of mixins.

Honestly it would be best to deploy that in our plugins directory. Mixins are now read from plugin jars in the plugins directory in 1.16.5+

ImMorpheus:

___

PR #3549, opened 2021-10-23
Add rudimentary detection of spaces in filenames on Java 8

dualspiral:
@Zidane There may be a better place for the code, but I figured we probably want to bail out immediately. I think this suffices, but it is in applaunch so just want to make sure it's in the right place before it goes in.

Not using an exception because all that'll happen is we'll print the message, then there will be a stack trace, then everyone won't read the message. If this is the last thing that is printed, and an error code that is non-zero is sent out, at least some panels might detect the crash, preven the auto-restart, and we have a fighting chance that people will read the output.


dualspiral:
As a visual - the red arrow on the prompt in my shell indicates a non-zero exit code (but I'd hope System.exit sans security manager worked in Java...):

image

Issue #3280, opened 2021-02-01
Crash with CPacket Windowclick (exploit)


Cutroo:
[14:43:01] [Server thread/WARN] [Sponge]: Could not find slot #2070 in Container net.minecraft.inventory.ContainerPlayer
--
  | [14:43:01] [Server thread/WARN] [Sponge]: Could not find slot #2070 in Container net.minecraft.inventory.ContainerPlayer
  | [14:43:01] [Server thread/WARN] [Sponge]: Could not find slot #2070 in Container net.minecraft.inventory.ContainerPlayer
  | [14:43:01] [Server thread/WARN] [Sponge]: Could not find slot #2070 in Container net.minecraft.inventory.ContainerPlayer
  | [14:43:01] [Server thread/WARN] [Sponge]: Could not find slot #2070 in Container net.minecraft.inventory.ContainerPlayer
  | [14:43:01] [Server thread/WARN] [Sponge]: Could not find slot #2070 in Container net.minecraft.inventory.ContainerPlayer
  | [14:43:01] [Server thread/WARN] [Sponge]: Could not find slot #2070 in Container net.minecraft.inventory.ContainerPlayer
  | [14:43:01] [Server thread/WARN] [Sponge]: Could not find slot #2070 in Container net.minecraft.inventory.ContainerPlayer
  | [14:43:01] [Server thread/WARN] [Sponge]: Could not find slot #2070 in Container net.minecraft.inventory.ContainerPlayer
  | [14:43:01] [Server thread/WARN] [Sponge]: Could not find slot #2070 in Container net.minecraft.inventory.ContainerPlayer
  | [14:43:01] [Server thread/WARN] [Sponge]: Could not find slot #2070 in Container net.minecraft.inventory.ContainerPlayer
  | [14:43:01] [Server thread/WARN] [Sponge]: Could not find slot #2070 in Container net.minecraft.inventory.ContainerPlayer
  | [14:43:01] [Server thread/WARN] [Sponge]: Could not find slot #2070 in Container net.minecraft.inventory.ContainerPlayer
  | [14:43:01] [Server thread/WARN] [Sponge]: Could not find slot #2070 in Container net.minecraft.inventory.ContainerPlayer
  | [14:43:01] [Server thread/WARN] [Sponge]: Could not find slot #2070 in Container net.minecraft.inventory.ContainerPlayer
[...]

PR #3578, opened 2022-01-08
Add another exploit patch for bad window clicks

Ve6om:
(This is my first PR so excuse anything not done properly, and let me know if I need to modify anything.)

There is a commit that's suppose to patch this issue, however it breaks inventories. https://github.com/SpongePowered/Sponge/commit/dededb4dd67574d11c065bb94f0ede630afcde09

Issue relating to this (Only one I could find): #3280

Exploit spams [this](https://paste.gg/p/anonymous/e929fe355a104d7eb63c8af575ac6708) error and eventually lags the server to death.

My fix simply checks how many slots there are and if the packet slot ID is greater than that, it cancels it and disconnects the player.

### DIFF
--- a/src/main/java/org/spongepowered/common/config/category/ExploitCategory.java
+++ b/src/main/java/org/spongepowered/common/config/category/ExploitCategory.java
@@ -92,7 +92,7 @@ public class ExploitCategory extends ConfigCategory {
     private boolean checkUnknownSlotDuringWindowClicks = false;
 
     @Setting(value = "check-slot-ids-during-window-clicks",
-            comment = "If enabled, window click with slot ids that are higher then the inventory size will be canceled.")
+            comment = "If enabled, window clicks with slot ids that are higher then the inventory size will be canceled.")
     private boolean checkSlotID = false;
 
     @Setting(value = "check-slot-size-during-window-clicks",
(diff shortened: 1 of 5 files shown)


gabizou:
Could you explain whether the fix is supposed to be a replacement for the other fix you linked? Or can they be enabled at the same time? From the server admin perspective, what is one fix going to be better over the other? What are the caveats this fix doesn't catch compared to the other fix?

Ve6om:
> Could you explain whether the fix is supposed to be a replacement for the other fix you linked? Or can they be enabled at the same time? From the server admin perspective, what is one fix going to be better over the other? What are the caveats this fix doesn't catch compared to the other fix?

I had the other fix disabled as it completely breaks containers. It would probably be worth removing that entirely. Also with that fix it doesn't patch the exploit.

gabizou:
> I had the other fix disabled as it completely breaks containers. It would probably be worth removing that entirely. Also with that fix it doesn't patch the exploit.

Then might as well just replace the old one and properly fix it with this PR. No sense in having an additional exploit patch that doesn't work still toggle able (imagine someone enabling both, what does that look like at runtime?)

Ve6om:
> > I had the other fix disabled as it completely breaks containers. It would probably be worth removing that entirely. Also with that fix it doesn't patch the exploit.
> 
> Then might as well just replace the old one and properly fix it with this PR. No sense in having an additional exploit patch that doesn't work still toggle able (imagine someone enabling both, what does that look like at runtime?)

Yeah that sounds good. I'll do that now.

PR #3579, opened 2022-01-09
Disable lighting updates when using BlockChangeFlags#NONE

Lignium:
To disable lighting computation, native flag must be set to `1`. In the API this value should be inverted and represented as `false`.
This fixes unwanted light computation when bulk update blocks when using this flag, which implies that lighting computation is turned off.


gabizou:
Very cool, thanks for the PR!

PR #3591, opened 2022-01-15
Integrate observer for System Observability

gabizou:
Starts integrating some of the metrics that we can easily add to aid in observability and debugging.

- [ ] Histogram for Timing tick updates
- [ ] Counter for Missed Transactions (PhaseTracker specific)
- [ ] Ideas??

Supersedes #3296 

@pschichtel let me know what you think of this

### DIFF
--- a/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
+++ b/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
@@ -49,4 +49,7 @@ public interface TrackerHooks {
     default void incrementBlocksRestored(final ServerLevel world, final BlockPos pos, final BlockState replaced) {
 
     }
+
+    default void updateChunkGauge(final ServerLevel level) {
+    }
 }
--- a/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
+++ b/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
@@ -65,4 +65,8 @@ public interface TrackerHooks {
     default void updateFakePlayerCount(final IntSupplier count) {
 
     }
+
+    default void updatePooledSnapshotBuilder(final IntSupplier count) {
+
+    }
 }
--- a/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
+++ b/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
@@ -28,6 +28,7 @@ import net.minecraft.core.BlockPos;
 import net.minecraft.server.level.ServerLevel;
 import net.minecraft.world.level.block.state.BlockState;
 
+import java.util.function.IntSupplier;
 import java.util.function.Supplier;
 
 public interface TrackerHooks {
@@ -58,6 +59,10 @@ public interface TrackerHooks {
     }
 
     default void updateWorldTps(final ServerLevel serverLevel) {
+
+    }
+
+    default void updateFakePlayerCount(final IntSupplier count) {
 
     }
 }
--- a/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
+++ b/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
@@ -69,4 +69,8 @@ public interface TrackerHooks {
     default void updatePooledSnapshotBuilder(final IntSupplier count) {
 
     }
+
+    default void setTransactedEventCount(final int size) {
+
+    }
 }
--- a/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
+++ b/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
@@ -24,6 +24,8 @@
  */
 package org.spongepowered.common.hooks;
 
+import java.util.function.Supplier;
+
 public interface TrackerHooks {
 
     default void run(
@@ -31,5 +33,9 @@ public interface TrackerHooks {
         final int chunkX, final int chunkZ
     ) {
         process.run();
+    }
+
+    default void incrementUnabsorbedTransaction(final Supplier<String> toGenericString) {
+
     }
 }
--- a/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
+++ b/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
@@ -1,0 +1,35 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.common.hooks;
+
+public interface TrackerHooks {
+
+    default void run(
+        final Runnable process, final String name,
+        final int chunkX, final int chunkZ
+    ) {
+        process.run();
+    }
+}
--- a/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
+++ b/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
@@ -38,4 +38,7 @@ public interface TrackerHooks {
     default void incrementUnabsorbedTransaction(final Supplier<String> toGenericString) {
 
     }
+
+    default void incrementIllegalThreadAccess(final Thread currentThread, final Thread sidedThread) {
+    }
 }
--- a/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
+++ b/src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
@@ -24,6 +24,10 @@
  */
 package org.spongepowered.common.hooks;
 
+import net.minecraft.core.BlockPos;
+import net.minecraft.server.level.ServerLevel;
+import net.minecraft.world.level.block.state.BlockState;
+
 import java.util.function.Supplier;
 
 public interface TrackerHooks {
@@ -40,5 +44,9 @@ public interface TrackerHooks {
     }
 
     default void incrementIllegalThreadAccess(final Thread currentThread, final Thread sidedThread) {
+    }
+
+    default void incrementBlocksRestored(final ServerLevel world, final BlockPos pos, final BlockState replaced) {
+
     }
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -29,6 +29,7 @@ val guavaVersion: String by project
 val junitVersion: String by project
 val mockitoVersion: String by project
 val timingsVersion: String by project
+val observerVersion: String by project
 val checkerVersion: String by project
 
 val commonManifest = the<JavaPluginConvention>().manifest {
@@ -155,8 +156,11 @@ dependencies {
     implementation("org.xerial:sqlite-jdbc:3.20.0")
     implementation("javax.inject:javax.inject:1")
 
-    // Timings
+    // Observability
     implementation("org.spongepowered:timings:$timingsVersion")
+    implementation("org.spongepowered:observer:$observerVersion") {
+        exclude(group = "org.checkerframework", module = "checker-qual") // We use our own version
+    }
 
     // ASM - required for generating event listeners
     implementation("org.ow2.asm:asm-util:$asmVersion")
@@ -202,6 +206,9 @@ dependencies {
 
     mixinsConfig(sourceSets.named("main").map { it.output })
     mixinsConfig("org.spongepowered:timings:$timingsVersion")
+    mixinsConfig("org.spongepowered:observer:$observerVersion") {
+        exclude(group = "org.checkerframework", module = "checker-qual") // We use our own version
+    }
     add(mixins.get().implementationConfigurationName, "org.spongepowered:spongeapi:$apiVersion")
 
     // Tests
--- a/forge/src/mixins/java/org/spongepowered/forge/mixin/observability/forge/FakePlayerFactoryMixin_Observability.java
+++ b/forge/src/mixins/java/org/spongepowered/forge/mixin/observability/forge/FakePlayerFactoryMixin_Observability.java
@@ -1,0 +1,68 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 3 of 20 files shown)


pschichtel:
Looks good as far as I can tell. At cube island I toyed with some per-chunk metrics (e.g. entities per chunk per world) in order to find hotspots in a world, but I guess these would still better be added by a plugin since these values can easily be observed through the API.

gabizou:
> per-chunk metrics

Can definitely add per-world gauges by labels, I was considering some counter based things with how many entities, how many block entities, etc. are being done. Would be an interesting metric to observe the number of updates vs the tps. The hard part about per-chunk metrics is: What does the gauge look like? Is it one gauge with labels for chunk coordinates? The cardinality of the coordinates is rather high normally, so it's somewhat difficult to consider prometheus would be happy with having that level of cardinality, compared to say the variations of counts per world. I'll admit it is far easier for sponge to add them than a plugin (since the mixins are fairly simple to add).

src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
```diff
@@ -0,0 +1,76 @@
...
+package org.spongepowered.common.hooks;
+
+import net.minecraft.core.BlockPos;
+import net.minecraft.server.level.ServerLevel;
+import net.minecraft.world.level.block.state.BlockState;
+
+import java.util.function.IntSupplier;
+import java.util.function.Supplier;
+
+public interface TrackerHooks {
```
dualspiral, line 34: This should really be a service that any plugin can provide, rather than a hooks that something will provide internally.

dualspiral: Or, rather - _something_ should be a service.

gabizou: The Observable implementation is a service a plugin provides. Sponge just makes use of the observer library to make some additional system metrics observable (without said plugin having to be mixing in to Sponge's internals to get the same metrics).

dualspiral: Okay, better question, how will a plugin integrate with this?

pschichtel: A plugin could just pull the observer library and create its own Meters

pschichtel:
> cardinality of the coordinates

that's the point I was also unsure about. If users use a dedicated prometheus instance for the server (and possibly push aggregated metrics up to their central one) then I think it would be fine. prometheus can handle millions of series without trouble from what I read. Haven't actually tested it myself with that amount of data though.

We could:

* provide the high "resolution metrics" behind an option (or as a plugin) for those that know they can handle it
* always provide them and expect admins to configure drop rules in prometheus if they can't handle them

src/main/java/org/spongepowered/common/hooks/TrackerHooks.java
dualspiral: These are all not really the answers I am looking for.

Given this library, what will a plugin need to do to make use of these metrics? How does it get the data, how does it create its own meters, for that matter, _what_ is a meter here (I'm not familiar with Prometheus). Does this require Prometheus, or can a user use another platform if they provide the hooks for it. What hooks are needed? Given no plugin, how will this metrics collection get out of the way, because lets face it, most servers will not use this.

My ultimate concern with this is that there seems to be code but no real planning on the Sponge side. I have no idea how this is going to work. Where does the plugin come into this? Are they basically receiving the data and passing it on to some front end? Or are they involved in the scraping?

I accept it's come from a plugin that was designed for the author, but I'd like a lot more clarity as to what this system will do, what integration points are needed and such so that we can consider how this fits in with the core platform and the API. This is why I was asking about services. We have a service manager that is somewhat locked down, but I could absolutely upgrade the service manager to add non-API facing services so that plugins provide their implementation. Right now, it seems to me that we're hurtling towards just doing whatever comes to mind without regards to the systems in the API that may be able to assist in this.

That's why I asked the question. I don't care about the implementation itself beyond its impact. I care about the extensible points.

gabizou: So maybe some clarifying details on the usage of the library:

The library by itself does the bare minimum to expose easy access to adding new metric counters etc. Where the implementation comes in is the implementation provides the MetricsCollector to be told of these counters updates etc. And that implementation plugin spins up a web server on a port where a separate Prometheus instance scrapes metrics since the last poll. So effectively, the implementing plugin is gathering the metric updates to be scraped by Prometheus. 

As for how other plug-ins can add/make use of observer is by adding their own metrics they’d intend to be used for observing the system’s activity, and if desired they’ve configured maybe alerting given some metrics specifications (like if tps drops below 15 for more than one minute, or missed transactions/errors occurred more than 1 per minute).

in short:
- the metrics by themselves give data to be scraped
- The implementing plugin exposes a port for Prometheus to poll for all metrics data
- A separate Prometheus instance aggregates the data for consuming/processing by other software (like AlertManager or Grafana)
- Additional plugins can freely add their own metrics using the same library with or without impunity. Their responsibility isn’t to serve the Prometheus endpoint itself, that’d be the Prometheus client serving plugin.

dualspiral: With the best will in the world, this feels like we're jumping on this and thinking up overcomplicated systems and use cases when we just don't need to do so. **As much of this as possible needs to be on the plugin side.** If a plugin wants to poll the TPS every minute and alert a user because the TPS has dropped, all Sponge needs to provide is the TPS which the plugin can poll... and we already do that. Why, then, do we need to provide metrics for that when a plugin can do it? Chunks loaded - same thing. Players online, same thing.

I'm _very_ wary about this and feel like a lot of the work here should be external. I feel like all this is going to do is increase the maintenance burden on us and lock us into a specific ecosystem. I am not interested in that.

Sponge is not a server health and alerting system. I do not mind providing instantaneous data getters to the API that these plugins can consume. I am not willing to go much further than that because plugins can, and should, be managing these things.

Zidane: I see where dual is coming from.

To me it appears the end result is some artificial number going up and down. Current TPS of a world, current loaded chunks of a world, last tick time of an entity or block entity, etc. Things like that. Those can be getters in the API and this system can sit outside of it and simply poll that data each frame (such as a scheduled task). That strategy keeps this system external, allows for other monitoring frameworks that exist out there to work (they all function on the same principle: polling numbers in intervals), and minimizes our maintenance burden (we need only track some numbers, not introduce an entire library). Hell we can axe the observability mixin set and bake it into core as it is useful telemetry.

pschichtel: nothing of the following is in any shape or form related to the PR directly, but I think still relevant to the discussion/understanding.

> These are all not really the answers I am looking for.

Sorry, I assumed some familiarity with the topic. sponge or plugin would be a producers of measurements by dumping values into "Meters" (a thing that measures another thing), plugins can be consumers of measurements by subscribing to "MetricsCollections". "Meter" and "MetricsCollection" are concepts of the observer library, a thin (mostly a few interfaces) abstraction between measurement producers and consumers (e.g. a prometheus or an InfluxDB)

> My ultimate concern with this is that there seems to be code but no real planning on the Sponge side.

I had a few discussions a while back which sparked significant interest in this. I did demonstrate the value of something like this with CubeEngine's observe module, however the observer library is not directly related to that, the prometheus plugin I did however is based on that code.

> I accept it's come from a plugin that was designed for the author, but I'd like a lot more clarity as to what this system will do

On my side this thing was initially driven by the frustration caused by the lack of insights into why our (at the time) 5-10 player API8 server would drop to 5 tps and lower on a regular basis. While I could get high level measurements like online players and chunks without issues, but these were not useful. More fine grained measurements (entity per type per world) would give some insights, but we occasionally had the suspicion that taking the measurements would itself cause measurable performance issues.

> what integration points are needed and such so that we can consider how this fits in with the core platform and the API.

I think @gabizou decision to move this into a library (from my former API8 PR) to avoid having it in the API to begin with. The observer library would be an implementation detail of sponge. If a plugin were to use observer without Sponge using it, everything would work fine. If Sponge used observer without any plugin using, still everything would be fine. If both Sponge and plugins using observer come together they interact by publishing measurements to a common place (or separate places if intended). A completely separate lifecycle from the SpongeAPI. From the PoV of Sponge or a simple API plugin, there is no such thing as prometheus or any other monitoring thingy. There is no impact on plugins that don't interact with observer (ignoring the minimal overhead in Sponge if it is enabled).

This whole thing is basically about Sponge (not API) exposing internal measurements to an outside observer in an efficient way **without** exposing them through APIs. I know, you could and to some degree should consider the defined Meters some form of API and breaking them can have impact on systems beyond the Sponge server. Since these measurements are internals from the implementation, the lifecycle of the Meters is also linked to the implementation (from the PoV of an operator I would expect metrics to change, if I update a webserver, even if the HTTP protocol hasn't changed). These Meters also don't need to talk about high level API concepts, the metrics are there to expose internal performance metrics of that specific application. SpongeForge would likely have wildly different metrics compared to SpongeVanilla (apart from a common subset).

The impact of something like this on Sponge implementations is limited to injecting measurement points which take the form of calls to predefined Meter instances. Depending on the complexity of the metrics the same Meter might be measured at several places (e.g. if several different places cause chunk loading and other places unloading), but this can usually be reduced to single lines and extracted out using runtime/compile-time injection tools like Mixin or AOP frameworks, the former of which is in use here anyway.

> Sponge is not a server health and alerting system.

You are absolutely right and it should not be, but to some degree it has to enable other health and alerting system to make statements about its health. I agree that a plugin could expose high-level metrics based on things it can observe using the API. And that is exactly what I'm doing in the plugin I have cooking. But I personally wouldn't extend the API *just* for a plugin to create some metric. This is blackbox monitoring and I'd keep it that way. The thing being monitored should not be aware that it is being monitored.

> That strategy keeps this system external, allows for other monitoring frameworks that exist out there to works

I think this approach we took is basically this. Don't confuse observer with something like for example micrometer (common in the Spring world). observer is a tiny interface as a common protocol for components to produce measurements and consume them. No monitoring system is pulled, no webservers are started.

> they all function on the same principle: polling numbers in intervals

Not quite, but I'll spare that discussion unless interest exists.

> Hell we can axe the observability mixin set and bake it into core as it is useful telemetry.

Do keep in mind that while most metrics or cheap, nothing is for free and there might at some point be a lot of measurements. With a game simulation loop already as strained as Minecraft's I would personally prefer to have a kill switch, even just for the sanity check that the lag *really* doesn't come from the monitoring itself.

dualspiral: I still have problems with this.

> Sorry, I assumed some familiarity with the topic. 

I am familiar with how things like this work, I was not familiar with how this library is meant to work, and the questions I was asking were not being answered.

> sponge or plugin would be a producers of measurements by dumping values into "Meters" (a thing that measures another thing), plugins can be consumers of measurements by subscribing to "MetricsCollections". "Meter" and "MetricsCollection" are concepts of the observer library, a thin (mostly a few interfaces) abstraction between measurement producers and consumers (e.g. a prometheus or an InfluxDB)

Again, why is it up to Sponge to do this? Why can't a plugin poll the situation and fill their own "meters" or whatever, away from Sponge? If we have nothing reading this data, then methods that are purely getters are dormant, and then there is very little to no impact on performance on a server not wanting these metrics.

The argument that we can use mixins to stick meters into the code is fine and all, but there is an inherent maintenance cost to these mixins. Simple getters aren't too bad, injecting code into the game loop to collect metrics is more invasive. 

I'm also really of the belief that Sponge shouldn't be collecting metrics at all, only providing data to allow plugins to poll it (or listen to events/callbacks).

> > I accept it's come from a plugin that was designed for the author, but I'd like a lot more clarity as to what this system will do
> 
> On my side this thing was initially driven by the frustration caused by the lack of insights into why our (at the time) 5-10 player API8 server would drop to 5 tps and lower on a regular basis. While I could get high level measurements like online players and chunks without issues, but these were not useful. More fine grained measurements (entity per type per world) would give some insights, but we occasionally had the suspicion that taking the measurements would itself cause measurable performance issues.

No, that's why you want it, not what it would do. I take no issue with why you want this, my issue is where the line of responsibility is, and I think more responsibility should be on the plugins, not Sponge and the core team. My question was probably badly phrased, but I was more interested in the technical what it does - and my understanding is that Sponge collects metrics and plugins poll for what Sponge collects, which might contain time series data or whatever.

For example, you say: 

> The impact of something like this on Sponge implementations is limited to injecting measurement points which take the form of calls to predefined Meter instances. 

I do not think we should be doing that. My argument is that Sponge should simply provide instantaneous raw values, and it's up to plugins to collect the metrics by polling the raw data and then plugins should manage the time-series of whatever itself. That is, the concept of `Metrics` should not exist here, but in the plugins, in the way they want to handle them. I don't mind the concept of this library being a shim over getters (and possibly timing key functions, which hasn't actually been mentioned but I think is probably wanted here - but that's a whole different kettle of fish as to deciding what should and shouldn't be measured and needs thought) that we don't want to expose in the API and so is able to break more easily, but really, that's all I think this should be.

> I think @gabizou decision to move this into a library (from my former API8 PR) to avoid having it in the API to begin with. The observer library would be an implementation detail of sponge. 

To just touch on this, I wasn't actually suggesting that it should be in the API, but there are constructs that _could_ be leveraged. I already have events that fire for non-API things - you can actually register Brig commands using the `RegisterCommandEvent` for example. If this was a candidate for services, it wouldn't need to be exposed in the API but could be registered using the `RegisterServiceEvent`, for example. That was what I was kind of angling at - we can still use the platform to effect.

---

The ultimate problem is that while I was aware of some ideas around observability, I haven't really seen any proposals as to what should even be "observable" or whatever _and why_. Indeed, if you look at the description of this PR it says nothing more than this:

> Histogram for Timing tick updates
> Counter for Missed Transactions (PhaseTracker specific)
> Ideas??

This is why I'm saying this seems to be hurtling forward in code there seems to otherwise be no real insight as to what should and should not have been provided. No real defined goal. I see no real compelling arguments for the responsibility that we would have to take on and I think this needs to pause for a moment so a proper problem statement, decisions on what we will and won't do etc. will be done. After all, we've been shaving off other libraries such as DB drivers and such to _reduce_ our responsibilties - something I accept is not directly comparable here but _we do need to think about it_.

As there is talk about "internal plugins" and libraries and such, these are all things we'd have to sort out infrastructure for, and things like that _also_ need discussions.

With all that said, I also need to say that this library is detracting from working on other things in Sponge that far more people are waiting on, as I said previously. this isn't something that most server owners are going to care about. I'm not really interested in telling people what they should and shouldn't do - but I'm concerned that as Sponge, that we're getting tied up in a system that isn't really a priority.

pschichtel: > My argument is that Sponge should simply provide instantaneous raw values, and it's up to plugins to collect the metrics by polling the raw data and then plugins should manage the time-series of whatever itself.

The proposed approach does exactly this. Sponge will not generate series. Sponge will just record/measure/whatever values. If nobody is there to actually take them, nothing happens.

dualspiral: >  Sponge will just record/measure/whatever values.

No, Sponge should not be recording anything, that's my point, it's up to plugins to do that.

pschichtel: > If this was a candidate for services, it wouldn't need to be exposed in the API but could be registered using the RegisterServiceEvent, for example. That was what I was kind of angling at - we can still use the platform to effect.

My original API8 PR did that.

pschichtel: > this isn't something that most server owners are going to care about.

if more server owners run into the same situation we did at cube island, then possibly this opinion changes or people might not even care enough. The lack of useful telemetry to diagnose performance issues has been an issue for many servers I know since I have been running MC servers.

> The ultimate problem is that while I was aware of some ideas around observability, I haven't really seen any proposals as to what should even be "observable" or whatever and why.

The why tends to be speculation at this point, since we don't actually know whats the cause for performance issues. The goal is to measure code paths and count events that could indicate issues. The point is that most metrics are cheap enough that it doesn't hurt to have them. I know one of the interesting parts was phase tracking. Something, from what I remember, tended to get blamed for causing performance issues. The better the knowledge of "problematic" areas is, the easier it will be to have targeted measurements. The is no problem in removing measurements when it becomes apparent that they are useless or redundant.

pschichtel: > No, Sponge should not be recording anything, that's my point, it's up to plugins to do that.

So you mean that the SpongeAPI should expose APIs for the entirety of the phase tracker to get call counter, call durations, error counters, ...

And who exactly keeps track of these values, so that plugins can fetch them through the SpongeAPI?

Or would there be APIs to inject wrappers into the phase tracker to measure execution time and and count errors? sounds like a mess.

dualspiral: > So you mean that the SpongeAPI should expose APIs for the entirety of the phase tracker to get call counter, call durations, error counters, ...

And where did I say that? I've said that it _could_ be a library that can afford breaking changes, I simply said Sponge should not record things.

> And who exactly keeps track of these values, so that plugins can fetch them through the SpongeAPI?

As I said, plugins get the instantaeous values, and they record them. It's up to the plugins for timings on when they get the data. On things like method timings, that would have to be done like akiar's timings impl does it - callbacks to the metrics provider after the call. Sponge should retain no knowledge beyond that.

> The why tends to be speculation at this point

If there is no why, then there shouldn't be metrics either. Everyone has an opinion, there are well thought reasons (which might be grounded in speculation) as to why you might target a specific method, etc. So while it tends to be speculation, there is reasoning involved.

> sounds like a mess.

Yes, which is why I want things to pause and to properly consider what the real goals are - because as I said, this all seems to be "let's add another library for observing metrics" without properly considering what that really needs to do - from my point of view, as it stands I do not like the current position of who has what responsibilities.

ImMorpheus:

___

PR #3612, opened 2022-01-29
Fire MoveEntityEvent for ridden entites

### LINKED ISSUE #3607: MoveEntityEvent does not fire for moving in vehicle
### Affected Product(s)

SpongeVanilla

### Version

spongevanilla-1.16.5-8.0.0-RC1055-universal

### Operating System

Windows

### Java Version

17.0.1

### Plugins/Mods

```shell
Nope
```

### Describe the bug

It seems that the MoveEntityEvent does not fire when a user moves in a vehicle (tested with boats and horses).

These is a [gist of latest.log](https://gist.github.com/pietelite/b7a117524c793aca46174b6d507e2271) with some comments.
I moved around next to the boat, got in the boat, moved around again, and got out, and moved around again. In the period of moving around, there were no MoveEntityEvent events fired.

### Link to logs

https://gist.github.com/pietelite/b7a117524c793aca46174b6d507e2271

dualspiral:
Also does some rotation stuff.

Untested as of yet, but any eyes over it would be nice.

Should fix #3607

### DIFF
--- a/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
+++ b/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
@@ -32,6 +32,7 @@ import net.minecraft.core.BlockPos;
 import net.minecraft.network.protocol.game.ServerboundPlayerActionPacket;
 import net.minecraft.server.level.ServerLevel;
 import net.minecraft.sounds.SoundEvent;
+import net.minecraft.util.Tuple;
 import net.minecraft.world.InteractionHand;
 import net.minecraft.world.damagesource.DamageSource;
 import net.minecraft.world.damagesource.EntityDamageSource;
@@ -136,6 +137,9 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public final class SpongeCommonEventFactory {
+
+    private static final double MOVEMENT_DELTA = 1.0d/256.0d;
+    private static final double ROTATION_DELTA = 0.15f * 0.15f;
 
     @SuppressWarnings("unchecked")
     public static <T extends net.minecraft.world.entity.Entity> CollideEntityEvent callCollideEntityEvent(
@@ -383,26 +387,23 @@ public final class SpongeCommonEventFactory {
             return;
         }
 
-        final double deltaX = entity.xOld - entity.getX();
-        final double deltaY = entity.yOld - entity.getY();
-        final double deltaZ = entity.zOld - entity.getZ();
-        final double deltaChange = Math.pow(deltaX, 2) + Math.pow(deltaY, 2) + Math.pow(deltaZ, 2);
-        if (deltaChange < 1f / 256) {
-            return;
-        }
-
         try (final CauseStackManager.StackFrame frame = PhaseTracker.getCauseStackManager().pushCauseFrame()) {
             frame.pushCause(entity);
             frame.addContext(EventContextKeys.MOVEMENT_TYPE, MovementTypes.NATURAL);
 
-            final MoveEntityEvent event = SpongeEventFactory.createMoveEntityEvent(frame.currentCause(), (Entity) entity,
-                    new Vector3d(entity.xOld, entity.yOld, entity.zOld), new Vector3d(entity.getX(), entity.getY(), entity.getZ()),
-                    new Vector3d(entity.getX(), entity.getY(), entity.getZ()));
+            final @Nullable Vector3d finalPosition =
+                    SpongeCommonEventFactory.createAndFireMoveAndRotateEvents(
+                            (Entity) entity,
+                            new Vector3d(entity.xOld, entity.yOld, entity.zOld),
+                            null,
+                            new Vector3d(entity.getX(), entity.getY(), entity.getZ()),
+                            null
+                    ).getA();
 
-            if (SpongeCommon.post(event)) {
+            if (finalPosition == null) {
                 entity.setPos(entity.xOld, entity.yOld, entity.zOld);
             } else {
-                entity.setPos(event.destinationPosition().x(), event.destinationPosition().y(), event.destinationPosition().z());
+                entity.setPos(finalPosition.x(), finalPosition.y(), finalPosition.z());
             }
         }
     }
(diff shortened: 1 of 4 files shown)


src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerGamePacketListenerImplMixin.java
```diff
@@ -362,10 +356,112 @@
...
+        // Use the position of the last movement with an event or the current player position if never called
+        // We need this because we ignore very small position changes as to not spam as many move events.
+        final Vector3d fromPosition = VecHelper.toVector3d(rootVehicle.position());
+
+        final Vector3d originalToPosition = new Vector3d(param0.getX(), param0.getY(), param0.getZ());
+        final Vector3d originalToRotation = new Vector3d(param0.getYRot(), param0.getXRot(), 0);
+
+        // common checks and throws are done here.
+        final Tuple<Vector3d, Vector3d> events = this.impl$createAndFireMoveAndRotationEvents(
+                player,
```
aromaa: This should be rootVehicle, the player isn't moving, its only the passenger.

src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerGamePacketListenerImplMixin.java
```diff
@@ -362,10 +356,112 @@
...
+            final org.spongepowered.api.entity.Entity movingEntity,
+            final @Nullable Vector3d fromPosition,
+            final @Nullable Vector3d fromRotation,
+            final @Nullable Vector3d toPosition,
+            final @Nullable Vector3d toRotation) {
+
+        final @Nullable Vector3d finalPosition;
+        final Cause cause = PhaseTracker.getCauseStackManager().currentCause();
+        // Call move & rotate event as needed...
+        if (ShouldFire.MOVE_ENTITY_EVENT && fromPosition != null && toPosition != null && fromPosition.distanceSquared(toPosition) > (1.0d/256.0d)) {
```
aromaa: I think we should apply the `api8/move-event-grid` trick for all move events as this can be eventually exploited as we only take account the last position. Sure, it would take long time to travel between big distances but players are pretty determined ;)

I can think of one easy way to fool simple region checks by moving slightly inside the region without triggering the move events, logging out and logging back in. This way the player was logged inside the region and could be seen as valid.

src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerGamePacketListenerImplMixin.java
aromaa: Also pushing the player as cause here could be helpful so plugins can determine if player is controlling vehicle and link the player with it.

dualspiral: You're right, this should be rootVehicle, good catch. Also, I think the player is on the stack already because they sent the packet.

src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerGamePacketListenerImplMixin.java
dualspiral: Oh, I know, but it's one step at a time.

src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerGamePacketListenerImplMixin.java
```diff
@@ -362,7 +362,76 @@
...
+                originalToRotation
+        );
+        if (toRotation == null) {
+            toRotation = fromRotation;
+        }
+
+        if (fromPosition.equals(toPosition)) {
+            // no point doing all that processing, just account for a potential rotation change.
+            if (!fromRotation.equals(toRotation)) {
+                rootVehicle.absMoveTo(rootVehicle.getX(), rootVehicle.getY(), rootVehicle.getZ(), (float) toRotation.y(), (float) toRotation.x());
```
whimxiqal, line 413: Why are rotations (in the API and here) not using floats on Sponge's side? I've noticed a lot of casting here and there when it seems we can just use Vector3f everywhere.

whimxiqal (commented):
I haven't tested yet but just some things I noticed

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
```diff
@@ -137,6 +137,9 @@
 
 public final class SpongeCommonEventFactory {
 
+    private static final double MOVEMENT_GRID_POINTS_PER_BLOCK = 8.0d;
```
whimxiqal, line 140: I meant to put this in my review, but my question was, why do 8? The delta change before the idea behind `api8/move-event-grid` had a distance of 1/16, so why change it to 1/8?

src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerGamePacketListenerImplMixin.java
```diff
@@ -362,7 +362,76 @@
...
+        packet.accessor$xRot((float) toRotation.y());
+
+        if (!toPosition.equals(originalToPosition)) {
+            // update the packet, let MC take care of the rest.
+            packet.accessor$x(toPosition.x());
+            packet.accessor$y(toPosition.y());
+            packet.accessor$z(toPosition.z());
+
+            // set the first and last good position now so we don't cause the "moved wrongly" warnings later down the line.
+            this.vehicleFirstGoodX = toPosition.x();
```
whimxiqal, line 430: Riding anything causes "move wrongly" errors. It doesn't seem like changing these would help with the problem in the first place, though, because the "move wrongly" message is thrown based on 

``` java
        var10 = packet.getX() - vehicleEntity.getX();
        var11 = packet.getY() - vehicleEntity.getY();
        if (var11 > -0.5D || var11 < 0.5D) {
          var11 = 0.0D;
        }

        var12 = packet.getZ() - vehicleEntity.getZ();
        var14 = var10 * var10 + var11 * var11 + var12 * var12;
        boolean var17 = false;
        if (var14 > 0.0625D) {
          var17 = true;
          LOGGER.warn("{} (vehicle of {}) moved wrongly! {}", var0.getName().getString(), this.player.getName().getString(), Math.sqrt(var14));
        }
```

Still new to packets though, so I'm not sure if this is the right way to think about it

src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerGamePacketListenerImplMixin.java
dualspiral: Because we specify doubles in most places in the API, simple as.

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
dualspiral: Why did I do 16 in move-event-grid? It's somewhat arbitary, trying to strike a balance between firing significant events, but not firing too many.

I may make this configurable for performance reasons.

src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerGamePacketListenerImplMixin.java
dualspiral: ...no?

`handleMoveVehicle` in `ServerGamePacketListenerImpl`, line 379 clearly shows `this.vehicleFirstGoodX` is used for `var10`, etc. By setting them to the co-ordinates we're proposing to teleport the entity to, these zero out, var14 will be zero, nullfying the if branch.

whimxiqal: Well line `379` sets `var10` and is checked with `var14` in `383` and `384`, but those are only for "moved too quickly" warning. Lines `396`, `397`, and `402` _reset_ the values to be used with `var14` in `403` and `405-408`, which is where the "moved wrong" warning shows up. I mean, I don't think there's any harm to adding the `vehicleFirstGood` variables, but they are set in lines `421-423` anyway at the end of the original method, which is run whether or not the vehicle "moved wrongly".

src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java
whimxiqal: Fair enough!

src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerGamePacketListenerImplMixin.java
```diff
@@ -288,42 +295,37 @@
...
 
         // At this point, we cancel out and let the "confirmed teleport" code run through to update the
         // player position and update the player's relation in the chunk manager.
-        if (cancelMovement) {
-            if (fromPosition.distanceSquared(toPosition) > 0) {
+        if (toPosition == null) {
+            if (fromPosition.distanceSquared(originalToPosition) > 0) {
                 // Set the location, as if the player was teleporting
                 this.awaitingTeleportTime = this.tickCount;
                 this.shadow$teleport(fromPosition.x(), fromPosition.y(), fromPosition.z(), (float) toRotation.x(), (float) toRotation.y());
```
whimxiqal, line 332: If `toRotation` and `toPosition` are both null, then this `toRotation` call here throws an NPE. Should there be something around line `321` that says "if `toRotation` is null, make it `fromLocation`"?

src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerGamePacketListenerImplMixin.java
dualspiral: You're right! I had somehow completely glossed over the reassignment of the variables.

I'll have to think of some other way to handle that.

src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerGamePacketListenerImplMixin.java
dualspiral: Yup, good spot.

Yeregorix:
Zidane has asked me to include the changes from my PR #3751 into this PR. I did this in a new branch and I'm not sure if I should open a new PR or not so for now I will just link my commit here: https://github.com/Yeregorix/Sponge/commit/34d5fee1121597e18b172f00548c7d1d2f562a1e. 

I added comments in my commit to explain why certain changes were needed.

Zidane:
@Yeregorix 

Thank you for your work. Feel free to close your last PR and open a new one with these combined changes. Closing this PR out and discussion can continue on your new one.

PR #3656, opened 2022-03-23
common: Should we sanitize enum names?

zml2008:
This resolves issues where mods add enum entries with names that are not valid resource locations

Are cases where there are enum entries differing only by case a concern we have?

Should other registries be converted?


Zidane:
I've taken a hard stance in API 8+ on mods adding stuff like this without valid names. However, this has been made a registry by us. Per Vanilla spec...there is no spec. We're adding it. So therefore, someone could snap back on us and state we're overstepping if we ask them on an issue to fix their name.

Therefore, if it is an Enum value which isn't in a registry by Vanilla, we will sanitize the name.

Issue #3657, opened 2022-03-23
Conflict between SpongeForge and Infernal Expansion


Rykita:
### Affected Product(s)

SpongeForge

### Version

1.16.5-36.2.5-8.0.0-RC1100

### Operating System

Linux

### Java Version

openjdk version "1.8.0_312"

### Plugins/Mods

```shell
Infernal Expansion
```

### Describe the bug

```
[22:23:53] [Server thread/FATAL] [mixin/]: Mixin apply failed infernal-expansion.mixins.json:common.MixinPlayerInteractionManager ->
 net.minecraft.server.management.PlayerInteractionManager: org.spongepowered.asm.mixin.injection.throwables.InvalidInjectionException
  @At("INVOKE") on net/minecraft/server/management/PlayerInteractionManager::IE_firePostRightClickBlockEventServer with priority 1000 cannot
  inject into
  net/minecraft/server/management/PlayerInteractionManager::func_219441_a(Lnet/minecraft/entity/player/ServerPlayerEntity;Lnet/minecraft/world
  /World;Lnet/minecraft/item/ItemStack;Lnet/minecraft/util/Hand;Lnet/minecraft/util/math/BlockRayTraceResult;)Lnet/minecraft/util/ActionResultType; 
  merged by org.spongepowered.common.mixin.tracker.server.level.ServerPlayerGameModeMixin_Tracker with priority 1102 [PREINJECT Applicator Phase -> 
  infernal-expansion.mixins.json:common.MixinPlayerInteractionManager -> Prepare Injections ->  -> 
  handler$zoo000$IE_firePostRightClickBlockEventServer(Lnet/minecraft/entity/player/ServerPlayerEntity;Lnet/minecraft/world
  /World;Lnet/minecraft/item/ItemStack;Lnet/minecraft/util/Hand;Lnet/minecraft/util/math/BlockRayTraceResult;
  Lorg/spongepowered/asm/mixin/injection/callback/CallbackInfoReturnable;)V -> Prepare]
```

zml2008:
For some extra context, this issue is encountered after running a build of https://github.com/SpongePowered/Sponge/pull/3656, which works around an issue with Infernal Expansion's additions to the MobCategory enum.

[...]

PR #3658, opened 2022-03-24
Lower the priority of tracker related mixins with overwrites.

gabizou:
Attempt at fixing #3657

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/tracker/server/level/ServerPlayerGameModeMixin_Tracker.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/tracker/server/level/ServerPlayerGameModeMixin_Tracker.java
@@ -24,54 +24,25 @@
  */
 package org.spongepowered.common.mixin.tracker.server.level;
 
-import net.minecraft.advancements.CriteriaTriggers;
-import net.minecraft.core.BlockPos;
 import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.server.level.ServerPlayerGameMode;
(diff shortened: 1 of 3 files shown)


Zidane:
@gabizou 

We could just lower the priority of all the tracker mixins (I'd prefer to not have these separations). Most of the overwrite code is with the tracker afterall.

dualspiral:
Rebased onto latest api-8 code for testing purposes. If we can lower all mixin priority then we should, though this is priority 1102 and the core is 1101 - does that mean we should lower the core priority too?

gabizou:
I think we'll have to evaluate a lot of our mixins lowering priority than default. Some things are "Wanted" at higher priority (like redirects/injections) while other things that are overwrites should be lower priority than the evaluator for redirects etc. At least, that's my view on it.

dualspiral:
> Some things are "Wanted" at higher priority (like redirects/injections) while other things that are overwrites should be lower priority than the evaluator for redirects etc. At least, that's my view on it.

This is pretty much the conclusion I've come to, namely:

* Overwrites need to be at lower priorities, as mod mixins might inject into them at default.
* Injections etc. should be at higher priorities, such that we can inject into (well crafted) mod overwrites.

I'd imagine that this will solve some as of yet unknown incompatibilities with mods that use mixins - most mods probably won't use the idea of mixin priorities.

Enterprise12nx01:
ive tried the latest spongeforge spongeforge-1.16.5-36.2.5-8.1.0-RC1152
has the issue been fixed yet or in an older version?

ImMorpheus:

___

PR #3664, opened 2022-04-16
Implement BlockEntityVolume methods for chunk

tyhdefu:
Some block entity methods were not implemented for chunks.
I haven't been able to test this yet as it doesn't currently seem to build (CI seems to be failing at the moment too)
```
error: incompatible types: Key<Value<? extends EntityType<?>>> cannot be converted to Key<? extends Value<EntityType>>
                .create((Key<? extends Value<org.spongepowered.api.entity.EntityType>>) Keys.ENTITY_TYPE)
```

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/api/minecraft/server/level/ServerLevelMixin_API.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/api/minecraft/server/level/ServerLevelMixin_API.java
@@ -263,7 +263,9 @@ public abstract class ServerLevelMixin_API extends LevelMixin_API<org.spongepowe
 
     @Override
     public void removeBlockEntity(final int x, final int y, final int z) {
-        this.shadow$removeBlockEntity(new BlockPos(x, y, z));
+        // *Just* removing the block entity is going to give is a glitchy state that should never happen.
+        // So just remove the whole block.
+        this.blockEntity(x, y, z).ifPresent(ignored -> this.removeBlock(x, y, z));
     }
 
     // UpdateableVolume
--- a/src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/level/LevelMixin_API.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/level/LevelMixin_API.java
@@ -27,6 +27,7 @@ package org.spongepowered.common.mixin.api.minecraft.world.level;
 import net.kyori.adventure.sound.Sound;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Registry;
+import net.minecraft.nbt.CompoundTag;
 import net.minecraft.network.protocol.game.ClientboundBlockUpdatePacket;
 import net.minecraft.network.protocol.game.ClientboundCustomSoundPacket;
 import net.minecraft.resources.ResourceKey;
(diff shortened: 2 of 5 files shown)


tyhdefu:
So, this was more complicated that initially thought. There were bugs in the other implementations in the world of these methods that I had to correct.

The whole idea of "adding" a block entity is a bit misleading. Minecraft creates them when you place the appropriate block. Adding a blockstate is really just implementation detail. You can't just tell minecraft to add a block entity to any block and expect it to work - it will only work if the block is appropriate. In my opinion the method should be called setBlockEntity, since its not like you can keep adding them in the same position. However I know the API is frozen so not much we can do about this now.
Furthermore, passing in a block entity makes it only really useable for pasting an existing block entity onto it. BlockEntities store where they are, so i've clone it and loaded it to prevent the bugs that appear otherwise.

"Removing" a block entity, similarly, is confusing - thats an illegal state for minecraft to be in, and doesn't make sense for us to offer this. I've opted to make it remove the entire block, but I think you could argue it would be fair to blank the tile entity data. **Please let me know if you want me to change it to wipe the block entity data instead of full removing the block**

Anyway i've added a test plugin and commented within the code to explain this a bit better.

The methods will also need documentation to make sense and understand what they do.

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/server/level/ServerLevelMixin_API.java
```diff
@@ -263,7 +263,9 @@ public Optional<org.spongepowered.api.entity.Entity> entity(final UUID uniqueId)
 
     @Override
     public void removeBlockEntity(final int x, final int y, final int z) {
-        this.shadow$removeBlockEntity(new BlockPos(x, y, z));
+        // *Just* removing the block entity is going to give is a glitchy state that should never happen.
+        // So just remove the whole block.
+        this.blockEntity(x, y, z).ifPresent(ignored -> this.removeBlock(x, y, z));
```
dualspiral, line 268: Seems reasonable to me, just want to check that we don't have any other clever things we want to do instead but I doubt we will.

gabizou: I don't know if this is correct, I want to mess with this a little bit locally:
- I don't like calling other API methods when implementing these interfaces (why they're all implemented with shadows/bridge calls)
- I'm worried about removeBlock having additional side effects that may be unintended (or rather a lack of controlled effects, like block drops etc.)

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/level/LevelMixin_API.java
```diff
@@ -308,7 +308,23 @@ public Collection<? extends BlockEntity> blockEntities() {
...
     public void addBlockEntity(final int x, final int y, final int z, final BlockEntity blockEntity) {
-        this.shadow$setBlockEntity(new BlockPos(x, y, z), (net.minecraft.world.level.block.entity.BlockEntity) Objects.requireNonNull(blockEntity, "blockEntity"));
+        // So, here we basically want to copy the given block entity into the location.
+        // BlockEntity stores its location, as well as it being mutable and stuff, so just setting what we've given here
+        // would cause unexpected bugs.
+        final net.minecraft.world.level.block.entity.BlockEntity mcOriginalBlockEntity = (net.minecraft.world.level.block.entity.BlockEntity) Objects.requireNonNull(blockEntity, "blockEntity");
+        // Save the nbt so we can copy it
+        final CompoundTag tag = mcOriginalBlockEntity.save(new CompoundTag());
+        // Ensure that where we are placing this blockentity is the right blockstate, so that minecraft will actually accept it.
+        this.world().setBlock(x, y, z, (org.spongepowered.api.block.BlockState) mcOriginalBlockEntity.getBlockState());
```
gabizou, line 318: Same with this, it's not a controlled set block, I'd like to have better control with using the block flags to set the block itself and ensure that lightning is enabled but drops are not enabled.

PR #3686, opened 2022-05-16
Add ServerWorld#saveAndFlush() method and fix memory leak related to block entities

Lignium:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2429) |  **Sponge**

- Implemented `ServerWorld#saveAndFlush()` method and fixed boolean flags (`skipSave` was enabled, `ServerWorld#save()` method actually did nothing).
- Fixed memory leak related to inability to unload block entities when the world is inactive (no players and forced chunks). When the world is inactive, updating and unloading entities (including block entities) takes only 300 ticks (15 seconds). Next, three special arrays for block entities are inflated and hold a huge number of block entities in memory. Now it is guaranteed that every world tick there will be cleanup of block entity arrays.

`ServerLevel#tick()`:
```java
boolean inactive = !this.players.isEmpty() || !this.getForcedChunks().isEmpty();
if (inactive) {
    this.emptyTime = 0;
}
if (inactive || this.emptyTime++ < 300) {
    ...
    this.tickBlockEntities();
    ...
}
```
`Level#tickBlockEntities()`:
```java
public final List<BlockEntity> blockEntityList = Lists.newArrayList();
public final List<BlockEntity> tickableBlockEntities = Lists.newArrayList();
protected final List<BlockEntity> blockEntitiesToUnload = Lists.newArrayList();

public void tickBlockEntities() {
    ...
    if (!this.blockEntitiesToUnload.isEmpty()) {
        this.tickableBlockEntities.removeAll(this.blockEntitiesToUnload);
        this.blockEntityList.removeAll(this.blockEntitiesToUnload);
        this.blockEntitiesToUnload.clear();
    }
    ...
}
```

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerLevelMixin.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerLevelMixin.java
@@ -441,6 +441,15 @@ public abstract class ServerLevelMixin extends LevelMixin implements ServerLevel
         this.impl$recentTickTimes[this.shadow$getServer().getTickCount() % 100] = postTickTime - this.impl$preTickTime;
     }
 
+    @Inject(method = "tick", at = @At("RETURN"))
+    private void impl$unloadBlockEntities(final BooleanSupplier param0, final CallbackInfo ci) {
+        if (!this.blockEntitiesToUnload.isEmpty()) {
+            this.tickableBlockEntities.removeAll(this.blockEntitiesToUnload);
+            this.blockEntityList.removeAll(this.blockEntitiesToUnload);
+            this.blockEntitiesToUnload.clear();
+        }
+    }
+
     private void impl$setWorldOnBorder() {
         ((WorldBorderBridge) this.shadow$getWorldBorder()).bridge$setAssociatedWorld(this.bridge$getKey());
     }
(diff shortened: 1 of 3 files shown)


src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerLevelMixin.java
```diff
@@ -441,6 +441,15 @@ public void save(@Nullable final ProgressListener progress, final boolean flush,
         this.impl$recentTickTimes[this.shadow$getServer().getTickCount() % 100] = postTickTime - this.impl$preTickTime;
     }
 
+    @Inject(method = "tick", at = @At("RETURN"))
+    private void impl$unloadBlockEntities(final BooleanSupplier param0, final CallbackInfo ci) {
+        if (!this.blockEntitiesToUnload.isEmpty()) {
```
Zidane, line 446: If Vanilla is doing this and see somehow overwrote it, fair. If not, we need to investigate if this causes breakages for game logic or mods. If the latter, where does Vanilla ever clear this?

src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerLevelMixin.java
```diff
@@ -441,6 +441,15 @@ public void save(@Nullable final ProgressListener progress, final boolean flush,
         this.impl$recentTickTimes[this.shadow$getServer().getTickCount() % 100] = postTickTime - this.impl$preTickTime;
     }
 
+    @Inject(method = "tick", at = @At("RETURN"))
```
Zidane, line 444: This is also out of scope for implementing the API portion of this, just so you know.

PR #3688, opened 2022-05-17
Add world save mode for ServerWorld#save() to support flush

Lignium:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2430) | **Sponge**

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/api/minecraft/server/level/ServerLevelMixin_API.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/api/minecraft/server/level/ServerLevelMixin_API.java
@@ -55,6 +55,7 @@ import org.spongepowered.api.scheduler.ScheduledUpdateList;
 import org.spongepowered.api.util.Ticks;
 import org.spongepowered.api.world.BlockChangeFlag;
 import org.spongepowered.api.world.SerializationBehavior;
+import org.spongepowered.api.world.WorldSaveMode;
 import org.spongepowered.api.world.border.WorldBorder;
 import org.spongepowered.api.world.chunk.WorldChunk;
 import org.spongepowered.api.world.generation.ChunkGenerator;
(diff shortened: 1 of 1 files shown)


src/mixins/java/org/spongepowered/common/mixin/api/minecraft/server/level/ServerLevelMixin_API.java
```diff
@@ -205,9 +208,17 @@ public Path directory() {
     }
 
     @Override
-    public boolean save() throws IOException {
+    public boolean save(WorldSaveMode saveMode) throws IOException {
+        Objects.requireNonNull(saveMode, "saveMode");
+        SerializationBehavior serializationBehavior =
+            ((PrimaryLevelDataBridge) serverLevelData).bridge$serializationBehavior().orElse(SerializationBehavior.AUTOMATIC);
+        if (!serializationBehavior.equals(SerializationBehavior.MANUAL) &&
```
ImMorpheus: Why ? This is changing behaviour and it decreasing coverage in the Overwrite below since the serialization behaviour will never be `AUTOMATIC` after this if.

src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerLevelMixin.java
```diff
@@ -324,13 +324,6 @@ public abstract class ServerLevelMixin extends LevelMixin implements ServerLevel
      */
     @Overwrite
     public void save(@Nullable final ProgressListener progress, final boolean flush, final boolean skipSave) {
-
-        final Cause currentCause = Sponge.server().causeStackManager().currentCause();
-
-        if (Sponge.eventManager().post(SpongeEventFactory.createSaveWorldEventPre(currentCause, ((ServerWorld) this)))) {
```
ImMorpheus: This is changing behaviour, why is the event not fired ? 

Note: it might make sense to not fire the event when serialization behaviour is `NONE` but it needs documentation and it does not seem to be what is happening here.

ImMorpheus (changes requested):
There are some changes in behaviour with serialization behaviour and events that seem unrelated to this PR.

src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerLevelMixin.java
Lignium: It turns out that this event is always fired, even if the save operation will not performed, and Post is not fired.

ImMorpheus: > Note: it might make sense to not fire the event when serialization behaviour is NONE but it needs documentation and it does not seem to be what is happening here.

On second thought, this is probably a breaking change so it needs to be fired regardless. We can change it in api-10 but if this targets api-8 the event needs to fire.

aromaa (changes requested):
I have made the fixes to the SerializationBehavior logic in 3a371bb06138134242ec0d706bccd8b7f909ba8a that were discussed on Discord.

We can take a look at how the events should be fired on api-10 but for api-8 it would be a breaking change.

PR #3697, opened 2022-05-31
feat(data): implement `SNBTDataFormat`

SettingDust:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2432)

### DIFF
--- a/src/main/java/org/spongepowered/common/data/persistence/SNBTDataFormat.java
+++ b/src/main/java/org/spongepowered/common/data/persistence/SNBTDataFormat.java
@@ -83,16 +83,16 @@ public final class SNBTDataFormat implements StringDataFormat {
 
     @Override
     public String write(final DataView data) throws IOException {
-        return NBTTranslator.INSTANCE.translate(data).getPrettyDisplay().getString();
+        return NBTTranslator.INSTANCE.translate(data).toString();
     }
 
     @Override
     public void writeTo(final Writer output, final DataView data) throws IOException {
-        output.write(NBTTranslator.INSTANCE.translate(data).getPrettyDisplay().getString());
+        output.write(NBTTranslator.INSTANCE.translate(data).toString());
     }
 
     @Override
     public void writeTo(final OutputStream output, final DataView data) throws IOException {
-        output.write(NBTTranslator.INSTANCE.translate(data).getPrettyDisplay().getString().getBytes(StandardCharsets.UTF_8));
+        output.write(NBTTranslator.INSTANCE.translate(data).toString().getBytes(StandardCharsets.UTF_8));
     }
 }
--- a/src/main/java/org/spongepowered/common/data/persistence/SNBTDataFormat.java
+++ b/src/main/java/org/spongepowered/common/data/persistence/SNBTDataFormat.java
@@ -1,0 +1,98 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/common/data/persistence/SNBTDataFormat.java
```diff
@@ -0,0 +1,98 @@
...
+
+        return new BufferedReader(reader);
+    }
+
+    @Override
+    public DataContainer read(final String input) throws InvalidDataException, IOException {
+        try {
+            return NBTTranslator.INSTANCE.translate(TagParser.parseTag(input));
+        } catch (CommandSyntaxException e) {
+            throw new RuntimeException(e);
```
dualspiral: No - we should not be throwing `RuntimeException`s like this. Use `InvalidDataException` instead (same goes for all the following methods).

testplugins/src/main/java/org/spongepowered/test/data/DataFormatTest.java
```diff
@@ -0,0 +1,79 @@
...
+                                }}).key("format").build(),
+                                Parameter.remainingJoinedStrings().key("value").build())
+                        .executor(context -> {
+
+                            StringDataFormat format = context.requireOne(Parameter.key("format", StringDataFormat.class));
+                            String value = context.requireOne(Parameter.key("value", String.class));
+                            try {
+                                context.cause().audience().sendMessage(Component.text(format.write(format.read(value))));
+                            } catch (IOException e) {
+                                throw new RuntimeException(e);
```
dualspiral: Don't throw `RuntimeExceptions` indiscriminately like this. In the case of commands, yes, it'll be caught, but the result is ugly. Gracefully handle it instead - we have `CommandException` that you should use instead if you must throw an exception.

src/main/java/org/spongepowered/common/data/persistence/SNBTDataFormat.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        return new BufferedReader(reader);
+    }
+
+    @Override
+    public DataContainer read(final String input) throws InvalidDataException, IOException {
+        try {
+            return NBTTranslator.INSTANCE.translate(TagParser.parseTag(input));
+        } catch (CommandSyntaxException e) {
```
dualspiral: `final` (same for all other catch blocks)

testplugins/src/main/java/org/spongepowered/test/data/DataFormatTest.java
```diff
@@ -0,0 +1,79 @@
...
+                        .addParameters(Parameter.choices(StringDataFormat.class, new HashMap<String, StringDataFormat>() {{
+                                    put("hocon", DataFormats.HOCON.get());
+                                    put("json", DataFormats.JSON.get());
+                                    put("snbt", DataFormats.SNBT.get());
+                                }}).key("format").build(),
+                                Parameter.remainingJoinedStrings().key("value").build())
+                        .executor(context -> {
+
+                            StringDataFormat format = context.requireOne(Parameter.key("format", StringDataFormat.class));
+                            String value = context.requireOne(Parameter.key("value", String.class));
```
dualspiral: Declare as final (it's generally what we do in our codebase, so I'd like to see it here too)

src/main/java/org/spongepowered/common/data/persistence/SNBTDataFormat.java
```diff
@@ -0,0 +1,98 @@
...
+            return NBTTranslator.INSTANCE.translate(TagParser.parseTag(
+                    new BufferedReader(new InputStreamReader(input, StandardCharsets.UTF_8)).lines().collect(Collectors.joining("\n"))));
+        } catch (final CommandSyntaxException e) {
+            throw new InvalidDataException(e);
+        }
+    }
+
+    @Override
+    public String write(final DataView data) throws IOException {
+        return NBTTranslator.INSTANCE.translate(data).getPrettyDisplay().getString();
```
dualspiral: Why not `toString` instead of `getPrettyDisplay().getString()`?

SettingDust: `getPrettyDisplay().getString()` is "pretty". I noticed the hocon data format seems like is "pretty". So I change that. It's `toString` before

PR #3699, opened 2022-06-01
SF: Delegate Vanilla registry population to Forge delegates

NickImpact:
The purpose of this PR is to fix registries on the SpongeForge side. With SpongeVanilla, we were successfully injecting and registering our own cache. With SpongeForge, Forge jumped in the middle, and expanded upon the vanilla registry system Sponge expected by creating two new registry types: NamespacedWrapper and NamespacedDefaultedWrapper.

 Both of these follow the MappingRegistry chain, and were receiving the expected mixin changes. However, the problem with registry registration was that Forge didn't actually write registry contents to these wrappers, but to a ForgeRegistry serving as a delegate to these wrappers. These calls were not tracked, so to Sponge, no registry contents existed. 

This set of mixins fixes the issue by registering the wrapper registry as a parent to the ForgeRegistry. This has been tested via debug and against a modded environment, both producing filled registries.

As for prefixes, I took a best guess but perhaps we will want them named differently.


forge/src/main/java/org/spongepowered/forge/bridge/minecraftforge/fml/RegistryDelegate.java
```diff
@@ -0,0 +1,35 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.forge.bridge.minecraftforge.fml;
+
+import org.spongepowered.common.bridge.core.RegistryBridge;
+
+public interface RegistryDelegate<T> {
```
dualspiral: `ForgeRegistryBridge` - this bridge only targets `ForgeRegistry` so should be named after it.

forge/src/main/java/org/spongepowered/forge/bridge/minecraftforge/fml/RegistryDelegate.java
```diff
@@ -0,0 +1,35 @@
...
+
+    RegistryBridge<T> sponge$parent();
```
dualspiral: `bridge$parent()`

forge/src/main/java/org/spongepowered/forge/bridge/minecraftforge/fml/RegistryDelegate.java
```diff
@@ -0,0 +1,35 @@
...
+
+    interface Parent<V> extends RegistryBridge<V> {}
```
dualspiral: What's the point of this? I can't quite work out what you're trying to do here from the below context. Can you not just use `RegistryBridge<?>` when it's needed?

If you _do_ need a different bridge, make this a different outer interface altogether - this isn't really related to this outer class. `NamespacedWrappedBridge`?

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/core/ForgeRegistryMixin.java
```diff
@@ -0,0 +1,68 @@
...
+import org.spongepowered.common.registry.SpongeRegistryType;
+import org.spongepowered.forge.bridge.minecraftforge.fml.RegistryDelegate;
+
+import java.util.Map;
+
+@Mixin(ForgeRegistry.class)
+public abstract class ForgeRegistryMixin<V extends IForgeRegistryEntry<V>> implements RegistryDelegate<V> {
+
+    @Shadow @Final Map<ResourceLocation, ?> slaves;
+    @Shadow @Final private net.minecraft.resources.ResourceKey<Registry<V>> key;
```
dualspiral:
```suggestion
    // @formatter:off
    @Shadow @Final Map<ResourceLocation, ?> slaves;
    @Shadow @Final private net.minecraft.resources.ResourceKey<Registry<V>> key;
    // @formatter:on
```

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/core/ForgeRegistryMixin.java
```diff
@@ -0,0 +1,68 @@
...
+    private Parent<V> parent;
+
+    @Inject(method = "add(ILnet/minecraftforge/registries/IForgeRegistryEntry;Ljava/lang/String;)I", at = @At("TAIL"))
+    public void sponge$writeToParent(int id, V value, String owner, CallbackInfoReturnable<Integer> cir) {
```
dualspiral: `forge$writeToParent`

Also, parameters should be final.

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/core/ForgeRegistryMixin.java
```diff
@@ -0,0 +1,68 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.common.registry.SpongeRegistryEntry;
+import org.spongepowered.common.registry.SpongeRegistryType;
+import org.spongepowered.forge.bridge.minecraftforge.fml.RegistryDelegate;
+
+import java.util.Map;
+
+@Mixin(ForgeRegistry.class)
+public abstract class ForgeRegistryMixin<V extends IForgeRegistryEntry<V>> implements RegistryDelegate<V> {
```
dualspiral: `ForgeRegistryMixin_Forge`

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/core/ForgeRegistryMixin.java
```diff
@@ -0,0 +1,68 @@
...
+                    .stream()
+                    .filter(key -> key.getValue() instanceof RegistryDelegate.Parent)
+                    .findFirst()
+                    .map(entry -> (RegistryDelegate.Parent) entry.getValue())
+                    .orElse(null);
+        }
+
+        final ResourceKey root = (ResourceKey) (Object) this.key.getRegistryName();
+        final ResourceKey location = (ResourceKey) (Object) this.key.location();
+        this.parent.bridge$register(new SpongeRegistryEntry<>(new SpongeRegistryType<>(root, location),
```
dualspiral: Potential NPE here if `this.parent` resolves to `null`.

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/core/ForgeRegistryMixin.java
```diff
@@ -0,0 +1,68 @@
...
+@Mixin(ForgeRegistry.class)
+public abstract class ForgeRegistryMixin<V extends IForgeRegistryEntry<V>> implements RegistryDelegate<V> {
+
+    @Shadow @Final Map<ResourceLocation, ?> slaves;
+    @Shadow @Final private net.minecraft.resources.ResourceKey<Registry<V>> key;
+    private Parent<V> parent;
+
+    @Inject(method = "add(ILnet/minecraftforge/registries/IForgeRegistryEntry;Ljava/lang/String;)I", at = @At("TAIL"))
+    public void sponge$writeToParent(int id, V value, String owner, CallbackInfoReturnable<Integer> cir) {
+        if(this.parent == null) {
```
dualspiral: If `this.parent` resolves to `null`, this will get called every time - is this expected, or should this only _ever_ evaluate once?

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/core/ForgeRegistryMixin.java
```diff
@@ -0,0 +1,68 @@
...
+            this.parent = this.slaves.entrySet()
+                    .stream()
+                    .filter(key -> key.getValue() instanceof RegistryDelegate.Parent)
+                    .findFirst()
```
dualspiral: Is there a possibilty that if more than one registry is returned, all are valid parents that should be registered in the Sponge registries (i.e., is this limit of one actually artificial)?

dualspiral (commented):
Thanks for doing this - few naming/logic things to query:

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/core/ForgeRegistryMixin.java
```diff
@@ -0,0 +1,68 @@
...
+import org.spongepowered.forge.bridge.minecraftforge.fml.RegistryDelegate;
+
+import java.util.Map;
+
+@Mixin(ForgeRegistry.class)
+public abstract class ForgeRegistryMixin<V extends IForgeRegistryEntry<V>> implements RegistryDelegate<V> {
+
+    @Shadow @Final Map<ResourceLocation, ?> slaves;
+    @Shadow @Final private net.minecraft.resources.ResourceKey<Registry<V>> key;
+    private Parent<V> parent;
```
dualspiral: `forge$parent`

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/core/ForgeRegistryMixin.java
NickImpact: Goal would be to only ever evaluate this once. From what I could tell, each forge registry did at least map to a proper vanilla registry. What I could do instead of this is, as the slave map is set, inject there instead for parent resolution. So in the event I'm wrong, we don't run this query for every registration.

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/core/ForgeRegistryMixin.java
NickImpact: This is something I did not quite consider, and I might need to fallback on this and rethink it. From my tests, I was really only focused on the minecraft:root/minecraft:items registry, which had a 1:1 mapping. I'll look around and see if any registries break that rule.

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/core/ForgeRegistryMixin.java
NickImpact: In the event our parent isn't available, should we just skip the operation? What would be your preference?

forge/src/main/java/org/spongepowered/forge/bridge/minecraftforge/fml/RegistryDelegate.java
NickImpact: Looking back at it with this feedback, the Parent isn't exactly required at all. I originally made this change thinking it'd be a good marker, but I did miss that the extensions already cover RegistryBridge. Is this question reconsidering the whole interface, or just parent?

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/core/ForgeRegistryMixin.java
dualspiral: If there  is no parent - we can't do anything anyway. However, are we ever expecting this to be the case? If not, then I'd be tempted to be writing a nasty error to the log, else I'd probably just slap a warning on saying we can't find it.

forge/src/main/java/org/spongepowered/forge/bridge/minecraftforge/fml/RegistryDelegate.java
dualspiral: Just the `Parent` interface. I see the point of this delegate.

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/core/ForgeRegistryMixin.java
dualspiral: I figured it was an edge case, but I'm also guessing that for the standard registries, there is only one "parent" registry - so removing the findFirst will have no additional effect. In this scenario, I'd rather end up sending it to all registries, because if we select just one, we might choose the "wrong" one.

forge/src/main/java/org/spongepowered/forge/bridge/minecraftforge/fml/RegistryDelegate.java
NickImpact: Renamed to ForgeRegistryBridge

forge/src/main/java/org/spongepowered/forge/bridge/minecraftforge/fml/RegistryDelegate.java
NickImpact: Name change applied

forge/src/main/java/org/spongepowered/forge/bridge/minecraftforge/fml/RegistryDelegate.java
NickImpact: Done, removed parent interface in favor of RegistryBridge

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/core/ForgeRegistryMixin.java
NickImpact: Added a no parent check, explanation will be in a separate comment

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/core/ForgeRegistryMixin.java
NickImpact: I've changed this solution for a scenario which wont attempt to invoke this parent establishment every registration. I'll comment further on that in a general comment as this code is no longer present.

forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/core/ForgeRegistryMixin.java
NickImpact: Ultimately changed this to a mapping of ResourceKey to Registry Bridge, now with the name forge$parents.

NickImpact:
Following the feedback, I've gone ahead and applied a general set of changes. Rather than attempt to set the parent each registration call if none had been set, this new solution only ever tries to append a parent at time of actual forge parent registration. I originally considered just using a Set of RegistryBridges, but figured it would be additionally smart (for debugging purposes) to have a ResourceKey mapped to that bridge for variable evaluation, and given it's available in forge's slave map. 

As for the other questions, at present, it looks like every ForgeRegistry simply has 1 parent, and I logged no cases where that number was higher or 0. I've added logging in the case of a registry with no parent established, which will only ever fire once if detected. And if there ever is a scenario where more than one parent exists, the code will now register to each registry it has established as a parent of itself.

dualspiral:
https://github.com/SpongePowered/Sponge/commit/6e4b474750c1dfdc28de44847077b804d0a25699

PR #3702, opened 2022-06-02
Fix many WorldManager issues

Yeregorix:
Affected API methods are : `World#save`, `WorldManager#loadProperties`, `WorldManager#loadWorld`, `WorldManager#copyWorld` and `WorldManager#moveWorld`. See the commit descriptions for more details.

I also added json indentation when writing templates so it's easier to edit as some server owners may want to edit generation settings.

~~One problem persists: renaming or copying from a vanilla world to a custom world does not work because custom worlds need a template and vanilla worlds don't have templates.~~

### DIFF
--- a/src/main/java/org/spongepowered/common/world/server/SpongeWorldManager.java
+++ b/src/main/java/org/spongepowered/common/world/server/SpongeWorldManager.java
@@ -726,12 +726,12 @@ public abstract class SpongeWorldManager implements WorldManager {
 
         if (this.isVanillaWorld(key)) {
             final LevelStem stem = this.server.getWorldData().worldGenSettings().dimensions().get(SpongeWorldManager.createStemKey(key));
-            JsonElement template = SpongeWorldManager.stemToJson(stem);
+            final JsonElement template = SpongeWorldManager.stemToJson(stem);
 
             try {
                 this.writeTemplate(template, movedKey);
             } catch (final IOException e) {
-                FutureUtil.completedWithException(e);
+                return FutureUtil.completedWithException(e);
             }
         } else {
             final Path dimensionTemplate = this.getDataPackFile(key);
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/world/server/SpongeWorldManager.java
```diff
@@ -773,46 +704,45 @@ public CompletableFuture<Boolean> moveWorld(final ResourceKey key, final Resourc
...
+            Files.move(configFile, movedConfigFile, StandardCopyOption.REPLACE_EXISTING);
         } catch (final IOException e) {
             return FutureUtil.completedWithException(e);
         }
 
-        final Path dimensionTemplate = this.getDataPackFile(key);
-        final Path copiedDimensionTemplate = this.getDataPackFile(movedKey);
+        if (this.isVanillaWorld(key)) {
+            final LevelStem stem = this.server.getWorldData().worldGenSettings().dimensions().get(SpongeWorldManager.createStemKey(key));
+            JsonElement template = SpongeWorldManager.stemToJson(stem);
```
Zidane: `final`

src/main/java/org/spongepowered/common/world/server/SpongeWorldManager.java
```diff
@@ -773,46 +704,45 @@ public CompletableFuture<Boolean> moveWorld(final ResourceKey key, final Resourc
...
+            }
+        } else {
+            final Path dimensionTemplate = this.getDataPackFile(key);
+            final Path movedDimensionTemplate = this.getDataPackFile(movedKey);
+
+            try {
+                Files.createDirectories(movedDimensionTemplate.getParent());
+                Files.move(dimensionTemplate, movedDimensionTemplate, StandardCopyOption.REPLACE_EXISTING);
+            } catch (final IOException e) {
+                FutureUtil.completedWithException(e);
```
Zidane: Did you forget a `return`? The original code did too.

PR #3709, opened 2022-06-25
Fix multiple RespawnPlayerEvents issues

Yeregorix:
1. New player instance was not captured properly (`Recreate#entity()` and `Recreate#recreatedPlayer()` were returning the same object).
2. `Recreate#setDestinationPosition(Vector3d)` was ignored (consequence of 1. ). 
3. `SelectWorld#setDestinationWorld()` was ignored.
4. `Post#originalDestinationWorld()` and `Post#destinationWorld()` were returning incorrect values.
5. API `ServerPlayer#respawn()` was not firing `SelectWorld` but only `Recreate` and `Post` events.

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/core/server/players/PlayerListMixin.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/core/server/players/PlayerListMixin.java
@@ -39,7 +39,6 @@ import net.minecraft.network.chat.TranslatableComponent;
 import net.minecraft.network.protocol.Packet;
 import net.minecraft.network.protocol.game.ClientboundDisconnectPacket;
 import net.minecraft.network.protocol.game.ClientboundLoginPacket;
-import net.minecraft.network.protocol.game.ClientboundPlayerInfoPacket;
 import net.minecraft.resources.ResourceKey;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.ServerScoreboard;
@@ -55,6 +54,7 @@ import net.minecraft.world.level.Level;
 import net.minecraft.world.level.border.BorderChangeListener;
 import net.minecraft.world.level.border.WorldBorder;
 import net.minecraft.world.level.dimension.DimensionType;
+import net.minecraft.world.phys.Vec3;
 import org.apache.logging.log4j.Logger;
 import org.objectweb.asm.Opcodes;
 import org.spongepowered.api.Sponge;
@@ -72,6 +72,7 @@ import org.spongepowered.api.service.ban.Ban;
 import org.spongepowered.api.service.permission.PermissionService;
 import org.spongepowered.api.service.permission.Subject;
 import org.spongepowered.api.world.server.ServerLocation;
+import org.spongepowered.api.world.server.ServerWorld;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Mutable;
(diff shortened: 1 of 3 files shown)


src/mixins/java/org/spongepowered/common/mixin/core/server/players/PlayerListMixin.java
```diff
@@ -148,8 +147,7 @@ public abstract class PlayerListMixin implements PlayerListBridge {
     // @formatter:on
 
     private boolean impl$isGameMechanicRespawn = false;
-    ResourceKey<Level> impl$newDestination = null;
-    ResourceKey<Level> impl$originalDestination = null;
+    private ServerWorld impl$originalRespawnDestination = null;
```
dualspiral: I don't like storing the ServerWorld in favour of the ResourceKey for the level in case something happens and we end up holding on to the reference. Can we stick with `ResourceKey` and call `this.server.getLevel` as required?

gabizou: I believe I made a function for making a Supplier<ServerLevel> or something that'll use a weak reference and/or uses the resource key as a fallback.

src/mixins/java/org/spongepowered/common/mixin/core/server/players/PlayerListMixin.java
```diff
@@ -521,56 +509,56 @@ public abstract class PlayerListMixin implements PlayerListBridge {
...
-        final org.spongepowered.api.world.server.ServerWorld destinationWorld = (org.spongepowered.api.world.server.ServerWorld) this.server.getLevel(this.impl$newDestination == null ? Level.OVERWORLD : this.impl$newDestination);
+        final RespawnPlayerEvent.SelectWorld event = SpongeEventFactory.createRespawnPlayerEventSelectWorld(PhaseTracker.getCauseStackManager().currentCause(),
+                        this.impl$originalRespawnDestination, (ServerWorld) player.getLevel(), this.impl$originalRespawnDestination, (ServerPlayer) player);
+        SpongeCommon.post(event);
+
+        return (ServerLevel) event.destinationWorld();
+    }
 
-        final RespawnPlayerEvent.Recreate event = SpongeEventFactory.createRespawnPlayerEventRecreate(PhaseTracker.getCauseStackManager().currentCause(), destinationPosition, originalWorld, originalPosition, destinationWorld, originalDestinationWorld, destinationPosition, (ServerPlayer) originalPlayer, (ServerPlayer) recreatedPlayer, this.impl$isGameMechanicRespawn, !keepAllPlayerData);
+    @Redirect(method = "respawn", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/level/ServerPlayer;getX()D", ordinal = 1))
```
gabizou: Can you use a slice instead of an ordinal here?

src/mixins/java/org/spongepowered/common/mixin/core/server/players/PlayerListMixin.java
```diff
@@ -521,56 +509,56 @@ public abstract class PlayerListMixin implements PlayerListBridge {
...
-        at = @At(
-            value = "INVOKE",
-            target = "Lnet/minecraft/server/network/ServerGamePacketListenerImpl;send(Lnet/minecraft/network/protocol/Packet;)V",
-            ordinal = 1
-        )
-    )
-    private void impl$callRespawnPlayerRecreateEvent(
-        final ServerGamePacketListenerImpl serverPlayNetHandler, final Packet<?> packetIn, final net.minecraft.server.level.ServerPlayer originalPlayer, final boolean keepAllPlayerData) {
-        final net.minecraft.server.level.ServerPlayer recreatedPlayer = serverPlayNetHandler.player;
+    @Redirect(method = "respawn", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/MinecraftServer;overworld()Lnet/minecraft/server/level/ServerLevel;"))
```
gabizou, line 517: Just to be sure, is this only one call in `respawn`? Or are there multiple `overworld` calls?

src/mixins/java/org/spongepowered/common/mixin/core/server/players/PlayerListMixin.java
```diff
@@ -521,56 +509,56 @@ public abstract class PlayerListMixin implements PlayerListBridge {
...
-        final org.spongepowered.api.world.server.ServerWorld destinationWorld = (org.spongepowered.api.world.server.ServerWorld) this.server.getLevel(this.impl$newDestination == null ? Level.OVERWORLD : this.impl$newDestination);
+        final ServerPlayer recreatedPlayer = (ServerPlayer) cir.getReturnValue();
+        final ServerWorld originalWorld = (ServerWorld) player.level;
 
-        final RespawnPlayerEvent.Post event = SpongeEventFactory.createRespawnPlayerEventPost(PhaseTracker.getCauseStackManager().currentCause(), destinationWorld, originalWorld, originalDestinationWorld, (ServerPlayer) cir.getReturnValue());
+        final RespawnPlayerEvent.Post event = SpongeEventFactory.createRespawnPlayerEventPost(PhaseTracker.getCauseStackManager().currentCause(),
+                recreatedPlayer.world(), originalWorld, this.impl$originalRespawnDestination, recreatedPlayer);
         SpongeCommon.post(event);
+
+        this.impl$originalRespawnDestination = null;
```
gabizou, line 569: I'm not fond of having this being the only reset for the respawn destination in the event of exceptions during respawns, I could see the field being leaked until the next player respawns, which means a world is leaked (unless the supplier is fixed).

gabizou (changes requested):

src/mixins/java/org/spongepowered/common/mixin/core/server/players/PlayerListMixin.java
Yeregorix: I searched but couldn't find such a structure. I'll probably end up using a `ResourceKey`.

src/mixins/java/org/spongepowered/common/mixin/core/server/players/PlayerListMixin.java
Yeregorix: There is only one.

PR #3711, opened 2022-06-26
Fixes for WorldEdit Forge

gabizou:


src/mixins/java/org/spongepowered/common/mixin/tracker/world/level/chunk/LevelChunkMixin_Tracker.java
```diff
@@ -89,15 +91,27 @@ public abstract class LevelChunkMixin_Tracker implements TrackedLevelChunkBridge
...
+            return;
+        }
+        final PhaseTrackerCategory trackerConfig = SpongeConfigs.getCommon().get().phaseTracker;
+        final boolean disableDirectAccess = trackerConfig.disableDirectChunkAccess;
+        final boolean log = trackerConfig.logDirectChunkAccess;
+        if (log) {
+            if (disableDirectAccess || !LevelChunkMixin_Tracker.tracker$hasLoggedDirectAccess) {
+                new PrettyPrinter(80).add("Direct Chunk Access")
+                    .hr()
+                    .add(new IllegalAccessException("It has been configured to log when Chunk.setBlock in a ServerWorld has been called, effectively bypassing Sponge's tracking."))
```
dualspiral, line 110: I think this is confusing. We should be a little bit more verbose about what's going on.

"A mod or plugin is attempting to directly edit blocks in the world, bypassing Sponge's tracking system. This is necessary for some plugins/mods that do massive changes to the world, however, this won't be tracked and some events will not fire, and Sponge does not guarantee the state of the world once these actions have completed.

The server is currently set to [ALLOW|DENY] this. You can change this behavior by changing the value of `disableDirectChunkAccess` in the Sponge config file."

src/mixins/java/org/spongepowered/common/mixin/tracker/world/level/chunk/LevelChunkMixin_Tracker.java
```diff
@@ -89,15 +91,27 @@ public abstract class LevelChunkMixin_Tracker implements TrackedLevelChunkBridge
...
+                    .log(PhaseTracker.LOGGER, Level.WARN);
```
dualspiral, line 111: You've not set `tracker$hasLoggedDirectAccess` to `true` here.

forge/src/mixins/java/org/spongepowered/forge/mixin/core/api/event/block/InteractBlockEvent_SecondaryMixin_Forge.java
```diff
@@ -0,0 +1,72 @@
...
+import org.spongepowered.common.util.VecHelper;
+import org.spongepowered.forge.hook.ForgeEventHooks;
+import org.spongepowered.forge.launch.bridge.event.SpongeEventBridge_Forge;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Optional;
+
+@Mixin(InteractBlockEvent.Secondary.class)
+public interface InteractBlockEvent_SecondaryMixin_Forge extends SpongeEventBridge_Forge {
```
dualspiral, line 46: We also need to mix in to the ~~Sponge~~ Forge event, like this, to enable syncing (if necessary):

https://github.com/SpongePowered/Sponge/commit/5f68c7b68e612720af4c0210e275f0dacc0c6831#diff-51b636ba70e8d0fd91fbd48a585c347dc64c6c14cab18b18157f82dc5fa3b3d6R45

It sucks, but it was the best way I could figure out to do it thanks to the fact we have to dual bus.

(edit: Forge, not Sponge)

src/applaunch/java/org/spongepowered/common/applaunch/config/common/PhaseTrackerCategory.java
```diff
@@ -142,6 +141,27 @@ public final class PhaseTrackerCategory {
...
+    public boolean logDirectChunkAccess = true;
+
+    @Setting("disable-direct-chunk-block-access")
+    @Comment("In certain situations, an administrator may wish to disable mods capabilities\n" +
+        "in performing block changes directly with a Chunk, effectively bypassing sponge's\n" +
+        "ability to track said changes. When enabled with `warn-on-direct-chunk-access`, a\n" +
+        "warning will be emitted each time a block is attempted to be changed on a Chunk.\n" +
+        "When logging is disabled, the changes are blocked regardless. Behaviors of these\n" +
+        "mods are not guaranteed.")
+    public boolean disableDirectChunkAccess = false;
```
dualspiral, line 163: I feel like we should put these into their own sub category so we don't have two walls of text saying the same thing.

In itself, I also think we need to simplify what we're trying to say here because we're just going to have admins glaze over what's going on here.

> Settings for controlling whether or not plugins/mods can directly edit chunks while bypassing the Sponge tracker.
> 
> Direct chunk access may be necessary for some plugins/mods that perform a large number of changes to operate effciently -- mostly those that perform large world manipulation on command. However, such access bypasses the Sponge tracker, meaning that some events that would normally fire will not, and so plugins that act upon those changes will not be able to do so. For servers that rely heavily on plugins that do something on every block change, and do not intend to perform large scale world editing, you may wish to turn direct chunk access off.
> 
> Sponge cannot guarantee the behavior of worlds where direct chunk access occurs.
> 
> By default, Sponge allows direct chunk access, and will print a warning in your server logs the first time it occurs during the game's lifetime. 
> 

and then, for the two options

> Controls whether Sponge will log when a direct chunk access occurs.
> 
> * If direct chunk access is allowed, only the first attempt will be logged.
> * If direct chunk access is denied, each attempt will be logged.
> 
> It is recommended that you keep this setting on.

> Controls whether Sponge will allow direct chunk access.

I also think we should remove the double negative here - `allow-direct-chunk-block-access = true` rather than `disable-direct-chunk-block-access = false`.

forge/src/mixins/java/org/spongepowered/forge/mixin/core/api/event/block/InteractBlockEvent_SecondaryMixin_Forge.java
gabizou: I don't think we really do, do we? In this particular case, it's the Sponge event getting thrown only and the forge event isn't being thrown.

dualspiral: Depends if we need to sync the event up - like cancellable status and such.

ImMorpheus:
@gabizou This seems to be a change in direction compared to past releases. Is there a concern this is going to be abused by mods ? What are the downsides of bypassing the phase tracker ? Is it just potentially losing events or are we going to deal with the tracker being in an inconsistent state ?

I think we should at the very least have this disabled by default if we want to go this route.

Yeregorix:
Superseded by #3844, #3847 and #3855.

PR #3713, opened 2022-06-27
Edits to World Edit Accomodations PR

Deltric:
- Set the boolean value to true for the one time log direct access mixin
- Adds a bridge mixin to the PlayerInteractEvent.RightClickBlock so that it no longer exceptions when right clicking blocks with items
- Fixes forge CommandEvent behavior by passing in the command value with a forward slash as it normally receives in a standalone environment.

### DIFF
--- a/forge/src/launch/java/org/spongepowered/forge/launch/command/ForgeCommandManager.java
+++ b/forge/src/launch/java/org/spongepowered/forge/launch/command/ForgeCommandManager.java
@@ -59,8 +59,10 @@ public final class ForgeCommandManager extends SpongeCommandManager {
         final CommandRegistrar<?> registrar = mapping.registrar();
         final boolean isBrig = registrar instanceof BrigadierBasedRegistrar;
         final ParseResults<CommandSourceStack> parseResults;
+        final ParseResults<CommandSourceStack> forgeResults;
         if (isBrig) {
             parseResults = this.getDispatcher().parse(original, (CommandSourceStack) cause);
+            forgeResults = new ParseResults<>(parseResults.getContext(), new SpongeStringReader("/" + original), parseResults.getExceptions());
         } else {
             // We have a non Brig registrar, so we just create a dummy result for mods to inspect.
             final CommandContextBuilder<CommandSourceStack> contextBuilder = new CommandContextBuilder<>(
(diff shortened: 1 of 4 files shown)


forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/event/entity/player/PlayerInteractEvent_RightClickBlock_Forge.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public void bridge$syncFrom(Event event) {
+        final InteractBlockEvent.Secondary spongeEvent = (InteractBlockEvent.Secondary) event;
+        ((net.minecraftforge.eventbus.api.Event) (Object) this).setCanceled(spongeEvent.isCancelled());
+    }
+
+    @Override
+    public void bridge$syncTo(Event event) {
+        final InteractBlockEvent.Secondary spongeEvent = (InteractBlockEvent.Secondary) event;
+        spongeEvent.setCancelled(((net.minecraftforge.eventbus.api.Event) (Object) this).isCanceled());
```
gabizou, line 45: I think you'll need to pass the tristate result back to the sponge event as well

Deltric: Is there an existing way of converting Tristates and Results back and forth? If not would be there be a recommended way to make a utility for it possibly?

forge/src/launch/java/org/spongepowered/forge/launch/command/ForgeCommandManager.java
```diff
@@ -73,10 +75,11 @@ protected CommandResult processCommand(final CommandCause cause, final CommandMa
                 contextBuilder.withArgument("parsed", new ParsedArgument<>(command.length(), original.length(), args));
             }
             parseResults = new ParseResults<>(contextBuilder, new SpongeStringReader(original), Collections.emptyMap());
+            forgeResults = new ParseResults<>(contextBuilder, new SpongeStringReader("/" + original), Collections.emptyMap());
```
dualspiral, line 78: This isn't really the right way to do this. If the problem is that we're dropping the `/` in the parse results in the stack somewhere, we should be fixing it there, rather than here. This is putting a plaster over the problem - which is not correct.

Let's fix it properly rather than just paper over the cracks.

Deltric: Understandable, will try taking a deeper look into this today after I've slept.

ImMorpheus:
I'll close this to keep discussion in https://github.com/SpongePowered/Sponge/pull/3711

If you ever want to readdress the underlying issue with the forward slash, open a new PR. Thanks!

PR #3731, opened 2022-07-30
Implement Keys#IS_ATTACHED for ladders

Lignium:
Ladder blocks always is attached to other blocks.


Faithcaio:
Should this not check if it actually is attached? e.g. with (LadderBlock#canSurvive)
Ladders can at least temporarily not be attached to anything.

Lignium:
@Faithcaio The implementation is similar to other blocks, such as a sign, banner or a torch, and I agree with this. This data is not tied to context and not tied to a world. In the correct world, the ladder will never hang in the air, because it will fall. Thus, data can be obtained outside the world, for example, by working with its own virtual buffer, and the data can be used to partially correct the block states in this buffer. This is as an example of my use case.

PR #3739, opened 2022-08-15
Replace printStackTrace() with a proper logger

Yeregorix:
I originally had an issue where my plugin failed to construct but I had no traces in the log files so I wanted to replace the printStackTrace() with a proper logger. I took the opportunity to do this in several places.

### DIFF
--- a/src/main/java/org/spongepowered/common/block/entity/SpongeBlockEntityArchetypeBuilder.java
+++ b/src/main/java/org/spongepowered/common/block/entity/SpongeBlockEntityArchetypeBuilder.java
@@ -102,7 +102,7 @@ public final class SpongeBlockEntityArchetypeBuilder extends AbstractDataBuilder
     public BlockEntityArchetype.Builder state(final BlockState state) {
         final net.minecraft.world.level.block.state.BlockState blockState = (net.minecraft.world.level.block.state.BlockState) state;
         if (!((BlockStateBridge) (blockState)).bridge$hasTileEntity()) {
-            SpongeBlockEntityArchetypeBuilder.LOGGER.warn("BlockState " + state + " does not provide TileEntities!", new IllegalArgumentException());
+            SpongeBlockEntityArchetypeBuilder.LOGGER.warn("BlockState {} does not provide BlockEntities!", state, new IllegalArgumentException());
         }
         if (this.blockState != state) {
             this.data = null;
--- a/src/main/java/org/spongepowered/common/block/entity/SpongeBlockEntityArchetypeBuilder.java
+++ b/src/main/java/org/spongepowered/common/block/entity/SpongeBlockEntityArchetypeBuilder.java
@@ -25,6 +25,8 @@
 package org.spongepowered.common.block.entity;
 
 import net.minecraft.nbt.CompoundTag;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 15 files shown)


gabizou (approved):
There's a fair few places we call `e.PrintStackTrace()`, mind fixing those up in one fell swoop?

Yeregorix:
There are still other `printStackTrace()` indeed.
Most of them are in the test plugins / unit tests, I thought it wasn't really important.
Some others are in asm class visitors so wasn't sure if it would cause classloading issues or not.

Yeregorix:
I have now replaced all `printStackTrace` except in test sourcesets, and except in `SpongeWorldManager` because it will be fixed by PR #3702.

src/main/java/org/spongepowered/common/block/entity/SpongeBlockEntityArchetypeBuilder.java
```diff
@@ -99,7 +102,7 @@ public BlockEntityArchetype.Builder from(final BlockEntityArchetype value) {
     public BlockEntityArchetype.Builder state(final BlockState state) {
         final net.minecraft.world.level.block.state.BlockState blockState = (net.minecraft.world.level.block.state.BlockState) state;
         if (!((BlockStateBridge) (blockState)).bridge$hasTileEntity()) {
-            new IllegalArgumentException("BlockState: "+ state + " does not provide TileEntities!").printStackTrace();
+            SpongeBlockEntityArchetypeBuilder.LOGGER.warn("BlockState " + state + " does not provide TileEntities!", new IllegalArgumentException());
```
Zidane: TileEntities -> BlockEntities

zml2008: you'll also want to replace string concat with string interpolation

```suggestion
            SpongeBlockEntityArchetypeBuilder.LOGGER.warn("BlockState {} does not provide TileEntities!", state, new IllegalArgumentException());
```

Zidane:
@Yeregorix @zml2008 

I am aware at the amount of effort of such a change but zml you are 100% right, the "global logger" is not helpful, not to know where whatever is coming from. I think it is time to likely axe it so `SpongeCommon.logger()` goes away and each class that logs creates it's own logger. Thoughts? Beyond being a massive change, I see no issue. This doesn't change giving each plugin a logger. If a plugin wants to do class loggers or not that is up to them. This is just a stylistic choice for us to keep doing this for them.

Yeregorix:
From a developer point of view, I prefer each class having its own logger.
From a server owner pov, it's easier to identify which mod is involded when you see `[Sponge]` at the beginning of the logged lines.

1. `[Sponge] Something went wrong.` (easy to identify that Sponge is involved)
vs
2. `[SomeClass] Something went wrong.` (most people have no idea what `SomeClass` is or to which mod it belongs)

Zidane:
@Yeregorix I agree. However classes such as DataUpdaterDelegate fall right into this issue in this PR, no? I'd have to look at our typical log files but someone wouldn't know based on the logging lines.

Yeregorix:
Yes, I'm just opposing the pros and cons of each convention.

Yeregorix:
For the moment, for each class I used the common logger if I saw that it was already used in the class, otherwise I used a specific logger.

zml2008 (approved):

PR #3743, opened 2022-08-18
Break explode overwrite into less overbearing injections

vectrixdevelops:
Fixes compatibility with the mod Road Runner's explosion optimization by breaking the `@Overwrite` into smaller injections that play nicely with the mod.

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
@@ -24,18 +24,12 @@
  */
 package org.spongepowered.common.mixin.core.world.level;
 
-import com.google.common.collect.Sets;
 import net.minecraft.core.BlockPos;
-import net.minecraft.util.Mth;
 import net.minecraft.world.damagesource.DamageSource;
 import net.minecraft.world.entity.Entity;
-import net.minecraft.world.entity.LivingEntity;
 import net.minecraft.world.entity.player.Player;
-import net.minecraft.world.item.enchantment.ProtectionEnchantment;
 import net.minecraft.world.level.ExplosionDamageCalculator;
 import net.minecraft.world.level.Level;
-import net.minecraft.world.level.block.state.BlockState;
-import net.minecraft.world.level.material.FluidState;
 import net.minecraft.world.phys.AABB;
 import net.minecraft.world.phys.Vec3;
 import org.spongepowered.api.event.Cause;
@@ -45,10 +39,12 @@ import org.spongepowered.api.world.explosion.Explosion;
 import org.spongepowered.api.world.server.ServerLocation;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
-import org.spongepowered.asm.mixin.Overwrite;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Constant;
 import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.ModifyConstant;
+import org.spongepowered.asm.mixin.injection.Redirect;
 import org.spongepowered.asm.mixin.injection.Surrogate;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 import org.spongepowered.common.SpongeCommon;
@@ -61,8 +57,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
 import java.util.StringJoiner;
 
 @Mixin(net.minecraft.world.level.Explosion.class)
@@ -90,6 +84,7 @@ public abstract class ExplosionMixin implements ExplosionBridge {
     private int impl$resolution;
     private float impl$randomness;
     private double impl$knockback;
+    private List<Entity> impl$affectedEntities;
 
     @Inject(
         method = "<init>(Lnet/minecraft/world/level/Level;Lnet/minecraft/world/entity/Entity;Lnet/minecraft/world/damagesource/DamageSource;Lnet/minecraft/world/level/ExplosionDamageCalculator;DDDFZLnet/minecraft/world/level/Explosion$BlockInteraction;)V",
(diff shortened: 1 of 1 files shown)


src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
```diff
@@ -118,158 +113,107 @@ public abstract class ExplosionMixin implements ExplosionBridge {
...
-     */
-    @Overwrite
-    public void explode() {
-
-        // Sponge Start - Do not run calculation logic on client thread
+    @Inject(method = "explode", at = @At("HEAD"), cancellable = true)
+    public void impl$explode_clientCheck(CallbackInfo ci) {
         if (this.level.isClientSide) {
-            return;
+            ci.cancel();
```
Zidane: This could remain a return imo or, better, inject right after this check. I only propose this in-case Mojang does something different in the future and we forget about the cancel. Which effectively enters overwrite territory there.

vectrixdevelops: Maybe I misunderstand what you mean, however this inject seemed strange to me anyway. I only added it since the overwrite had it, but is there a particular reason why `explode` should only run on the server? The `explode` method from Mojang does not have a client check itself. If the intention is to not fire the event on the client, would it be better to move this check next to the `ShouldFire.EXPLOSION_EVENT_DETONATE`?

gabizou: Part of the premise of not calling explode on the client is when Sponge is running on the client, it's expecting to only ever call this function on the server (which if you look at the usages of the injecting method, it is by default always called on the server game thread.

src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
```diff
@@ -118,158 +113,107 @@ public abstract class ExplosionMixin implements ExplosionBridge {
...
+    }
 
-        final float f3 = this.radius * 2.0F;
-        final int k1 = Mth.floor(this.x - (double) f3 - 1.0D);
-        final int l1 = Mth.floor(this.x + (double) f3 + 1.0D);
-        final int i2 = Mth.floor(this.y - (double) f3 - 1.0D);
-        final int i1 = Mth.floor(this.y + (double) f3 + 1.0D);
-        final int j2 = Mth.floor(this.z - (double) f3 - 1.0D);
-        final int j1 = Mth.floor(this.z + (double) f3 + 1.0D);
+    @ModifyConstant(method = "explode", constant = @Constant(intValue = 16, ordinal = 2))
```
gabizou: Why the second 16 constant and not the first?

src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
```diff
@@ -118,158 +113,107 @@ public abstract class ExplosionMixin implements ExplosionBridge {
...
-        // Sponge Start - Only query for entities if we're to damage them
-        final List<Entity> list = this.impl$shouldDamageEntities ? this.level.getEntities(this.source,
-                new AABB((double) k1, (double) i2, (double) j2, (double) l1, (double) i1, (double) j1)) : Collections.emptyList();
-        // Sponge End
+    @Redirect(
+            method = "explode", at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/world/level/Level;getEntities(Lnet/minecraft/world/entity/Entity;Lnet/minecraft/world/phys/AABB;)Ljava/util/List;"
+    ))
+    public List impl$explode_onlyDamageableEntities(Level instance, Entity entity, AABB aabb) {
```
gabizou: Generic parameter missing for the `List`. Also missing `final` on the parameters.

src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
```diff
@@ -118,158 +113,107 @@ public abstract class ExplosionMixin implements ExplosionBridge {
...
 
-        // Sponge Start - Only query for entities if we're to damage them
-        final List<Entity> list = this.impl$shouldDamageEntities ? this.level.getEntities(this.source,
-                new AABB((double) k1, (double) i2, (double) j2, (double) l1, (double) i1, (double) j1)) : Collections.emptyList();
-        // Sponge End
+    @Redirect(
+            method = "explode", at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/world/level/Level;getEntities(Lnet/minecraft/world/entity/Entity;Lnet/minecraft/world/phys/AABB;)Ljava/util/List;"
+    ))
```
gabizou:
```suggestion
    @Redirect(
        method = "explode",
        at = @At(
            value = "INVOKE",
            target = "Lnet/minecraft/world/level/Level;getEntities(Lnet/minecraft/world/entity/Entity;Lnet/minecraft/world/phys/AABB;)Ljava/util/List;"
        )
    )
```
Keep the spacing consistent.

src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
```diff
@@ -118,158 +113,107 @@ public abstract class ExplosionMixin implements ExplosionBridge {
...
+    }
 
+    @Inject(
+            method = "explode",
+            at = @At(
+                    value = "NEW",
+                    target = "net/minecraft/world/phys/Vec3",
+                    ordinal = 0
+            ),
+            cancellable = true
```
gabizou: 4 space indent, not 8.

src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
```diff
@@ -118,158 +113,107 @@ public abstract class ExplosionMixin implements ExplosionBridge {
...
-        }
+    }
+
+    @Redirect(
+            method = "explode",
+            at = @At(
+                    value = "INVOKE",
+                    target = "Lnet/minecraft/world/phys/Vec3;add(DDD)Lnet/minecraft/world/phys/Vec3;",
+                    ordinal = 0
+            )
```
gabizou: same about indents

src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
```diff
@@ -118,158 +113,107 @@ public abstract class ExplosionMixin implements ExplosionBridge {
...
+    )
+    public Vec3 impl$explode_applyKnockbackModifier(Vec3 instance, double param0, double param1, double param2) {
```
gabizou: Missing final on parameters

gabizou: Also the injection method should be private.

src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
```diff
@@ -118,158 +113,107 @@ public abstract class ExplosionMixin implements ExplosionBridge {
...
+        return instance.add(param0 * this.impl$knockback, param1 * this.impl$knockback, param2 * this.impl$knockback);
+    }
+
+    @Redirect(
+            method = "explode",
+            at = @At(value = "NEW", target = "net/minecraft/world/phys/Vec3", ordinal = 1)
+    )
+    public Vec3 impl$explode_recordKnockbackModifier(double param0, double param1, double param2) {
```
gabizou: final parameters and private method declaration. Give the parameters names more appropriate than just `param0` etc.

src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
```diff
@@ -118,158 +113,107 @@ public abstract class ExplosionMixin implements ExplosionBridge {
...
+                    ordinal = 0
+            )
+    )
+    public Vec3 impl$explode_applyKnockbackModifier(Vec3 instance, double param0, double param1, double param2) {
+        return instance.add(param0 * this.impl$knockback, param1 * this.impl$knockback, param2 * this.impl$knockback);
+    }
+
+    @Redirect(
+            method = "explode",
+            at = @At(value = "NEW", target = "net/minecraft/world/phys/Vec3", ordinal = 1)
```
gabizou: indentation of 4 not 8.

src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
```diff
@@ -118,158 +113,107 @@ public abstract class ExplosionMixin implements ExplosionBridge {
...
-     * @author gabizou
-     * @author zidane
-     * @reason Fire ExplosionEvent.Detonate
-     */
-    @Overwrite
-    public void explode() {
-
-        // Sponge Start - Do not run calculation logic on client thread
+    @Inject(method = "explode", at = @At("HEAD"), cancellable = true)
+    public void impl$explode_clientCheck(CallbackInfo ci) {
```
gabizou: private method declaration and final parameter declaration.

gabizou (changes requested):

src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
```diff
@@ -118,158 +113,109 @@ public abstract class ExplosionMixin implements ExplosionBridge {
...
+    }
 
-        final float f3 = this.radius * 2.0F;
-        final int k1 = Mth.floor(this.x - (double) f3 - 1.0D);
-        final int l1 = Mth.floor(this.x + (double) f3 + 1.0D);
-        final int i2 = Mth.floor(this.y - (double) f3 - 1.0D);
-        final int i1 = Mth.floor(this.y + (double) f3 + 1.0D);
-        final int j2 = Mth.floor(this.z - (double) f3 - 1.0D);
-        final int j1 = Mth.floor(this.z + (double) f3 + 1.0D);
+    @ModifyConstant(method = "explode", constant = @Constant(intValue = 16, ordinal = 1))
```
ImMorpheus, line 123: This is a bit dangerous, I'd add a slice to limit the location of the constant

src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
```diff
@@ -118,158 +113,109 @@ public abstract class ExplosionMixin implements ExplosionBridge {
...
+            ordinal = 0
+        )
+    )
+    private Vec3 impl$explode_applyKnockbackModifier(final Vec3 instance, final double motX, final double motY, final double motZ) {
+        return instance.add(motX * this.impl$knockback, motY * this.impl$knockback, motZ * this.impl$knockback);
+    }
+
+    @Redirect(
+        method = "explode",
+        at = @At(value = "NEW", target = "net/minecraft/world/phys/Vec3", ordinal = 1)
```
ImMorpheus, line 215: Use a slice here, `ordinal` is going to break down really quick during updates

src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
```diff
@@ -118,158 +113,109 @@ public abstract class ExplosionMixin implements ExplosionBridge {
...
-                }
-            }
-        }
+    }
+
+    @Redirect(
+        method = "explode",
+        at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/world/phys/Vec3;add(DDD)Lnet/minecraft/world/phys/Vec3;",
```
ImMorpheus, line 205: Same as below, use a slice

src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java
```diff
@@ -90,6 +84,7 @@ public abstract class ExplosionMixin implements ExplosionBridge {
     private int impl$resolution;
     private float impl$randomness;
     private double impl$knockback;
+    private List<Entity> impl$affectedEntities;
```
ImMorpheus, line 87: Add another injection at "RETURN" to clear this list. We need to make sure we're not leaking entities: if `ShouldFire.EXPLOSION_EVENT_DETONATE` is false the list is never cleared.
Keep in mind `ShouldFire` flags can change while the server is running (e.g. plugins adding/removing listeners).

ImMorpheus (changes requested):
Thanks for the contribution! I've left a few comments.

ImMorpheus:

___

PR #3749, opened 2022-08-20
Implement listener registration with provided lookup

A248:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2441) | **Sponge**

### DIFF
--- a/src/main/java/org/spongepowered/common/event/manager/SpongeEventManager.java
+++ b/src/main/java/org/spongepowered/common/event/manager/SpongeEventManager.java
@@ -325,8 +325,13 @@ public abstract class SpongeEventManager implements EventManager {
 
     @Override
     public SpongeEventManager registerListeners(final PluginContainer plugin, final Object listener,
-                                                final MethodHandles.Lookup lookup) {
+                                                MethodHandles.Lookup lookup) {
         Objects.requireNonNull(lookup, "lookup");
+        try {
+            lookup = MethodHandles.privateLookupIn(listener.getClass(), lookup);
+        } catch (IllegalAccessException ex) {
+            throw new IllegalArgumentException("Provider lookup lacks sufficient privilege", ex);
+        }
         this.registerListener(plugin, listener, lookup);
         return this;
     }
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/common/event/manager/SpongeEventManager.java
```diff
@@ -318,7 +319,15 @@ public <E extends Event> EventManager registerListener(final EventListenerRegist
...
     public SpongeEventManager registerListeners(final PluginContainer plugin, final Object listener) {
-        this.registerListener(plugin, listener);
+        this.registerListener(plugin, listener, null);
+        return this;
+    }
+
+    @Override
+    public SpongeEventManager registerListeners(final PluginContainer plugin, final Object listener,
+                                                final MethodHandles.Lookup lookup) {
+        Objects.requireNonNull(lookup, "lookup");
```
Zidane, line 329: I think you are missing a call to our internal lookup for private methods:

https://github.com/SpongePowered/Sponge/blob/1434c20ed1455db0a0df9039c96401c8a425947b/src/main/java/org/spongepowered/common/event/manager/ClassEventListenerFactory.java#L90

src/main/java/org/spongepowered/common/event/manager/SpongeEventManager.java
```diff
@@ -325,8 +325,13 @@ public SpongeEventManager registerListeners(final PluginContainer plugin, final
...
     @Override
     public SpongeEventManager registerListeners(final PluginContainer plugin, final Object listener,
-                                                final MethodHandles.Lookup lookup) {
+                                                MethodHandles.Lookup lookup) {
         Objects.requireNonNull(lookup, "lookup");
+        try {
+            lookup = MethodHandles.privateLookupIn(listener.getClass(), lookup);
+        } catch (IllegalAccessException ex) {
+            throw new IllegalArgumentException("Provider lookup lacks sufficient privilege", ex);
+        }
```
gabizou: I'm a fan of having final variables, I'd suggest not unfinalizing the one here, and providing a new variable for the `usedLookup` instead.

I'd also suggest to include a string message for the plugin's container information to help identify what plugin was causing the issue or supplying the inaccessible lookup.

zml2008 (approved):

PR #3754, opened 2022-08-26
MoveEntityEvent and RotateEntityEvent implementation improvements

Yeregorix:
This PR includes changes from #3612 and from #3751.

To summarize: 
- Fire events for vehicles.
- Avoid spamming events for small movements, based on a grid system.
- Notify the client when the destination is changed.
- Fix "moved too quickly" errors when the destination is changed.


Yeregorix:
~~I'm currently testing the vehicle move event. I can barely move when I'm riding a horse. PR is not ready yet.~~

PR #3755, opened 2022-08-27
Fix multiple missing slot transaction tracking + backport from 1.19

vectrixdevelops:
Fixes the reported cases in https://github.com/SpongePowered/Sponge/issues/3680.

Backports https://github.com/SpongePowered/Sponge/commit/6ebd3ea977f6c24ae44f5e11edca9d0e32051801. 

Includes:
- Captures player inventory transactions during an event (i.e mod adds an item to the players inventory on the forge left click event). i.e in the Storage Drawers mod where the player will interact with the block to take an item out of it (without using a GUI).
- Captures slot transactions during a command correctly, using the example provided here https://github.com/SpongePowered/Sponge/issues/3680#issuecomment-1220822916.
- Captures inventory open and slot transactions during an open gui, this fixes cases where a mod will add items to the container as it is opening. i.e in the Storage Drawers mod where the player will shift + interact to open the GUI.
- Fixes the slot transaction tracking to include what happens during the call to `useItem`. i.e interacting with an empty map to create a map.

### DIFF
--- a/forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/fml/NetworkHooksMixin_Forge.java
+++ b/forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/fml/NetworkHooksMixin_Forge.java
@@ -1,0 +1,56 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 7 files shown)


forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/fml/NetworkHooksMixin_Forge.java
```diff
@@ -0,0 +1,52 @@
...
+public abstract class NetworkHooksMixin_Forge {
+    @Redirect(
+        method = "openGui(Lnet/minecraft/server/level/ServerPlayer;Lnet/minecraft/world/MenuProvider;Ljava/util/function/Consumer;)V",
+        at = @At(
+                value = "INVOKE",
+                target = "Lnet/minecraft/world/inventory/AbstractContainerMenu;addSlotListener(Lnet/minecraft/world/inventory/ContainerListener;)V"
+        )
+    )
+    private static void forge$openGui(AbstractContainerMenu instance, ContainerListener containerListener) {
+        try (final PhaseContext<?> context = BlockPhase.State.RESTORING_BLOCKS.createPhaseContext(PhaseTracker.SERVER).source(containerListener);) {
```
Faithcaio: Maybe add some comments explaining that this effectively ignores any changes during `addSlotListener` for our events. (in case a mod does something on open)

Faithcaio (approved):
Have not tested, but looks good

PR #3763, opened 2022-09-29
Implement translations for item name and lore

Yeregorix:
Translate the components directly on item nbt tag right before sending to the client.
Undo the modifications when receiving an item from the client.
The nbt tag is not modified if there is nothing to translate.

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/core/network/FriendlyByteBufMixin.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/core/network/FriendlyByteBufMixin.java
@@ -24,17 +24,23 @@
  */
 package org.spongepowered.common.mixin.core.network;
 
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.nbt.ListTag;
+import net.minecraft.nbt.StringTag;
+import net.minecraft.world.item.ItemStack;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.spongepowered.api.util.locale.Locales;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.ModifyVariable;
+import org.spongepowered.asm.mixin.injection.Redirect;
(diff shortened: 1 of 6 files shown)


SawFowl:
I would like to see similar functionality for entities.

zml2008:
@SawFowl what functionality doesn't work on entities? most areas of the server support custom translations already, inventory stuff was just an exception

SawFowl:
> @SawFowl what functionality doesn't work on entities? most areas of the server support custom translations already, inventory stuff was just an exception

Can I have an example of a translated entity name? I tried to create a villager with a translatable name.  I couldn't do it.

Yeregorix:
I just tried setting a translation as an entity name and it works fine. The only problem is that entity names are not updated when the player change his language. If you need an example to setup server-side translations, look at the ChatTest plugin.

src/mixins/java/org/spongepowered/common/mixin/core/network/FriendlyByteBufMixin.java
```diff
@@ -43,7 +49,87 @@ public abstract class FriendlyByteBufMixin implements FriendlyByteBufBridge {
...
     @ModifyVariable(method = "writeComponent", at = @At("HEAD"), argsOnly = true)
     private Component localizeComponent(final Component input) {
-        return NativeComponentRenderer.apply(input, this.impl$locale == null ? Locales.EN_US : this.impl$locale);
+        return NativeComponentRenderer.apply(input, this.impl$locale == null ? Locales.DEFAULT : this.impl$locale);
+    }
+
+    @Redirect(method = "writeItem", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/item/ItemStack;getTag()Lnet/minecraft/nbt/CompoundTag;"))
+    public CompoundTag renderItemComponents(final ItemStack stack) {
+        CompoundTag tag = stack.getTag();
+        if (tag != null && tag.contains(Constants.Item.ITEM_DISPLAY, 10)) {
```
Zidane: To remove branching, might be worth checking for the inverse and returning the tag right away.

PR #3768, opened 2022-10-05
Add FluidState to LocatableBlock

gabizou:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2443) | **Sponge**


ImMorpheus (dismissed):
LGTM! Make sure the linter does not complain about the spaces.

@gabizou Same as the API: please target this to api-10 and let me know if you need someone to take this up

PR #3774, opened 2022-10-12
Fix forwarded profile being unused when using legacy forwarding

derklaro:
The ServerLoginPacketListenerImpl#handleHello method added a check if the current server is running in singleplayer and assigns the player profile to the "singleplayer profile" which is set in the minecraft server instance. The current mixin therefore only works if running in singleplayer. The simple fix is to move the mixin one down to the second assignment which is executed when the server runs in multiplayer:

```java
...
    GameProfile $$1 = this.server.getSingleplayerProfile();
    if ($$1 != null && $$0.name().equalsIgnoreCase($$1.getName())) {
      this.gameProfile = $$1; // current inject target
      this.state = ServerLoginPacketListenerImpl.State.READY_TO_ACCEPT;
    } else {
      this.gameProfile = new GameProfile((UUID)null, $$0.name());
...
```


Zidane:
Is this applicable to api-8/1.16.5? If so, could you target that instead?

Thanks for helping out regardless! Appreciated.

derklaro:
This extra field assign was added in 1.19, so it doesn't apply to 1.16/1.18 :)

PR #3783, opened 2022-10-23
Fix spotless check on windows

vectrixdevelops:
Ensures spotless is looking for the correct line endings on windows.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -304,6 +304,7 @@ allprojects {
             indentWithSpaces(4)
             trimTrailingWhitespace()
             importOrderFile(rootProject.file("SpongeAPI/extra/eclipse/sponge_eclipse.importorder"))
+            targetExclude("build/generated/**/*") // exclude generated content
         }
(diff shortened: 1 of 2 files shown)


zml2008 (commented):
Thanks for poking at this!

 A few things:

1. Can we just unilaterally exclude generated contents from spotless checks? It's not really worth the time to fight generators for this sort of thing.
2. Can you do a `gitattributes` line of `*.gradle.kts eol=lf` rather than just the specific files, to avoid duplication

vectrixdevelops:
@zml2008 Thanks for the review. I've made some changes for what you've suggested.

1. Is the regex pattern I got good, or would I need to be more specific?
2. Doing `*.gradle.kts` didn't work as `build.gradle.kts` for example is supposed to be crlf on windows. Since the `spongeforge.settings.gradle.kts` is created, I removed the manually specified `\n` and made it add the seperator depending on the platform. The other gradle files are not affected as they're tracked and changed by git.

zml2008 (approved):

PR #3798, opened 2022-12-08
Filter cancelled events even when the declared class is not cancellable

Yeregorix:
Currently, the IsCancelled filter isn't generated when the event parameter declared by the listener is not cancellable. The problem is that subclasses may be cancellable. This PR fixes this. The IsCancelled annotation can now be used on any event. A `event instanceof Cancellable` check is added into the generated bytecode when needed.


zml2008 (approved):
Sorry for the delay in reviewing this!

PR #3806, opened 2023-01-07
Remove the internal usage of optional from Registry#value

aromaa:
Removes the usage of Optional from the `Registry#value` in preference to just have a null check. This method can actually show up in profiler when hot methods end up calling `DefaultedRegistryReference#get` multiple times in a single tick.

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/api/minecraft/core/RegistryMixin_API.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/api/minecraft/core/RegistryMixin_API.java
@@ -52,6 +52,7 @@ public abstract class RegistryMixin_API<T> implements Registry<T> {
     @Shadow public abstract net.minecraft.resources.ResourceKey<? extends net.minecraft.core.Registry<T>> shadow$key();
     @Shadow @Nullable public abstract ResourceLocation shadow$getKey(T p_177774_1_);
     @Shadow public abstract Optional<T> shadow$getOptional(@Nullable net.minecraft.resources.ResourceLocation param0);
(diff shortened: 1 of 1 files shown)


src/mixins/java/org/spongepowered/common/mixin/api/minecraft/core/RegistryMixin_API.java
```diff
@@ -91,8 +92,13 @@ public <V extends T> Optional<V> findValue(final ResourceKey key) {
 
     @Override
     public <V extends T> V value(final ResourceKey key) {
-        return (V) this.shadow$getOptional((ResourceLocation) (Object) Objects.requireNonNull(key, "key"))
-            .orElseThrow(() -> new ValueNotFoundException(String.format("No value was found for key '%s'!", key)));
+        V value = (V) this.shadow$get((ResourceLocation) (Object) Objects.requireNonNull(key, "key"));
+        if (value != null)
+        {
```
zml2008:
```suggestion
        final V value = (V) this.shadow$get((ResourceLocation) (Object) Objects.requireNonNull(key, "key"));
        if (value != null) {
```

zml2008 (changes requested):

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/core/RegistryMixin_API.java
aromaa: Oops. Habits :)

zml2008 (approved):

PR #3813, opened 2023-01-11
Fix block entity memory leak (API 9)

Lignium:
Adaptation of my early PR #3689 for api-8. In version 1.18.2, the associated code has changed slightly, leaving only one field with a collection of ticking block entities. The code is tested, the leak is gone.

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerLevelMixin.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerLevelMixin.java
@@ -541,7 +541,7 @@ public abstract class ServerLevelMixin extends LevelMixin implements ServerLevel
          * is without players.
          */
         if (this.emptyTime >= 300 && !this.blockEntityTickers.isEmpty()) {
-            blockEntityTickers.removeIf(TickingBlockEntity::isRemoved);
+            this.blockEntityTickers.removeIf(TickingBlockEntity::isRemoved);
         }
     }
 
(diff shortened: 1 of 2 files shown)


aromaa:
This can only happen when a plugin manipulates the chunks while the world is seen as empty? Which would mean this can't be reported to Mojang directly to get this fixed? Could you add more details to the comment so this is not overlooked when porting to the newer versions of Minecraft.

Lignium:
@aromaa Quite right. I have added some comments.

src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerLevelMixin.java
```diff
@@ -520,6 +522,29 @@ public void save(@Nullable final ProgressListener progress, final boolean flush,
...
+        /*
+         * This code fixes block entity memory leak when the level hasn't online players
+         * and forced chunks. For the first 300 ticks the level can still clean up removed
+         * block entities on its own (ticks are performed for block entities). After it
+         * this mixin code is responsible for the subsequent unloading of block entities.
+         * Such a memory leak occurs when a plugin writes a lot of blocks, but the level
+         * is without players.
+         */
+        if (this.emptyTime >= 300 && !this.blockEntityTickers.isEmpty()) {
+            blockEntityTickers.removeIf(TickingBlockEntity::isRemoved);
```
aromaa:
```suggestion
            this.blockEntityTickers.removeIf(TickingBlockEntity::isRemoved);
```

aromaa (changes requested):
Otherwise LGTM

aromaa (approved):
Thank you!

PR #3815, opened 2023-01-13
[API-8] Improve unit tests

Yeregorix:
This PR adds the ability to run JUnit 5 tests in a ModLauncher context where Sponge and a DedicatedServer is initialized (but not ticking). This is done using [modlauncher-injector-junit](https://github.com/SpongePowered/modlauncher-injector-junit/) the successor of [LaunchWrapperTestSuite](https://github.com/SpongePowered/LaunchWrapperTestSuite/).

I also added JaCoCo so we can measure code coverage. This should help us improve the test suite in the future.

Some of the launch logic is shared between SpongeVanilla and Sponge.test. ~~I moved a few parts (as few as possible) of SpongeVanilla.applaunch and SpongeVanilla.launch into Sponge to avoid duplicating the code.~~ Sponge.test is depending on SpongeVanilla.launch and SpongeVanilla.applaunch-base

The only issues identified so far with this setup are:
- ~~JaCoCo ignores classes loaded by ModLauncher, so the measured coverage is much lower than expected.~~ Fixed.
- `@ParameterizedTest` cannot be executed in a ModLauncher context. However `@TestFactory` is a wonderful alternative and is working perfectly.
- All unit tests are running in the same ModLauncher context. Any value registered into Sponge by a test should be unregistered after the test so it doesn't interfere with other tests. In the future, using a custom class loader might be a solution to create multiple isolated ModLauncher instances. It's not planned at the moment, because we don't need it.

I have updated ~~almost~~ all tests in the `invalid/` directory.
~~The only test that I have not updated is the `TileEntityRegistrationTest`. I don't know how to update it and I'm not sure the test is relevant anymore.~~

Some tests are failing and need fixes but this it out of the scope of this PR, so these tests are currently `@Disabled`. Anyone willing to fix the implementation to make pass these tests is welcome.


zml2008:
I'd like to keep the modlauncher workings outside of SpongeCommon -- maybe stick them into a separate subproject that's a testImplementation dep of common and the appropriate configurations in SV

Yeregorix:
Would a new sourceset in SpongeCommon be enough or do you really want a subproject?
Also, another approach would be to put the code back in SpongeVanilla and then import the applaunch and launch sourcesets into Sponge.test.

Yeregorix:
Jacoco is instrumenting classes at runtime which is conflicting with Mixin. The consequence is that classes loaded by Modlauncher are ignored by Jacoco, resulting in a coverage way below the reality.

I have managed to fix this by using [Jacoco offline instrumentation](https://www.jacoco.org/jacoco/trunk/doc/offline.html) which is not supported by the official gradle plugin. While it is possible to have a workaround in SpongeCommon kts, I have preferred to implement this in a plugin, so that it can be used in the future by SpongeVanilla, SpongeForge and hybrid plugins as well. I wrote a PR for gradle https://github.com/gradle/gradle/pull/23656 so that it can officially supports offline instrumentation. It's a fairly requested feature, so I hope it finds its way. In the mean time (and forever if it doesn't get merged), I will try to a publish the fork as a standalone plugin.

Yeregorix:
I have forked the official Jacoco plugin into https://github.com/Yeregorix/jacoco-offline and published both to Maven central and Gradle plugin portal.

PR #3817, opened 2023-01-18
Fix Error related to Banned users with Expiration dates

### LINKED ISSUE #3788: Error related to banned users when given an expiration date
### Affected Product(s)

SpongeVanilla

### Version

1.16.5-8.1.0-RC1235

### Operating System

Windows 11

### Java Version

JVM: 17/64-bit (Eclipse Adoptium)

### Plugins/Mods

```shell
Luckperms
Strax (my plugin)
```

### Describe the bug

I'm banning myself using the ban service, and the entry gets added to banned-players.json
<https://gist.github.com/DrZoddiak/a111816e411ff9ab1eb8eea051ab1811>

[When I join the server]
However, once this entry is added I get an error that the object can't be formatted to a Date
<https://gist.github.com/DrZoddiak/757300994c14abe421e11eeb0a211f31>

Then once the expiration date has passed I get this error instead
<https://gist.github.com/DrZoddiak/b3033a177cf0e45ec0f71a2b37491bc2>

If the entry is removed everything is normal.

### Link to logs

_No response_

DrZoddiak:
Fixes <https://github.com/SpongePowered/Sponge/issues/3788>

The issue above was actually 2 separate issues.
First: BAN_DATE_FORMAT#format required a Date instead of an Instant, causing the first error.

~~Second: StoredUserListMixin was actually causing the ClassCastException because the List it redirects is actually a list of GameProfiles and not of Strings, which it later uses to remove an entry from a map wherein the Key is a GameProfile~~


ImMorpheus:
@DrZoddiak Thank you for your contribution!

Changes look good but it would be best to create separate PRs since they are two different issues.

> First: BAN_DATE_FORMAT#format required a Date instead of an Instant, causing the first error.

Note for reviewers: there's an overloaded method for Object but it does an instance check under the hood and throw an exception if the object is not a Date or an Number

PR #3820, opened 2023-01-21
Add movement threshold configuration

vectrixdevelops:
Replaces the existing enable/disable movement checks with a customizable threshold. Using negative numbers will disable the movement checks still if desired.

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/movementcheck/server/network/ServerGamePacketListenerImplMixin_MovementCheck.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/movementcheck/server/network/ServerGamePacketListenerImplMixin_MovementCheck.java
@@ -31,7 +31,6 @@ import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Constant;
 import org.spongepowered.asm.mixin.injection.ModifyConstant;
-import org.spongepowered.asm.mixin.injection.Redirect;
 import org.spongepowered.asm.mixin.injection.Slice;
 import org.spongepowered.common.config.SpongeGameConfigs;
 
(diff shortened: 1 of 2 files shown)


ImMorpheus (commented):
Code looks fine! Not sure if changing existing settings should go into a new major version. cc @Zidane thoughts ?

src/mixins/java/org/spongepowered/common/mixin/movementcheck/server/network/ServerGamePacketListenerImplMixin_MovementCheck.java
```diff
@@ -40,53 +39,59 @@ public abstract class ServerGamePacketListenerImplMixin_MovementCheck {
...
-            to = @At(value = "INVOKE", target = "Lnet/minecraft/server/network/ServerGamePacketListenerImpl;isSingleplayerOwner()Z", ordinal = 0))
+            from = @At(value = "INVOKE", target = "Lnet/minecraft/world/entity/Entity;move(Lnet/minecraft/world/entity/MoverType;Lnet/minecraft/world/phys/Vec3;)V", ordinal = 0),
+            to = @At(value = "INVOKE", target = "Lorg/apache/logging/log4j/Logger;warn(Ljava/lang/String;Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;)V")
+        )
     )
-    private double movementCheck$onVehicleMovedTooQuicklyCheck(final double val) {
-        if (SpongeGameConfigs.getForWorld(this.player.level).get().movementChecks.player.vehicleMovedTooQuickly) {
-            return val;
+    private double movementCheck$onVehicleMovedWronglyCheck(final double value) {
+        final double threshold = SpongeGameConfigs.getForWorld(this.player.level).get().movementChecks.movedWronglyThreshold;
```
aromaa: This is using the player's one instead of the vehicle's. You forgot to introduce `vehicle-moved-wrongly-threshold`

aromaa (changes requested):
As mentioned earlier, this most likely should at least respect the old configuration knobs for api-8. Or try to migrate them to the new ones.

vectrixdevelops:
Good point. I'll add back the old configuration knobs, where if they are false it will act as though the threshold is negative (disable the check), otherwise if it's true it will enable and use the configured threshold (or the default).

src/mixins/java/org/spongepowered/common/mixin/movementcheck/server/network/ServerGamePacketListenerImplMixin_MovementCheck.java
vectrixdevelops: A previous configuration knob did not exist seperating player and vehicle and Paper/Spigot doesn't have a seperate one for each either I think. If you still think this is useful though I can add it.

aromaa: I think it would be useful. It doesn't hide the fact that the same knob is used for both vehicle and players as the other one is separated like that. Also considering players and vehicles have different physics.

PR #3824, opened 2023-01-29
Fix login NPE with fluid states

ImMorpheus:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2452) | **Sponge**

Closes https://github.com/SpongePowered/Sponge/issues/3823


LordRalex:
As a note, is it outside the scope that https://github.com/SpongePowered/Sponge/blob/6407bf9b13e606ee9329d23d7bac04bd826b1ed0/src/main/java/org/spongepowered/common/world/server/SpongeLocatableBlock.java#L196-L198 is not updated to also check the FluidState equality as they can be outside the reference the block state has?

ImMorpheus:
Superseded by https://github.com/SpongePowered/Sponge/commit/0f73bdd845f661538b2d7c68d0d67b73c3615ec1

PR #3827, opened 2023-02-04
Add XP Orb Merge Optimization

Ve6om:
This Pull Request is focused on adding the XP Orb merging feature to Sponge 7. This will aid in optimizing the game performance by reducing the amount of EntityXPOrbs on the ground, thereby reducing the EntityXPOrbs tick time.

The PR implements two settings: xp-merge-radius and xp-orb-merge. The xp-merge-radius setting determines the distance at which nearby EntityXPOrbs will merge into a single entity. The xp-orb-merge setting, when set to true, enables the merging feature.

This feature will benefit not only the server performance but also the client performance for players in scenarios where there are numerous EntityXPOrbs on the ground. With the merging feature, the amount of entities being rendered will be reduced, improving the overall player experience.

EDIT:
Sorry for the old PR commits below, not sure why those are there lol.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,7 +5,7 @@ url=https://www.spongepowered.org
 organization=SpongePowered
 
 minecraftVersion=1.12.2
-recommendedVersion=6-SNAPSHOT
+recommendedVersion=8-SNAPSHOT
 mcpMappings=snapshot_20180808
 
 # We use VanillaGradle instead of an official ForgeGradle plugin.
--- a/src/main/java/org/spongepowered/common/config/category/ExploitCategory.java
+++ b/src/main/java/org/spongepowered/common/config/category/ExploitCategory.java
@@ -92,7 +92,7 @@ public class ExploitCategory extends ConfigCategory {
     private boolean checkUnknownSlotDuringWindowClicks = false;
 
     @Setting(value = "check-slot-ids-during-window-clicks",
-            comment = "If enabled, window click with slot ids that are higher then the inventory size will be canceled.")
+            comment = "If enabled, window clicks with slot ids that are higher then the inventory size will be canceled.")
     private boolean checkSlotID = false;
 
     @Setting(value = "check-slot-size-during-window-clicks",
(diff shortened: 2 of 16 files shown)


Yeregorix:
Modifying the world from inside the constructor is a problem. Imagine a plugin that instantiates an orb but never spawns it. All nearby orbs in the world are removed, but the new merged one is never spawned. I think the merge logic should be moved to either on entity tick, either on entity spawn.

I'm personally against checking on each tick, this would add a lot of unnecessary processing, which may counter the potential performance gain. Also, for farms, this would continuously merge all incoming XP into one big orb, which may not be not desirable for all players. I'm sure some people love XP showers.

I think the correct way to do it would be to merge orbs on spawn. This would keep a good balance. At this point, this optimization can be implemented by a small plugin using just API and I don't think it needs to be added into Sponge directly.

Ve6om:
@Yeregorix Yeah I do agree putting it under the entity spawn would be better. Why not include it in Sponge? There's already optimizations added that could be done through a plugin. I made this PR for the general community out there. Making more plugins for each performance issue is not the solution.

Yeregorix:
We used to put optimizations directly into Sponge but this has increased the maintenance cost and the risk of incompatibility with other mods on Forge, especially optimization mods which are often bundled by default in most packs.

Sponge should only be the base. Functionalities and optimizations should be reserved to plugins. Exception made for some optimizations that are heavily depending on Sponge internal code.

This is my own opinion and doesn't engage the Sponge team. I'm curious to see what other people might think of it.

ImMorpheus:
Hey! You seem to be running an unsupported version (API 7) of SpongeAPI. Support for API 7 ended in January 2022.

API-7 is now end-of-life. We will not be issuing any more API updates for version 7 of our API[^1]

As part of our legacy cleanup we are closing issues relating to API7/1.12.2[^2]
___

I have to reiterate, API 7 is not supported and EOL. If you wish to move this forward you need to reopen this and target the current branch.

[^1]: https://forums.spongepowered.org/t/status-update-8th-january-2022-spongeapi-8-released/39861#minecraft-1122-spongeapi-7-10
[^2]: https://forums.spongepowered.org/t/sponge-status-update-28th-may-2022/40135#api-71122-community-support-8

PR #3831, opened 2023-02-27
Prevent invalid status responses with empty/dummy player UUIDs

stephan-gh:
Right now `SpongeGameProfile.toMcProfile()` converts all SpongeAPI `GameProfile`s with **empty** UUIDs (`= new UUID(0, 0)`) to authlib `GameProfile`s with a **null** UUID.

This is fine in most cases but problematic for the sample/player list in the status response. Empty UUIDs can be used there to create dummy entries, to display custom messages in the server list instead of the online players. (See e.g. https://github.com/Minecrell/ServerListPlus/wiki/Status-Configuration#player-hover-messages)

The Minecraft client completely rejects status responses that contain GameProfiles with null UUIDs and/or null names. This is why it is important that the empty UUID is not converted to a null UUID.

`[Client thread/ERROR]: Can't ping localhost: Internal Exception: io.netty.handler.codec.DecoderException: java.lang.IllegalArgumentException: Invalid UUID string: `

Prevent sending invalid responses to the client by making sure both UUID and name are non-null in the player samples of the status response.

This fixes a regression in API 8 when using the ServerListPlus plugin on Sponge. The problem was likely introduced in some `GameProfile` refactoring (API 7 worked fine).

There is similar code in Paper to avoid this problem: https://github.com/PaperMC/Paper/blob/ad3e1bc12104f1f8ea8bcf9248a606133bfc7780/patches/server/0182-Implement-extended-PaperServerListPingEvent.patch#L122-L128

I would appreciate if you could pick/merge this into all newer branches as well, not sure how Sponge development works nowadays :)

### DIFF
--- a/src/main/java/org/spongepowered/common/profile/SpongeGameProfile.java
+++ b/src/main/java/org/spongepowered/common/profile/SpongeGameProfile.java
@@ -25,6 +25,7 @@
 package org.spongepowered.common.profile;
 
 import com.google.common.base.Charsets;
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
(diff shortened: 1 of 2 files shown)


ImMorpheus:
Thanks!

I've merged your changes in the other branches but afaics api 10 doesn't seem to be affected. The mojang->sponge->mojang gameprofile conversion is gone.

PR #3832, opened 2023-03-06
removed unnecessary read lock in SpongeScheduler

sunmisc:
synchronization block (read-only) is not needed for ConcurrentHashMap

to all this, the lock there has no meaning on reads, without a write lock (it is not there)

a very unpleasant bug can also turn out, this is when the uuid of the prng tasks will match, then the previous task will simply be removed from the map) this is a very rare case, but in theory it can happen

might be worth looking at the existing api - java.util.concurrent.ScheduledExecutorService


ImMorpheus:
Thanks!

Issue #3835, opened 2023-04-02
Problems with Create mod


Inscrutable:
### Affected Product(s)

SpongeForge

### Version

SpongeForge 1.16.5-RC1296

### Operating System

Win10

### Java Version

Oracle 1.8.0_361-b09

### Plugins/Mods

```shell
create-mc1.16.5_v0.3.2g.jar
flywheel-1.16-0.2.5
```

### Describe the bug

Rotational power is not correctly transferred in many cases. The connection between shafts and gearboxes fails vertically when they are placed. Bashing most things with a wrench seems to liven them up, but it's not intended behavior (i.e. it doesn't happen without SpongeForge present). 
Gearshifts (reversing rotational; direction) are completely insensitive to switches, and this really breaks things. I can't get the steam turbine to work either. Basically, something's broken with the way it updates the transmitted power in Create blocks. 
There was also an issue when placing cogs; Create shows a preview for placing them when you aim at another cog. This feature is broken, and the cogs end up misplaced. 
No relevant console errors were seen. Just a few ignored slot transactions when I got sawed in half.
On the bright side, water wheels and windmills definitely work now. Even the stress gauge yields data.

Server log file: [latest.log](https://github.com/SpongePowered/Sponge/files/11131327/latest.log)

Inscrutable:
OK, the guts of this issue have been fixed in PR #3838. The only thing left is the weird cog positioning that results from using Create's "Placement Assist". 
Addendum: During testing I also noted that interaction with Create's wireless redstone links is broken. Normal behavior allows them to be visibly tagged with 2 items (to make a unique channel). They can't be tagged at all at the moment, so only the default wireless channel works.

Yeregorix:
Fixed in #3844 and #3838

PR #3838, opened 2023-04-11
Fix issues with modded tile entities and lights

Yeregorix:
Forge adds specific methods to check whether a block has a tile entity or is emitting light. We need to use these methods instead of the vanilla ones to properly support modded blocks. Previously this was done correctly in API 7, but got lost in the update process.

This fixes issues with Create's gearbox and gearshit (#3835) and probably issues for many other mods as well.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/context/transaction/effect/UpdateLightSideEffect.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/context/transaction/effect/UpdateLightSideEffect.java
@@ -26,6 +26,7 @@ package org.spongepowered.common.event.tracking.context.transaction.effect;
 
 import net.minecraft.server.level.ServerLevel;
 import net.minecraft.world.level.block.state.BlockState;
+import org.spongepowered.common.bridge.world.level.block.state.BlockStateBridge;
 import org.spongepowered.common.event.tracking.context.transaction.pipeline.BlockPipeline;
 import org.spongepowered.common.event.tracking.context.transaction.pipeline.PipelineCursor;
 import org.spongepowered.common.world.SpongeBlockChangeFlag;
@@ -51,7 +52,6 @@ public final class UpdateLightSideEffect implements ProcessingSideEffect {
         if (!flag.updateLighting()) {
             return EffectResult.NULL_PASS;
         }
-        final int originalOpactiy = oldState.opacity;
         final ServerLevel serverWorld = pipeline.getServerWorld();
         final BlockState currentState = pipeline.getAffectedChunk().getBlockState(oldState.pos);
         // local variable notes:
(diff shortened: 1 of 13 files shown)


gabizou (approved):
LGTM

Inscrutable:
This fixes all the mechanical issues I had with Create vs SpongeForge. Tested on a server with forge-1.16.5-36.2.39. Nice work!
The Placement Assist for cogs is still wacky, so that's gotta be down to something else.

src/main/java/org/spongepowered/common/event/tracking/context/transaction/effect/UpdateLightSideEffect.java
```diff
@@ -68,8 +68,8 @@ public EffectResult processSideEffect(
         //         )
         //     ) {
         if (oldState.state != currentState
-            && (currentState.getLightBlock(serverWorld, oldState.pos) != originalOpactiy
-            || currentState.getLightEmission() != oldState.state.getLightEmission()
+            && (currentState.getLightBlock(serverWorld, oldState.pos) != oldState.opacity
+            || ((BlockStateBridge) currentState).bridge$getLightValue(serverWorld, oldState.pos) != oldState.light
```
ImMorpheus, line 72: This doesn't look the same as the old code.

For vanilla this will call `getLightBlock` instead of `getLightEmission`. I think that's not intended.

ImMorpheus (changes requested):

src/main/java/org/spongepowered/common/event/tracking/context/transaction/effect/UpdateLightSideEffect.java
Yeregorix: `BlockStateMixin#bridge$getLightValue` is calling `BlockStateBase#getLightEmission`
`BlockStateMixin_Forge#bridge$getLightValue` is calling `IForgeBlockState#getLightValue`
I think this is correct.

ImMorpheus: Oh my bad, I was looking at api-10. Current code is correct for api-8, the issue is api-10.

See https://github.com/SpongePowered/Sponge/blob/237e0b803d5faca0d76be66dd04dae9524fd7c7f/src/mixins/java/org/spongepowered/common/mixin/core/world/level/block/state/BlockStateMixin.java#L65

The method changed in api-10 and this will change vanilla behaviour.

Yeregorix:
I just amended the last commit because a forgot two `final` keywords.

ImMorpheus (approved):
LGTM! I fixed the issue with api-10 in another commit

PR #3839, opened 2023-04-12
Fixes for Forge Worldedit

carbotaniuman:
I've taken the code from #3711 and reworked some of it, including the exact minutiae of the logging for direct block access here. I've also added `PlayerInteractEvent_RightClickBlockMixin_Forge`, which I'm very unsure of how much I need to sync, but it seems that many things work fine enough without syncing.

There's also a change in `ForgeCommandManager` to get WE to recognize the Forge `CommandEvent`s without the first character being cut off for no reason.

### DIFF
--- a/forge/src/launch/java/org/spongepowered/forge/launch/command/ForgeCommandManager.java
+++ b/forge/src/launch/java/org/spongepowered/forge/launch/command/ForgeCommandManager.java
@@ -28,6 +28,7 @@ import com.google.common.base.Throwables;
 import com.google.inject.Inject;
 import com.google.inject.Provider;
 import com.mojang.brigadier.ParseResults;
+import com.mojang.brigadier.StringReader;
 import com.mojang.brigadier.context.CommandContextBuilder;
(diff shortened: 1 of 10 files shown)


aromaa:
There are still few open questions and some investigation needs to be done before we want to go with the route to allow direct block access and which is why the #3711 has not been merged yet.

Also more investigation needs to be done on why we are stripped the slash from the commands which was expressed in [#3713](https://github.com/SpongePowered/Sponge/pull/3713/files#r907249213)

ImMorpheus:
Thank you for your contribution!

Unfortunately as @aromaa pointed out, there's more discussion/investigation that needs to happen before this is merged. I'll close this to keep the discussion in https://github.com/SpongePowered/Sponge/pull/3711 which seems to be the older one.

A small note on the command changes since they are not addressed in the other PR:

The changes you made for the command slash were already proposed here (https://github.com/SpongePowered/Sponge/pull/3713#discussion_r907249213) and rejected because they are not addressing the underlying issue.

PR #3840, opened 2023-04-12
[api-10] Prevent invalid status responses with empty/dummy player UUIDs

stephan-gh:
Port of #3831 for api-10. The fix is still needed, the code just moved to a different place. (cc @ImMorpheus)


ImMorpheus (approved):
*sigh* something something too tired

LGTM!

PR #3841, opened 2023-04-14
Expose protocol version for status pings

stephan-gh:
See https://github.com/SpongePowered/SpongeAPI/pull/2456

Also includes some minor fixes for the status handling after the refactoring for api-10, see first commit.


src/main/java/org/spongepowered/common/SpongeMinecraftVersion.java
```diff
@@ -30,7 +30,7 @@
 
 import java.util.OptionalInt;
 
-public final class SpongeMinecraftVersion implements ProtocolMinecraftVersion, MinecraftVersion {
```
ImMorpheus, line 33: Why remove the final modifier ?

stephan-gh: I was considering reusing it as base class for `SpongeStatusResponse.SpongeStatusVersion` but that didn't work out and I just forgot to add it back. Fixed, thanks!

ImMorpheus (approved):

PR #3844, opened 2023-04-15
Fix forge interactions

Yeregorix:
This PR fixes several issues with Forge interactions:
1. `PlayerInteractEvent.RightClickBlock` was never fired.
2. `onItemUseFirst` and `doesSneakBypassUse` were never called.
3. Before firing `InteractItemEvent.Primary` we must ensure a block or an entity is not reachable to ensure it's indeed an item interaction. The reach was invalid, ignoring both `ForgeMod#REACH_DISTANCE` attribute and the gamemode.

~~For 1. , the approach here is different from what gabizou did in #3711. Instead of creating a custom `EventContextKey` to compensate the fact the Sponge event doesn't contain as much information as the Forge event. I decided to first create the Forge event first, then create the Sponge event from it. Also, I've made a few modifications to the dual bus system because I need to read the result of both events to properly support Forge `PlayerInteractEvent#setCancellationResult`.~~

All other interactions seem to be working properly but are not using the dual bus.


forge/src/main/java/org/spongepowered/forge/hook/ForgeEventHooks.java
```diff
@@ -25,19 +25,48 @@
...
+import org.spongepowered.api.util.Tuple;
+import org.spongepowered.common.event.tracking.PhaseTracker;
 import org.spongepowered.common.hooks.EventHooks;
+import org.spongepowered.forge.launch.bridge.event.ForgeEventBridge_Forge;
+import org.spongepowered.forge.launch.event.ForgeEventManager;
 
 public final class ForgeEventHooks implements EventHooks {
 
+    @Override
+    public Tuple<InteractBlockEvent.Secondary, InteractionResult> callInteractBlockEventSecondary(
```
Yeregorix: As discussed on discord, this hook can be removed if we add the "cancellation result" to the API.

forge/src/mixins/java/org/spongepowered/forge/mixin/core/server/level/ServerPlayerMixin_Forge.java
```diff
@@ -41,8 +43,13 @@
...
 import org.spongepowered.forge.mixin.core.world.entity.LivingEntityMixin_Forge;
 
-@Mixin(ServerPlayer.class)
-public abstract class ServerPlayerMixin_Forge extends LivingEntityMixin_Forge {
+@Mixin(value = ServerPlayer.class, priority = 1300)
+public abstract class ServerPlayerMixin_Forge extends LivingEntityMixin_Forge implements ServerPlayerBridge {
+
+    @Override
+    public double bridge$reachDistance() {
+        return shadow$getAttribute(ForgeMod.REACH_DISTANCE.get()).getValue();
```
aromaa:
```suggestion
        return this.shadow$getAttribute(ForgeMod.REACH_DISTANCE.get()).getValue();
```

aromaa (approved):

gabizou (approved):

PR #3849, opened 2023-04-23
Remove unused or broken config options

ImMorpheus:
Keeping unused or non-working options in the config file is confusing.

This PR prunes unused/broken config options for now.


zml2008:
Thanks for doing this cleanup pass!

For config options that are gone for good, we should add a version migration to clean them up from end user config files.
For config options that are removed because they haven't been reimplemented yet, we should open issues to keep track of things, otherwise they'll just get forgotten.

ImMorpheus:
Thanks for mentioning that! Updated :+1:

Issue #3811, opened 2023-01-11
WorldEdit mod malfunctions with SpongeForge present


Inscrutable:
### Affected Product(s)

SpongeForge

### Version

spongeforge-1.16.5-36.2.5-8.1.0-RC1256-universal

### Operating System

Win10

### Java Version

1.8.0_351 (Oracle)

### Plugins/Mods

```shell
spongeforge-1.16.5-36.2.5-8.1.0-RC1256-universal
forge-1.16.5-36.2.39
And I tested:
worldedit-mod-7.2.5
worldedit-mod-7.2.6-SNAPSHOT-dist
worldedit-sponge-api8.1.0-7.2.11
```

### Describe the bug

After editing the mods.toml in the WorldEdit jar file to remove the Sponge mod dependency data, the plugin loaded but did not function properly. Results were the same for WE versions 7.2.5 and 7.2.6. Several issues were found:

(1) The right-click event for the wand tool does not fire. Left click works OK.
(2) The majority of the commands don't work. e.g. //pos1, //pos2, //set, /jumpto, ...  There's lots of ` Root command /unknown:replacenear does not have an associated plugin!` and similar in the log.
(3) I could get a few commands to work but the results were very broken. //extinguish fires but won't put out fires, and the brush tool activates but no changes occur, Instead, the phase tracker says `Illegal Direct Chunk Access`.  This happens only with SpongeForge installed.
(4) The Sponge Version of WorldEdit (worldedit-sponge-api8.1.0-7.2.11) won't load in Forge at all, it just gives `[com.sk89q.worldedit.world.registry.LegacyMapper/]: Failed to load the built-in legacy id registry java.lang.NoClassDefFoundError: net/minecraft/world/level/block/state/BlockState`

Enginehub folks would vastly prefer that the Sponge version of WorldEdit worked on both SpongeVanilla and SpongeForge, The ideal situation would be that SpongeVanilla used the same obfuscation environment as SpongeForge, or at least allowed a way for plugins to work with Forge obfuscation mappings applied - this would allow the Sponge WorldEdit plugin to work on SpongeForge too.

Here's the server log from trying the brush tool, with WE version 7.2.5:
[2023-01-10-1.log](https://github.com/SpongePowered/Sponge/files/10387743/2023-01-10-1.log)

Inscrutable:
For reference, PR #3711 is where 1-3 of this went to roost.

Yeregorix:
(1) Fixed in #3844

[...]

PR #3855, opened 2023-04-29
Track direct chunk access

Yeregorix:
Track block changes made with direct calls to `LevelChunk#setBlockState`.
Fixes the third point in #3811.

I made some tests with worldedit-forge and a small test plugin I wrote: 

```java
@Listener
public void onBlockChange(ChangeBlockEvent.All e, @First ServerPlayer p) {
	Sponge.server().sendMessage(Component.text(e.transactions().size() + " block changes by ").append(TestUtil.debugStack(e.cause())));
	if (e.transactions().size() % 2 == 0) {
		Sponge.server().sendMessage(Component.text("CANCELLED"));
		e.setCancelled(true);
	}
}
```

The event is fired correctly and all transactions are grouped into a single event with the worldedit command being the root cause. Cancelling the event, properly cancels block changes.

To have an idea of the performance impact, I selected a 100x100x100 area (1 million blocks) and I measured the average time for command `//set 33%granite,33%stone,33%andesite` to execute on my computer. 
With Sponge: ~5.5s
Without Sponge: ~3.2s

It is a very imprecise benchmark but it shows the order of magnitude of the tracking overhead. 
The performance impact is significant but IMO tolerable for now.

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/tracker/world/level/chunk/LevelChunkMixin_Tracker.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/tracker/world/level/chunk/LevelChunkMixin_Tracker.java
@@ -34,10 +34,10 @@ import net.minecraft.world.level.chunk.LevelChunk;
 import net.minecraft.world.level.chunk.LevelChunkSection;
 import net.minecraft.world.level.chunk.ProtoTickList;
 import net.minecraft.world.phys.AABB;
-import org.apache.logging.log4j.Level;
 import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
 import org.checkerframework.checker.nullness.qual.NonNull;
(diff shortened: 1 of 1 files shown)


gabizou (approved):
Looks fine. I did intend to have logged warnings in the other PR so that there's some visibility into possible overhead with certain cases. There's also the possibility of simply enabling WorldEdit to throw the events itself and call the direct chunk setter without Sponge's tracking by itself.

But that may blow up the scope of the PR for now.

Yeregorix:
I think I will just merge this PR as it is. Warnings can be added in a new PR.

As for worldedit-forge throwing the events, that would require depending on Sponge. At this point, they could just use the API like worldedit-sponge does. I think worldedit-forge should stay forge only. People looking for perfect compatibility should use worldedit-sponge (when it will be fixed).

PR #3858, opened 2023-05-09
Fixes for inventory offer

### LINKED ISSUE #3857: Invalid successful inventory transaction
### Affected Product(s)

SpongeVanilla

### Version

1.16.5-8.1.0-RC1324

### Operating System

Linux/6.2.12-zen1-1-zen

### Java Version

18.0.1/64-bit

### Plugins/Mods

```shell
-
```

### Describe the bug

It seems that when you offer a stack of an item to an inventory that can only fit less than a stack of that item, the transaction result will be successful.

For example, offering 64 oak wood to the inventory in the following image results in a succesful transaction. The rejected items list will also be empty.
image

### Link to logs

_No response_

aromaa:
Offering a stack of stone to a slot with half stack of stone would result in the transaction being marked as success and the offered item stack quantity being set to 32. Now the the transaction is marked as failure and `#rejectedItems` has half stack of stone.

Also changed that offering any amount of stone to full stack would result to transactions in `#slotTransactions`.

Fixes #3857

@Faithcaio

### DIFF
--- a/src/main/java/org/spongepowered/common/inventory/adapter/impl/AdapterLogic.java
+++ b/src/main/java/org/spongepowered/common/inventory/adapter/impl/AdapterLogic.java
@@ -202,7 +202,7 @@ public abstract class AdapterLogic {
                 remaining -= push;
                 Slot slot = ((SlotAdapter) lens.getSlotLens(fabric, ord).getAdapter(fabric, null));
                 result.transaction(new SlotTransaction(slot, ItemStackUtil.snapshotOf(old), ItemStackUtil.snapshotOf(lens.getStack(fabric, ord))));
-            } else if (!old.isEmpty() && ItemStackUtil.compareIgnoreQuantity(old, stack)) {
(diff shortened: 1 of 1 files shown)


Yeregorix:
`SlotAdapter#offer(ItemStack...)` needs to be updated too.

If I understand correctly, in API 8 we no longer decrement the quantity of the stack passed as argument as we used to do in API 7. Am I right ?

aromaa:
Right. I didn't happen to check what we have for API 7. Looks like we no longer document this in API 8 so I would expect it to work like this. So at least the documentation should be fixed at minimum. Imo the API contract feels really weird that we mutate parameters..

Yeregorix:
What you did is correct. Mutating parameters was documented in API 7 and now it's not, because we no longer do that in API 8.

Faithcaio:
Fixed in
https://github.com/SpongePowered/Sponge/commit/48a3a90efa700368e55d834208454dc712a18118
and
https://github.com/SpongePowered/Sponge/commit/b567c47f8a4f984ea1a55e5a3bdb28f4824968a2

I also fixed the equivalent code in `SlotAdapter`
Thanks @aromaa for the initial work.

PR #3859, opened 2023-05-13
Complete packet futures on disconnect

aromaa:
Not really sure about this and the exception we are throwing. Fixes that plugin message channel futures don't complete on disconnection or when the player has disconnected long time ago.

### DIFF
--- a/src/main/java/org/spongepowered/common/network/channel/PacketSender.java
+++ b/src/main/java/org/spongepowered/common/network/channel/PacketSender.java
@@ -24,7 +24,6 @@
  */
 package org.spongepowered.common.network.channel;
 
-import io.netty.util.concurrent.Future;
 import net.minecraft.network.Connection;
 import net.minecraft.network.PacketSendListener;
 import net.minecraft.network.protocol.Packet;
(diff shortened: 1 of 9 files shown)


src/main/java/org/spongepowered/common/network/channel/PacketSender.java
```diff
@@ -45,32 +44,32 @@ public static void sendTo(final EngineConnection connection, final Packet<?> pac
         PacketSender.sendTo(connection, packet, (Consumer) null);
     }
 
-    public static void sendTo(final EngineConnection connection, final Packet<?> packet, final @Nullable Consumer<Future<? super Void>> listener) {
+    public static void sendTo(final EngineConnection connection, final Packet<?> packet, final @Nullable Consumer<Throwable> listener) {
```
Yeregorix:
```suggestion
    public static void sendTo(final EngineConnection connection, final Packet<?> packet, final @Nullable Consumer<@Nullable Throwable> listener) {
```

src/main/java/org/spongepowered/common/network/channel/PacketSender.java
```diff
@@ -45,32 +44,32 @@ public static void sendTo(final EngineConnection connection, final Packet<?> pac
...
         });
     }
 
     public static final class SpongePacketSendListener implements PacketSendListener {
         private final BlockableEventLoop<?> executor;
-        private final Consumer<Future<? super Void>> listener;
+        private final Consumer<Throwable> listener;
 
-        public SpongePacketSendListener(final EngineConnectionSide<? extends EngineConnection> side, final Consumer<Future<? super Void>> listener) {
+        public SpongePacketSendListener(final EngineConnectionSide<? extends EngineConnection> side, final Consumer<Throwable> listener) {
```
Yeregorix:
```suggestion
        public SpongePacketSendListener(final EngineConnectionSide<? extends EngineConnection> side, final Consumer<@Nullable Throwable> listener) {
```

src/main/java/org/spongepowered/common/network/channel/PacketSender.java
```diff
@@ -45,32 +44,32 @@ public static void sendTo(final EngineConnection connection, final Packet<?> pac
...
             this.executor = (BlockableEventLoop<?>) (side == EngineConnectionSide.CLIENT ? Sponge.client() : Sponge.server());
             this.listener = listener;
         }
 
-        public void accept(final Future nettyFuture) {
-            this.executor.execute(() -> this.listener.accept(nettyFuture));
+        public void accept(final Throwable throwable) {
```
Yeregorix:
```suggestion
        public void accept(final @Nullable Throwable throwable) {
```

src/mixins/java/org/spongepowered/common/mixin/core/network/ConnectionMixin.java
```diff
@@ -148,7 +155,33 @@ public abstract class ConnectionMixin extends SimpleChannelInboundHandler<Packet
...
+    }
+
+    @Inject(method = "send(Lnet/minecraft/network/protocol/Packet;Lnet/minecraft/network/PacketSendListener;)V", at = @At(value = "HEAD"), cancellable = true)
+    private void impl$onSend(final Packet<?> $$0, final @Nullable PacketSendListener $$1, final CallbackInfo ci) {
+        if (this.disconnectionHandled) {
+            if ($$1 instanceof final PacketSender.SpongePacketSendListener spongeListener) {
+                spongeListener.accept(new IOException("Connection has been closed."));
+            }
+
+            ci.cancel();
```
Yeregorix, line 169: Cancelling packet send when disconnected sounds logic to me but vanilla doesn't do that. I don't know what the implications are.

aromaa: Vanilla doesn't really have any concerns but mods might but that seems highly unlikely to me. The connection is effectively closed at this point and we are just piling up the packets in the `#queue` field which doesn't get touched after disconnection. So this is pretty much just saving up some allocations.

Yeregorix: Alright, that should be fine then.

src/main/java/org/spongepowered/common/network/channel/PacketSender.java
```diff
@@ -45,32 +44,32 @@ public static void sendTo(final EngineConnection connection, final Packet<?> pac
...
-                future.completeExceptionally(sendFuture.cause());
+                future.completeExceptionally(throwable);
             }
         });
     }
 
     public static final class SpongePacketSendListener implements PacketSendListener {
         private final BlockableEventLoop<?> executor;
-        private final Consumer<Future<? super Void>> listener;
+        private final Consumer<Throwable> listener;
```
Yeregorix:
```suggestion
        private final Consumer<@Nullable  Throwable> listener;
```

PR #3864, opened 2023-06-11
Add missing data conversion for BlockStateKeys

Lignium:
Primitive types and mixins can cast freely, but types like `Direction` and `Axis` are not mixins and cannot cast on their own.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/provider/block/state/BlockStateDataProvider.java
+++ b/src/main/java/org/spongepowered/common/data/provider/block/state/BlockStateDataProvider.java
@@ -140,12 +140,8 @@ public final class BlockStateDataProvider {
         registrator.asImmutable(BlockState.class)
                 .create(key)
                     .supports(bs -> bs.getOptionalValue(property).isPresent())
-                    .get(bs -> {
-                        return ((org.spongepowered.api.block.BlockState) bs).stateProperty((StateProperty<T>) property).orElse(null);
-                    })
-                    .set((bs, v) -> {
-                        return (BlockState) ((org.spongepowered.api.block.BlockState) bs).withStateProperty((StateProperty<T>) property, v).orElse(null);
-                    });
+                    .get(bs -> ((org.spongepowered.api.block.BlockState) bs).stateProperty((StateProperty<T>) property).orElse(null))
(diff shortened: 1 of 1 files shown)


aromaa (changes requested):
We already seem to have similar logic in `StateHolderMixin_API`. This should be easier to do by casting to our `BlockState` to achieve the same?

Lignium:
> We already seem to have similar logic in `StateHolderMixin_API`. This should be easier to do by casting to our `BlockState` to achieve the same?

Yes, this also works, moreover, I did not take into account all the data initially.

src/main/java/org/spongepowered/common/data/provider/block/state/BlockStateDataProvider.java
```diff
@@ -134,13 +135,17 @@ public static void register(final DataProviderRegistrator registrator) {
...
                 .create(key)
                     .supports(bs -> bs.getOptionalValue(property).isPresent())
-                    .get(bs -> (T) bs.getOptionalValue(property).orElse(null))
-                    .set((bs, v) -> bs.setValue(property, (V) v));
+                    .get(bs -> {
+                        return ((org.spongepowered.api.block.BlockState) bs).stateProperty((StateProperty<T>) property).orElse(null);
+                    })
+                    .set((bs, v) -> {
+                        return (BlockState) ((org.spongepowered.api.block.BlockState) bs).withStateProperty((StateProperty<T>) property, v).orElse(null);
+                    });
```
aromaa:
```suggestion
                .get(bs -> ((org.spongepowered.api.block.BlockState) bs).stateProperty((StateProperty<T>) property).orElse(null))
                .set((bs, v) -> (BlockState) ((org.spongepowered.api.block.BlockState) bs).withStateProperty((StateProperty<T>) property, v).orElse((org.spongepowered.api.block.BlockState) bs));
```
We generally return the last `BlockState` if the offered value is invalid. The contract says that we return empty optional if the key is not supported.

Lignium: > We generally return the last `BlockState` if the offered value is invalid. The contract says that we return empty optional if the key is not supported.

Exactly, I missed that

aromaa (approved):
Thank you!!

PR #3865, opened 2023-06-11
Resend player permission level after update Keys#SKIN_PROFILE_PROPERTY

Lignium:
Fixes a nasty bug that caused some client-side functions to stop working, such as fast switching to spectator mode (F3+N). The client was losing data about the permissions (`permissionLevel`) and thought that there were no permissions.
An example of the code that reproduces the bug:

```java
@Listener
private void onJoinPlayer(ServerSideConnectionEvent.Join event) {
    Objects.requireNonNull(event, "event");

    ServerPlayer player = event.player();
    Value<ProfileProperty> profileProperty = player.requireValue(Keys.SKIN_PROFILE_PROPERTY);

    player.offer(profileProperty);
}
```

After trying to write the same data, the permission level on the client is reset.


aromaa (approved):
LGTM. Thanks!

PR #3866, opened 2023-06-11
Throw FileNotFoundException when resource doesn't exist

Lignium:
According to the documentation, the `ResourceManager#load(ResourcePath)` method should throw a `FileNotFoundException` if the resource file is not found. However, an unexpected `NoSuchElementException` is being thrown.


aromaa (changes requested):
`ResourceProvider` seems to have a method `getResourceOrThrow`

aromaa (approved):
Thanks!!

PR #3873, opened 2023-07-07
[API-10] Improve unit tests

Yeregorix:
Continuing from #3815.

This PR adds the ability to run JUnit 5 tests in a ModLauncher context where Sponge and a DedicatedServer is initialized (but not ticking). This is done using [modlauncher-injector-junit](https://github.com/SpongePowered/modlauncher-injector-junit/) the successor of [LaunchWrapperTestSuite](https://github.com/SpongePowered/LaunchWrapperTestSuite/).

I also added JaCoCo so we can measure code coverage. This should help us improve the test suite in the future.

Some of the launch logic is shared between SpongeVanilla and Sponge.test. Sponge.test is depending on SpongeVanilla.launch and SpongeVanilla.applaunch-base

The only issues identified so far with this setup are:
- `@ParameterizedTest` cannot be executed in a ModLauncher context. However `@TestFactory` is a wonderful alternative and is working perfectly.
- All unit tests are running in the same ModLauncher context. Any value registered into Sponge by a test should be unregistered after the test so it doesn't interfere with other tests. In the future, using a custom class loader might be a solution to create multiple isolated ModLauncher instances. It's not planned at the moment, because we don't need it.

I have updated all tests in the `invalid/` directory.

Some tests are failing and need fixes but this it out of the scope of this PR, so these tests are currently `@Disabled`. Anyone willing to fix the implementation to make pass these tests is welcome.


Yeregorix:
Need rework on latest versions because of breaking changes in ModLauncher.

Issue #3862, opened 2023-06-02
Cancelling ClickContainerEvent.Primary removes the item in the clicked slot


avaruus1:
### Affected Product(s)

SpongeVanilla

### Version

1.19.4-10.0.0-RC1341

### Operating System

Linux/6.3.3-zen1-1-zen

### Java Version

18.0.1

### Plugins/Mods

```shell
-
```

### Describe the bug

Cancelling `ClickContainerEvent.Primary` removes the item in the clicked slot. The first time a slot is clicked the interaction is actually cancelled correctly, but everything clicked after that is removed.

```java
@Listener
public void onClick(ClickContainerEvent.Primary event) {
    event.setCancelled(true);
}
```

PR #3875, opened 2023-07-16
Cancelling ClickContainerEvent.Primary removes the item in the clicked slot

AlexandreArcil:
This PR is aimed to fix the issue #3862 
Tested with SpongeAPI 10.0.0-SNAPSHOT and SpongeVanilla 1.19.4-10.0.0-RC0.

### The bug
When the `ClickContainerEvent.Primary` event is cancelled, the first click on an item is cancelled, but a second click causes the item to disappear. It's not a visual bug, leaving and then reconnecting to the server will not make the item to reappear.

### Why this bug happens
During the restore of the container after the first click, many packets are sent to the client to replace the item in the slot and remove the item in the cursor. At the end, the `stateId` of the container will be 4. But during the second click, the client will send a `stateId` of 3, leading to the execution of the method `AbstractContainerMenu.broadcastFullState` instead of `AbstractContainerMenu#broadcastChanges`, where the items in the container are set without transactions.

This difference of `stateId` is due to the fact that after the restoration, the `stateId` of the container is 3, but the tick system will detect an inconsistency between the "carried" item and the "remote carried" item. It will fix it and then send a packet with a `stateId` of 4 to tell to the client that the "carried" item has been updated. But it has already received the updated item through the cursor restore, so I think that it ignores the packet and so keep the `stateId` of the container to 3.

### The fix
The fix simply sets the "remote carried" and not only the "carried" item during the cursor restore to avoid the inconsistency. A test plugin has been made that cancels the event ClickContainerEvent.Primary to see the bug and that this PR fixes it.

### DIFF
--- a/src/main/java/org/spongepowered/common/bridge/world/inventory/container/TrackedContainerBridge.java
+++ b/src/main/java/org/spongepowered/common/bridge/world/inventory/container/TrackedContainerBridge.java
@@ -28,5 +28,5 @@ public interface TrackedContainerBridge {
 
     boolean bridge$capturePossible();
 
-    void bridge$detectAndSendChanges(boolean captureOnly);
+    void bridge$detectAndSendChanges(boolean capture, boolean synchronize);
 }
--- a/src/mixins/java/org/spongepowered/common/mixin/inventory/event/world/inventory/AbstractContainerMenuMixin_Inventory.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/inventory/event/world/inventory/AbstractContainerMenuMixin_Inventory.java
@@ -81,6 +81,8 @@ public abstract class AbstractContainerMenuMixin_Inventory implements TrackedCon
     @Shadow protected abstract void shadow$updateDataSlotListeners(int $$0, int $$1);
     @Shadow protected abstract void shadow$synchronizeDataSlotToRemote(int $$0, int $$1);
     @Shadow protected abstract void shadow$synchronizeCarriedToRemote();
+
+    @Shadow public abstract void shadow$sendAllDataToRemote();
     //@formatter:on
 
     private boolean impl$isClicking; // Menu Callbacks are only called when clicking in a container
@@ -168,6 +170,10 @@ public abstract class AbstractContainerMenuMixin_Inventory implements TrackedCon
             return;
         }
         this.bridge$detectAndSendChanges(false, false);
+        this.impl$broadcastDataSlots();
+        this.shadow$sendAllDataToRemote();
+        this.impl$captureSuccess = true; // Detect mod overrides
+        ci.cancel();
     }
 
     // broadcastChanges
@@ -230,17 +236,19 @@ public abstract class AbstractContainerMenuMixin_Inventory implements TrackedCon
                 for (final ContainerListener listener : this.containerListeners) {
                     listener.slotChanged(((AbstractContainerMenu) (Object) this), i, oldStack);
                 }
-                final ItemStack remoteStack = this.remoteSlots.get(i);
-                if (!ItemStack.matches(remoteStack, newStack)) {
-                    this.remoteSlots.set(i, newStack.copy());
-                    if (this.synchronizer != null) {
-                        this.synchronizer.sendSlotChange(((AbstractContainerMenu) (Object) this), i, newStack);
+                if(synchronize) {
+                    final ItemStack remoteStack = this.remoteSlots.get(i);
+                    if (!ItemStack.matches(remoteStack, newStack)) {
+                        this.remoteSlots.set(i, newStack.copy());
+                        if (this.synchronizer != null) {
+                            this.synchronizer.sendSlotChange(((AbstractContainerMenu) (Object) this), i, newStack);
+                        }
                     }
                 }
             }
         }
 
-        if(synchronize) {
+        if (synchronize) {
             this.shadow$synchronizeCarriedToRemote();
             this.impl$broadcastDataSlots();
         }
--- a/src/mixins/java/org/spongepowered/common/mixin/inventory/event/world/inventory/AbstractContainerMenuMixin_Inventory.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/inventory/event/world/inventory/AbstractContainerMenuMixin_Inventory.java
@@ -267,7 +267,7 @@ public abstract class AbstractContainerMenuMixin_Inventory implements TrackedCon
         }
     }
 
-    private void impl$broadcastDataSlots(boolean synchronize) {
+    private void impl$broadcastDataSlots(final boolean synchronize) {
         for(int j = 0; j < this.dataSlots.size(); ++j) {
             final DataSlot dataSlot = this.dataSlots.get(j);
             if (dataSlot.checkAndClearUpdateFlag()) {
--- a/src/mixins/java/org/spongepowered/common/mixin/inventory/event/world/inventory/AbstractContainerMenuMixin_Inventory.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/inventory/event/world/inventory/AbstractContainerMenuMixin_Inventory.java
@@ -131,7 +131,7 @@ public abstract class AbstractContainerMenuMixin_Inventory implements TrackedCon
         if (((LevelBridge) player.level).bridge$isFake() || player.level.isClientSide || !(thisContainer.getSlot(slotId) instanceof ResultSlot)) {
             return result;
         }
-        this.bridge$detectAndSendChanges(true);
+        this.bridge$detectAndSendChanges(true, true);
         final PhaseContext<@NonNull ?> context = PhaseTracker.SERVER.getPhaseContext();
         TrackingUtil.processBlockCaptures(context); // ClickContainerEvent -> CraftEvent -> PreviewEvent
         // result is modified by the ClickMenuTransaction
@@ -162,6 +162,14 @@ public abstract class AbstractContainerMenuMixin_Inventory implements TrackedCon
 
     }
 
+    @Inject(method = "broadcastFullState", at = @At("HEAD"))
+    private void impl$broadcastFullStateWithTransactions(final CallbackInfo ci) {
+        if (!PhaseTracker.SERVER.onSidedThread()) {
+            return;
+        }
+        this.bridge$detectAndSendChanges(false, false);
+    }
+
     // broadcastChanges
 
     /**
@@ -176,14 +184,14 @@ public abstract class AbstractContainerMenuMixin_Inventory implements TrackedCon
         if (!PhaseTracker.SERVER.onSidedThread()) {
             return;
         }
-        this.bridge$detectAndSendChanges(false);
+        this.bridge$detectAndSendChanges(false, true);
         this.impl$captureSuccess = true; // Detect mod overrides
         ci.cancel();
     }
 
 
     @Override
-    public void bridge$detectAndSendChanges(final boolean captureOnly) {
+    public void bridge$detectAndSendChanges(final boolean capture, final boolean synchronize) {
         // Code-Flow changed from vanilla completely!
 
         final SpongeInventoryMenu menu = ((MenuBridge)this).bridge$getMenu();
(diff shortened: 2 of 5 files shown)


AlexandreArcil:
This fix reveals the existence of an another bug: if you spam click 3 times, the event will not be cancelled. This can be used to duplicate items. This problem is fairly the same as this one: the difference of stateId between the client and the server.

After spam clicking 3 times, the client sends not 3 but 4 "click container" packets. The third and fourth packets have the same stateId, whereas the server has updated the container after the third packet and so the stateId. 
After spam clicking 2 times, the client sends not 2 but 3 "click container" packets, but this time, the stateId is correct. That's why the item is moved 3 times instead of 2.

I don't know why this happens, if it's a normal behaviour, which happens after some conditions or not, and if I'm the only one who he's affected. I'm still investigating.

EDIT: This mysterious behaviour is just... a double click. The third packet is sent after a mouse release and the other packets after a mouse click. The response for the third packet is received after the fourth packet is sent, which explain why they have the same stateId. I'm still thinking about how to handle this behaviour to avoid this bug.

AlexandreArcil:
I fixed the bug mentioned above.

I don't know if you are ok with the modification of TrackerContainerBridge and the other methods, tell me if you prefer an another way.

testplugins/src/main/java/org/spongepowered/test/interact/ClickContainerEventCancelledTest.java
```diff
@@ -0,0 +1,78 @@
...
+        private void onInteractItem(final ClickContainerEvent.Primary event) {
+            event.setCancelled(true);
+            Sponge.server().broadcastAudience().sendMessage(Identity.nil(), Component.text("/*************"));
+            Sponge.server().broadcastAudience().sendMessage(Identity.nil(), Component.text().append(Component.text("/* Event: ")).append(Component.text(event.getClass().getSimpleName())).build());
+            Sponge.server().broadcastAudience().sendMessage(Identity.nil(),
+                    Component.text().append(Component.text("/* Hand: "))
+                            .append(Component.text(event.context().get(EventContextKeys.USED_HAND).map(h -> RegistryTypes.HAND_TYPE.keyFor(Sponge.game(), h).formatted()).orElse("UNKNOWN")))
+                            .build()
+            );
+            Sponge.game().systemSubject().sendMessage(Identity.nil(), Component.text().append(Component.text("/ Cause: ")).append(Component.text(event.cause().all().toString())).build());
```
gabizou, line 69: Out of curiosity, why the switch from `broadcastAudience` to `systemSubject`?

src/main/java/org/spongepowered/common/bridge/world/inventory/container/TrackedContainerBridge.java
```diff
@@ -28,5 +28,5 @@ public interface TrackedContainerBridge {
 
     boolean bridge$capturePossible();
 
-    void bridge$detectAndSendChanges(boolean captureOnly);
+    void bridge$detectAndSendChanges(boolean capture, boolean synchronize);
```
gabizou, line 31: Just to be clear, the synchronize argument is to synchronize any changes back to the client, yes?

testplugins/src/main/java/org/spongepowered/test/interact/ClickContainerEventCancelledTest.java
AlexandreArcil: This code is a copy of **InteractTest** with some modifications. I will change to `systemSubject` because `broadcastAudience` doesn't print anything.

src/main/java/org/spongepowered/common/bridge/world/inventory/container/TrackedContainerBridge.java
AlexandreArcil: Yes. If true, it will send the carried item and data slots to the client, but `broadcastFullState` already do that so it doesn't need to do it again.

src/mixins/java/org/spongepowered/common/mixin/inventory/event/world/inventory/AbstractContainerMenuMixin_Inventory.java
```diff
@@ -162,6 +162,14 @@ public abstract class AbstractContainerMenuMixin_Inventory implements TrackedCon
 
     }
 
+    @Inject(method = "broadcastFullState", at = @At("HEAD"))
+    private void impl$broadcastFullStateWithTransactions(final CallbackInfo ci) {
+        if (!PhaseTracker.SERVER.onSidedThread()) {
+            return;
+        }
+        this.bridge$detectAndSendChanges(false, false);
```
aromaa, line 171:
```suggestion
        this.bridge$detectAndSendChanges(false, false);
        this.impl$broadcastDataSlots();
        this.shadow$sendAllDataToRemote();
        this.impl$captureSuccess = true; // Detect mod overrides
        ci.cancel();
```

src/mixins/java/org/spongepowered/common/mixin/inventory/event/world/inventory/AbstractContainerMenuMixin_Inventory.java
```diff
@@ -232,8 +240,10 @@ public abstract class AbstractContainerMenuMixin_Inventory implements TrackedCon
             }
         }
 
-        this.shadow$synchronizeCarriedToRemote();
-        this.impl$broadcastDataSlots();
+        if(synchronize) {
```
aromaa:
```suggestion
        if (synchronize) {
```

aromaa (changes requested):
Thank you for looking at this and sorry for the delay of getting this reviewed. This looks about right, just suggesting few minor changes. Indeed the problem was that we were not capturing the transactions for the `broadcastFullState` which happens on mismatched state id, this fixes it.

The reason why we get mismatched state id seems to be due to the client not updating it's state id when the cursor is updated and the vanilla server logic does increment it. This means that when the cursor is updated the client will always send mismatched state id to the server which causes the full inventory resync.

We avoid incrementing the state id on cursor restore so setting the remote carrier there too is correct to avoid vanilla re-sending it yet again.

src/mixins/java/org/spongepowered/common/mixin/inventory/event/world/inventory/AbstractContainerMenuMixin_Inventory.java
AlexandreArcil: `broadcastFullState` calls `triggerSlotListeners`, `updateDataSlotListeners` and `sendAllDataToRemote` but your change calls  `synchronizeDataSlotToRemote`, `updateDataSlotListeners` and `sendAllDataToRemote` which is not the same, why ?

aromaa: I tried to avoid doing some extra work that was unnecessary but right its sending some changed slots and then doing full broadcast on top of that. You could avoid that by guarding the `sendSlotChange` on the `bridge$detectAndSendChanges` with the synchronize flag and then do the same for `impl$broadcastDataSlots`.

AlexandreArcil: Ok so if I understood correctly, I add a parameter to `impl$broadcastDataSlots` in order to only call `updateDataSlotListeners`, but I didn't understand for `sendSlotChange`. 
Otherwise, I can simply shadow `triggerSlotListeners` and create a method similar to the first for loop of `broadcastFullState` which will call `triggerSlotListeners` to have a behaviour similar to the game. I don't think it's a big extra work, while avoiding problems.

aromaa: >  but I didn't understand for sendSlotChange.

At the moment we are sending individual slot changes in the `bridge$detectAndSendChanges` (L237) even when the `synchronize` is set to false which I suggested to fix. Then after this and the `impl$broadcastDataSlots` fix the methods are performing identical logic to vanilla.

AlexandreArcil: Oh I didn't see that `slotChanged` was called in `bridge$detectAndSendChanges`.
I will make the changes !

AlexandreArcil: Done !

src/mixins/java/org/spongepowered/common/mixin/inventory/event/world/inventory/AbstractContainerMenuMixin_Inventory.java
AlexandreArcil: Done

AlexandreArcil:
Modifications made !

src/mixins/java/org/spongepowered/common/mixin/inventory/event/world/inventory/AbstractContainerMenuMixin_Inventory.java
```diff
@@ -162,6 +164,18 @@ public abstract class AbstractContainerMenuMixin_Inventory implements TrackedCon
 
     }
 
+    @Inject(method = "broadcastFullState", at = @At("HEAD"))
+    private void impl$broadcastFullStateWithTransactions(final CallbackInfo ci) {
+        if (!PhaseTracker.SERVER.onSidedThread()) {
+            return;
+        }
+        this.bridge$detectAndSendChanges(false, false);
+        this.impl$broadcastDataSlots();
```
aromaa: You forgot to add the flag for `shadow$synchronizeDataSlotToRemote` that we discussed.

src/mixins/java/org/spongepowered/common/mixin/inventory/event/world/inventory/AbstractContainerMenuMixin_Inventory.java
```diff
@@ -222,18 +236,22 @@ public abstract class AbstractContainerMenuMixin_Inventory implements TrackedCon
                 for (final ContainerListener listener : this.containerListeners) {
                     listener.slotChanged(((AbstractContainerMenu) (Object) this), i, oldStack);
                 }
-                final ItemStack remoteStack = this.remoteSlots.get(i);
-                if (!ItemStack.matches(remoteStack, newStack)) {
-                    this.remoteSlots.set(i, newStack.copy());
-                    if (this.synchronizer != null) {
-                        this.synchronizer.sendSlotChange(((AbstractContainerMenu) (Object) this), i, newStack);
+                if(synchronize) {
```
aromaa:
```suggestion
                if (synchronize) {
```

src/mixins/java/org/spongepowered/common/mixin/inventory/event/world/inventory/AbstractContainerMenuMixin_Inventory.java
```diff
@@ -81,6 +81,8 @@ public abstract class AbstractContainerMenuMixin_Inventory implements TrackedCon
     @Shadow protected abstract void shadow$updateDataSlotListeners(int $$0, int $$1);
     @Shadow protected abstract void shadow$synchronizeDataSlotToRemote(int $$0, int $$1);
     @Shadow protected abstract void shadow$synchronizeCarriedToRemote();
+
```
aromaa:
```suggestion
```

Might as well remove this empty line while at it.

aromaa (changes requested):
You forgot to add the flag for `shadow$synchronizeDataSlotToRemote` that we discussed.

src/mixins/java/org/spongepowered/common/mixin/inventory/event/world/inventory/AbstractContainerMenuMixin_Inventory.java
```diff
@@ -250,13 +267,15 @@ public abstract class AbstractContainerMenuMixin_Inventory implements TrackedCon
         }
     }
 
-    private void impl$broadcastDataSlots() {
+    private void impl$broadcastDataSlots(boolean synchronize) {
```
aromaa:
```suggestion
    private void impl$broadcastDataSlots(final boolean synchronize) {
```
Sorry 💀

aromaa (approved):
LGTM. Thank you for your contribution!

PR #3884, opened 2023-08-22
Add BlockType#hasBlockEntity

Lignium:
[[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2462)|**Sponge**]

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/api/minecraft/block/AbstractBlockMixin_API.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/api/minecraft/block/AbstractBlockMixin_API.java
@@ -25,6 +25,7 @@
 package org.spongepowered.common.mixin.api.minecraft.block;
 
 import net.minecraft.world.item.Item;
+import net.minecraft.world.level.block.EntityBlock;
 import net.minecraft.world.level.block.SoundType;
 import net.minecraft.world.level.block.state.BlockBehaviour;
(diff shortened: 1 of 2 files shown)


src/mixins/java/org/spongepowered/common/mixin/api/minecraft/block/AbstractBlockMixin_API.java
```diff
@@ -57,4 +58,8 @@ public BlockSoundGroup soundGroup() {
         return (BlockSoundGroup) this.soundType;
     }
 
+    @Override
+    public boolean hasBlockEntity() {
+        return this instanceof EntityBlock;
```
gabizou, line 63: Honestly could make the same mixin in EntityBlockMixin_API and default this returning false?

Lignium: > Honestly could make the same mixin in EntityBlockMixin_API and default this returning false?

The `EntityBlock` interface does not extend any other interfaces, it is not related directly to `Block`. If simply create a mixin extending `BlockType` and implement this method with return `true`, it simply won't be called. The default method in `AbstractBlockMixin_API` is always called.

Faithcaio:
merged: https://github.com/SpongePowered/Sponge/commit/ba7223a43b8f64b4cefca869eb6a4bbeb4f325b6

Issue #3871, opened 2023-06-30
Players are being disconnected when they click on NPCs from pixelmon


kasdo:
### Affected Product(s)

SpongeForge

### Version

1.16.5-36.2.5-8.1.0-RC1352-universal

### Operating System

Linux

### Java Version

1.8.0_212/64-bit

### Plugins/Mods

```shell
Mods: pixelmon, BiomesOPlenty, BooshelfForge, chiselandhabits, WAWLA, costumes, ironchests, placebo, selene. clumps, fastfurnace, fastworkbench, ferritecore, notenoughcrashes, spark, starlight, storagedrawers, treechop, worldedit

Plugins: ChatManager, CommandPack, GTS, Impactor, LocaleAPI, LuckPerms, RegionGuard, TablistBoard
```

### Describe the bug

The video of the issue:
https://www.youtube.com/watch?v=Xkf1djfgDMU&t=1s&ab_channel=KonanMayer

When players click on TCG trader from pixelmon mod, they are being disconnected from the server. I have the same modpack on mohist server and there's no an issue like that. I made a request to fix that to pixelmon but they said it's sponge issue

### Link to logs

https://paste.anfanik.me/oyipapoqej.sql

ImMorpheus:
Hey! You seem to be running an unsupported version (API 8) of SpongeAPI. Support for API 8 ended in October 2023.

API-8 is now end-of-life. We will not be issuing any more API updates for version 8 of our API[^1]

As part of our legacy cleanup we are closing issues relating to API8/1.16
___

If you wish to move this forward, please try to reproduce the issue on a supported version of Sponge.

[^1]: https://forums.spongepowered.org/t/sponge-status-update-17th-october-2023/42212#spongeapi-82-1

Issue #3828, opened 2023-02-06
Latest Spongeforge Does Not Allow You To Talk To Pixelmon TCG NPCs - Kicks You


DecaNFT:
### Affected Product(s)

SpongeForge

### Version

1.16.5-36.2.5-8.1.0-RC1278

### Operating System

Windows

### Java Version

Java 8 - jre1.8.0_361

### Plugins/Mods

```shell
Pixelmon, SpongeForge
```

### Describe the bug

I just found out that latest spongeforge version messes with the TCG NPC in pixelmon reforged 1.16.5. Using an older spongeforge I was able to get it to stop but it was kicking me from my server any time I tried to talk to the TCG NPC. Now it works fine since I started using the first 8.1.0 version of spongeforge rather than latest.

### Link to logs

https://paste.gg/p/anonymous/999bf6c69fa346c190bbe4c9c13ee676

ImMorpheus:
Hey! You seem to be running an unsupported version (API 8) of SpongeAPI. Support for API 8 ended in October 2023.

API-8 is now end-of-life. We will not be issuing any more API updates for version 8 of our API[^1]

As part of our legacy cleanup we are closing issues relating to API8/1.16
___

If you wish to move this forward, please try to reproduce the issue on a supported version of Sponge.

[^1]: https://forums.spongepowered.org/t/sponge-status-update-17th-october-2023/42212#spongeapi-82-1

PR #3888, opened 2023-08-31
Click on merchants can kick players

AlexandreArcil:
### Version
Sponge: 8.0.0
SpongeForge: 1.16.5-36.2.5-8.1.0-RC1352-universal

### The bug
When a player clicks on a merchant, an error can be thrown and the player kicked off the server.
This bug typically happens when clicking on a TCG Trader from Pixelmon mod.
This bug was reported by #3871 and #3828.

### Description
When the packet _ClientboundMerchantOffersPacket_ is sent, the itemstacks and its NBTs are written into the byte buf. Sponge manipulates the NBTs to do server-side translation and for this, it needs to deserialize the name of the item and its lore.
The lines of the lore of the itemstacks sold by the TCG traders are a simple string and not a JSON in a string, so the deserialization throws an _JsonParseException_ which kicks the player.

### The fix
The fix is to support simple strings as lore lines. The are several ways to do it, and the solution of this PR is to use a  _TextComponent_ if the deserialization fails.

An another solution would be to use `fromJsonLenient` instead of `fromJson` but I don't know if this can cause "corrupted" merchant because of JSON deserializable with SpongeForge but not with Minecraft.

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/core/network/FriendlyByteBufMixin.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/core/network/FriendlyByteBufMixin.java
@@ -24,11 +24,13 @@
  */
 package org.spongepowered.common.mixin.core.network;
 
+import com.google.gson.JsonSyntaxException;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.nbt.ListTag;
 import net.minecraft.nbt.StringTag;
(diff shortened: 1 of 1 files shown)


aromaa:
How does the mod add these lores? Looking at the Vanilla's code, it also uses `Component.Serializer.fromJson` and if that fails it just removes the whole lore?

ImMorpheus:

___

PR #3890, opened 2023-09-03
some improvements in task scheduler

sunmisc:
Removed the extra AtomicBoolean from the AsyncScheduler class, we do not need it, since we work with the stateChanged field always inside the lock

and set from in class SpongeScheduler added AtomicLong to sequenceNumber

Since the addition of tasks can occur from different threads, we must make a safe increment, from a structural point of view, nothing was broken, but the information was given incorrect

Then I made small optimizations in AsyncScheduler with timeout calculation, we don’t have to write it in the field, since we calculate the value again at each tick, I moved it to preTick at the same time removed the extra lock on this method (separate)

other changes: such as adding an extra method to avoid copying the map with tasks on every tick

### DIFF
--- a/src/main/java/org/spongepowered/common/scheduler/AsyncScheduler.java
+++ b/src/main/java/org/spongepowered/common/scheduler/AsyncScheduler.java
@@ -31,13 +31,9 @@ import org.spongepowered.api.util.Functional;
 import org.spongepowered.common.SpongeCommon;
 import org.spongepowered.common.util.PrettyPrinter;
 
-import java.util.Set;
-import java.util.concurrent.Callable;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.*;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -47,15 +43,13 @@ public final class AsyncScheduler extends SpongeScheduler {
     // Locking mechanism
     private final Lock lock = new ReentrantLock();
     private final Condition condition = this.lock.newCondition();
-    private final AtomicBoolean stateChanged = new AtomicBoolean(false);
     // The dynamic thread pooling executor of asynchronous tasks.
-    private final ExecutorService executor = Executors.newCachedThreadPool(new ThreadFactoryBuilder()
-                                                                                   .setNameFormat("Sponge-AsyncScheduler-%d")
-                                                                                   .build());
-    private volatile boolean running = true;
+    private final ExecutorService executor = Executors.newCachedThreadPool(
+            new ThreadFactoryBuilder()
+                    .setNameFormat("Sponge-AsyncScheduler-%d")
+                    .build());
+    private boolean stateChanged;
 
-    // Adjustable timeout for pending Tasks
-    private long minimumTimeout = Long.MAX_VALUE;
     private long lastProcessingTimestamp;
 
     public AsyncScheduler() {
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/scheduler/AsyncScheduler.java
```diff
@@ -31,13 +31,9 @@
...
-import java.util.Set;
-import java.util.concurrent.Callable;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.*;
```
gabizou, line 36: Please follow the styleguide, we don't want to use star imports.

src/main/java/org/spongepowered/common/scheduler/AsyncScheduler.java
```diff
@@ -69,52 +63,54 @@ public AsyncScheduler() {
...
     }
 
-    private void recalibrateMinimumTimeout() {
-        this.lock.lock();
-        try {
-            final Set<ScheduledTask> tasks = this.tasks();
-            this.minimumTimeout = Long.MAX_VALUE;
+    private long recalibrateMinimumTimeout() {
+        long minimumTimeout = Long.MAX_VALUE;
+        final Collection<ScheduledTask> tasks = pendingTasks();
```
gabizou, line 73: nit: `this.pendingTasks()`

src/main/java/org/spongepowered/common/scheduler/AsyncScheduler.java
```diff
@@ -172,35 +170,35 @@ protected void onTaskCompletion(final SpongeScheduledTask task) {
...
     public <T> CompletableFuture<T> submit(final Callable<T> callable) {
         return Functional.asyncFailableFuture(callable, this.executor);
     }
 
     public void close() {
-        this.running = false;
-        // Cancel all tasks
-        final Set<ScheduledTask> tasks = this.tasks();
-        tasks.forEach(ScheduledTask::cancel);
+        boolean terminated = executor.isTerminated();
```
gabizou, line 181:
```suggestion
        final boolean terminated = this.executor.isTerminated();
```

src/main/java/org/spongepowered/common/scheduler/AsyncScheduler.java
```diff
@@ -130,15 +126,17 @@ protected void addTask(final SpongeScheduledTask task) {
...
-        this.lock.lock();
         try {
+            this.lock.lockInterruptibly();
+
             // If we have something that has indicated it needs to change,
             // don't await, just continue.
-            if (!this.stateChanged.get()) {
-                this.condition.await(this.minimumTimeout, TimeUnit.NANOSECONDS);
+            if (!stateChanged) {
+                this.condition.awaitNanos(recalibrateMinimumTimeout());
```
gabizou, line 135:
```suggestion
                this.condition.awaitNanos(this.recalibrateMinimumTimeout());
```

src/main/java/org/spongepowered/common/scheduler/AsyncScheduler.java
```diff
@@ -130,15 +126,17 @@ protected void addTask(final SpongeScheduledTask task) {
...
     protected void preTick() {
-        this.lock.lock();
         try {
+            this.lock.lockInterruptibly();
+
             // If we have something that has indicated it needs to change,
             // don't await, just continue.
-            if (!this.stateChanged.get()) {
-                this.condition.await(this.minimumTimeout, TimeUnit.NANOSECONDS);
+            if (!stateChanged) {
```
gabizou, line 134:
```suggestion
            if (!this.stateChanged) {
```

src/main/java/org/spongepowered/common/scheduler/AsyncScheduler.java
```diff
@@ -172,35 +170,35 @@ protected void onTaskCompletion(final SpongeScheduledTask task) {
...
-        final Set<ScheduledTask> tasks = this.tasks();
-        tasks.forEach(ScheduledTask::cancel);
+        boolean terminated = executor.isTerminated();
+        if (!terminated) {
+            // Shut down the executor
+            this.executor.shutdown();
 
-        // Shut down the executor
-        this.executor.shutdown();
+            List<ScheduledTask> tasks = drainTasks();
```
gabizou, line 186:
```suggestion
            final List<ScheduledTask> tasks = this.drainTasks();
```

src/main/java/org/spongepowered/common/scheduler/SpongeScheduler.java
```diff
@@ -30,25 +30,18 @@
...
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
-import java.util.UUID;
-import java.util.concurrent.Callable;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.Future;
-import java.util.concurrent.FutureTask;
-import java.util.concurrent.TimeUnit;
+import java.util.*;
```
gabizou, line 41: Again, style guide on imports.

src/main/java/org/spongepowered/common/scheduler/SpongeScheduler.java
```diff
@@ -315,4 +318,18 @@ public Future<?> execute(final Runnable runnable) {
             return null;
         });
     }
+
+    protected List<ScheduledTask> drainTasks() {
+        List<ScheduledTask> list = new LinkedList<>();
```
gabizou, line 323:
```suggestion
        final List<ScheduledTask> list = new LinkedList<>();
```

src/main/java/org/spongepowered/common/scheduler/SpongeScheduler.java
```diff
@@ -315,4 +318,18 @@ public Future<?> execute(final Runnable runnable) {
...
+        tasks.values().removeIf(x -> {
```
gabizou, line 324:
```suggestion
        this.tasks.values().removeIf(x -> {
```

src/main/java/org/spongepowered/common/scheduler/SpongeScheduler.java
```diff
@@ -315,4 +318,18 @@ public Future<?> execute(final Runnable runnable) {
...
+            if (x.cancel())
+                list.add(x);
```
gabizou, line 326:
```suggestion
            if (x.cancel()) {
                list.add(x);
            }
```

src/main/java/org/spongepowered/common/scheduler/SpongeScheduler.java
gabizou: Also, how is it guaranteed this method is not going to concurrently be accessed during ticks or adding tasks? It doesn't seem threadsafe in comparison if this method gets called in a few different ways.

gabizou (changes requested):
Some nits. Some concerns about introduced thread safety.

src/main/java/org/spongepowered/common/scheduler/SpongeScheduler.java
sunmisc: This method is also thread-safe, the collections that ConcurrentHashMap returns are thread-safe, also this method has protected access, which will not allow it to be used outside, also this method is possible under race (during pool closure), but it will not break anything because before that we terminate the pool, perhaps the close method is not very thought out, but It wasn't much better

ImMorpheus:

___

PR #3893, opened 2023-09-04
Support changing the item of certain projectiles like Snowballs

avaruus1:
Add support for changing the display item of `ThrowableItemProjectile`s like snowballs. 

This also removes the unnecessary requirement, that the display item of the thrown potion entity can only be splash or lingering potion. I assume this was necessary in the past, but it isn't anymore.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/provider/entity/ThrowableItemProjectileData.java
+++ b/src/main/java/org/spongepowered/common/data/provider/entity/ThrowableItemProjectileData.java
@@ -1,0 +1,27 @@
+package org.spongepowered.common.data.provider.entity;
+
+import net.minecraft.world.entity.projectile.ThrowableItemProjectile;
+import org.spongepowered.api.data.Keys;
+import org.spongepowered.common.data.provider.DataProviderRegistrator;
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/common/data/provider/entity/ThrowableItemProjectileData.java
```diff
@@ -0,0 +1,51 @@
...
+    private ThrowableItemProjectileData() {
+    }
+
+    // @formatter:off
+    public static void register(final DataProviderRegistrator registrator) {
+        registrator
+                .asMutable(ThrowableItemProjectile.class)
+                    .create(Keys.ITEM_STACK_SNAPSHOT)
+                        .get(h -> ItemStackUtil.snapshotOf(h.getItem()))
+                        .setAnd((h, v) -> {
```
Yeregorix: I believe `setAnd` was necessary to change the resulting boolean. Now that it is always `true` you can use `set` and simplify the code.

avaruus1: Done!

aromaa (approved):
LGTM. Thank you!

PR #3897, opened 2023-09-10
Fix loot table not being removed when RandomizableContainerBlockEntity is cleared

Lignium:
To remove/replace a chest in the world without loot drop, you need to get its `BlockEntity` using the API, check if it is a `CarrierBlockEntity`, and then clear the inventory. This works well when chests are filled with specific loot. But when they are not yet generated, meaning they have a loot table, the clear method (`Inventory#clear()`/`RandomizableContainerBlockEntity#clearContent()`) will not work properly. This PR fixes this issue.

https://bugs.mojang.com/browse/MC-136917

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -9,7 +9,7 @@ projectDescription=The SpongeAPI implementation targeting vanilla Minecraft and 
 
 mixinConfigs=mixins.sponge.accessors.json,mixins.sponge.api.json,mixins.sponge.concurrent.json,mixins.sponge.core.json,\
   mixins.sponge.entityactivation.json,mixins.sponge.exploit.json,mixins.sponge.inventory.json,mixins.sponge.movementcheck.json,\
-  mixins.sponge.tracker.json,mixins.sponge.ipforward.json,mixins.sponge.optimization.json,mixins.sponge.bugfix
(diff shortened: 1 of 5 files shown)


gabizou:
Is this a bugfix of Minecraft or a bugfix of the way our Inventory API is implemented?

Lignium:
> Is this a bugfix of Minecraft or a bugfix of the way our Inventory API is implemented?

I think Minecraft in general and the API in particular. Some vanilla commands also rely on this method, and accordingly will also drop loot from non-generated containers.

gabizou:
Out of curiosity, can this be tracked to a Mojira ticket? So in the future we can be sure this isn't out of necessity of functionality but a bugfix. I'd prefer we keep bug fixes in a category of mixins so we know the group of things to review as we migrate to newer Minecraft versions.

Lignium:
> Out of curiosity, can this be tracked to a Mojira ticket? So in the future we can be sure this isn't out of necessity of functionality but a bugfix. I'd prefer we keep bug fixes in a category of mixins so we know the group of things to review as we migrate to newer Minecraft versions.

Yes, such a ticket already exists - https://bugs.mojang.com/browse/MC-136917. I made some changes.

gabizou (approved):

PR #3898, opened 2023-09-10
Update SpongeForge to Minecraft 1.19.4

Yeregorix:
### Context
ModLauncher has added 4 [ModuleLayer](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ModuleLayer.html)s. Each with specific roles.

ModLauncher layers

We can choose which layer we want our classes to be loaded in, per jar by declaring the correct mod type in the mod manifest or the Java manifest.

Only the `GAME` layer is transformable.

### Notable changes
- Gradle plugin `architectury-loom` has been updated from `0.7.2-SNAPSHOT` to `1.1-SNAPSHOT` (we are still using a fork)
- Instead of remapping mojmap AW to srg AW, loom is now converting it to srg AT
- `applaunch` is now packaged in a separate jar because `IModLocator`s must be loaded in the `SERVICE` layer.
- A new sourceset `lang` has been added and is packaged in a separate jar because language providers must be loaded in the `PLUGIN` layer.
- We can no longer use Mixin on FML. `PluginContainer` is now implemented via a wrapper instead of a mixin.
- Libraries that must be transformed must now have a `GAMELIBRARY` mod type instead of `LIBRARY`. Eg: adventure.
- Everything is packaged with a custom made jar-in-jar strategy. The root jar is the service jar. It contains a mod jar and a lang jar.
- We used to register dummy mod containers for `sponge` and `spongeapi`. We now register proper mods so we can declare everything into a single `mods.toml`. As required by FML.
- `PluginModContainer` no longer extends `FMLModContainer`.
- We are now transforming any class containing the `@Listener` annotation. We add 3 instructions in `<clinit>` to obtain a privileged `MethodHandles.Lookup`.

### TODO
- ~~No longer shade AW transformer in SF because we are now using ATs.~~ This will be done later when updating ModLauncher in SpongeVanilla.


forge/src/mixins/java/org/spongepowered/forge/mixin/core/minecraftforge/registries/RegistryManagerMixin_Forge.java
```diff
@@ -22,12 +22,30 @@
...
 
+    @SuppressWarnings("UnstableApiUsage")
+    @Inject(method = "takeSnapshot", at = @At("RETURN"))
+    public void forge$dontSendSpongeDataToClient(boolean savingToDisc, CallbackInfoReturnable<Map<ResourceLocation, ForgeRegistry.Snapshot>> cir) {
+        if (savingToDisc) {
+            return;
+        }
+
+        for (ForgeRegistry.Snapshot snapshot : cir.getReturnValue().values()) {
+            snapshot.ids.keySet().removeIf(loc -> loc.getNamespace().equals("sponge"));
```
Yeregorix, line 48: I'm not sure filtering sponge data here is the correct solution. Also, this currently doesn't support any data registered by plugins.

Issue #3904, opened 2023-10-18
The item serializer cannot load enchantments by default.


SawFowl:
### Affected Product(s)

SpongeVanilla

### Version

SpongeVanilla: 1.19.4-10.0.0-RC1397

### Operating System

OS: Linux/5.16.15-arch1-1 (amd64)

### Java Version

JVM: 17.0.8.1/64-bit

### Plugins/Mods

```shell
My test plugins.
```

### Describe the bug

You need to create an item and add enchantments to it. Then write it to the config that has default serializers assigned to it. After that you need to load the item from the config and try to get enchantments from it. The enchantments list will be empty.

The first screenshot shows a test of loading item data using the default serializer.
![image](https://github.com/SpongePowered/Sponge/assets/29023558/29ad47c9-ad70-4af9-a25c-d958edf11a52)
The second screenshot shows the output of the item data using my serializer.  There was no change in the amount of data added to the item.
![image](https://github.com/SpongePowered/Sponge/assets/29023558/bf14e58a-caad-482d-aa9b-7e51f34fe9ed)

PR #3906, opened 2023-10-18
Enchantement ids were not retrieved correctly during deserialization

AlexandreArcil:
### Version
Minecraft: 1.19.4
SpongeAPI: 10.1.0-SNAPSHOT
Sponge: 1.19.4-10.1.0-SNAPSHOT
SpongeVanilla: 1.19.4-10.0.0-RC0

### The bug
The deserialization of an enchanted item in a config file produces an item without its enchantments.  
Reported by #3904.

### Description
During the deserilization of an enchanted item, it will first create a `CompoundTag` from `UnsafeData` and then fix the enchantment data. The problem is that the enchantment ids in the `CompoundTag` are String which represent the resource location of an enchantment but they were retrieved as short. As a result, the modified `CompoundTag` contained enchantment ids with "0s" as value.
After that, if the enchantments were retrieved from the itemstack, the Optional was empty because it couldn't retrieved the enchantment type from the incorrect value "0s".

### The fix
The fix is to retrieve and set as a string the enchantment id in the nbt and not as a short.

### DIFF
--- a/src/main/java/org/spongepowered/common/item/SpongeItemStack.java
+++ b/src/main/java/org/spongepowered/common/item/SpongeItemStack.java
@@ -356,10 +356,8 @@ public final class SpongeItemStack  {
             for (int i = 0; i < nbttaglist.size(); ++i)
             {
                 final CompoundTag nbttagcompound = nbttaglist.getCompound(i);
-                final String id = nbttagcompound.getString(Constants.Item.ITEM_ENCHANTMENT_ID);
                 final short lvl = nbttagcompound.getShort(Constants.Item.ITEM_ENCHANTMENT_LEVEL);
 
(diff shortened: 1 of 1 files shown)


SawFowl:
#3904

aromaa (changes requested):
Should be fine by just removing the enchantment id altogether as its a string and does not need fixup for its type.
```java
final CompoundTag nbttagcompound = nbttaglist.getCompound(i);
final short lvl = nbttagcompound.getShort(Constants.Item.ITEM_ENCHANTMENT_LEVEL);

nbttagcompound.putShort(Constants.Item.ITEM_ENCHANTMENT_LEVEL, lvl);
```

AlexandreArcil:
Done !

aromaa (approved):
LGTM. Thank you!

PR #3908, opened 2023-10-27
Fix minor issues with show_entity hover in messages

Anna-28:
The entity name was being used instead of the entity type, causing the type to not be shown to the user unless it happened to be the same as the entity name.

And in some cases such as an entity type (or name, before these changes) being "Player" it would cause an null pointer exception and not send the message at all.


ImMorpheus:
Thanks!

API 7 is EOL but the old Text API is no longer available in API 8+ because we're now using kyori.

This looks ok so I guess we can merge it. 

Note: this is untested.

PR #3910, opened 2023-10-27
Convert arrays from NBT to JSON.

Anna-28:
Arrays are not an Iterable instance like lists are.

Also accept long arrays when translating NBT.

This fixes plugins having errors when trying to convert NBT with int/long/byte arrays to JSON. Which usually happens with firework items (the Colors and FadeColors are int arrays).

However upon converting this JSON back to NBT, it will create lists of ints instead of int arrays etc.

### DIFF
--- a/src/main/java/org/spongepowered/common/data/persistence/JsonDataFormat.java
+++ b/src/main/java/org/spongepowered/common/data/persistence/JsonDataFormat.java
@@ -51,6 +51,7 @@ import java.io.StringWriter;
 import java.io.Writer;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/common/data/persistence/JsonDataFormat.java
```diff
@@ -211,7 +212,29 @@ private static void write(JsonWriter writer, @Nullable Object value) throws IOEx
...
+            writer.beginArray();
+            for (Object item : (Iterable<?>) value) {
+                write(writer, item);
+            }
+            writer.endArray();
+        } else if (value instanceof byte[]) {
+            writer.beginArray();
+            for (Object item : (byte[]) value) {
+                write(writer, item);
+            }
```
Yeregorix, line 223:
```java
for (byte item : (byte[]) value) {
    writer.write(item);
}
```
This would probably improve performance as it avoid boxing, recursion and instanceof.
Same goes for `int[]` and `long[]` below.

Anna-28: Thanks, not sure why I was still using `Object` in there.

ImMorpheus:
Hey! You seem to be running an unsupported version (API 7) of SpongeAPI. Support for API 7 ended in January 2022.

API-7 is now end-of-life. We will not be issuing any more API updates for version 7 of our API[^1]

As part of our legacy cleanup we are closing issues relating to API7/1.12.2[^2]
___

[^1]: https://forums.spongepowered.org/t/status-update-8th-january-2022-spongeapi-8-released/39861#minecraft-1122-spongeapi-7-10
[^2]: https://forums.spongepowered.org/t/sponge-status-update-28th-may-2022/40135#api-71122-community-support-8

Issue #3907, opened 2023-10-18
ServerSideConnectionEvent.Auth called twice per login


avaruus1:
### Affected Product(s)

SpongeVanilla

### Version

1.20.2-11.0.0-RC1386

### Operating System

Linux

### Java Version

18.0.1/64-bit

### Plugins/Mods

```shell
-
```

### Describe the bug

`ServerSideConnectionEvent.Auth` is called twice per login.

ImMorpheus:
PR merged

PR #3911, opened 2023-10-29
Auth event called twice

AlexandreArcil:
## Version
Minecraft: 1.20.2
SpongeAPI: 11.0.0-SNAPSHOT
Sponge: 1.20.2-11.0.0-SNAPSHOT
SpongeVanilla: 1.20.2-11.0.0-RC0

## The bug
During the login of a player, the event `ServerSideConnectionEvent.Auth` was called twice.
Reported by #3907.

## Description
When a player logs in, the event `ServerSideConnectionEvent.Auth` is called for the first time by `ServerLoginPacketListenerImplMixin` and then by `ServerLoginPacketListenerImpl_1Mixin` if the server is on online mode, but and not called if it is on offline mode.
The reason for this is due to how the mixin `ServerLoginPacketListenerImpl#impl$fireAuthEventOffline` was injected: at the assignment of the field _state_ with the value KEY. In API-10 (1.19.4), this was done in singleplayer or if the server is in offline mode, but on API-11 (1.20.2), this is done if the server is on online mode. Later in the login process, the mixin `ServerLoginPacketListenerImpl_1Mixin` will trigger the event if the player was authenticated on an online server.

## The fix
As the name suggests, `fireAuthEventOffline` should only call the event when the server is in offline mode.
The fix changes the injection point to the method `startClientVerification` because it's called when the server is in offline mode. It then checks whether if the game profile given as argument is an "offline profile". This is done by comparing it with the result of a call to the method `createOfflineProfile` because the profile argument comes from that method (ie: `this.startClientVerification(createOfflineProfile(this.requestedUsername))`).

`startClientVerification` is also called in other cases: when it is on a singleplayer, but we were in the context of a server, and when the player is authenticated to the server, but the check skips the event trigger.

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerLoginPacketListenerImplMixin.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerLoginPacketListenerImplMixin.java
@@ -185,21 +185,23 @@ public abstract class ServerLoginPacketListenerImplMixin implements ServerLoginP
         return event.isCancelled();
     }
 
-    @Inject(method = "handleHello(Lnet/minecraft/network/protocol/login/ServerboundHelloPacket;)V",
-        at = @At(
-            value = "FIELD",
-            target = "Lnet/minecraft/server/network/ServerLoginPacketListenerImpl$State;KEY:Lnet/minecraft/server/network/ServerLoginPacketListenerImpl$State;",
-            opcode = Opcodes.GETSTATIC),
-        cancellable = true)
-    private void impl$fireAuthEventOffline(final CallbackInfo ci) {
+    @Inject(method = "startClientVerification(Lcom/mojang/authlib/GameProfile;)V", at = @At("HEAD"), cancellable = true)
+    private void impl$fireAuthEventOffline(final GameProfile gameProfile, final CallbackInfo ci) {
         // Move this check up here, so that the UUID isn't null when we fire the event
         // TODO broken
         /*if (!this.authenticatedProfile.isComplete()) {
             this.authenticatedProfile = this.shadow$createFakeProfile(this.authenticatedProfile);
         }*/
 
-        if (this.bridge$fireAuthEvent()) {
-            ci.cancel();
+        if(gameProfile.equals(createOfflineProfile(gameProfile.getName()))) {
+            if (this.bridge$fireAuthEvent()) {
+                ci.cancel();
+            }
         }
     }
+
+    @Shadow protected static GameProfile createOfflineProfile(String $$0) {
+        throw new UnsupportedOperationException("Shadowed createOfflineProfile");
+    }
+
 }
--- a/src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerLoginPacketListenerImplMixin.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerLoginPacketListenerImplMixin.java
@@ -195,7 +195,6 @@ public abstract class ServerLoginPacketListenerImplMixin implements ServerLoginP
 
         if (this.bridge$fireAuthEvent()) {
             ci.cancel();
-            System.out.println("hello");
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerLoginPacketListenerImplMixin.java
```diff
@@ -185,21 +184,23 @@ public abstract class ServerLoginPacketListenerImplMixin implements ServerLoginP
...
-        if (this.bridge$fireAuthEvent()) {
-            ci.cancel();
+        if(gameProfile.equals(createOfflineProfile(gameProfile.getName()))) {
+            if (this.bridge$fireAuthEvent()) {
+                ci.cancel();
+            }
         }
     }
+
+    @Shadow protected static GameProfile createOfflineProfile(String $$0) {
```
ImMorpheus: Can you group this with the other shadow methods at the top of the class ?

src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerLoginPacketListenerImplMixin.java
```diff
@@ -185,21 +184,23 @@ public abstract class ServerLoginPacketListenerImplMixin implements ServerLoginP
...
     }
 
-    @Inject(method = "handleHello(Lnet/minecraft/network/protocol/login/ServerboundHelloPacket;)V",
-        at = @At(
-            value = "FIELD",
-            target = "Lnet/minecraft/server/network/ServerLoginPacketListenerImpl$State;KEY:Lnet/minecraft/server/network/ServerLoginPacketListenerImpl$State;",
-            opcode = Opcodes.GETSTATIC),
-        cancellable = true)
-    private void impl$fireAuthEventOffline(final CallbackInfo ci) {
+    @Inject(method = "startClientVerification(Lcom/mojang/authlib/GameProfile;)V", at = @At("HEAD"), cancellable = true)
```
ImMorpheus: I don't think we need to change the injection point.

In 1.19 this was injecting to right before retrieving the static field `ServerLoginPacketListenerImpl.State.READY_TO_ACCEPT` which was removed in 1.20 and replaced with the call to `startClientVerification`.

When updating to 1.20 the injection target was changed to `ServerLoginPacketListenerImpl.State.KEY` which is accessed when the server is online.

I think we should still inject in `handleHello` but update the `at` target to call before `startClientVerification`

src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerLoginPacketListenerImplMixin.java
```diff
@@ -185,21 +184,23 @@ public abstract class ServerLoginPacketListenerImplMixin implements ServerLoginP
...
+    private void impl$fireAuthEventOffline(final GameProfile gameProfile, final CallbackInfo ci) {
         // Move this check up here, so that the UUID isn't null when we fire the event
         // TODO broken
         /*if (!this.authenticatedProfile.isComplete()) {
             this.authenticatedProfile = this.shadow$createFakeProfile(this.authenticatedProfile);
         }*/
 
-        if (this.bridge$fireAuthEvent()) {
-            ci.cancel();
+        if(gameProfile.equals(createOfflineProfile(gameProfile.getName()))) {
```
ImMorpheus: See above, we should remove this condition and update the injection point to the correct method

ImMorpheus (changes requested):

AlexandreArcil:
Ok I changed the injection point to what you wanted !

ImMorpheus:
Thanks! Can you rebase on api-11 ?

AlexandreArcil:
Done!

ImMorpheus:
I don't think that's it, the PR has ~30 commits.

Please see https://docs.spongepowered.org/stable/en/contributing/howtogit.html

AlexandreArcil:
Ok I did some force push and now everything is correct, no ?

ImMorpheus:
Yep, thank you!

PR #3930, opened 2023-12-07
WIP Allow overriding offerSingle and removeSingle

aromaa:
This is a very rough draft just to show some outlines.

Tries to solve the problem that offerSingle retrieves the whole list, modifies it and then offers it back as a set. This works fine for most things but it gets a bit awkward for potion effects as they can contain attributes which are reset, for example absorption. This causes you to get new set of absorption hearts every time new potion effects are offered. This is also a performance improvement in a sense.

I tried to go the alternative route which was to compare if the same effects are offered but it feels a bit awkward and clumsy.

Some open questions are:
- Should we expose this in the API instead of doing `instanceof` checks?
- The generic mess.
- Is this worth it?


aromaa:
Resolved with 561da39 which has more robust implementation and can be extended with the API and is not just internal implementation detail.

PR #3953, opened 2024-02-26
Use ANSI serializer for console suggestion tooltips

jpenilla:


build.gradle.kts
```diff
@@ -169,6 +169,14 @@ dependencies {
         exclude(group = "org.jetbrains", module = "annotations")
         exclude(group = "org.checkerframework", module = "checker-qual")
     }
+    implementation(libs.adventure.serializerAnsi) {
+        exclude(group = "org.jetbrains", module = "annotations")
+        exclude(group = "org.checkerframework", module = "checker-qual")
+    }
+    implementation(libs.ansi) {
```
zml2008: Do we need to declare `ansi` directly? Shouldn't it be brought in as a transitive dep?

zml2008 (approved):

PR #3964, opened 2024-03-05
Replace Preconditions.checkState with Preconditions class

ImMorpheus:
get rid of guava checkstate preconditions and use stdlib function


Faithcaio:
I'd prefer one-liners for those checks

ImMorpheus (commented):
discussed with faith. I'll redo this adding a checkstate util method

aromaa:
It would be great if the JDK included these one-liners. This is the approach the .NET has taken and it has added these one-liners in the BCL itself.

PR #3965, opened 2024-03-05
Remove tab-list-remove-delay config option and fix human playerlistpacket

ImMorpheus:


aromaa:
This will not work if the client does not at least tick once before removing the tab list entry. Otherwise their skin will not load.

ImMorpheus:
Messed up the skin test. Thanks!

PR #3987, opened 2024-03-31
fix: Update vanish support

gabizou:


src/mixins/java/org/spongepowered/common/mixin/core/world/entity/ai/sensing/NearestLivingEntitySensorMixin.java
```diff
@@ -0,0 +1,58 @@
...
+import org.spongepowered.asm.mixin.injection.Redirect;
+import org.spongepowered.common.bridge.data.VanishableBridge;
+
+import java.util.List;
+import java.util.function.Predicate;
+
+@Mixin(NearestLivingEntitySensor.class)
+public class NearestLivingEntitySensorMixin<T extends LivingEntity> extends SensorMixin<T> {
+
+    @Redirect(method = "doTick", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/level/ServerLevel;getEntitiesOfClass(Ljava/lang/Class;Lnet/minecraft/world/phys/AABB;Ljava/util/function/Predicate;)Ljava/util/List;"))
```
gabizou, line 43: This is the most "vanilla compatible" way of avoiding targeting vanished entities since it only will affect "Senses AI" vs the more "surefire" way in LevelMixin. I'm debating on it for mod compatibility however, since mods can have their own AI implementations.

Thoughts @Zidane ?

src/mixins/java/org/spongepowered/common/mixin/core/world/level/LevelMixin.java
```diff
@@ -238,4 +243,29 @@ public abstract class LevelMixin implements LevelBridge, LevelAccessor {
 
         return (E) entity;
     }
+
+    @Inject(method = {
+        "getEntities(Lnet/minecraft/world/entity/Entity;Lnet/minecraft/world/phys/AABB;Ljava/util/function/Predicate;)Ljava/util/List;",
+        "getEntities(Lnet/minecraft/world/level/entity/EntityTypeTest;Lnet/minecraft/world/phys/AABB;Ljava/util/function/Predicate;)Ljava/util/List;",
+    }, at = @At("RETURN"))
+    private void impl$IgnoreTargetingOfVanishedEntities(
```
gabizou, line 249: This is a little drastic, and I may want to take advantage of the PhaseTracker to handle when vanishing logic comes into play (specifically for AI ticks, or overall).

ImMorpheus (approved):

PR #3995, opened 2024-04-23
[NO MERGE] Contextual Data API

aromaa:
**Sponge** | [SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2496)

Implemented data directions:
- [x] Entity -> Entity
- [x] Entity -> Team
- [ ] Entity -> World
- [x] Team -> Entity
- [x] Team -> Team
- [ ] Team -> World
- [ ] World -> Entity
- [ ] World -> Team
- [ ] World -> World

Implemented data operations:
- [x] offer
- [ ] offerSingle
- [ ] offerAll
- [ ] removeSingle
- [ ] removeAll
- [x] remove


src/main/java/org/spongepowered/common/data/contextual/ContextualDataDelegate.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @Override
+    public @Nullable PerspectiveContainer<?, ?> getDataPerception(final DataPerspective perspective) {
+        return this.perspectives.get(perspective);
+    }
+
+    @Override
+    public PerspectiveContainer<?, ?> createDataPerception(final DataPerspective perspective) {
+        return this.perspectives.computeIfAbsent(perspective, p -> {
+            if (p instanceof final Entity entity) {
```
Zidane, line 48: Maybe add a bridge to all the perspectives implement that returns the container? Feels cleaner than this if chain that may keep growing over time.

aromaa: Could be done but it can also be super confusing because the `PerspectiveContainer` have different behavior depending on is the entity itself the holder as in `ContextualDataHolder` or some other type like `Team` which uses `ContextualDataDelegate` that propagates the updates differently.

The difference is in the implementation of `PerspectiveContainer#offer`

PR #3997, opened 2024-04-25
Rework ChunkEvent

aromaa:
**Sponge** | [SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2497)

### DIFF
--- a/src/main/java/org/spongepowered/common/world/level/chunk/storage/SpongeChunkEntities.java
+++ b/src/main/java/org/spongepowered/common/world/level/chunk/storage/SpongeChunkEntities.java
@@ -1,0 +1,233 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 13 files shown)


src/main/java/org/spongepowered/common/world/level/chunk/storage/SpongeChunkEntities.java
```diff
@@ -0,0 +1,233 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+public final class SpongeChunkEntities implements EntityChunk {
```
Zidane: `SpongeEntityChunk`

src/main/java/org/spongepowered/common/world/level/chunk/storage/SpongeChunkEntities.java
```diff
@@ -0,0 +1,233 @@
...
+    private final Level level;
+    private final Vector3i chunkPosition;
+    private final Stream<net.minecraft.world.entity.Entity> entities;
+
+    private @MonotonicNonNull SpongeChunkLayout chunkLayout;
+    private @MonotonicNonNull Vector3i blockMin;
+    private @MonotonicNonNull Vector3i blockMax;
+    private @MonotonicNonNull List<net.minecraft.world.entity.Entity> newEntities;
+
+    public SpongeChunkEntities(final Level chunkLayout, final Vector3i chunkPosition, final Stream<net.minecraft.world.entity.Entity> entities) {
```
Zidane: `chunkLayout` -> `level`

Zidane: Also should be `ServerLevel` as this would never be client-side (unless you are seeing something here I'm not seeing).

aromaa:
Merged manually

PR #4006, opened 2024-05-09
SpongeForge for Minecraft 1.20.6

Yeregorix:


forge/src/mixins/java/org/spongepowered/forge/mixin/core/server/network/ServerLoginPacketListenerImplMixin_Forge.java
```diff
@@ -24,51 +24,27 @@
...
-        }
+    @Shadow private ServerLoginPacketListenerImpl.State state;
+    // @formatter:on
 
-        if (this.bridge$fireAuthEvent()) {
-            ci.cancel();
+    @Inject(method = "tick", at = @At("HEAD"))
+    private void impl$onTick(final CallbackInfo ci) {
+        // In SpongeVanilla we do channel registration during this state, not sure if we need to do anything in SpongeForge
+        if (this.state == ServerLoginPacketListenerImpl.State.NEGOTIATING) {
```
Zidane, line 46: Why does this change exist?

src/mixins/java/org/spongepowered/common/mixin/core/network/ConnectionMixin.java
```diff
@@ -156,13 +155,6 @@ public abstract class ConnectionMixin extends SimpleChannelInboundHandler<Packet
         this.impl$version = new SpongeMinecraftVersion(String.valueOf(version), version);
     }
 
-    @Inject(method = "lambda$doSendPacket$10", at = @At(value = "INVOKE", target = "Lio/netty/util/concurrent/Future;isSuccess()Z"))
```
Zidane: Why was this removed?

forge/src/mixins/java/org/spongepowered/forge/mixin/core/server/network/ServerLoginPacketListenerImplMixin_Forge.java
Yeregorix: See https://github.com/SpongePowered/Sponge/blob/api11/forge/src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerLoginPacketListenerImplMixin.java#L146
and https://github.com/SpongePowered/Sponge/blob/api11/forge/vanilla/src/mixins/java/org/spongepowered/vanilla/mixin/core/server/network/ServerLoginPacketListenerImplMixin_Vanilla.java#L80

After auth, we switch to state `NEGOTIATING` which is unused by vanilla MC. We use this state in SpongeVanilla to register channels and then we go to `VERIFYING` which will be handled by vanilla MC.

I had to replicate the `NEGOTIATING` -> `VERIFYING` switch in SpongeForge otherwise the login stay forever in state `NEGOTIATING`.

src/mixins/java/org/spongepowered/common/mixin/core/network/ConnectionMixin.java
Yeregorix: Moved to `ConnectionMixin_Vanilla` and `ConnectionMixin_Forge` due to injection error.

PR #4009, opened 2024-05-14
Add Keys.MAX_GROWTH_STAGE

MrHell228:
I removed Keys.GROWTH_STAGE from saplings because it used `STAGE` property instead of `AGE` properties.
And those 2 has to be differed because there is `BambooStalkBlock` that use both `STAGE` and `AGE` properties.
So I think it would be better to keep only age-related blocks in this key.


aromaa:
Saplings should still implement the `Keys.GROWTH_STAGE`. While the underlying property is `STAGE`, its still used for growth unlike where for bamboo its used to to limit growth.

MrHell228:
Makes sense, done

src/main/java/org/spongepowered/common/data/provider/block/state/GrowthData.java
```diff
@@ -24,27 +24,54 @@
...
                 .asImmutable(BlockState.class)
                     .create(Keys.GROWTH_STAGE)
-                        .constructValue((h, v) -> BoundedUtil.constructImmutableValueInteger(v, Keys.GROWTH_STAGE, ((CropBlockAccessor) h.getBlock()).invoker$getAgeProperty()))
-                        .get(h -> h.getValue(((CropBlockAccessor) h.getBlock()).invoker$getAgeProperty()))
-                        .set((h, v) -> BoundedUtil.setInteger(h, v, ((CropBlockAccessor) h.getBlock()).invoker$getAgeProperty()))
-                        .supports(h -> h.getBlock() instanceof CropBlock);
+                        .constructValue((h, v) -> BoundedUtil.constructImmutableValueInteger(v, Keys.GROWTH_STAGE, propertyFunction.apply(h)))
+                        .get(h -> h.getValue(propertyFunction.apply(h)))
+                        .set((h, v) -> BoundedUtil.setInteger(h, v, propertyFunction.apply(h)))
+                        .supports(cropClass::isInstance)
```
aromaa:
```suggestion
                        .supports(h -> cropClass.isInstance(h.getBlock()))
```

src/main/java/org/spongepowered/common/data/provider/block/state/GrowthData.java
```diff
@@ -24,27 +24,54 @@
...
+                    .create(Keys.MAX_GROWTH_STAGE)
+                        .get(h -> DataUtil.maxi(propertyFunction.apply(h)))
+                        .supports(cropClass::isInstance);
```
aromaa:
```suggestion
                        .supports(h -> cropClass.isInstance(h.getBlock()));
```

aromaa (changes requested):
This wasn't actually working because the supports condition is wrong. Otherwise LGTM. Nice simplification too :)

MrHell228:
Oh I think I did wrong thing by making new commit instead of using "commit suggestion" button and now I can't press it.

aromaa (approved):
No worries, doesn't really matter which way you do it. Thank you :)

PR #4020, opened 2024-05-31
TeleportDuration Key for Display Entities

DrZoddiak:
`teleport_duration` was added in MC 1.20.2

https://github.com/SpongePowered/SpongeAPI/pull/2508 | Sponge


src/main/java/org/spongepowered/common/data/provider/entity/DisplayEntityData.java
```diff
@@ -91,6 +91,15 @@ public static void register(final DataProviderRegistrator registrator) {
...
                         })
+                    .create(Keys.TELEPORT_DURATION)
+                    .get(h -> Ticks.of(h.invoker$getTeleportDuration()))
+                    .setAnd((h, v) -> {
+                        if (v.isInfinite()) {
+                            return false;
+                        }
+                        h.invoker$setTeleportDuration(SpongeTicks.toSaturatedIntOrInfinite(v));
+                        return true;
+                    })
```
aromaa: Formatting
```suggestion
                        .get(h -> Ticks.of(h.invoker$getTeleportDuration()))
                        .setAnd((h, v) -> {
                            if (v.isInfinite()) {
                                return false;
                            }
                            h.invoker$setTeleportDuration(SpongeTicks.toSaturatedIntOrInfinite(v));
                            return true;
                        })
```

testplugins/src/main/java/org/spongepowered/test/entity/DisplayEntityTest.java
```diff
@@ -224,8 +226,19 @@ public void onRegisterCommand(final RegisterCommandEvent<Parameterized> event) {
...
-                        textDisplay = createEntity(player.world(), EntityTypes.TEXT_DISPLAY, centerPos, forwardDir, col6, 0);
-                        textDisplay.offer(Keys.DISPLAY_NAME, Component.text("Look at this interpolation").color(NamedTextColor.RED));
+                        blockDisplay = createEntity(player.world(), EntityTypes.BLOCK_DISPLAY, centerPos, forwardDir, col6, 0);
+                        blockDisplay.offer(Keys.BLOCK_STATE, BlockTypes.OBSIDIAN.get().defaultState());
+                        blockDisplay.offer(Keys.TRANSFORM, Transform.of(blockCenterOffset)); // set initial value before spawning
+                        blockDisplay.offer(Keys.SHADOW_RADIUS, 2d); // set initial value before spawning
+                        blockDisplay.offer(Keys.SHADOW_STRENGTH, 5d); // set initial value before spawning
+                        player.world().spawnEntity(blockDisplay);
+                        blockDisplay.offer(Keys.TELEPORT_DURATION,Ticks.of(20));
+                        blockDisplay.offer(Keys.INTERPOLATION_DELAY,Ticks.of(20));
```
aromaa: Might as well fix this while at it.
```suggestion
                        blockDisplay.offer(Keys.TELEPORT_DURATION, Ticks.of(20));
                        blockDisplay.offer(Keys.INTERPOLATION_DELAY, Ticks.of(20));
```

aromaa (commented):
Looking at Minecraft's sources they seem to clamp this between 0 and 59 on serialization. Either we should do that too or serialize them internally separately.

In general, where is this useful? What can you do with this that is not possible with the transformation interpolation?

DrZoddiak:
> In general, where is this useful? What can you do with this that is not possible with the transformation interpolation?

It is my understanding that Transformations don't actually move the entity.

In this case the teleportation is explicitly moving the entity and interpolating it from one spot to the other.

So with transformations you'd have to actively update the entities position to match behavior if I understand correctly.

PR #4023, opened 2024-06-02
Add Viewer#sendBlockProgress

MrHell228:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2513) | **Sponge**

1. I think getting progress stage from progress should be done in some different place but I'm not sure which one
2. Maybe `SpongeServer#getOrCreateBlockDestructionId()` should be changed to just `SpongeServer#createBlockDestructionId()`


src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/level/LevelMixin_API.java
```diff
@@ -281,24 +284,33 @@ public void stopMusicDisc(final Vector3i position) {
...
-                .map(ServerPlayer.class::cast)
-                .forEach(p -> p.connection.send(packet));
+    @Override
+    public void sendBlockProgress(final int x, final int y, final int z, final float progress) {
+        if (progress < 0 || progress >= 1) {
+            throw new IllegalArgumentException("Progress must be between 0 and 1");
+        }
+
+        final BlockPos pos = new BlockPos(x, y, z);
+        final int id = ((SpongeServer) SpongeCommon.server()).getOrCreateBlockDestructionId(pos);
```
Yeregorix: Get the server from the player
Avoid static getter whenever possible

src/main/java/org/spongepowered/common/SpongeServer.java
```diff
@@ -47,6 +49,10 @@ public interface SpongeServer extends SpongeEngine, Server {
 
     UsernameCache getUsernameCache();
 
+    @Nullable Integer getBlockDestructionId(BlockPos pos);
```
Faithcaio, line 52: Shouldn't that be on the world?

MrHell228: Minecraft renderer makes no difference between worlds regarding breaking progress so making this per-world may lead to some weird interferences. For example:
1. Player in World №1. We send progress to block №1.
2. Player travels to World №2. We send progress to block №2.
3. Player travels back to World №1. And 2 results can happen:
3.1. If packet in (2) had the same id as in (1) (pretty random thing), progress on block №1 will be reset and player will not see it.
3.2. If packets had different ids, progress on block №1 will remain and player will see it.

This is some kind of indeterminacy and generally I don't like it. But maybe we don't even want to somehow support this described use case so I'm not sure.

Faithcaio:
squashed and rebased in https://github.com/SpongePowered/Sponge/commit/dbe01eaaa536ea85caf9c00f02e41444cbb0034a

PR #4030, opened 2024-06-14
Some viewer-related fixes

MrHell228:
Viewer#sendWorldType():
- DimensionType effects are contained in `ClientLevel#effects` so we need to modify it instead of modifying actual DimensionType
- Removed `implements LevelBridge` from `ClientLevelMixin` because `LevelMixin` already implements it

BookItemStackData:
- Don't know if we want to expose `WrittenBookContent` or keep current keys but for now I just made it working with current stuff

BookUtil:
- There is no more need to manually localize book content because it's done on client
- `bookSlot` was changed from `inventory.items.size() + inventory.selected` to `inventory.selected` and container id was changed from `0` to `-2` because the only reason it works now is that (picture) size of red area (`inventoryMenu` indexing) luckily equals to size of blue area (`inventory.items.size()`). `-2` stands for `Player#inventory` and first 9 items in it is hotbar
image

### DIFF
--- a/src/main/java/org/spongepowered/common/data/provider/item/stack/BookItemStackData.java
+++ b/src/main/java/org/spongepowered/common/data/provider/item/stack/BookItemStackData.java
@@ -55,8 +55,7 @@ public final class BookItemStackData {
                             return LegacyComponentSerializer.legacySection().deserialize(content.author());
                         })
                         .set((h, v) -> {
-                            final WrittenBookContent content = h.get(DataComponents.WRITTEN_BOOK_CONTENT);
-                            // TODO handle missing data?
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/common/data/provider/item/stack/BookItemStackData.java
```diff
@@ -71,11 +70,10 @@ public static void register(final DataProviderRegistrator registrator) {
                             return content.generation();
                         })
                         .setAnd((h, v) -> {
-                            if (v < 0) {
+                            if (v < 0 || v > 3) {
```
Faithcaio, line 73: Is it actually a limitation or just that you cant craft it in vanilla gameplay?

MrHell228: Actual limitation. Also transaltion keys exist only for those generation values
image
image

PR #4042, opened 2024-06-22
Add new EntityTypes mixins

ImMorpheus:


src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/entity/animal/armadillo/ArmadilloMixin_API.java
Faithcaio, line 1: already exists but in wrong package

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/entity/projectile/BreezeWindChargeMixin_API.java
Faithcaio, line 1: already exists (but is not extending the AbstractWindCharge)
also this is the wrong package

Faithcaio (approved):

PR #4054, opened 2024-06-23
Implement ForwardingViewer & Viewer cleanup

MrHell228:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2526) | **Sponge**

This PR should be merged after these PRs (so I could resolve here all the merge conflicts with this PR that will likely appear):
- [Add Viewer#sendBlockProgress](https://github.com/SpongePowered/Sponge/pull/4023)
  - [x] Merged
  - [x] Bumped to api-12
  - [x] Conflicts resolved here
- [Some viewer-related fixes](https://github.com/SpongePowered/Sponge/pull/4030)
  - [x] Merged
  - [x] Bumped to api-12
  - [x] Conflicts resolved here
- [Update particles](https://github.com/SpongePowered/Sponge/pull/4032)
  - [x] Merged
  - [x] Conflicts resolved here

Some general points:
- `ViewerPacketUtil` was made to collect all viewer-related packets and some auidience-related packets in one place and simplify work with them instead of having a lot of the same code in different `Viewer` impls.
- `SpongeViewer` is default `Viewer` impl, `SpongeForwardingViewer` is default `ForwardingViewer` impl. 
- All viewer methods just send some packet to _someone_. So `ViewerBridge` comes into play here with `send(Packet)` methods. This allows to move most `Viewer` impl to `SpongeViewer`.
- `ViewerBridge` has different impls for `ServerPlayer` and `LocalPlayer` (to mostly support both sided `Viewer` calls) and  `ForwardingViewer`
- `SpongeForwardingViewer` has some audience methods overriden with `SpongeViewer.super.someMethod()` to make single packet and broadcast it instead of making new packet for each viewer


src/main/java/org/spongepowered/common/bridge/effect/ViewerBridge.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.spongepowered.common.bridge.effect;
+
+import net.minecraft.network.protocol.Packet;
+import net.minecraft.network.protocol.game.ClientGamePacketListener;
+
+public interface ViewerBridge {
+
+    // Used to send sponge custom packets
+    default void bridge$sendToViewer(final org.spongepowered.api.network.channel.packet.Packet packet) {
```
Faithcaio:
```suggestion
    default void bridge$sendSpongePacketToViewer(final org.spongepowered.api.network.channel.packet.Packet packet) {
```
makes it easier to understand that this is different that the other one

MrHell228:
Resolved conflicts with sub PRs.

Currently block progress methods work only on server.
If we want them to work with client side plugins too (which is one of the goals of this whole PR), we might move `blockDescructionId` methods from `SpongeServer` to `SpongeEngine`.
Collisions between client-sent and server-sent ids in theory could be minimized by having different id counters: On server from `0` going down and on client from `Integer.MIN_VALUE` going up.
Does this sound fine?

src/main/java/org/spongepowered/common/effect/SpongeViewer.java
```diff
@@ -0,0 +1,148 @@
...
+        ((ViewerBridge) this).bridge$sendToViewer(ViewerPacketUtil.blockUpdate(x, y, z, state));
+    }
+
+    @Override
+    default void resetBlockChange(final int x, final int y, final int z) {
+    }
+
+    @Override
+    default void sendBlockProgress(final int x, final int y, final int z, final double progress) {
+        final Engine engine;
```
Yeregorix: Your current implementation may pick the `Server` engine when calling the method on a `ClientPlayer`.
IMO Engine should be determined by the viewer type. Eg:
`ServerPlayer` -> `Server` engine
`ClientPlayer` -> `Client` engine
You could add a `engine()` method in this interface, and implement it in the subclasses.
Or alternatively, directly implement `sendBlockProgress` in the subclasses.

MrHell228: Settled on direct implementation in subclasses. Luckily we already have `World#engine()`

PR #4110, opened 2024-07-13
Fix client crash with portals

MrHell228:
Fixes `java.lang.ClassCastException: class net.minecraft.client.multiplayer.ClientLevel cannot be cast to class net.minecraft.server.level.ServerLevel` when player tries to enter portal with Sponge on client (running it in ide).


src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
```diff
@@ -424,6 +424,10 @@ public Entity changeDimension(final DimensionTransition transition) {
     @Inject(method = "setAsInsidePortal", at = @At(value = "FIELD", opcode = Opcodes.PUTFIELD, shift = At.Shift.AFTER,
             target = "Lnet/minecraft/world/entity/Entity;portalProcess:Lnet/minecraft/world/entity/PortalProcessor;"))
     public void impl$onCreatePortalProcessor(final Portal $$0, final BlockPos $$1, final CallbackInfo ci) {
+        if (this.shadow$level().isClientSide()) {
```
ImMorpheus: should check if level is fake, see bridge$isFake

ImMorpheus (changes requested):

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/EntityMixin.java
MrHell228: done

ImMorpheus (approved):

PR #4111, opened 2024-07-15
Implement scaling for display entities

DrZoddiak:
Implements `#setScale` for Display Entities since they allow for scaling in 3 directions.


boxbeam (commented):
yes good

avaruus1:
How does this differ from setting the scale through the Transform or is it just a different way of doing the same thing?

DrZoddiak:
> How does this differ from setting the scale through the Transform or is it just a different way of doing the same thing?

Effectively it's the same thing. Interpolation keys still effect the entity. This PR simply implements a method that already exists but has no implementation. 

<https://github.com/SpongePowered/Sponge/blob/994fb277df1885974c03289f33563a617c2d355b/src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/entity/EntityMixin_API.java#L147-L155>

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/entity/DisplayMixin_API.java
```diff
@@ -24,11 +24,30 @@
...
+    // @formatter:on
+
+    @Override
+    public void setScale(Vector3d scale) {
+        entityData.set(DATA_SCALE_ID, new Vector3f((float) scale.x(), (float) scale.y(), (float) scale.z()));
+    }
+
+    @Override
+    public Vector3d scale() {
+        Vector3f scale = entityData.get(DATA_SCALE_ID);
```
gabizou: Two small nits: explicitly reference the field by `DisplayMixin_API.DATA_SCALE_ID`

gabizou (approved):
One small nit, and I'm good with merging.

PR #4117, opened 2024-08-03
fix: Issues related to fireworks and fused explosives

### LINKED ISSUE #4116: Firework effects of a firework rocket are not synchronized with client
### Affected Product(s)

SpongeVanilla

### Version

1.20.6-11.0.1-RC1778

### Operating System

Linux (Fedora 40)

### Java Version

openjdk 21.0.3

### Plugins/Mods

```shell
- Custom plugin (called 'lightshow')
```

### Describe the bug

Whenever `FIREWORK_EFFECTS` is modified on a firework rocket entity, the data is not synchronized with the client. Modified data can bee seen using the `/data` command, but the client doesn't show any firework effects.

### Reproducing

When using registering the following command:

```java
final Command.Parameterized fireworkCommand = Command.builder()
                .executor(context -> {
                    final ServerPlayer player = context.cause().first(ServerPlayer.class).orElseThrow();
                    final ServerWorld world = player.world();
                    final FireworkRocket rocket = world.createEntity(EntityTypes.FIREWORK_ROCKET, player.position());
                    rocket.offer(Keys.FIREWORK_EFFECTS, List.of(FireworkEffect.builder().color(Color.RED).shape(FireworkShapes.BURST).build()));
                    rocket.offer(Keys.FIREWORK_FLIGHT_MODIFIER, Ticks.of(1));
                    world.spawnEntity(rocket);
                })
                .build()
```

and then executing the command as a player, the rocket is spawned and flies for a bit, but when it should explode with the specified firework effects, it just disappears instead.

### Link to logs

https://paste.gg/p/anonymous/af2f41e0bfec4802904a23312cbcc07d

pandier:
Fixes #4116 

- synchronizes firework entity data changes with client
- implements `IS_PRIMED` for all fused explosives
- fixes `TICKS_REMAINING` throwing an error for anything other than `PrimedTNT`
- fixes `FUSE_DURATION` always rejecting
- ~~`detonate` method of Creeper only igniting, not exploding it (other methods explode)~~
- allows to modify flight duration of a firework item with `FIREWORK_FLIGHT_MODIFIER` (not sure if this is wanted)

### DIFF
--- a/src/main/java/org/spongepowered/common/util/FireworkUtil.java
+++ b/src/main/java/org/spongepowered/common/util/FireworkUtil.java
@@ -44,7 +44,7 @@ import java.util.function.Function;
 
 public final class FireworkUtil {
     public static boolean setFireworkEffects(final FireworkRocketEntity firework, final List<? extends FireworkEffect> effects) {
-        return updateFireworkRocketItem(firework, item -> setFireworkEffects(item, effects));
+        return FireworkUtil.updateFireworkRocketItem(firework, item -> FireworkUtil.setFireworkEffects(item, effects));
     }
 
     public static boolean setFireworkEffects(final ItemStack item, final List<? extends FireworkEffect> effects) {
@@ -69,7 +69,7 @@ public final class FireworkUtil {
     }
 
     public static Optional<List<FireworkEffect>> getFireworkEffects(final FireworkRocketEntity firework) {
-        return getFireworkEffects(getItem(firework));
+        return FireworkUtil.getFireworkEffects(FireworkUtil.getItem(firework));
     }
 
     public static Optional<List<FireworkEffect>> getFireworkEffects(final ItemStack item) {
@@ -94,7 +94,7 @@ public final class FireworkUtil {
     }
 
     public static boolean removeFireworkEffects(final FireworkRocketEntity firework) {
-        return updateFireworkRocketItem(firework, FireworkUtil::removeFireworkEffects);
+        return FireworkUtil.updateFireworkRocketItem(firework, FireworkUtil::removeFireworkEffects);
     }
 
     public static boolean removeFireworkEffects(final ItemStack item) {
@@ -117,7 +117,7 @@ public final class FireworkUtil {
     }
 
     public static boolean setFlightModifier(final FireworkRocketEntity firework, final int modifier) {
-        if (updateFireworkRocketItem(firework, item -> setFlightModifier(item, modifier))) {
+        if (FireworkUtil.updateFireworkRocketItem(firework, item -> FireworkUtil.setFlightModifier(item, modifier))) {
             int lifetime = 10 * modifier + ((EntityAccessor) firework).accessor$random().nextInt(6) + ((EntityAccessor) firework).accessor$random().nextInt(7);
             ((FireworkRocketEntityAccessor) firework).accessor$lifetime(lifetime);
             return true;
@@ -138,7 +138,7 @@ public final class FireworkUtil {
     }
 
     public static OptionalInt getFlightModifier(final FireworkRocketEntity firework) {
-        return getFlightModifier(getItem(firework));
+        return FireworkUtil.getFlightModifier(FireworkUtil.getItem(firework));
     }
 
     public static OptionalInt getFlightModifier(final ItemStack item) {
@@ -150,7 +150,7 @@ public final class FireworkUtil {
     }
 
     public static boolean updateFireworkRocketItem(final FireworkRocketEntity firework, final Function<ItemStack, Boolean> function) {
-        final ItemStack item = getItem(firework).copy();
+        final ItemStack item = FireworkUtil.getItem(firework).copy();
         if (function.apply(item)) {
             firework.getEntityData().set(FireworkRocketEntityAccessor.accessor$DATA_ID_FIREWORKS_ITEM(), item);
             return true;
--- a/src/main/java/org/spongepowered/common/util/FireworkUtil.java
+++ b/src/main/java/org/spongepowered/common/util/FireworkUtil.java
@@ -32,37 +32,47 @@ import net.minecraft.world.item.component.FireworkExplosion;
 import net.minecraft.world.item.component.Fireworks;
 import org.spongepowered.api.item.FireworkEffect;
 import org.spongepowered.api.item.ItemTypes;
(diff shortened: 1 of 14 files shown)


src/main/java/org/spongepowered/common/util/FireworkUtil.java
```diff
@@ -32,37 +32,47 @@
...
 
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.OptionalInt;
+import java.util.function.Function;
 
 public final class FireworkUtil {
+    public static boolean setFireworkEffects(final FireworkRocketEntity firework, final List<? extends FireworkEffect> effects) {
+        return updateFireworkRocketItem(firework, item -> setFireworkEffects(item, effects));
```
aromaa: Keep the type qualifiers, there are few others.

```suggestion
        return FireworkUtil.updateFireworkRocketItem(firework, item -> FireworkUtil.setFireworkEffects(item, effects));
```

src/main/java/org/spongepowered/common/util/FireworkUtil.java
```diff
@@ -103,23 +116,54 @@ public static boolean removeFireworkEffects(final Object object) {
         return false;
     }
 
-    public static ItemStack getItem(final FireworkRocketEntity firework) {
-        ItemStack item = firework.getEntityData().get(FireworkRocketEntityAccessor.accessor$DATA_ID_FIREWORKS_ITEM());
+    public static boolean setFlightModifier(final FireworkRocketEntity firework, final int modifier) {
```
aromaa, line 119: I think it would be better to leave the item alone here and just change the fireworks lifetime unconditionally, this way you can change it for any item.

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/entity/monster/CreeperMixin_API.java
```diff
@@ -36,19 +36,21 @@
...
     // @formatter:off
-    @Shadow public void shadow$ignite() { } // explode
+    @Shadow protected abstract void shadow$explodeCreeper();
     // @formatter:on
 
+
     @Override
     public void detonate() {
-        this.shadow$ignite();
+        this.shadow$explodeCreeper();
```
aromaa: This is going to be a breaking change due to the behavior change. We would want to have another API method to allow igniting the creeper. You can revert it for now and we can do this in another PR to get this merged.

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/item/PrimedTntMixin.java
```diff
@@ -129,7 +136,8 @@ public abstract class PrimedTntMixin extends EntityMixin implements PrimedTntBri
 
     @Inject(method = "tick()V", at = @At("RETURN"))
     private void impl$updateTNTPushPrime(final CallbackInfo ci) {
-        if (this.shadow$getFuse() == this.bridge$fuseDuration - 1 && !this.shadow$level().isClientSide) {
+        if (!impl$postPrimeTriggered && !this.shadow$level().isClientSide) {
```
aromaa: What was this change for? Also remember the this qualifier (and other places).

aromaa (changes requested):
Overall looks good, few comments.

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/item/PrimedTntMixin.java
pandier: I made this change because every time you change the fuse duration, the post prime event would get called

src/mixins/java/org/spongepowered/common/mixin/api/minecraft/world/entity/monster/CreeperMixin_API.java
pandier: Just to clarify, the original behavior wasn't implemented correctly based on the API. The `detonate` method should trigger an immediate explosion ("Instructs this explosive to detonate as soon as possible.") as seen with the TNT minecart (you can only make it explode, you cannot ignite it with the API). I will still revert this change.

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/item/PrimedTntMixin.java
aromaa: Ah, right. You also changed that the current fuse time is changed when then duration is modified.

I think the general idea here is that you can set the fuse duration independently of the current remaining ticks. So lets say we have already ticked 5 times and now you change the fuse duration to 10. That would now mean we explode after 5 more ticks instead of 10.

aromaa: Wouldn't you also hit this when you set the remaining ticks to the fuse duration? So keeping the `impl$postPrimeTriggered` around would still be beneficial?

pandier: Oh yeah, that's true! Sorry!

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/monster/CreeperMixin.java
```diff
@@ -77,14 +77,19 @@ public abstract class CreeperMixin extends MonsterMixin implements FusedExplosiv
         this.explosionRadius = radius == null ? Constants.Entity.Creeper.DEFAULT_EXPLOSION_RADIUS : radius;
     }
 
+    @Override
+    public boolean bridge$isPrimed() {
+        return shadow$isIgnited() || shadow$getSwellDir() == Constants.Entity.Creeper.STATE_PRIMED;
```
aromaa:
```suggestion
        return this.shadow$isIgnited() || this.shadow$getSwellDir() == Constants.Entity.Creeper.STATE_PRIMED;
```

aromaa (approved):
LGTM! Thank you!!

Another thing I noticed was that the FIREWORK_FLIGHT_MODIFIER is `Ticks` but we more of treat it as `int` which seems problematic.. There is also no direct way to change the lifetime as actual `Ticks` which would be helpful, but all of those are unrelated here.

PR #4120, opened 2024-08-10
Fix channels on SpongeForge

### LINKED ISSUE #4115: Attempting to send custom clientbound packet errors & kicks player
### Affected Product(s)

SpongeForge

### Version

1.20.6-50.0.22-11.0.1-RC1778

### Operating System

Windows

### Java Version

21.0.4/64-bit (Eclipse Adoptium)

### Plugins/Mods

```shell
Just SpongeForge and my mod :-)
```

### Describe the bug

Attempting to send a custom packet to a player results in an exception being thrown and the player disconnected, on SpongeForge. I did briefly test on SpongeVanilla and did not see any exceptions, (although I can't 100% confirm if I was actually receiving those packets, will update tomorrow as I test more). Repro code looks something like:

```java
public class PluginChannel {
	private RawDataChannel channel;

	@Listener
	public void onRegisterChannels(RegisterChannelEvent event) {
		channel = event.register(ResourceKey.resolve("crowdcontrol:version-request", RawDataChannel.class);
	}

	public void sendMessage(ServerPlayer player, PluginPacket packet) {
		channel.play().sendTo(player, buf -> buf.writeString("hello world"));
	}
}
```

Stacktrace is attached below. Not super familiar with Sponge's internal structure; I do see [`ClientboundCustomPayloadPacketMixin_Vanilla`](https://github.com/SpongePowered/Sponge/blob/a091b4f49ee921021f8e4a5a10e28869d2d66122/vanilla/src/mixins/java/org/spongepowered/vanilla/mixin/core/network/protocol/common/ClientboundCustomPayloadPacketMixin_Vanilla.java#L46-L51) appears to register the SpongeChannelPayload, but it only seems to exist in the Vanilla package and not the Forge package? Could explain why the error occurs only on SF but I'm not sure :P

### Link to logs

https://gist.github.com/qixils/b36964dd3d7ba8a20705ecf8f68f6477

ginyai:
Fix channels on SpongeForge
Fix [#4115](https://github.com/SpongePowered/Sponge/issues/4115)

### DIFF
--- a/forge/src/mixins/java/org/spongepowered/forge/mixin/core/client/ClientPacketListenerMixin_Forge.java
+++ b/forge/src/mixins/java/org/spongepowered/forge/mixin/core/client/ClientPacketListenerMixin_Forge.java
@@ -1,0 +1,60 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.forge.mixin.core.client;
+
+import net.minecraft.client.Minecraft;
+import net.minecraft.client.multiplayer.ClientCommonPacketListenerImpl;
+import net.minecraft.network.Connection;
+import net.minecraft.network.protocol.common.ClientboundCustomPayloadPacket;
+import net.minecraft.network.protocol.game.ClientGamePacketListener;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.network.EngineConnectionState;
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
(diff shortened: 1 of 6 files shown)


forge/src/mixins/java/org/spongepowered/forge/mixin/core/client/ClientPacketListenerMixin_Forge.java
```diff
@@ -0,0 +1,60 @@
...
+@Mixin(ClientCommonPacketListenerImpl.class)
+public abstract class ClientPacketListenerMixin_Forge implements ClientGamePacketListener {
+
+    // @formatter: off
+    @Shadow @Final protected Minecraft minecraft;
+    @Shadow @Final protected Connection connection;
+    // @formatter: on
+
+    @Inject(method = "handleCustomPayload(Lnet/minecraft/network/protocol/common/ClientboundCustomPayloadPacket;)V", cancellable = true, at = @At(value = "HEAD"))
+    private void vanilla$handleCustomPayload(final ClientboundCustomPayloadPacket pkt, final CallbackInfo ci) {
```
aromaa, line 53: Should be fine to move to common module, remove the vanilla impls. Remember to adjust the method prefix.

forge/src/mixins/java/org/spongepowered/forge/mixin/core/server/network/ServerCommonPacketListenerImplMixin_Forge.java
```diff
@@ -0,0 +1,59 @@
...
+@Mixin(value = ServerCommonPacketListenerImpl.class, priority = 999)
+public abstract class ServerCommonPacketListenerImplMixin_Forge {
+
+    // @formatter: off
+    @Shadow @Final protected MinecraftServer server;
+    @Shadow @Final protected Connection connection;
+    // @formatter: on
+
+    @Inject(method = "handleCustomPayload", cancellable = true, at = @At(value = "HEAD"))
+    private void vanilla$onHandleCustomPayload(final ServerboundCustomPayloadPacket packet, final CallbackInfo ci) {
```
aromaa, line 52: Same as above.

forge/src/mixins/java/org/spongepowered/forge/mixin/core/server/network/ServerGamePacketListenerImplMixin_Forge.java
```diff
@@ -63,4 +71,13 @@ public abstract class ServerGamePacketListenerImplMixin_Forge implements ServerG
             player.containerMenu.broadcastChanges();
         }
     }
+
+    @Inject(method = "handleCustomPayload", cancellable = true, at = @At(value = "HEAD"), require = 0)
+    private void vanilla$onHandleCustomPayload(final ServerboundCustomPayloadPacket packet, final CallbackInfo ci) {
```
aromaa, line 76: Same as above.

forge/src/mixins/java/org/spongepowered/forge/mixin/core/network/protocol/common/ClientboundCustomPayloadPacketMixin_Forge.java
```diff
@@ -0,0 +1,60 @@
...
+@SuppressWarnings({"rawtypes"})
+@Mixin(ClientboundCustomPayloadPacket.class)
+public abstract class ClientboundCustomPayloadPacketMixin_Forge {
+
+    // @formatter: off
+    @Shadow @Final private static int MAX_PAYLOAD_SIZE;
+    // @formatter: on
+
+    @Inject(method = "lambda$static$0", cancellable = true, at = @At("HEAD"))
+    private static void forge$getCustomPayloadCodecGame(ResourceLocation arg, CallbackInfoReturnable<StreamCodec> cir) {
```
aromaa, line 50: This is most likely going to cause awkward issues if both the Forge and Sponge register the same channel. Preferably we should let Forge pick first and then fallback to Sponge to avoid breaking mods.

aromaa (commented):
In general, I think this is fine as long as we let Forge handle payloads first. The interop with Forge channels needs more structural changes to actually make them work together without causing them to fight each other. So I think this is a fine stopgap.

The other thing is how would we inform plugin developers if both Forge and Sponge registered a channel on the same identifier which would fail silently.

aromaa:
After thinking more about this, we might get some functionality to work with Forge channels too if type check for the Forge wrapped object too. Obviously that is a bit of hack too :)

PR #4121, opened 2024-08-13
Support (mutable) MAX_STACK_SIZE for ItemStack

avaruus1:
Also fix incorrect supports condition for MAX_DURABILITY


aromaa (approved):
LGTM. Thanks!

PR #4132, opened 2024-09-27
Support reloadable registries

MrHell228:
Changes `RegistryAccess` for server from default access to full access (from ReloadableResources, full access contains registries from default access plus few additional ones). The only additional registries are 3 related to loot tables registries.

Also added handler for registries on data pack reload. Currently we dont update registries even though they are actually changed (replaced with completely new registry) because of data packs add/removal. Reloadable registries include 3 loot registries and some registries from default access.

About `GeneratorMain`: I don't really know what deleted line is supposed to do, but I widened access here too, nothing broke and now it can generate loot tables which wasn't possible before.


aromaa:
Superseded by #4179.

PR #4133, opened 2024-09-27
Support plugin provided guice module

aromaa:
This allows plugins to provide a Guice module to expose services to their dependencies and dependents which are injected to their plugin instance.

Example 1, plugin A can provide implementation for a service which plugin B can consume.

```java
@Plugin("plugina")
public final class PluginA {

    //One can also use `PrivateModule` to configure internal services
    //and choose which ones are exposed to the public.
    public static final class Module extends AbstractModule {

        @Override
        protected void configure() {
            //Bind to your own implementation classes.
            this.bind(PluginAService.ExampleA.class).to(ServiceImpl.class);

            //Binding to the plugin instance is also legal, it points to the same instance.
            this.bind(PluginAService.ExampleB.class).to(PluginA.class);
        }
    }
}

@Plugin("pluginb")
public final class PluginB {
    //Dependencies: PluginA { LoadOrder.AFTER }

    @Inject
    public PluginB(final PluginA pluginA, final PluginAService.ExampleA exampleA,
        final PluginAService.ExampleB.class exampleB) {
        //PluginA instance can be directly accessed.
        //The service from plugin A is injected.
        //pluginA == exampleB: true
    }

    public static final class Module extends AbstractModule {

        @Override
        protected void configure() {
            Multibinder.newSetBinder(this.binder(), PluginBService.class);
        }
    }
}
```

Example 2, plugin C can discovered services provided by other plugins.
```java
@Plugin("pluginc")
public final class PluginC {

    @Inject
    public PluginC(final Set<Service> services) {
        //services has PluginD
    }

    public static final class Module extends AbstractModule {

        @Override
        protected void configure() {
            //Bind empty binder in case there are no other plugins.
            Multibinder.newSetBinder(this.binder(), Service.class);
        }
    }
}

@Plugin("plugind")
public final class PluginD {
    //Dependencies: PluginC { LoadOrder.BEFORE }

    public static final class Module extends AbstractModule {

        @Override
        protected void configure() {
            Multibinder.newSetBinder(this.binder(), Service.class).addBinding().to(PluginD.class);
        }
    }
}
```


gabizou:
Is there a guarantee about ordering of event listeners provided by a plugin instance like the description? Or is the guarantee that you _can_ inject plugin instances only without effects to listener ordering?

aromaa:
The requirement is that if you wish to inject dependencies to your plugin then you have to declare them as `LoadOrder.AFTER`.

We could relax it but it requires changing the actual load algorithm, which can be independent from the ordering of `ConstructPluginEvent`. Changing the load algorithm is also observable side-effect. We are now actually initializing plugin A first even after plugin B declared it as `LoadOrder.BEFORE`. Even if we call `ConstructPluginEvent` for plugin B first, the constructor of plugin A could have caused observable side-effects. I would call this myself as undefined behavior but its something to take in to the consideration if we want to change it.

aromaa:
I have extended the support for indirect dependencies, where plugin C declares itself as having dependency on plugin B but as `LoadOrder.BEFORE`. This allows for "service discoverability" where the plugin B can get list of all services by its dependents.

For example the following works:
```java
@Plugin("pluginb")
public final class PluginB {

    @Inject
    public PluginB(final Set<Service> service) {
        //services has PluginC
    }

    public static final class Module extends AbstractModule {

        @Override
        protected void configure() {
            Multibinder.newSetBinder(this.binder(), Service.class);
        }
    }
}

@Plugin("pluginc")
public final class PluginC {

    public static final class Module extends AbstractModule {

        @Override
        protected void configure() {
            Multibinder.newSetBinder(this.binder(), Service.class).addBinding().to(PluginC.class);
        }
    }
}
```

PR #4136, opened 2024-10-12
Move deprecated ItemStackLike related methods to Common

MrHell228:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2565) | **Sponge**

Put all mixins in specific folder so it can be easily nuked in API 13.

Also fixed `Ingredient` API mixin (missed this while migrating to `ItemStackLike`)


aromaa (approved):
Thank you for addressing this promptly!

PR #4137, opened 2024-10-13
fix: Exceptions and unexpected behavior related to `HIDE_CAN_PLACE` key

pandier:
Issue appears on both API 11 and 12


Yeregorix:
Fixed by db4e24e409bd2162468f27c4308fc269b5729079

PR #4142, opened 2024-10-24
Use a supplier to avoid costly String#format in PhaseStack#push

SamB440:
This was showing up in spark as mentioned on Discord a while ago, so it's a nice little performance improvement.


aromaa (approved):
Thank you, I have also seen this in a profile.

PR #4148, opened 2024-11-11
Update ServerLevelMixin::save to match the vanilla impl again

nelind3:
As discussed on discord here's a PR updating ServerLevelMixin::save to once again match the vanilla version. However I'm honestly a bit unsure how/if SerializationBehaviour should factor into this part.


aromaa (approved):
Dealing with the SerializationBehaviour should be left up to the entity manager, so that is fine as is.

LGTM, thank you!

PR #4151, opened 2024-11-14
Replace RunAroundLikeCrazyGoalMixin::tick overwrite with injects

nelind3:
Updated another outdated overwrite that I noticed while working on Loofah.

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/core/world/entity/ai/goal/RunAroundLikeCrazyGoalMixin.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/core/world/entity/ai/goal/RunAroundLikeCrazyGoalMixin.java
@@ -24,19 +24,18 @@
  */
 package org.spongepowered.common.mixin.core.world.entity.ai.goal;
 
-import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.ai.goal.RunAroundLikeCrazyGoal;
 import net.minecraft.world.entity.animal.horse.AbstractHorse;
-import net.minecraft.world.entity.player.Player;
 import org.spongepowered.api.entity.living.animal.horse.HorseLike;
 import org.spongepowered.api.event.CauseStackManager;
 import org.spongepowered.api.event.SpongeEventFactory;
 import org.spongepowered.api.event.cause.entity.DismountTypes;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
-import org.spongepowered.asm.mixin.Mutable;
-import org.spongepowered.asm.mixin.Overwrite;
 import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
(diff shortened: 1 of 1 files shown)


aromaa:
This should be easy to change to just use redirects instead and avoid the overwrite?

nelind3:
Yea the reason i was even working on this is because I'm rewriting it like that since the overwrite is conflicting with Architechtury API on fabric. I can update this PR with the rewritten version once I finish that after dinner if wanted.

aromaa:
We would like to get rid of overwrites as much as possible.

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/ai/goal/GoalMixin.java
```diff
@@ -42,6 +43,8 @@
 @Mixin(Goal.class)
 public abstract class GoalMixin implements GoalBridge {
 
+    @Shadow protected abstract int adjustedTickDelay(int $$0);
```
gabizou: I'd recommend explicitly naming the method shadowed `shadow$adjustedTickDelay` so it's clearer

nelind3:
There we go now it's replacing the overwrites with injects instead

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/ai/goal/RunAroundLikeCrazyGoalMixin.java
```diff
@@ -45,51 +46,41 @@
...
+    @Inject(
+        method = "tick",
+        at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/world/entity/animal/horse/AbstractHorse;tameWithName(Lnet/minecraft/world/entity/player/Player;)Z"
+        ),
+        cancellable = true,
+        locals = LocalCapture.CAPTURE_FAILHARD
+    )
+    private void impl$throwTameEntityEvent(CallbackInfo ci, Entity entity) {
```
aromaa:
```suggestion
    private void impl$throwTameEntityEvent(final CallbackInfo ci, final Entity entity) {
```
We tend to use final where applicable. (There are two others)

src/mixins/java/org/spongepowered/common/mixin/core/world/entity/ai/goal/RunAroundLikeCrazyGoalMixin.java
```diff
@@ -45,51 +46,41 @@
...
+        try (CauseStackManager.StackFrame frame = PhaseTracker.getCauseStackManager().pushCauseFrame()) {
+            frame.pushCause(entity);
```
aromaa: We also prefer to avoid capturing locals if possible due to their brittle nature. One could change this to `this.horse.getFirstPassenger()`.

aromaa (approved):
Added few nits, other than that LGTM.

Thank you again, these are very much appreciated :)

nelind3:
There you go :)

PR #4177, opened 2025-02-22
Fix ItemStackLike#maxStackQuantity

MrHell228:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2581) | **Sponge**

Implements `ItemStackLike#maxStackQuantity` because it's no longer defaulted in API

Also adds ability to remove max durability & max stack size components


aromaa:
Should add the check for `MAX_DURABILITY` too? The API changes should be unnecessary with the new changes.

MrHell228:
Because `get` changed to return component value, it may return empty optional, meaning `require` in API will throw anyway.
Fixed `Keys.MAX_DURABILITY` (getter effectively unchanged but now follows general pattern for components)

aromaa:
Ah I didn't notice you changed that. Yeah, that should be left as `ItemStack::getMaxStackSize`.

MrHell228:
Haven't we agreed in discord that making it return component value is fine?

aromaa:
The whole point of removing the supports was that it would return the default value if the component was missing, right? :)

MrHell228:
`MAX_STACK_SIZE` component presence doesn't really correlate with `MAX_DAMAGE` component presence.
The only limitation is that if `MAX_DAMAGE` is present, `MAX_STACK_SIZE` must be either absent or equal to 1. And I guess it's impl detail that there is no behaviour difference between absence and equality to 1. But devs should be able to see if component is actually missing, or it's value set to 1

aromaa:
>  But devs should be able to see if component is actually missing

This is a whole separate issue that we have not yet decided up on how to deal with it. In that case we might want to delay this.

MrHell228:
Sad, but okay. Closed API PR, reverted changes related to API (including MAX_STACK_QUANTITY getter)

aromaa (approved):
Thanks!

PR #4182, opened 2025-03-08
Fixes for DataHolder & ItemStackSnapshot

MrHell228:
`ItemStackSnapshot` `getKeys()` & `getValues()` now return all values & supported keys instead of just custom
`DataHolder#getKeys` now returns all supported keys instead of just present


aromaa (approved):
LGTM, thank you!

PR #4185, opened 2025-03-09
Allow including external projects

Faithcaio:
Adding `userplugins=true` to `project.properties` enables 
`testplugins/userPluginBuilds` file to add paths to additional gradle projects
```
/path/to/my/plugin/project
```
`testplugins/userPlugins` file to add plugins as gradle dependencies
```
com.example.plugin:plugin1
com.example.plugin:plugin2
```

Sponge Plugins added this way are automatically loaded and can be hot-swapped.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,3 +85,6 @@ buildSrc/src/api
 testplugins.settings.gradle.kts
 spongeforge.settings.gradle.kts
 projects.properties
+# External plugin projects
+/userplugins/userPluginBuilds
+/userplugins/userPlugins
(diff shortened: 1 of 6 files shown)


aromaa:
Personally I would rather not alter the build scripts just for this and avoid the extra LOC. I don't think this even saves a lot compared to doing it the traditional way of appending the classpath in the IDE itself.

Yeregorix:
I rarely debug plugins this way so I don't know if this is useful or not but if it is then I'm fine with adding such a thing.

PR #4194, opened 2025-03-26
introduce scoped registry holder using PhaseTracker

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2584) | **Sponge**

Intruduces `SpongeCommon#scopedHolder` which attempts to find a `RegistryHolder` in the current cause.
`SpongeCommon#vanillaRegistry` now uses this.
Fallback to Server -> Game depending on availability.

During Startup `SpongeLifecycle#processerverRegistries` the current RegistryHolder is pushed on the stack allowing plugins during registration to access registries earlier.
Later when `DedicatedServer#initServer` is called the Server is pushed on the stack.

### DIFF
--- a/src/main/java/org/spongepowered/common/SpongeCommon.java
+++ b/src/main/java/org/spongepowered/common/SpongeCommon.java
@@ -37,7 +37,9 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.event.Event;
 import org.spongepowered.common.applaunch.config.core.SpongeConfigs;
+import org.spongepowered.common.event.tracking.PhaseTracker;
 import org.spongepowered.common.launch.Launch;
+import org.spongepowered.common.registry.SpongeRegistryHolder;
 import org.spongepowered.common.scheduler.AsyncScheduler;
 import org.spongepowered.common.scheduler.ServerScheduler;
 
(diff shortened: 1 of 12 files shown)


src/main/java/org/spongepowered/common/SpongeCommon.java
```diff
@@ -100,8 +102,21 @@ public static RegistryAccess.Frozen vanillaRegistryAccess() {
         return SpongeCommon.server().registryAccess();
     }
 
+    public static SpongeRegistryHolder scopedHolder() {
+        var holder = PhaseTracker.getInstance().currentCause().first(SpongeRegistryHolder.class);
```
Zidane: I still strongly feel that this is violating Cause. We need a static area to store this field, separate from an engine per my original idea. So, it is perfectly fine to put it on the `PhaseTracker` as an optional field. But not in the Cause. Cause is meant to be inspected, and plugin developers are told to take it as face value. Not for us to use it as a dumping ground.

Zidane: Not to mention, you are creating frames where they were not needed before. If you put the field on the tracker directly, no need for these pointless frames.

Faithcaio: the frames are created 6 times on async threads

Faithcaio: But I guess now that I realized I never needed to push the Server at `#initServer` I could just put a try finally block around there

PR #4196, opened 2025-03-30
Implement Viewer#playTotemOfUndyingEffect

MrHell228:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2586) | **Sponge**

Since vanilla packet make client looking for "totem" item only in hands, the same trick is used as with `Audience#openBook`: send fake item in hand, send death protection animation and then reset item in hand to original


Yeregorix:
Merged by f8b5721ed291ad49251e8c872e74c473aee2227e

Issue #4200, opened 2025-04-02
SpongeNeo | Boss Progress Bar under 0 causes log spam


MarGar12:
### Affected Product(s)

SpongeForge

### Version

1.21.1-21.1.35-12.0.2-RC2149

### Operating System

Linux

### Java Version

openjdk version "21.0.6" 2025-01-21

### Plugins/Mods

```shell
Curios API 9.4.0+1.21.1 (curios)
GeckoLib 4 4.7.5.1 (geckolib)
Iron's Spells 'n Spellbooks 1.21.1-3.10.1 (irons_spellbooks)
Player Animator 2.0.1+1.21.1 (playeranimator)
SpongeNeo 1.21.1-21.1.35-12.0.2-RC2149 (spongeneo)
```

### Describe the bug

This happens when the Dead King boss from Iron's Spells and Spellbooks reaches 200HP. The Boss Bar supposedly goes back to full after reaching 200HP, sorta like half half. But Sponge causes massive log spam and lags the server lots. The Boss Progress bar also doesn't change in-game.

![Image](https://github.com/user-attachments/assets/27e6a3df-36ac-4251-aec0-b7a4893ccbd1)

### Link to logs

https://paste.gg/p/anonymous/686a28ecbcd54f9280fcc565b691647e

PR #4203, opened 2025-04-03
Clamp mod-provided progress for bossbars

MrHell228:
Should solve #4200 

The progress restriction of [0, 1] is coming from Adventure. Minecraft doesn't have this restriction so mods don't really care about clamping the value they provide. So I think it would be better to just clamp all mod-provided progress

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/core/world/BossEventMixin.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/core/world/BossEventMixin.java
@@ -90,7 +90,7 @@ public abstract class BossEventMixin implements BossEventBridge {
 
     @Redirect(method = "setProgress", at = @At(value = "FIELD", target = "Lnet/minecraft/world/BossEvent;progress:F"))
     private void adventurePercent(final BossEvent $this, final float percent) {
-        this.bridge$asAdventure().progress(percent);
+        this.bridge$asAdventure().progress(Math.clamp(percent, 0, 1));
     }
 
     @Redirect(method = "setColor", at = @At(value = "FIELD", target = "Lnet/minecraft/world/BossEvent;color:Lnet/minecraft/world/BossEvent$BossBarColor;"))


aromaa:
This is going to need more investigation on the rules that vanilla and mods use to render the progress and whatever there is something that Sponge can do here.

The first question is whatever the mod has custom render logic to handle the display of progress?

MrHell228:
[This line](https://github.com/iron431/irons-spells-n-spellbooks/blob/1.21/src/main/java/io/redspace/ironsspellbooks/entity/mobs/dead_king_boss/DeadKingBoss.java#L267) caused exception mentioned in the issue. It's simply 1 tick when entity is hurt below half health (meaning negative bar progress) and on the next tick bossbar is set to full again (so in this case client basically won't notice anything). Easy fix from mod side.
But I'm not sure why mod would want to do that if game doesn't require progress to be in [0, 1] range.
And unless we want to mixin into Adventure's BossBar and cancel range check, I don't see any other general solution than clamping progress from Sponge side

aromaa:
Thank you, thats all I wanted to know. Its just a bug in the mod and not that they have special rendering.

MrHell228:
Not really a bug if it works as intended in normal circumstances I guess.
It's Sponge that injects whole library that adds constraints on existing object

aromaa:
Minecraft is known to have weird edge cases around values out of the vanilla bounds, so anything unusual is unexpected and could be considered a bug. I'm fairly certain sending out of bounds values for the progress used to produce weird rendering bugs.

Anyways, thaths besides the point. The changes LGTM.

MrHell228:
Yeah, it surely is used. But any solution other than clamping will probably break Adventure's BossBar contract for progress which is not good

aromaa (approved):
My point being that we don't usually consider values out of the vanilla bounds to (ever) be valid and try to sanitize them as mods and plugins tend to not do so themselves and lead to all sorts of weird behavior. So this is just another new place to do so, nothing unusual for us.

That being said, thank you again!

Issue #4216, opened 2025-07-16
[SpongeNeo 1.21.1] Saving entity NBT error for pixelmon


hund35:
### Affected Product(s)

SpongeNeo

### Version

1.21.1-21.1.35-12.0.2-RC2257

### Operating System

Windows

### Java Version

java 21.0.7 2025-04-15 LTS

### Plugins/Mods

```shell
Pixelmon-1.21.1-9.3.6-server.jar
```

### Describe the bug

Throws the following exception when trying to send a pokemon out with the R button or tries to throw a pokeball and nothing happens expect for getting the exception
[co.pi.pi.Pixelmon/]: Error handling packet net.minecraft.ReportedException: Saving entity NBT

I have tried to report as a pixelmon bug and they said it is sponge causing it

### Link to logs

https://gist.github.com/hund35/122beee2593a0d9b0fd67e840bf29c76

PR #4218, opened 2025-07-23
[Issue #4216] Saving entity NBT error for pixelmon

AlexandreArcil:
Issue #4216 

### Bug
When throwing an empty pokeball, the transaction **SpawnEntityTransaction** will call `Entity#saveWithoutId` that will call `Entity#addAdditionalSaveData` overrided by **ThrowableItemProjectile** to save the **ItemStack** linked to the throwable. To retrieve this **ItemStack**, it will call `ThrowableItemProjectile#getItem` that is overrided by **PokeBallEntity**, but it will return an empty **ItemStack**, which is not serializable, hence the _Cannot encode empty ItemStack_ error.

### Fix
`Entity#saveWithoutId` is only call if the entity is made to be saved/serializable. It's something clearly set by the developer and so intended, and a way to tell that a call to the method may cause issues while remaining more permissive than `Entity#save`.

### Impact
In the game, 4 entities are marked as not serializable: leash knot, lightning bolt, player and fishing robber. We can also add some entities of other mods in the list. So the change has a significant impact because it will stop serialize a lot of entities for which it worked, but it seems that the `entityTag` field is not used in Sponge or even shared to mods, so I don't understand in what way it is important. Am I missing something here ?

#### Else clause
I didn't set a value in the case where the entity is not serializable because I didn't know what a good one could be. An another value can be an empty **CoumpoundTag**, which is a way to say that `captureState()` has been called but the state cannot be captured and allows to avoid NPE, but then you cannot differentiate between something that don't have something to save.

#### Condition
I though of another solutions but that I like less:
- Surround the method with a try/catch: reduce the impact to the minimum, only the problematic ones, but may hide potential bugs as the exception will not be logged anymore or difficult to see if logged at a lower level
- Do not call the method if the entity is a throwable item with an empty item: too specific to the bug, an exception may also happen with other type of entities or for other reasons, and so the condition will become bigger and bigger. This is less future-proof and increase cost maintenance.

### DIFF
--- a/src/main/java/org/spongepowered/common/event/tracking/context/transaction/world/SpawnEntityTransaction.java
+++ b/src/main/java/org/spongepowered/common/event/tracking/context/transaction/world/SpawnEntityTransaction.java
@@ -91,7 +91,9 @@ public final class SpawnEntityTransaction extends WorldBasedTransaction<SpawnEnt
     @Override
     protected void captureState() {
         super.captureState();
-        this.entityTag = this.entityToSpawn.saveWithoutId(new CompoundTag());
+        if(this.entityToSpawn.getType().canSerialize()) {
+            this.entityTag = this.entityToSpawn.saveWithoutId(new CompoundTag());
+        }
     }
 
     @Override


Yeregorix:
Thanks for identifying the cause of the issue!
Indeed it seems `entityTag` is not used anywhere. That is either an old code that could be cleaned up or something missing to implement.
@gabizou Do you happen to know what this field is supposed to be used for? It is copied from `SpawnEntityTransaction#entityTag` to `DummySnapshot#entityTag` which is then unused. I can literally delete all members of `DummySnapshot` and Sponge still compiles.

aromaa:
Looks like it was introduced in da8ff03 and seems to be unused even there.

There is also another unused tag in `EntityPerformingDropsTransaction`. Not to mention all the other APIs like snapshots and archetypes that call these.

aromaa:
Also more out of curiosity, what does happen if you try to teleport these across dimensions? Vanilla also does use these methods outside of serialization, they only guard the `Entity#getEncodeId` which is used as part of actual serialization, the other methods are used more freely.

Yeregorix:
I ended up removing the unused tag snapshot: 0a75cbc. Thanks @AlexandreArcil for identifying the problem.

Issue #4122, opened 2024-08-14
Hunger not reverted for client on cancelled cake eating


SamB440:
### Affected Product(s)

SpongeVanilla

### Version

1.21.1-12.0.0-RC1784

### Operating System

Linux

### Java Version

21

### Plugins/Mods

```shell
Plugins (18)
Minecraft v1.21.1
SpongeAPI v12.0.0-SNAPSHOT
Sponge v1.21.1-12.0.0-SNAPSHOT
SpongeVanilla v1.21.1-12.0.0-RC1784
LemonadeStand v1.0.1-SNAPSHOT
WorldEdit v7.3.7-SNAPSHOT+6896-6e37730
LemonSqueezer v1.0.1-SNAPSHOT
LuckPerms v5.4.139
spark-sponge8 v1.10.93
Treaty v1.0.0
NuVotifier v3.0.0-SNAPSHOT+9d69da2
ViaVersion v1.0.1
Orbis v1.0.0-SNAPSHOT
PacketEvents v2.5.0+e0b9d6813-SNAPSHOT
Spelare v1.0.1-SNAPSHOT
Pharus v1.0.0-SNAPSHOT
Masquerade v1.0.2-SNAPSHOT
Battlegrounds v3.1.2
```

### Describe the bug

When cancelling a player from eating a cake with `InteractBlockEvent.Secondary`, the player's hunger visually increases however this is purely clientside. It seems that the food data of the player needs to be resynced. I also attempted to use the other methods such as `setUseBlockResult` however this had no effect.

PR #4221, opened 2025-08-10
Sync health and food to clients when interact block secondary event is cancelled

AlexandreArcil:
Reported by #4122

When right-clicking on a cake and `InteractBlockEvent.Secondary` is cancelled, the player's hunger level visually increases while on the server the value remains unchanged.
This PR fixes this by resynchronizing the food level when the event is cancelled. 

The code is quite similar to what minecraft does in `ServerPlayer#doTick`.

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/tracker/server/level/ServerPlayerGameModeMixin_Tracker.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/tracker/server/level/ServerPlayerGameModeMixin_Tracker.java
@@ -31,12 +31,14 @@ import com.llamalad7.mixinextras.sugar.Cancellable;
 import com.llamalad7.mixinextras.sugar.Share;
 import com.llamalad7.mixinextras.sugar.ref.LocalBooleanRef;
 import com.llamalad7.mixinextras.sugar.ref.LocalRef;
+import net.minecraft.network.protocol.game.ClientboundSetHealthPacket;
 import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.server.level.ServerPlayerGameMode;
 import net.minecraft.world.InteractionHand;
 import net.minecraft.world.InteractionResult;
 import net.minecraft.world.ItemInteractionResult;
 import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.food.FoodData;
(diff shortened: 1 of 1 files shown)


src/mixins/java/org/spongepowered/common/mixin/tracker/server/level/ServerPlayerGameModeMixin_Tracker.java
```diff
@@ -103,10 +105,18 @@ public abstract class ServerPlayerGameModeMixin_Tracker {
...
+        final FoodData foodData = player.getFoodData();
+        final float prevHealth = player.getHealth();
+        final int prevFoodLevel = foodData.getFoodLevel();
+        final float prevSaturationLevel = foodData.getSaturationLevel();
 
         ((ServerPlayerGameModeBridge) this).bridge$setInteractBlockRightClickCancelled(event.isCancelled());
         if (event.isCancelled()) {
             this.player.inventoryMenu.sendAllDataToRemote();
+            // Cancel health or food level changes to the client, for example after eating a cake
+            if (player.getHealth() != prevHealth || foodData.getFoodLevel() != prevFoodLevel || foodData.getSaturationLevel() != prevSaturationLevel) {
```
Yeregorix: I don't think this condition will ever be true.
Since `bridge$setInteractBlockRightClickCancelled` is a pure setter and `sendAllDataToRemote` is only synchronizing the inventory, I don't see any reason for the health and food data values to change.
I tested the PR and as expected it doesn't resolve the issue.

Yeregorix: Note that removing the condition and always sending the packet solves the issue. Ideally we would only send this packet when needed but I don't think we can detect that the client health or food data have changed without playing the interaction and then undoing everything.

AlexandreArcil: You're right, I don't see why I added this condition.
I didn't find a way to check if the interaction will change the food level, so I send it every time, like inventory update does. I could send it if the block snapshot is a cake (I don't think they are other blocks that do that), but the bug might reappear if a new block with the same behaviour is added.

PR #4226, opened 2025-09-09
Use LauncherInterceptor for JUnit tests

aromaa:
We are now guaranteed to be in the transforming class loader during unit test loading and their execution, ensuring class loading happens from the correct `ClassLoader`.

### DIFF
--- a/bootstrap/src/main/java/org/spongepowered/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/org/spongepowered/bootstrap/Bootstrap.java
@@ -26,6 +26,7 @@ package org.spongepowered.bootstrap;
 
 import org.spongepowered.bootstrap.dev.DevClasspath;
 
+import java.io.File;
 import java.lang.module.Configuration;
 import java.lang.module.ModuleFinder;
 import java.net.URL;
@@ -38,6 +39,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public abstract class Bootstrap<Jar> {
     public static final boolean DEBUG = Boolean.getBoolean("sponge.bootstrap.debug");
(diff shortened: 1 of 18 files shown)


bootstrap/src/main/java/org/spongepowered/bootstrap/Bootstrap.java
```diff
@@ -117,15 +121,31 @@ public void boot(final List<Path[]> classpath, final boolean isolated) throws Ex
...
+            final String resources = System.getProperty("sponge.resources");
+            if (resources != null) {
+                final List<Jar> spongeJars = new ArrayList<>();
+                for (final String entry : resources.split(File.pathSeparator)) {
+                    final Path[] paths = Stream.of(entry.split("&")).map(Path::of).toArray(Path[]::new);
+                    spongeJars.add(this.createJar(paths));
+                }
+                final ModuleFinder spongeFinder = this.createModuleFinder(spongeJars);
+                final ModuleFinder resourceFinder = this.createModuleFinder(resourceJars);
+                parentLoader = new FilteringPassthroughClassLoader(contextLoader,
```
aromaa, line 137: This ensures that the classes that are supposed to be inside the module layers are not loaded through the bootstrap layer as we are now exposing a hole that can access the application class loader and thus gain access to the original classpath. Classes that we don't end up loading inside the module layers are still accessible and fallback to the parent.

bootstrap/src/main/java/org/spongepowered/bootstrap/Bootstrap.java
```diff
@@ -117,15 +121,31 @@ public void boot(final List<Path[]> classpath, final boolean isolated) throws Ex
...
-        if (!resourceJars.isEmpty()) {
-            final URL[] urls = new URL[resourceJars.size()];
+        if (!resourcePaths.isEmpty()) {
+            final URL[] urls = new URL[resourcePaths.size()];
             for (int i = 0; i < urls.length; i++) {
-                urls[i] = resourceJars.get(i).toUri().toURL();
+                urls[i] = resourcePaths.get(i).toUri().toURL();
             }
-            parentLoader = new URLClassLoader("BOOTSTRAP-RESOURCES", urls, parentLoader);
+            parentLoader = new ResourceClassLoader("BOOTSTRAP-RESOURCES", urls, parentLoader);
```
aromaa, line 148: Same here, this could cause class loading from the resources we just provided outside of the module layers.

bootstrap/src/main/java/org/spongepowered/bootstrap/ResourceClassLoader.java
```diff
@@ -0,0 +1,44 @@
...
+        ClassLoader.registerAsParallelCapable();
+    }
+
+    public ResourceClassLoader(final String name, final URL[] urls, final ClassLoader parent) {
+        super(name, urls, parent);
+    }
+
+    @Override
+    protected Class<?> findClass(final String name) throws ClassNotFoundException {
+        throw new ClassNotFoundException(name);
```
aromaa, line 42: This rejects all class loading that failed to be fetched from the parent.

Yeregorix:
LGTM!
We expose the same JUnit services in SV production so that any plugin dev can easily setup unit tests. So I just copied your changes there.

See https://github.com/Yeregorix/AutoPickup/blob/feature/unit-tests/build.gradle#L71 for a POC. In the long term, I plan to pre-configure everything in SpongeGradle.

Yeregorix (approved):

PR #4229, opened 2025-09-11
Make Enchantment a ComponentLike

MrHell228:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2600) | **Sponge**

Implementation uses the same method that vanilla use to create enchantment component


Yeregorix:
Merged!

PR #4236, opened 2025-10-03
Push scoped holder for RegisterRegistryEvent default values populators

MrHell228:


src/main/java/org/spongepowered/common/event/lifecycle/AbstractRegisterRegistryEvent.java
```diff
@@ -81,10 +82,20 @@ public <T> RegistryType<T> register(final ResourceKey key, final boolean isDynam
...
+        holder.createRegistry(type, actualHolder -> this.wrapWithScopedHolder(holder, () -> defaultValues.apply(actualHolder)), isDynamic, dependencies);
         return type;
     }
 
+    private <T> Map<ResourceKey, T> wrapWithScopedHolder(final SpongeRegistryHolder holder, final Supplier<Map<ResourceKey, T>> defaultValues) {
+        return PhaseTracker.getInstance().switchStartupRegistryHolder(
+            holder, defaultValues,
+            (exception) -> {
+                // If we don't catch it, vanilla will shut the server with "Failed to load datapacks"
+                exception.printStackTrace();
```
aromaa: Use logger and print something useful. Like `SpongeCommon.logger().error("Could not supply values for register {}")`.

MrHell228: Done

aromaa (approved):
Thank you!

PR #4240, opened 2025-10-06
fix SpongeShapelessRecipe.matches (API 14)

MrHell228:
Current approach doesn't account for stacks that does not match any ingredient.
This results in recipe to match even if there are items that shouldn't be in it.


aromaa:
This should target API 12. Can we instead just use the vanilla `StackedContents`? Doing some quick testing, this seems to work.

MrHell228:
`StackedContents` approach seems to work in API 14.
In API 12 `StackedContents` is unparametrized and works with item integer ids so it's not possible to use it there.
I guess there should be 2 PRs, one to 12 with simple fix, and another to 14 with using `StackedContents`

MrHell228:
Reworked `matches` method to use `StackedContents`

MrHell228:
To think about this more, it seems better to `WrapOperation` `StackedItemContents#canCraft` in `ShapelessRecipe#matches` instead of overriding in `SpongeShapelessRecipe`.
I would like to do this after merging recipe placing PR to not mess with merge conflicts a lot, and maybe it would be worth to use `SpongeStackedItemContents` in `CraftingInputMixin`

PR #4241, opened 2025-10-08
Make recipe placing from recipe book working with custom ingredients

MrHell228:


aromaa:
This is a lot of Mixins to achieve one goal and is hard to justify to be this invasive. Can we simply this by modifying the second parameter of the `Inventory#findSlotMatchingCraftingIngredient` inside the `ServerPlaceRecipe#moveItemToGrid`?

MrHell228:
Most of the injections in `ServerPlaceRecipeMixin` are simply maintaining the list of applicable `ItemStack`s, which in the end is used exactly for `ServerPlaceRecipe#moveItemToGrid` and `Inventory#findSlotMatchingCraftingIngredient` (for their modified versions in `RecipeBookUtil`, to be more accurate).

I used as much vanilla logic as I could, and no custom logic is performed if recipe is regular, so I'm not sure what does "this invasive" mean.

aromaa:
We don't want to duplicate vanilla code in our code base unless we really have to, its too fragile and heavily increases maintenance burden and iteration times, that is a no go.

If we can cut this down to few  mixins instead of having a dozen, that would be a lot better.

MrHell228:
I wouldn't copy these 2 vanilla methods if I could really avoid it, it's just the bare minimum to make it work (as far as I figured out).

Vanilla recipe placing logic is heavily bounded to `Item`s, and assuming `Ingredient#test` can not even depend on `Item` (but on `ItemStack` data, for example), I honestly don't see the way to reduce the amount if injections here.

aromaa:
I'm suggesting that we do the final filtering in `ServerPlaceRecipe#moveItemToGrid`.

MrHell228:
You mean something like this? It does reduce duplicated code to just one simple method

aromaa:
I mean that we literally do all the filtering in just the `ServerPlaceRecipe#moveItemToGrid`. It means that we need to find the items again, but thats small price to pay compared to the amount of code we would have to roll.

MrHell228:
Wouldn't it remove like... 3 `Inject`s and 1 `ModifyArg` and add the requirement to duplicate vanilla code that filters the items?

aromaa:
I'm expecting us to nuke 90% of this PR with just that change, am I missing something? Do we need to do anything else than just filter the items?

MrHell228:
While decomposing all the steps I understood that `RecipeBridge` and relevant mixins are unnecessary because it can simply be done on `PlacementInfo`. And I think `PlacementInfoMixin` will exist anyway because of the PR that fixes shapeless recipes.

About everything else:

We won't be able to capture items without duplicating if we don't use `SpongeStackedItemContents` where it's used now.
That's unnecessary (and might be costly) for regular recipes so that's why `PlacementInfoBridge#hasCustomIngredients` exists.
And if we don't capture items there, `StackedItemContents#canCraft` (that is used 3 times) will literally return random booleans which will make placing for some sponge recipes simply not working at all

3 `Inject`s and `ModifyArg` that maintain applicable `ItemStack` list are simply following the vanilla `List<Holder<Item>>` behaviour.
If we don't maintain the list here, we would need to duplicate... I guess this code from `ServerPlaceRecipe#placeRecipe`?
```
if ($$1.canCraft(var10001, $$6, $$7::add)) {
  int $$8 = clampToMaxStackSize($$6, $$7);
  if ($$8 != $$6) {
    $$7.clear();
    var10001 = $$0.value();
    Objects.requireNonNull($$7);
    if (!$$1.canCraft(var10001, $$8, $$7::add)) {
      return;
    }
  }
...
```
Because of the `$$7::add` that accumulates mentioned list.
And I feel like few injects are better than copying all of this.

`RecipeBookUtil#findSlotMatchingCraftingIngredient` seems to remain anyway because I don't see how vanilla method could be used there.

So... I don't understand how you suppose to reduce the PR by this much.

aromaa:
Yes, I fully expect us to throw away the `List<Holder<Item>>` for custom recipes. We would be pulling the actual `ItemStack`s inside the `ServerPlayerRecipe#moveItemToGrid` from a field. The easiest thing I see is to:

- Override the methods with `StackedContents<Holder<Item>>` calls in `SpongeStackedItemContents` to `StackedContents<ItemStack>` (Don't duplicate the vanilla logic).
- Redirect the output to say, `ArrayDequeue` (Give the `SpongeStackedItemContents` reference to it so you are good to go).
- Poll that from `ServerPlayerRecipe#moveItemToGrid`.

MrHell228:
2nd and 3rd points done. Didn't exactly understand what you meant in 1st (beacuse I think it's already done).
Managed to reduce the amount of injections to 4

src/main/java/org/spongepowered/common/item/recipe/crafting/SpongeStackedItemContents.java
```diff
@@ -0,0 +1,129 @@
...
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.crafting.PlacementInfo;
+import net.minecraft.world.item.crafting.Recipe;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.spongepowered.common.bridge.world.item.crafting.PlacementInfoBridge;
+
+import java.util.List;
+import java.util.function.Function;
+
+public class SpongeStackedItemContents extends StackedItemContents {
```
aromaa:
```suggestion
public final class SpongeStackedItemContents extends StackedItemContents {
```

src/main/java/org/spongepowered/common/item/recipe/crafting/SpongeStackedItemContents.java
```diff
@@ -0,0 +1,129 @@
...
+            return ItemStack.hashItemAndComponents(o);
+        }
+
+        @Override
+        public boolean equals(final @Nullable ItemStack a, final @Nullable ItemStack b) {
+            return (a == b) || (a != null && b != null && ItemStack.isSameItemSameComponents(a, b));
+        }
+    };
+
+    private final Object2ReferenceMap<ItemStack, ItemStack> stackInterner = new Object2ReferenceOpenCustomHashMap<>(STACK_HASH_STRATEGY);
```
aromaa:
```suggestion
    private final Object2ReferenceMap<ItemStack, ItemStack> stackInterner = new Object2ReferenceOpenCustomHashMap<>(SpongeStackedItemContents.STACK_HASH_STRATEGY);
```

src/main/java/org/spongepowered/common/item/recipe/crafting/SpongeStackedItemContents.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Override
+    public boolean canCraft(
+        final List<? extends StackedContents.IngredientInfo<Holder<Item>>> ingredients,
+        final StackedContents.@Nullable Output<Holder<Item>> output
+    ) {
+        // By default, this method is not called in the context the instance of this class is created.
+        // If this happens, it's either error in Sponge impl or
+        // mixin from some mod (which should be inspected instead of silently doing something that impl does not expect).
+        throw new UnsupportedOperationException("This method should not have been called, please report about it");
```
aromaa: Lets just implement this

src/mixins/java/org/spongepowered/common/mixin/core/recipebook/ServerPlaceRecipeMixin.java
```diff
@@ -0,0 +1,126 @@
...
+    private static StackedItemContents impl$useCustomStackedItemContents(
+        final Operation<StackedItemContents> original,
+        final ServerPlaceRecipe.CraftingMenuAccess<?> menu,
+        final int gridWidth, final int gridHeight,
+        final List<Slot> inputGridSlots, final List<Slot> slotsToClear,
+        final Inventory inventory, final RecipeHolder<?> recipe,
+        final boolean useMaxItems, final boolean isCreative
+    ) {
+        return ((PlacementInfoBridge) recipe.value().placementInfo()).bridge$hasCustomIngredients()
+            ? new SpongeStackedItemContents()
```
aromaa: No need to wrap the `placeRecipe`.
```suggestion
            ? new SpongeStackedItemContents(this.impl$stackList::add, this.impl$stackList::clear)
```

src/mixins/java/org/spongepowered/common/mixin/core/recipebook/ServerPlaceRecipeMixin.java
```diff
@@ -0,0 +1,126 @@
...
+        } else {
+            original.call(recipe, contents);
+        }
+    }
+
+    @Inject(
+        method = "placeRecipe(Lnet/minecraft/world/item/crafting/RecipeHolder;Lnet/minecraft/world/entity/player/StackedItemContents;)V",
+        at = @At(
+            value = "INVOKE",
+            target = "Ljava/util/List;clear()V"
```
aromaa: This is too wide, lets just clear in the canCraft.

src/mixins/java/org/spongepowered/common/mixin/core/recipebook/ServerPlaceRecipeMixin.java
```diff
@@ -0,0 +1,126 @@
...
+            target = "Lnet/minecraft/world/entity/player/Inventory;findSlotMatchingCraftingIngredient(Lnet/minecraft/core/Holder;Lnet/minecraft/world/item/ItemStack;)I"
+        )
+    )
+    private int impl$adjustMatchingSlotFinder(
+        final Inventory instance, final Holder<Item> exemplaryItem, final ItemStack craftInputStack, final Operation<Integer> original
+    ) {
+        return this.impl$stackList == null
+            ? original.call(instance, exemplaryItem, craftInputStack)
+            : RecipeBookUtil.findSlotMatchingCraftingIngredient(inventory, this.impl$stackList.removeFirst(), craftInputStack);
+    }
```
aromaa, line 103: `ModifyArg` the `ItemStack` to return `this.impl$stackList.removeFirst()` when its empty.

src/mixins/java/org/spongepowered/common/mixin/core/recipebook/ServerPlaceRecipeMixin.java
```diff
@@ -0,0 +1,126 @@
...
+            ? new SpongeStackedItemContents()
+            : original.call();
+    }
+
+    @WrapMethod(method = "placeRecipe(Lnet/minecraft/world/item/crafting/RecipeHolder;Lnet/minecraft/world/entity/player/StackedItemContents;)V")
+    private void impl$handleApplicableStacks(
+        final RecipeHolder<?> recipe, final StackedItemContents contents, final Operation<Void> original
+    ) {
+        if (contents instanceof final SpongeStackedItemContents spongeContents) {
+            this.impl$stackList = new ArrayList<>();
```
aromaa:
```suggestion
            this.impl$stackList = new ArrayDequeue<>();
```
Removing the first element from a list is O(n)

src/mixins/java/org/spongepowered/common/mixin/core/recipebook/ServerPlaceRecipeMixin.java
MrHell228: The original and modified methods have different signatures. `exemplaryItem` of the original is `Holder<Item>`, we need to use `ItemStack`

aromaa: The `findSlotMatchingCraftingIngredient` has a parameter for `ItemStack`

MrHell228: This is `ItemStack` that is located in the crafting grid, it may be empty. Modified method uses it too (so 2 different stacks in total)

MrHell228: Ah, i got it, sorry, let met test

src/main/java/org/spongepowered/common/item/recipe/crafting/SpongeStackedItemContents.java
MrHell228: `List<? extends StackedContents.IngredientInfo<Holder<Item>>> ingredients` must be used on `StackedContents<Holder<Item>>`.

We only have `StackedContents<ItemStack>`, contents from `StackedItemContents` are not used currently, it may be changed and we can delegate to this (which in the context of this class doesn't really makes much sense but hides the exception).

In vanilla the only class that implements `StackedContents.IngredientInfo` is `Ingredient`, but I don't think it's safe to cast all of this to ingredients (because if this method is even called, it's done by some mod and such it can provide `IngredientInfo` other than `Ingredient`).

So, what way to implement this should be used?

aromaa: I would expect us to just match the type because thats explicitly asked, we cant do anything else and the mod can't do anything more.

MrHell228:
Fixed all the reviewed things

src/mixins/java/org/spongepowered/common/mixin/core/recipebook/ServerPlaceRecipeMixin.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @ModifyArg(
+        method = "moveItemToGrid",
+        at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/world/entity/player/Inventory;findSlotMatchingCraftingIngredient(Lnet/minecraft/core/Holder;Lnet/minecraft/world/item/ItemStack;)I"
+        )
+    )
+    private ItemStack impl$adjustMatchingSlotFinder(final ItemStack craftInputStack) {
+        return this.impl$stackList.isEmpty() ? craftInputStack : this.impl$stackList.poll();
```
aromaa: This still needs to account for the non empty `craftInputStack`. My initial suggestion was to just use it instead when its non empty, but now that I thought about it more, it actually just pulls more wrong items inside the slot. I think we should just bail when `!ItemStack.isSameItemSameComponents(craftInputStack, this.impl$stackList.poll())` because theres nothing more we can do.

We should hit here when the players inventory is full and they have more stacks in the crafting input than they can fit in their inventory.

I think the following works as intended:
```java
if (this.impl$stackList.isEmpty()) {
    return original.call(...);
} else {
    final ItemStack input = his.impl$stackList.poll();
    if (!ItemStack.isSameItemSameComponents(craftInputStack, input)) {
        return -1:
    }
    return original.call(..., input);
}
```

MrHell228: I think the condition to fail should be `!craftInputStack.isEmpty() && !ItemStack.isSameItemSameComponents(craftInputStack, input)`, or we won't be able to move items to empty crafting slots

MrHell228:
Made tests, found a bug in mixin and a bug in vanilla.
It tests different layouts for all placeable recipes (with regular only and custom only ingredients).
I think testing not only for recipe match but also for the expected layout (for the amount of placed items) is necessary because it's actually the place where bugs can happen (found bugs mentioned above and there were other bugs related to this while doing this PR).
Test for custom shapeless recipe is temporarily commented out unless they are fixed (tested it on different branch with recipe fix, it works too). 
I was doing tests for the first time so I hope it's not that bad but at least everything works.

src/test/java/org/spongepowered/common/recipe/RecipePlaceTest.java
```diff
@@ -0,0 +1,431 @@
...
+            stack -> stack.type() == smallPearl.type()
+                && stack.maxStackQuantity() == smallPearl.maxStackQuantity(),
+            pearl);
+        final Ingredient bigPearlIngredient = Ingredient.of(ResourceKey.sponge("big_pearl"),
+            stack -> stack.type() == bigPearl.type()
+                && stack.maxStackQuantity() == bigPearl.maxStackQuantity(),
+            pearl);
+
+        return Stream.of(
+            RecipePlaceTest.populateTests(
```
aromaa: Let's not cram in this much parameters in to one single method, its quite hard to tell whats going on. I would suggest fluent API.

These tests are going to need to have more identifiable names so you can look at the test output and Ctrl + F the source code.

Something along the lines:
```java
test("big_pearl_recipe", ...)
    .success("Matching input", ...)
    .fail("Missing pearl", ...)
    .fail("Has stone", ...)
    .fail("Asked where the bathroom is", ...)
```

src/test/java/org/spongepowered/common/recipe/RecipePlaceTest.java
```diff
@@ -0,0 +1,431 @@
...
+                4, 8,
+                Collections.nCopies(8, bigSnowball), List.of(),
+                List.of(snowball4), List.of(snowball4)
+            )
+        ).flatMap(Function.identity());
+    }
+
+    @TestFactory
+    public Stream<DynamicTest> testRecipes() {
+        final ServerPlayer player = new FakePlayer(SpongeCommon.server().overworld(), new GameProfile(UUID.randomUUID(), "Player"));
```
aromaa: Preferably we don't share mutable state across tests.

src/test/java/org/spongepowered/common/recipe/RecipePlaceTest.java
```diff
@@ -0,0 +1,431 @@
...
+        }
+
+        final List<Integer> quantities = input.slots().stream()
+            .map(Slot::peek)
+            .map(ItemStack::quantity)
+            .distinct()
+            .toList();
+        final int max = quantities.stream().max(Integer::compare).get();
+        final int min = quantities.stream().filter(quantity -> quantity != 0).min(Integer::compare).orElse(context.expectedCrafts());
+        assertTrue(context.expectedCrafts() == max && min == max,
```
aromaa: This seems rather odd, I would expect us to compare the input and output `ItemStack`s directly that they match and their quantities are expected. You can just maintain a `List<ItemStack>` and increment their quantities each time a recipe is placed.

src/test/java/org/spongepowered/common/recipe/RecipePlaceTest.java
```diff
@@ -0,0 +1,431 @@
...
+        for (int i = 0; i < initialInput.size(); ++i) {
+            input.set(i, initialInput.get(i));
+        }
+
+        final List<String> inputs = new ArrayList<>();
+        final List<String> inventories = new ArrayList<>();
+        inputs.add(RecipePlaceTest.inventoryToString(input, false));
+        inventories.add(RecipePlaceTest.inventoryToString(playerInventory, true));
+        for (int i = 0; i < context.clickAmount(); ++i) {
+            menu.handlePlacement(context.shiftClick(), true, context.recipe(), player.serverLevel(), player.getInventory());
```
aromaa: We should short-circuit here as soon as something goes wrong. This gives the best developer experience as you can breakpoint on the exception and then force return on the debugger to re-run the test on the fly and step-by-step manually.

That should make the history log obsolete, it is a clever idea but its not as helpful as the debugger.

src/test/java/org/spongepowered/common/recipe/RecipePlaceTest.java
MrHell228: Replaced method with a lot of parameters with `TestPopulator`.
As an identifiable name for Ctrl+F I suggest `TestContext#asTestName`.
Updated everything else as well.

aromaa: This looks a lot better now! Something I didn't happen to mention explicitly and just implicitly in the code snippet is that ideally you should be able to have dynamic number of successful and failure cases.

Given this is a bit of complex surface area as ingredients and their permutations can be quite large, it would make sense to have this written more flexibly to potentially add more edge cases or programmatically generate them in the future.

MrHell228: Something like this? Edge cases could be added via `TestPopulator#test`, those tests are applied before "match" test and thus can cancel it if needed.

aromaa: Nothing wrong with the actual `TestEntry` being fixed, but rather that the inputs like the provided inventories and their expected outputs should be easily definable so you can give any number of inventories that should success or fail.

So more like `record TestEntry(boolean shiftClick, int clicks, List<ItemStack> inventory, List<ItemStack> input, List<ItemStack> expectedInputs)`

This makes all the `partialInventory` and `badInventory` obsolete and they can be defined using the same method. Nothing wrong with providing additional helpers to create those test cases but it makes this a lot more straightforward from consumer stand point and when people are glancing at them and wondering where and why did things go work.

MrHell228: I don't quite understand what should all of these parameters except `expectedInputs` do in `TestEntry`.
I assume it's `TestContext`'s role to make a layout for the actual tests.
If `TestEntry` is not supposed to do anything else than matching expected input then.. why was the previous approach with `List<List<ItemStack>> expectedInputs` not enough?

MrHell228: Wait, did you originally mean that instead of testing for _single_ input after each placement, we should test for one of _some_ allowed inputs? So like `List<List<List<ItemStack>>>` instead of `List<List<ItemStack>>`?

aromaa: Yes, the problem is the large number of permutations we can have.

Let's just take an example of simple recipe which has two ingredients, A and B. A requires either dirt or stone and B requires either iron or diamond. We now have four unique combinations that work, dirt & iron, dirt & diamond, stone & iron, stone & diamond.

This means that there are four successful combinations and there are even more options when we run out resources of one of the results. Lets say you have two dirt and iron in the crafting input but you only have dirt and diamonds left in your inventory, we might want to test that. We can also test what if their inventory is full and there are items in the crafting slot, just like how one of our mixins need to account for it.

As you can expect, the amount of potential combinations explode and the number of successful and failure cases are not just singular, so we should keep that in mind.

MrHell228: Does this even makes sense to implement?
I mean, if we know the initial layout, shouldn't we know the _exact_ behaviour after each click?
For example, case with full inventory is currently doable with just `TestContext` if needed

aromaa: Implement what, exactly? My point is that implementing exhaustive tests would require that one can provide the inputs in a sensible manner, either manually or programmatically. For the example I provided above, one could write something like this:

```java
test("custom_recipe", ...)
    .entry("test dirt and iron", inventory = List.of(dirt, iron), input = List.of(dirt, iron))
    .entry("test dirt and diamond", inventory = List.of(dirt, diamond), input = List.of(dirt, diamond))
    .entry("test stone and iron", inventory = List.of(stone, iron), input = List.of(stone, iron))
    .entry("test stone and diamond", inventory = List.of(stone, diamond), input = List.of(stone, diamond))
    .entry("test dirt only", inventory = List.of(dirt), input = List.of())
    .entry("test diamond only", inventory = List.of(diamond), input = List.of())
    .entry("test dirt in slot and diamond in inventory" inventory = List.of(diamond), craftingSlots = List.of(dirt), input = List.of(dirt, diamond))
    .entry("test full inventory unable to craft", inventory = List.of(...), craftingSlots = List.of(iron), input = List.of(iron))
```

MrHell228: Isn't it.. already doable? WIth something like
```
Stream.of(new TestContext("custom_recipe", ...)).flatMap(test -> Stream.of(
    test.name("test dirt and iron").inventory(List.of(dirt, iron)).expectInput(List.of(dirt, iron))
    test.name("test dirt and diamond").inventory(List.of(dirt, diamond)).expectInput(List.of(dirt, diamond))
    ...
    test.name("test full inventory unable to craft").inventory(List.of(...)).input(List.of(iron)).expectInput(List.of(iron))
));
```
Am I missing something?

MrHell228: Added `SimpleTestPopulator` to behave _exactly_ how you suggested.
Also added tests for full inventory and surprisingly vanilla just drops all the items to world if there is not enough space in inventory. So that check in mixin is actually not required.

MrHell228: Well, it dropped items because it was said to place recipe in creative mode.
Without creative mode it retains item in crafting grid.
I think this can be a flag in `RecipeContext`

aromaa: > Isn't it.. already doable? WIth something like

Ah, right, that's actually totally fine!

Whats actually throwing me off is the fact you have crammed in the `badInventory` and the `partialInventory` in the same class, forcing you to define one successive and two failing cases at once. We should just generate one case for single entry. If you can just split them off to their own individual cases, like following the code snipped you just gave me, then that's fine! Feel free to use helper method to generate them.

Its better for tests to be verbose and expressive as they are meant to be easily understood.

MrHell228: Moved tests for bad layouts (which involved `badInventory`) to separate `TestPopulator`.

aromaa: Looks like theres still some `partialInventory` and `partialInput` and `badInput` leftover in the `DefaultedTestPopulator`

src/test/java/org/spongepowered/common/recipe/RecipePlaceTest.java
```diff
@@ -400,11 +412,49 @@ public Stream<TestContext> populate() {
...
+        public BadLayoutTestPopulator badInput(final List<ItemStack> items) {
+            this.badInput = items;
+            return this;
+        }
+
+        @Override
+        public Stream<TestContext> populate() {
+            return Stream.of(this.base)
+                .flatMap(context -> Stream.of(
+                    context.name("Empty input").input(List.of()),
```
aromaa: These should be directly asked for by generating the tests with the the appropriate `badInventory` and `badInput`.

src/test/java/org/spongepowered/common/recipe/RecipePlaceTest.java
MrHell228: I just don't understand what you want me to do honestly. I changed it again but at this point I'm not even sure if this is what I had to change

aromaa: Feel free to ask before you change anything. I understand that this is a bit new to you, and I'm honestly giving you a bit too vague feedback because I'm leaving gaps and just pointing to a direction and expecting some existing knowledge, that's on me.

The general gist of unit tests are that they should be very simple and not overly complex and easy to follow (and setup) and have most (if not all) relevant information on few lines. The general problem is that the shape was overly complex, you had to understand the program flow, had multiple inputs and tests with variety of different situations with different inputs (not very obvious) crammed in together.

It's important that test code is easy to understand so people don't have to spend a whole lot of time just to see what the test is doing. Its common to see these files to be very long and have repetitive copy & paste with just few changes for this exact reason.

So instead of having something like:
```java
test(recipe).inventory(...).input(...).badInventory(...).badInput(...)
```
its more natural to write something like:
```java
test(recipe).inventory(...).input(...).state(success)
test(recipe).inventory(...).input(...).state(success)
test(recipe).inventory(...).input(...).state(fail)
test(recipe).inventory(...).input(...).state(fail)
test(recipe).inventory(...).input(...).state(fail)
```
Now the test cases are obvious and their inputs and their expected outcomes can be read and understood in a single line without having to look around. It's also very easy to understand how to add new tests cases.

That said, yes, the `badLayoutTests` method is something I prefer. Its direct and obvious and hard to miss.

PR #4248, opened 2025-10-20
Adapt modifications of GameProfile PropertyMap to the new immutable design

avaruus1:
PropertyMap is now immutable.


aromaa (approved):
Thank you!

PR #4249, opened 2025-10-25
Fix registering to builtin registries with dependencies

MrHell228:
So, my use case is:
```
public void registerRegistryValues(RegisterRegistryValueEvent.GameScoped e) {
  e.registry(RegistryTypes.ITEM_TYPE, (holder, step) -> ..., customGameScopedRegistry);
}
```
Currently it just crashes the server. If no dependencies are provided, it's registered fine.
I understand that generally builtin registries are not supposed to be extended by plugins, but it doesn't crash without dependencies, so I assume current behaviour with dependencies as a bug.


aromaa (approved):
Thank you! This was the original intent.

PR #4250, opened 2025-10-25
fix SpongeShapelessRecipe.matches (API 12)

MrHell228:
As I said [here](https://github.com/SpongePowered/Sponge/pull/4240#issuecomment-3447217230), this is a simple fix just to make it work as intended


aromaa (approved):
Thank you!

PR #4251, opened 2025-11-01
Fix spectating of other entities not working

abola2:
Requires https://github.com/SpongePowered/Sponge/pull/4252

### DIFF
--- a/src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerPlayerMixin.java
+++ b/src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerPlayerMixin.java
@@ -619,7 +619,7 @@ public abstract class ServerPlayerMixin extends PlayerMixin implements SubjectBr
 
         }
 
-        thisPlayer.setCamera(thisPlayer);
+        thisPlayer.setCamera(thisPlayer.getCamera());
         thisPlayer.stopRiding();
         if (thisPlayer.isSleeping()) {
             thisPlayer.stopSleepInBed(true, true);


src/mixins/java/org/spongepowered/common/mixin/core/server/level/ServerPlayerMixin.java
```diff
@@ -619,7 +619,7 @@ public boolean teleportTo(final ServerLevel world,
 
         }
 
-        thisPlayer.setCamera(thisPlayer);
+        thisPlayer.setCamera(thisPlayer.getCamera());
```
Yeregorix: This is equivalent to not calling `setCamera` at all. The line is here on purpose to stop spectating when changing dimension. The real problem is why is this line executed when the player is teleporting without changing dimension. I'm gonna investigate a bit.

Yeregorix: Turns out the method names `impl$fireDimensionTransitionEvents` and `bridge$changeDimension` are misleading. They are called for all teleports but have special treatment for actual dimension changes.

Yeregorix: Your fix causes positioning issues when spectating an entity and then teleporting. Whereas in pure vanilla it just stops spectating.

aromaa: Something worth to mention is that recent versions have changed the behavior of how passengers and spectators are teleported, and make sure we follow vanilla.

abola2: > Your fix causes positioning issues when spectating an entity and then teleporting. Whereas in pure vanilla it just stops spectating.

Nice catch! After digging into this, it seems the issue is caused because TeleportCommandMixin’s overwrite didn’t follow vanilla logic. In vanilla, the player’s camera would be set back to the player in teleportTo. I’ve opened https://github.com/SpongePowered/Sponge/pull/4252, another PR that restores the logic to match vanilla behavior.

Yeregorix:
I went down the rabbit hole and discovered many more issues related to our teleport implementation. So I did a full refactor which fixed this issue and many others. Thank you for identifying the spectator issue.

Issue #4253, opened 2025-11-13
Console spam and crash when the server fails to find a valid spawn


eirikh1996:
### Affected Product(s)

SpongeVanilla

### Version

spongevanilla-1.21.10-17.0.0-RC2410-universal

### Operating System

Windows 11

### Java Version

openjdk version "24" 2025-03-18

### Plugins/Mods

```shell
[19:33:00] [Server thread/INFO]: Plugins (5)
[19:33:00] [Server thread/INFO]:     Minecraft v1.21.10
[19:33:00] [Server thread/INFO]:     SpongeVanilla v1.21.10-17.0.0-RC2410
[19:33:00] [Server thread/INFO]:     Sponge v1.21.10-17.0.0-RC2410
[19:33:00] [Server thread/INFO]:     SpongeAPI v17.0.0-SNAPSHOT
[19:33:00] [Server thread/INFO]:     SpongeTestPlugin v1.0-SNAPSHOT
```

### Describe the bug

When trying to join the server after booting up SpongeVanilla, the console gets spammed with error messages. After a while, the server will crash

## To reproduce
1. Install SpongeVanilla and set up the server
2. Try joining the server
3. Watch the spam of error messages in the console and subsequent crash

### Link to logs

https://gist.github.com/eirikh1996/13665dac6e6f19a2978430af4a80a75c

zKOSMOSz:
The same problem exists in SpongeForge.
https://github.com/SpongePowered/Sponge/issues/4257

Yeregorix:
I think this is a false positive of the phase tracker runaway detection. As a workaround until we have a proper fix, you can disable it by editing `config/sponge.conf` and setting `phase-tracker { verbose=false }`.

The technical explanation:
[...]

PR #4258, opened 2025-11-28
Avoid recursion as much as possible when searching for a spawn

Yeregorix:
Fix [MC-304426](https://bugs.mojang.com/browse/MC/issues/MC-304426) and #4253.

I'm not sure if we want to avoid the overwrite here because there are some important control flow changes.


src/mixins/java/org/spongepowered/common/mixin/core/server/level/PlayerSpawnFinderMixin.java
```diff
@@ -0,0 +1,160 @@
...
+                } catch (Exception exception) {
+                    error = exception;
+                }
+            }
+
+            if (error != null) {
+                this.crash(x, z, index, error);
+            }
+        }, server);
+        return true; // exit loop
```
aromaa: Might want to handle the race, too?
```suggestion
        final AtomicReference<OperationState> state = new AtomicReference<>(OperationState.UNINITIALIZED);
        future.whenCompleteAsync((chunk, error) -> {
            if (error == null) {
                try {
                    Optional<Vec3> result = supplier.get();
                    if (result.isPresent()) {
                        this.finishedFuture.complete(result.get());
                    } else if (state.getAndSet(OperationState.EXECUTED) == OperationState.WAITING) {
                        this.scheduleNext();
                    }
                } catch (Exception exception) {
                    error = exception;
                }
            }

            if (error != null) {
                this.crash(x, z, index, error);
            }
        }, server);
        
        if (state.compareAndSet(OperationState.UNINITIALIZED, OperationState.WAITING)) {
            return true; // exit loop
        }
        
        return false;
```

src/mixins/java/org/spongepowered/common/mixin/core/server/level/PlayerSpawnFinderMixin.java
```diff
@@ -0,0 +1,160 @@
...
+    @Shadow @Final private int radius;
+    @Shadow @Final private int candidateCount;
+    @Shadow private int nextCandidateIndex;
+    @Shadow @Final private int offset;
+    @Shadow @Final private int coprime;
+
+    @Shadow
+    private static Vec3 fixupSpawnHeight(CollisionGetter $$0, BlockPos $$1) {
+        throw new IllegalStateException();
+    }
```
aromaa: We usually just squash these like this so the LOC's stay small.
```suggestion
    @Shadow private static Vec3 fixupSpawnHeight(CollisionGetter $$0, BlockPos $$1) { throw null; }
```

aromaa (approved):
LGTM, modulo the two comments.

While the `Overwrite` is a bit invasive, I don't think there is any other good alternative. We can always revisit if something comes up.

src/mixins/java/org/spongepowered/common/mixin/core/server/level/PlayerSpawnFinderMixin.java
Yeregorix: I don't think it's necessary to stay on the loop when the future has already completed. Yes it's a unnecessary recursion but this one doesn't overflow the stack since it's a recursion between different threads.

aromaa: > since it's a recursion between different threads.

Not entirely sure which different thread you are referring to? This is just a classical race condition, the `future#isDone` returns 'false' on the line just above and then it happens to complete before or while executing `future#whenCompleteAsync` and we are still in the same thread.

But yeah, I don't have strong feelings about it too, just wanted to point it out.

PR #4259, opened 2025-11-30
Fix dimension query

MrSlimeDiamond:
This fixes getting an offline player's location via `User#worldKey()`

At some point, it was renamed to Dimension instead of WorldKey I suppose

`SpongeEntityArchetype` was changed but seemingly nothing is different in my testing?


Yeregorix:
Thank you for spotting this. IRC back in the days, the `dimension` field was just a int and Sponge had to store a custom field `WorldKey` for the full string id. Then Mojang refactored their dimensions system and added a `Dimension` field with the string id.

PR #4260, opened 2025-12-08
Fix incorrect minimum height check in teleportHelper

abola2:
The Minecraft Overworld’s minimum world height is –64, and mods or datapacks can change it. The current check prevents teleporting below Y 0.


Yeregorix:
Thank you!

PR #4272, opened 2026-01-15
Implement Adventure ClickEvent.callback

avaruus1:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2609) | **Sponge**

Implements Adventure's ClickEvent.callback & moves Sponge's existing callback implementation to use the new system.


gabizou (approved):

Yeregorix:
I agree with this change but please note that it may not stay very long in place as I plan to replace all our adventure implementation with the official adventure-platform-mod as soon as I have time to work on it.

aromaa:
Just gonna mention it here too that this will not satisfy our contracts. The permission checks are done against the `Audience` instead of the underlying `Subject`. These objects are not equal and each can hold distinct set of permissions. I'm worried that this will be surprising behavior as it will go against the default contract we have for commands.

Issue #4269, opened 2025-12-30
Unbanning a player does not work


abola2:
### Affected Product(s)

SpongeVanilla

### Version

1.21.10-17.0.0-RC2468

### Operating System

linux

### Java Version

21

### Plugins/Mods

```shell
none
```

### Describe the bug

Unbanning a player does not work. This can be reproduced by banning a player `/ban example` and then attempting to unban them with `/pardon example`

### Link to logs

https://gist.github.com/abola2/f279f918562e5ee910b38123d4d3f7bd

abola2:
fixed by https://github.com/SpongePowered/Sponge/pull/4275

PR #4275, opened 2026-02-27
#4269: Fix pardon command

AlexandreArcil:
Issue #4269: Unbanning a player does not work

`/pardon` command wasn't working due to a silent **ClassCastException**: the key element type of the list `SpongeUserBanList` is `NameAndId` but a `GameProfile` was given instead.

The fix just change the element type to be removed by a `NameAndId`.


Yeregorix:
Thank you!

PR #4280, opened 2026-03-20
Fix world border issues and cleanup

Vilsu221:
Fix WORLD_BORDER key, border loading

Remove old per-world border
Vanilla now implements per-world world border


aromaa (approved):
LGTM. Thank you!

Issue #4284, opened 2026-03-26
Abnormal explosion logic of TNT, Creeper, etc


MC-Nirvana:
### Affected Product(s)

SpongeNeo

### Version

1.21.1-21.1.211-12.0.4-RC2569

### Operating System

Windows、Linux

### Java Version

openjdk 21.0.8 2025-07-15 LTS

### Plugins/Mods

```shell
Mod List:
                Name Version (Mod Id)

                LuckPerms 5.5.36 (luckperms)
                Minecraft 1.21.1 (minecraft)
                MiniPlaceholders 3.1.0 (miniplaceholders)
                NeoForge 21.1.221 (neoforge)
                SpongeNeo 1.21.1-21.1.211-12.0.4-RC2569 (spongeneo)
                WorldEdit 7.3.8+6939-7d32b45 (worldedit)
```

### Describe the bug

The explosion logic of TNT, Creeper, etc. is abnormal. There are no particle effects or sounds

### Link to logs

https://mclo.gs/SsFTsSV

Yeregorix:
Fixed by 6cec0960d0d87c639a1c425833fdcaaa96367d6b.

PR #4293, opened 2026-04-26
#4284: Fix abnormal explosion logic of TNT, Creeper

AlexandreArcil:
#4284: Abnormal explosion logic of TNT, Creeper, etc

After an explosion, the `ClientboundExplodePacket` packet was not sent, whereas it is needed to play the explosion sound and particules effects on client (See `ClientPacketListener#handleExplosion`).

I used `ServerLocation` because it is created with the original x, y and z:
https://github.com/SpongePowered/Sponge/blob/582aacf1acb9125b6b7d612878f78ab1009a909e/src/mixins/java/org/spongepowered/common/mixin/core/world/entity/item/PrimedTntMixin.java#L123-L132


Yeregorix:
Thank you!

PR #4304, opened 2026-08-26
Fix HumanEntity registration missed during udate

avaruus1:
Previously, Minecraft registered the EntityTypes in the `EntityType` class, but the registration was moved to EntityTypes class.

As a result of this, I was getting disconnected due to a "Network Protocol Error" when trying to join the server.

EntityType class diff: https://mcsrc.dev/2/diff/26.1.2/26.2/net/minecraft/world/entity/EntityType


aromaa (approved):
Thank you!

PR #4305, opened 2026-08-27
Add missing keys for hiding tool tip data components

GiganttinenEtana:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2615) | Sponge

Add missing keys for hiding tool tip data components. Map.of replaced with ImmutableMap.Builder due to too many parameters.


aromaa (approved):
Thank you!

PR #4307, opened 2026-09-01
Fix entities not spawning in structures on Forge

XMxmx111:
Sponge replaces MinecraftServer#loadLevel with this world loading path, which never reaches MinecraftServer#prepareLevels.
Forge injects StructureSpawnManager.gatherEntitySpawns() at the very start of prepareLevels to populate its structure spawn registry (structuresWithSpawns, including the vanilla defaults pre-filled by the gather event).

### DIFF
--- a/forge/build.gradle.kts
+++ b/forge/build.gradle.kts
@@ -176,6 +176,13 @@ configurations.configureEach {
     if (name != "minecraft") { // awful terrible hack sssh
         exclude(group = "com.mojang", module = "minecraft")
     }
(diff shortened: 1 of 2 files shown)


Yeregorix:
Ah I didn't notice that this was about API 8 when I suggested to open a PR. We usually no longer work with this outdated version but I guess I will make an exception this time because the changes are quite small. However we no longer merge API 8 changes upward, but this shouldn't be a problem in this case because the `loadLevel` override is gone in API 12.

